Issue #10, opened 2015-12-07
Cancel on next checkpoint


philsmd:
Attention: this issue is one of those imported from the old hashcat trac ticketing system.

OP: unix-ninja
Date of report: 2015-10-14
Original post:
"We should add a feature to the hashcat family of apps where we can press a hotkey to quit after the next "checkpoint" rather than immediately. Maybe (c) for cancel at checkpoint?"

This ticket was accepted by 'atom' on the trac ticketing system. 
Thx

jsteube:
OK, merged. Good work, thanks!

PR #11, opened 2015-12-07
implements the enhancement mentioned in issue #10 (cancel on next checkpoint)

philsmd:
The goal of this pull request is to implement the enhancement mentioned in issue #10: an user should be allowed to press a hotkey (in this case 'c') to tell oclHashcat that it should stop right after the next restore point was reached (and also saved to the .restore file).

This feature should work whenever --restore-disable was not set.

oclHashcat will continue cracking until the next restore point was reached, after that it will stop cracking with status code "aborted" (2).

Please review. Thx

### DIFF
--- a/include/shared.h
+++ b/include/shared.h
@@ -1883,15 +1883,16 @@ extern hc_thread_mutex_t mux_display;
  * status
  */
 
-#define STATUS_STARTING       0
-#define STATUS_INIT           1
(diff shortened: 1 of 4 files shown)


epixoip:
Ok to commit

Xanadrel:
Ok for merging.

PR #12, opened 2015-12-07
allow both enabling and disabling of restore point update (implemented with issue #10)

philsmd:
This is just an additional small feature to issue #10: also allow the user to disable a restore point monitoring check, i.e. with this extra feature the user can do both create a new check or disable the check if it is already there (but the user does not want it anymore).
Thx


Xanadrel:
Ok for merging.

Issue #14, opened 2015-12-09
hashcat-deps (dependencies) should not be in /opt/ and therefore not require root permission


philsmd:
Currently the dependencies for oclHashcat (and for hashcat too) are all within the /opt/ directory (/opt/hashcat-deps to be exact).

I suggest that we should move the hashcat-deps directory directly into the root path of the github clone such that no root permissions are required and that the dependencies are local to the cloned repository.

Thx

jsteube:
Merged, thanks!

PR #15, opened 2015-12-09
issue #14: move /opt/hashcat-deps/ to [git_clone_dir]/hascat-deps/

philsmd:
This commit fixes issue #14, remove the dependency on the /opt/ folder (which can only be modified by root users under linux).

(P.S.: We should do the same for hashcat)


epixoip:
Looks good, I'll provide my signoff for this

PR #16, opened 2015-12-09
issue #16: use just "deps" instead of "hashcat-deps" as directory name

philsmd:
I suggest that we rename the "hashcat-deps" folder (intruduced by solving issue #16) to just "deps".
Thx


Xanadrel:
Ok for merging.

PR #21, opened 2015-12-10
Add an .editorconfig file enforcing some of the code style.

magnumripper:
This helps a lot when working in projects with contradicting code styles. See http://editorconfig.org


jsteube:
I don't even know what this is, but I know that you know what you're doing. Still, need one more sign-off by the board members to merge.

Xanadrel:
This smells good, can be merged.

PR #22, opened 2015-12-10
BUILD.md incorrectly referred to docs/deps.sh as opposed to tools/deps.sh

magnumripper:


Xanadrel:
Ok for merging obviously.

philsmd:
Yes, this is correct. The link is currently not correct (I think this problem was introduced because deps.sh was moved to /tools/ right before the source was released and the link somehow was not updated at all).
Please merge

jsteube:
Thanks!

PR #30, opened 2015-12-22
fix OpenCL memory problem (exhaustion): -4 error should be gone

philsmd:
There was a slight memory problem (especially noticeable when using --benchmark with all default algos). errors of type -4 etc should not occur anymore.
Thx


epixoip:
Ok to merge

PR #34, opened 2015-12-28
SHA1 optimization

Merenon:
Short description:
https://crypto.beer/additional-slight-optimization-of-jens-steubes-sha1-message-expansion-attack/

### DIFF
--- a/amd/m00100_a3.cl
+++ b/amd/m00100_a3.cl
@@ -444,7 +444,9 @@ static void m00100s (u32 w[16], const u32 pw_len, __global pw_t *pws, __global g
 
     const u32x w0s04___w0s06 = w0s04 ^ w0s06;
     const u32x w0s04___w0s08 = w0s04 ^ w0s08;
+    const u32x w0s07___w0s12 = w0s07 ^ w0s12;
     const u32x w0s08___w0s12 = w0s08 ^ w0s12;
+    const u32x w0s12___w0s16 = w0s12 ^ w0s16;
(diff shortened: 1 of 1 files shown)


philsmd:
Are you sure this is related to md5 at all (the title says "Md5 optimization")?
Even if just the title is wrong and it is all about SHA1, how did you test the speed gains? Did you benchmark it? (I know the optimization is by Jens aka atom, yeah this project leader!).

One of the main problems could be register pressure etc. Implementing optimizations only make sense if it leads to a general speedup. Not all optimizations work on all platforms (e.g. cpu/gpu). Do you have some benchmarking numbers?

gm4tr1x:
Hi, the "Word Expansion / XOR groups" optimization has been made by me 3 years ago.
What @Merenon found is well known but it was not applied for problems related to the performance. As @philsmd say, please do some benchmarks and post the results here.

Merenon:
Hi, of course, you're right, it's related to SHA1 (I changed the title).
I will do some benchmark tests and post them. Thanks for your comments.

magnumripper:
You could obviously do something like this (given there is some macro(s) usable for selecting platforms where this does improve performance).
```c
#if __this_or_that__
const u32x w0s07___w0s12 = w0s07 ^ w0s12;
#else
#define w0s07___w0s12 w0s07 ^ w0s12
#endif
```
It'd still not be a 100% no-op for the else clause (due to different ordering) but that should not be much of a problem.

jsteube:
There was alot of changes due to due Get rid of CUDA branch. Please send in a new PR based on the current master

Issue #38, opened 2016-01-01
nvml lib is not needed anymore (for compiling)


philsmd:
Since the GetRidOfCuda (GROC) branch did also implement the dynamic loading of the nvml library, we do not need to install the NVIDIA sdk anymore.
We should update the src/Makefile, tools/deps.sh and docs/BUILD.md to address this issue.
Thx

PR #39, opened 2016-01-01
issue #38: NVIDIA SDK is not needed anymore

philsmd:
The NVML lib is dynamically loaded on Nvidia systems, therefore there is no need (anymore) that the devs install it on the dev system (the required library ships with the driver anyway).

Thanks


epixoip:
Ok to merge

Issue #40, opened 2016-01-03
native compilation of oclHashcat also requires paths to libOpenCL.so


philsmd:
Currently the native compilation of oclHashcat ("make") does not have a link to the libraries in the deps/amd-app-sdk/ folder. This is required because otherwise the user will get this error:
/usr/bin/ld: cannot find -lOpenCL

I suggest that we need to introduce a OCLLIBPATH_NATIVE variable that will be set to the correct 64-bit or 32 bit libOpenCL.so parent directory.

jsteube:
I just want to comment the real background of this issue here. AMD's Catalyst driver doesn't really install it's own 'libOpenCL.so' correctly in /usr/lib because it installs it as 'libOpenCL.so.1' and forgets to symlink back to 'libOpenCL.so'. NVidias ForceWare is more clever and installs it, so this problem never occurs in NV installations anyway.

jsteube:
As an alternative to all this we could think of using:

https://github.com/KhronosGroup/OpenCL-Headers
https://github.com/KhronosGroup/OpenCL-ICD-Loader

The first one provides the CL/cl.h and the second one the libOpenCL.so. This is the Khronos OSS reference implementation.

Ob Ubuntu/Debian (mayber others, too) those exist as installable packages. Both without any depencies:

opencl-headers
ocl-icd-libopencl1

We should look into them and maybe switch. This would also enable us to drop the depency on amd-app-sdk which is weird anyway for NVidia users

jsteube:
While working on it it turns out on ubuntu 14.04 it needs another package:

ocl-icd-opencl-dev

But this package is somehow broken in 15.04 (wants to uninstall xorg-server and such). We need to make sure in deps.sh that this package is selected only in case the libOpenCL.so wasn't created

PR #41, opened 2016-01-03
issue #40: native compilation fixed with OCLLIBPATH_NATIVE variable

philsmd:
this is a fix for issue #40. The problem was that whenever the user tried to compile with the native compilation targets (the default one, "make") it happened that libOpenCL was not found (error: cannot find -lOpenCL).
This happened because the libOpenCL.so is within the deps/amd-app-sdk/ directory, but there was no link to it when using native compilation.


jsteube:
Replaced with https://github.com/hashcat/oclHashcat/commit/dc13c2fc71c8c9e3af66ceaa9dfc8ff34cb56bfb

PR #42, opened 2016-01-03
fixed tab completion to work with 64.bin, 32.bin and just oclHashcat binaries

philsmd:
Since we now allow the user to use "sudo make install", tab-completion had to be updated to work also with just "oclHashcat ..." in the command line.
This patch addresses this problem and in addition to that it changes the name of the bash complete script to "oclHashcat.sh" (was oclHashcat64.bin) such that it is more clear that this script works for all three variants: oclHashcat64.bin, oclHashcat32.bin and oclHashcat
Thx

### DIFF
--- a/extra/tab_completion/howto.txt
+++ b/extra/tab_completion/howto.txt
@@ -1,4 +1,4 @@
 run:
 source ./install
 OR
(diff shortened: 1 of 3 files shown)


Xanadrel:
Ok for merging.

PR #57, opened 2016-01-04
The changes need to be commite for compiling the package in Arch Linux

yhfudev:
It includes three patches for compile and run correctly on Arch Linux.
1. the bool is re-defined.
2. the debug message, there're several similar 'ERROR:' messages in the source code, so we add more detail in it so that we can locate the error point instantly. We also add one function to convert CL error code to its error messages.
3. add --session-dir argument. The oclhashcat would complain the system dir '/usr/bin/' is not writable if we install the binaries, and we have to set it to other location.

Example:
/oclhashcat -m 2500 test-01-hs.hccap -a 3 ?d?d?d?d?d?d?d?d?d?d --outfile-check-dir=$(pwd)/aaa -session-dir=$(pwd)/bbb --gpu-platform 2

For the complete package of the oclhashcat on ArchLinux, please see https://aur.archlinux.org/packages/oclhashcat-git/

### DIFF
--- a/include/shared.h
+++ b/include/shared.h
@@ -2240,7 +2240,7 @@ void hc_signal (void c (int));
 
 #endif
 
-typedef int bool;
+//typedef int bool;
 
 bool class_num   (char c);
 bool class_lower (char c);
--- a/src/ext_OpenCL.c
+++ b/src/ext_OpenCL.c
@@ -5,6 +5,25 @@
 
 #include <ext_OpenCL.h>
 
+const char *
+val2cstr_cl (cl_int CL_err)
+{
+#define CLERR(a) case a: return #a
+    switch (CL_err) {
+    CLERR(CL_INVALID_PROGRAM);
+    CLERR(CL_INVALID_VALUE);
+    CLERR(CL_INVALID_DEVICE);
+    CLERR(CL_INVALID_BINARY);
+    CLERR(CL_INVALID_BUILD_OPTIONS);
+    CLERR(CL_INVALID_OPERATION);
+    CLERR(CL_COMPILER_NOT_AVAILABLE);
+    CLERR(CL_BUILD_PROGRAM_FAILURE);
+    CLERR(CL_OUT_OF_RESOURCES);
+    CLERR(CL_OUT_OF_HOST_MEMORY);
+    }
+    return "(unknown CL error)";
+}
+
 void hc_clEnqueueNDRangeKernel (cl_command_queue command_queue, cl_kernel kernel, cl_uint work_dim, const size_t *global_work_offset, const size_t *global_work_size, const size_t *local_work_size, cl_uint num_events_in_wait_list, const cl_event *event_wait_list, cl_event *event)
 {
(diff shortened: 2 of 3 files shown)


epixoip:
There are several issues with this PR:

1. You did not follow the "Solve only one problem in each pull request" rule.

2. The --session-dir stuff is already obsolete based on other recent changes.

3. Getting rid of the typedef bool is not the way to fix that problem.

yhfudev:
What't the alternative method to get the session-dir work?

There's multiple definition of bool, what's your solution?
include/ext_ADL.h:13:typedef int bool;
include/shared.h:2243:typedef int bool;

yhfudev:
github will combine the request even I post 3 times.

jsteube:
I think you're a tad bit late for this change. Please see the changes discussed here:

https://github.com/hashcat/oclHashcat/issues/20

This issue addressed all the general problems maintainer could have when trying to build packages for oclHashcat and should solve it with current master branch.

You, as a package maintainer, are one of the people why we did that change. It would be really nice if you could please find out if this change helps you in integrating oclHashcat in your packaged version or if we are still missing something (like the the bool thing from above)

yhfudev:
Thank you for your explain. Do you have any plan to use autoconf style make system to compile the source code? It can handle the dir problems easily, and it's cross platform and can eliminates package maintainer's pain. I can try to create one if you're interest.

gm4tr1x:
Hi @yhfudev, check it 
https://github.com/gm4tr1x/hashcat/tree/autotools

yhfudev:
@gm4tr1x 
Great job!

Are you still developing the Makefile.am? It seems the other files not installed, for example the files in /usr/share/?

And the bin name hashcatocl would be better so that he can use 'tab' key to explore which command he should use if he installed both hashcat and oclhashcat.

gm4tr1x:
@yhfudev yes, what you see it's only the "initial commit".
First i like to see windows binaries work, if you have time for test it ... you are welcome ;)

jsteube:
Patch 1) I liked some parts of your patch, especially the fix for the bool type. I've modified a bit and pushed to master as part of this patch:

https://github.com/hashcat/oclHashcat/commit/0ff49c5b44bbe0eab4daa3b76389c66c306e4669

Patch 2) I liked the idea of having a human-readable error message for the OpenCL error codes. Unfortionally the reference OpenCL API does not have something like strerror (), therefore I think we actually need to use such a switch () composition. However I'd like to have it in full, not just for clBuildProgram (). Maybe you can provide a full switch block and patch all the wrapped functions?

Patch 2) I didn't like the idea of having an description which variable/option a failed fopen () belongs to. That's very untypical for unix-type tools, for example:

```
$ cat x
cat: x: No such file or directory
```

Patch 3) With latest master version of oclHashcat and using the "normal" make install version of oclHashat there's no more need for such a parameter. The session folder will be somewhere in $HOME/.hashcat/sessions/xxx which is something a user should have write-permissions to.

yhfudev:
The error message is for locating errors quickly. I don't think it is just for 'good looking'. The problem of the error message is there're same message printed in different location in the source code. I met this during compile and use the software and have to dig into the source to find how to use it.

PR #59, opened 2016-01-05
the parameter --opencl-platform takes a number, not a string

philsmd:
This makes it much more clearer that the option --opencl-platform takes a integer number (1, 2 or 3 etc) and that it is not a string argument.


src/oclHashcat.c
```diff
@@ -384,7 +384,7 @@ const char *USAGE_BIG[] =
   "       --bitmap-max=NUM              Maximum number of bits allowed for bitmaps",
   "       --cpu-affinity=STR            Locks to CPU devices, seperate with comma",
   "  -d,  --opencl-devices=STR          OpenCL devices to use, separate with comma",
-  "       --opencl-platform=STR         OpenCL platform to use, in case multiple OpenCL platforms are present",
+  "       --opencl-platform=NUM         OpenCL platform to use, in case multiple OpenCL platforms are present",
   "  -w,  --workload-profile=NUM        Enable a specific workload profile, see references below",
```
magnumripper: Perhaps you should also shorten it to *OpenCL platform to use, in case multiple platforms are present*

jsteube: ok, will do after merge

PR #118, opened 2016-01-12
fixed % and C (degrees celsius) in status display

philsmd:
this is just a cosmetic fix.
For some reason the util, temp and fan variables were not in the correct order and therefore the %, c, %/fan was not matching.
This should be fixed now (by correcting the order of the variables).


Xanadrel:
Ok for merging.

PR #119, opened 2016-01-12
HMS fix: only copy the ADL adapter info for GPU devices

philsmd:
... for CPU devices we should leave it AS-IS (zeroed out).

We only need the ADL info for GPU devices, if we encounter a non-GPU devices (e.g. CPU) we leave the current data.hm_device[] struct with all-zero-values (i.e. fan support turned off etc).
This prevents that we get errors about not working fan reading etc.


Xanadrel:
Ok for merging.

PR #128, opened 2016-01-13
set default iteration count for -m 2100 = DCC2 to 10240

philsmd:
this pull request makes sure that the iteration count for -m 2100 = DCC2 hashes is always correctly set.
In some cases, user reported that DCC2 is very slow or they see 0 H/s. The problem was that the atoi () function for some hashes (without iteration count) returned 0 and therefore salt->salt_iter was not correct.
Hereby, I suggest to add an extra check to make sure that the value returned by atoi () is at least 1. If this is not true, we should set the value to the default iteration count (ROUNDS_DCC2).

Thx


epixoip:
This looks sane, ok to merge

PR #132, opened 2016-01-15
Remove some memsets

neheb:
Redundant braces are for removing clang's warnings.


philsmd:
this looks good. Can be merged :)
Note: it's just a different notation. compilers will optimize the memset () anyway. so this is just a nice code re-write (no changes in logic or performance).

jsteube:
thanks!

PR #134, opened 2016-01-16
human-readable error message for the OpenCL error codes.

yhfudev:


philsmd:
this PR seems good. Can be merged.
Thanks

btw: @yhfudev please stick to the formatting rules (line breaks, indentation etc). It would be also great if you write just a little bit more what your (future) commits do and in this case for instance if it contains really every switch case that OpenCL defines (and where we can find the complete list - that e.g. you used as a reference - to double-check etc). This PR is great, but some more description and explanation about how complete it is would also be a PLUS. Thx again

PR #135, opened 2016-01-17
Fixed a buffer overflow in ascii_digest for hash type 8300

matrix:


philsmd:
wow, this is really a great find.
Thx matrix

But we need to fix it a little bit differently. The facts:
-  given the example hash 7b5n74kq8r441blc2c5qbbat19baj79r:.lvdsiqfj.net:33164473:1 (from hashcat.net/wiki), we know that the digest/hash is 32 base32 chars
- within the nsec3_parse_hash () function we convert (base32 decode) the human readable hash like this: base32_decode (itoa32_to_int, (const u8 *) hashbuf_pos, 32, tmp_buf);  (see https://github.com/hashcat/oclHashcat/blob/master/src/shared.c#L13950 )
- the result are 20 bytes because 32 / 8 \* 5 = 20 (we can see that the 20 bytes are copied here: https://github.com/hashcat/oclHashcat/blob/master/src/shared.c#L13952)
- this means, that our encode (the opposite of the decode in the parsing function) must convert 20 bytes to 32 human readable base32 chars
- that means, the only problem within this line (the encoding within the ascii_digest () function): base32_encode (int_to_itoa32, (const u8 *) digest_buf, 32, (u8 *) digest_buf_c); (see https://github.com/hashcat/oclHashcat/blob/master/src/shared.c#L7116) is the number 32
- we need to fix the value 32 to 20 (as input to the base32_encode () function ! (and not increase any buffer so it should stay char digest_buf_c[33]; because we want 32 + 1 additional byte for the NULL-byte

In other words, the fix should be as simple as this:

```
     -    base32_encode (int_to_itoa32, (const u8 *) digest_buf, 32, (u8 *) digest_buf_c);
     +    base32_encode (int_to_itoa32, (const u8 *) digest_buf, 20, (u8 *) digest_buf_c);
```

(within the hash_mode == 8300 section of the ascii_digest () function)

@gm4tr1x could you please submit this change to this PR (just commit it to the same branch)
Thx (and again nice find ;) )

matrix:
@philsmd, yes but we need also "digest_buf_c[34]"
Bye ;)

philsmd:
oh the extra 1 byte is for the padding character "=".

This is now okay to merge. thx

Issue #120, opened 2016-01-12
support building without ADL (and without CUDA)


rhertzog:
I would like make a proper Debian package of oclHashcat for Kali Linux, unfortunately it's not currently possible because it's not possible to build oclHashcat without ADL and ADL's license makes it impossible to ship that SDK even in the non-free section of Kali (you need a special authorization to ship it as part of an operating system).

I wanted to build the package with only OpenCL support but actually building the OCL version fails because it cannot find adl_sdk.h... the codebase seems to entangle the header files of the various SDK.

Ideally we can build each version independently (./configure --with-opencl or --with-adl or --with-cuda if you added autoconf support). But at the very least we should have makefile targets to build each version.

jsteube:
Hey, first for thinking of building a proper package for Kali. We really need this!

My comments on this so far:

1) Why do you want to ship the SDK at all? You don't need to ship the SDK. All you need is the ADL library itself (the .so files) and those are part of the Catalyst driver package. That's why it's enough for endusers to simply install Catalyst and oclHashcat will work. Users really need this feature to avoid burning their hardware.

2) With oclHashcat v2.10 there's no more CUDA at all. NVidia devices will use OpenCL runtime the same way as AMD devices. You don't need to worry about CUDA anymore! This also enabled us to use OpenCL CPU drivers. See here for details: https://github.com/hashcat/oclHashcat/issues/4 and https://github.com/hashcat/oclHashcat/issues/3

3) There's no ./configure in oclHashcat because we do not use autotools. There are people working on it but I think it will take some time. Anyway even with autotools we'd need to change the code so that it accepts an build without hardware-management. But in regards to 1) I don't think we'd need such a solution.

4) oclHashcat v2.10 had lots of changes to make it more easy for maintainer to integrate it into a packaging system. Basically you simply set the PREFIX in Makefile as you'd do it with ./configure and the files were installed on there after running make install. Files created for sessions or kernel cache are stored in home directory. See here for details: https://github.com/hashcat/oclHashcat/issues/20

jsteube:
Just to make this complete: Both you as the maintainer and the package endusers do _not_ need any of those SDK:
- CUDA SDK 
- AMD APP SDK

Those actually create more problems than they do good. All you both need is the NVidia ForceWare driver or the AMD Catalyst driver. They include all the libraries oclHashcat need to run.

Only you as the maintainer who builds a native binary needs the following additional SDK:
- ADL SDK
- GDK

[...]

Issue #63, opened 2016-01-07
OpenCL/OSX support


lloeki:
Is this reasonable to ask? May be tied to https://github.com/hashcat/oclHashcat/issues/3

Reference: https://developer.apple.com/opencl/

jsteube:
Sure, why not. I just hope someone can provide patches for it since I do not have an apple computer so I can't code it.

magnumripper:
I'm extremely improductive at the moment but I'll have a look sooner or later unless someone beats me to it. The recent OpenCL changes have paved the way, should be very easy now.

Right now though, Apple's OpenCL drivers are very crappy. They were perfectly fine with JtR prior to El Capitan but now we see segfaults, incorrect results and whatnot.

lloeki:
> They were perfectly fine with JtR prior to El Capitan but now we see segfaults, incorrect results and whatnot.

I admit I didn't write an OpenCL program myself since Mavericks, but I recently ran JtR on some Raw-SHA512-opencl task under El Cap. It was on an Intel card though, so maybe there's a flaky Nvidia or AMD driver down the line.

magnumripper:
Sure, many combinations of formats/devices does work fine, but far from all of them. There was a major regression for El Capitan.

lloeki:
I was merely providing a data point :-) I'll keep this in mind anyway as I may need OpenCL for other things, thanks.

lloeki:
That's more work but maybe we can work around OpenCL: it seems Metal can be used for [GPGPU computations](http://memkite.com/blog/2014/12/15/data-parallel-programming-with-metal-and-swift-for-iphoneipad-gpu/) but I don't know how convoluted it would be to call it from C.

matrix:
Beta is out, please test it ;)

https://hashcat.net/beta/oclHashcat-2.10b2-osx.7z

lloeki:
Thanks @gm4tr1x. Cursory test on an Intel Iris (3GHz Core i7, Mid 2014 MBP 13"):

```
$ ./example0.sh
oclHashcat v2.10 starting...

Device #1: skipped                      
Device #2: Iris, 384/1536 MB allocatable, 1200Mhz, 40MCU

Hashes: 6494 hashes; 6494 unique digests, 1 unique salts
Bitmaps: 16 bits, 65536 entries, 0x0000ffff mask, 262144 bytes, 5/13 rotates
Applicable Optimizers:
* Zero-Byte
* Precompute-Init
[...]

PR #144, opened 2016-01-20
About issue #120 and issue #63

matrix:
Added support for build without ADL/NVML/NVAPI (issue #120).
Added support for build OSX native binaries (issue #63).

I do not think it is useful split into 2 PR because each one depends on the other.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ oclHashcat
 *.dictstat
 *.pot
 *.log
+*.dSYM
 deps/**
 kernels/**
 lib/*.a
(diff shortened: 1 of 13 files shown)


matrix:
Thanks bro ;)

magnumripper:
Good stuff! Works fine on my macbook.

matrix:
Thanks for the test :)

philsmd:
Can be merged!

(BTW: it would be great if you @gm4tr1x update #63 (I guess we shouldn't close it yet) and mention the progress, what is currently not working yet e.g. --benchmark, for which algorithm the tests currently fail and also most importantly why you had to change some parts in this particular manner, e.g. why there is no --cpu-affinity etc. I mean, this are very critical changes and imply that we can't tell the users @ forum/irc that they can "just use --cpu-affinity" to just mention an example. We need some clues/documentation why it was changed like this and what is still missing).
Thx

PR #151, opened 2016-01-24
Avoid sprintf

matrix:
Avoided sprintf usage


philsmd:
seems good, can be merged

PR #152, opened 2016-01-24
Optimized memset calls and also some initializations

matrix:
Optimized memset calls and also some initializations

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -10951,7 +10951,7 @@ int main (int argc, char **argv)
             {
               user_t **user = &hashes_buf[hashes_cnt].hash_info->user;
 
-              *user = (user_t *) malloc (sizeof (user_t));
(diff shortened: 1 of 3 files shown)


philsmd:
can be merged, thx

(Btw: some of these fixes only work because the mymalloc () function already calls memset () , these changes are just cosmetic, they shouldn't change any logic, performance etc... the compiler will always optimize these things, especially with the -O2 compiler option... so there should be no noticeable changes besides the optical ones for devs/contributors)

philsmd:
@gm4tr1x while you are trying to change this mymalloc () calls, let me suggest this fix too: here https://github.com/hashcat/oclHashcat/blob/master/src/oclHashcat.c#L10957 (*user = (user_t *) malloc (sizeof (user_t));)  we have a call to just "malloc ()", we should probably call mymalloc here too... Can you fix that with a new pull request?

PR #156, opened 2016-01-24
Updated test.sh (support for osx and some fixes)

matrix:


philsmd:
I don't understand what this --gpu-temp-disable change is trying to accomplish. The commit messages says something about OSX ("Re-enable gpu-temp-disable argument for osx build") but I don't see any OSX related changes (with #ifdef OSX or #ifndef OSX).
When HAVE_HWMON is true, --gpu-temp-disable probably should be available, otherwise it should not be in the --help and not be available in general. I can't understand what this has to do with OSX.
Maybe the problem is just that I only see "No description provided." here and can't really figure out why it was changed the way it was... but without having more details and documentation of the reasons and goals of this change I can't understand and sign it.

matrix:
In my previous patch i removed this option for builds without HAVE_HWMON,
with that i re-enable (test.sh use this option by default).

matrix:
You can find the ifdef here for example
https://github.com/hashcat/oclHashcat/blob/master/src/oclHashcat.c#L5634

philsmd:
this doesn't seem like a good solution for me, i.e. enabling the option just that the test.sh script runs without error and otherwise it shouldn't be used or if it is used it will be ignored.
It would make much more sense to either check within tools/test.sh if the option is available and only use it when it is available or, we need to discuss this, disable the option in test.sh completely.

Changing it this way which is completely unlogical and which would suggest that --gpu-temp-disable would also be available for the OSX users and users that have HAVE_HWMON set to 0.

In general, it is much more important that the oclHashcat --help output and the options that the user can use are correct, than the tools/test.sh options (which only some devs will run).  Of course, both should be working and hence I suggest that we add --gpu-temp-disable in test.sh if and only if the binary allows that.

matrix:
For now OSX don't use hwmon but in near future can so i don't understand what's your goal.
My is find a way to support all options, also for OSX :)

philsmd:
All I'm trying to say is that this makes no sense:

```
"       --gpu-temp-disable            Disable temperature and fanspeed readings and triggers",
#ifdef HAVE_HWMON
...
```

it should stay as it was:

```
#ifdef HAVE_HWMON
"       --gpu-temp-disable            Disable temperature and fanspeed readings and triggers",
...
```

why should --gpu-temp-disable be available when hardware monitoring was disabled?
If hardware monitoring was disabled, also --gpu-temp-disable should not be available (like it was before this strange changes).

matrix:
For now hwmon for OSX is disabled,
also a lot of algo for now are disabled on OSX due to uncommited change to kernel code.
Do you want remove all code related because today is not ready/supported ?

jsteube:
I also don't see what sense it makes to have --gpu-temp-disable if we have HAVE_HWMON. It may confuse the user because being able to disable something implies that it's turned on by default, which it is not since it's not supported.

I think what's the real reason behind this is that we want to use --gpu-temp-disable in test.sh but if we would run test.sh on OSX oclHashcat would stop instantly because it would report --gpu-temp-disable as an unknown parameter. From that view it might make sense to add it back.

philsmd:
> From that view it might make sense to add it back.

No. No. No.

It makes absolutely no sense to break the oclHashcat binary just that 2-3 devs can run tools/test.sh (which of course not every oclHashcat user uses 24/7) on OSX/without HWMON too. I don't support this idea to be that lazy and to not **fix test.sh** (instead) to use --gpu-temp-disable only when it is available... but instead break the binary with some very confusing changes that make no sense at all.

matrix:
The "unknown parameter" is the key, also because OSX support is under development and will be released if finally work else not.
For me is the same, tell me if re-enable --gpu-temp-disable globally or push another mod in test.sh

@philsmd, i don't see any break

jsteube:
OK, that's correct. It's only a few users using tools/test.sh, even less on osx, and those could remove the flag from test.sh manually. Now I understand what @philsmd means with lazy and I agree. Please remove that one patch about --gpu-temp-disable but send in a new one for the other changes from this PR

philsmd:
I still don't understand why @gm4tr1x  says HAVE_HWMON is directly related to OSX builds. The fact ist that users/maintainers can set HAVE_HWMON to 0 also on non-OSX builds. Therefore this is **not** a valid fix

```
       elif [ ${OPTARG} == "osx" ]; then
+        OPTS=$(echo $OPTS | sed -e 's/--gpu-temp-disable //g')
```

The correct fix would check for the --gpu-temp-disable switch in the --help output for each platform (linux/windows/osx), as suggest previously here https://github.com/hashcat/oclHashcat/pull/156#issuecomment-174543275

matrix:
@philsmd when I add HAVE_HWMON i to for osx, not for fix the issue.

I think you don't see what do in the code, before this PR 156 --gpu-temp-disable for OSX is disabled,
i write in the comment "Re-enable" for that. If the initial commit for oclHashcat.c is correct why you write a lot of comment about that?

"The correct fix would check for the --gpu-temp-disable switch in the --help output for each platform (linux/windows/osx)"
The correct fix is the fix I do before "Keep disable --gpu-temp-disable on osx in oclHashcat.c", the same that you want :)

matrix:
Now I remove the last patch, tell me what is wrong now

jsteube:
I think the problem is this one:

https://github.com/gm4tr1x/oclHashcat/commit/479327fc20b1678d8d25c9cd6f422bf685fef769

matrix:
I re-check the patch, if you see in master --gpu-temp-disable is not present if HAVE_HWMON is not defined. With that patch we can restore the parameter.
OSX hwmon support is also under development ;)

PR #157, opened 2016-01-24
Improved osx support with custom kernel loops and gpu accel (issue #63)

matrix:
Added custom OSX kernel loops and gpu accel default values
Disabled some not yet worked hash types in OSX


jsteube:
What happens to OSX when using the default values?

matrix:
Trap 6

jsteube:
I'm mostly OK with this PR but there's a few things I don't like. 

Please fix that first:
- KERNEL_LOOPS_\* and KERNEL_ACCEL_\* macro names should stay as they are. Please use #ifdef and create a full set of all algorithms for OSX for both accel and loops
- You can do the same for the benchmark list. Don't forget to update NUM_DEFAULT_BENCHMARK_ALGORITHMS if you remove some algorithms
- Do not use something like _TUNED

Issue #84, opened 2016-01-08
--keyspace  should not require valid OpenCL runtime environment


philsmd:
This feature request was originally reported on the (now obsolete and offline) **trac** ticketing system of hashcat.net.

---

| Ticket details:                    |               |
| -------------------------------- | ---------- |
| Original reporter (OP):      | epixoip |
| Title:                                  | `keyspace' flag should not require valid OpenCL runtime environment |
| Ticket number:                  | 562 |
| Date reported:                  | 2015-01-26T20:39:39+01:00 |
| **Description**:                 | I would like to propose that when using [the] \`keyspace' flag, oclHashcat should simply calculate & print the keyspace, then exit. No OpenCL checks or device enumeration should be performed, as OpenCL runtime & devices are not required for this functionality.<br><br>The purpose for this request is three-fold:<br>- Reduce the overall runtime of oclHashcat when --keyspace is used. This helps reduce overhead when calling oclHaschat programmatically in a distributed environment. Presently, running `oclHashcat --keyspace' takes about 3 seconds on a system with eight GPUs. Ideally this operation should complete in milliseconds.<br>- Enable the keyspace flag to be used on systems without GPUs or an OpenCL runtime environment.<br>- Eliminate the need for the standalone keyspace binary in hashcat-utils, which is now out of date due to the refactoring in 1.32  |

---

| Comments:               |                |               |
| -------------------------- | ----------- | ---------- |
| 1.                               | atom       | That might be possible, but it's not so easy. For example, for oclHashcat you will need at least the libOpenCL.so/.dll with correct exports otherwise it won't start. You may can replace it with a dummy library.<br><br>However, the standalone keyspace binary in hashcat-utils should be in sync with the current version. If not, we better patch it. Do you have an example [to] reproduce [it]?  |

jsteube:
Done with latest beta version

PR #162, opened 2016-01-25
OpenCL runtime environment is no longer required with --keyspace option

matrix:
Solved issue #84

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -46,6 +46,8 @@
 #include <mach-o/dyld.h>
 #endif
 
+typedef void *OCL_LIB;
+
 #ifdef HAVE_HWMON
 typedef void *HM_LIB;
 #endif
@@ -77,6 +79,8 @@ typedef UINT64 uint64_t;
 #ifdef HAVE_HWMON
 typedef HINSTANCE HM_LIB;
 #endif
+
+typedef HINSTANCE OCL_LIB;
 
 #define mkdir(name,mode) mkdir (name)
 
--- a/include/ext_OpenCL.h
+++ b/include/ext_OpenCL.h
@@ -1,5 +1,7 @@
 /**
- * Author......: Jens Steube <jens.steube@gmail.com>
+ * Authors.....: Jens Steube <jens.steube@gmail.com>
+ *               Gabriele Gristina <matrix@hashcat.net>
+ *
  * License.....: MIT
  */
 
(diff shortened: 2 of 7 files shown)


philsmd:
It doesn't look like it works for me:
strace ./oclHashcat --keyspace -a 3 ?a?a?a?a?a?a?a?a 2>&1 | egrep 'libamd|libOpenCL'

As far as I understood, there should be no requirement for loading those (and maybe other) libs when running with --keypace option. Your fix doesn't change this.

matrix:
Check here

http://pastebin.com/pZNagJ1P

philsmd:
If I try without the libraries I get this error:
./oclHashcat --keyspace -a 3 ?a?a?a?a?a?a?a?a
./oclHashcat: error while loading shared libraries: libOpenCL.so.1: cannot open shared object file: No such file or directory

This is what the issue #84 is trying to fix (in my opinion), i.e. using --keyspace also when the libraries are not present

matrix:
You can't never build without OpenCL library but now you can use --keyspace switch without have a compatible hardware.
I think is the max we can do without write another standalone tool for calculate only the keyspace.

philsmd:
the tool for calculating the --keyspace already exist in hashcat-utils: https://github.com/hashcat/hashcat-utils/blob/master/src/keyspace.c
But again the goal that issue #84 should solve is to allow the user to run --keyspace without loading those libraries that are not needed to show the --keyspace value.

matrix:
I read that in original issue description:

"I would like to propose that when using [the] `keyspace' flag, oclHashcat should simply calculate & print the keyspace, then exit. No OpenCL checks or device enumeration should be performed, as OpenCL runtime & devices are not required for this functionality."

If you don't like that talk also with others team member about closing this pull request and also issue 84.

jsteube:
I think what we really need to make it "perfect" is a dlopen() loader for OpenCL as we already do for ADL and NVML. That would be great!

magnumripper:
That is very tedious to achieve but obviously has some advantages. Hashkill does that. I had a look (long ago) at nicking that code but opted not to.

matrix:
Very tedious but i like @jsteube idea.
Work in progress

matrix:
@jsteube look this patch ;)

jsteube:
WOW! Good work, you must have been busy. I'll merge it but maybe rename some variable names.

jsteube:
Don't forget to create a PR for it

PR #168, opened 2016-01-27
Do not create a kernel cache file if build failed (actually if size ended up as zero).

### LINKED ISSUE #164: NTLM kernel empty file
So I noticed that today after I compiled 0102aef I get the following error when cracking NTLM hashes (not present with other algorithms) upon further investigation I see that for what ever reason hashcat is not populating the ~/.hashca/kernels folder (going further you may want to change the directory name to hashcat) with the m01000_a0.ef202b9a.kernel, the result being when I try and crack NTLM hashes I get the following error:

```
Device #1: Kernel /home/user/.hashca/kernels/m01000_a0.ef202b9a.kernel (0 bytes)


ERROR: clCreateProgramWithBinary() : -30 : CL_INVALID_VALUE
```

when I cat  /home/user/.hashca/kernels/m01000_a0.ef202b9a.kernel   I see it is indeed a empty

magnumripper:
Also disregard any existing cache files with size of zero. Should close #164.

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -13547,7 +13547,7 @@ int main (int argc, char **argv)
 
         struct stat cst;
 
-        if (stat (cached_file, &cst) == -1)
+        if ((stat (cached_file, &cst) == -1) || cst.st_size == 0)
         {
           cached = 0;
         }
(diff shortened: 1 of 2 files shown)


jsteube:
Cool, many Thanks! I'll close #164

PR #169, opened 2016-01-27
Drop dependencies on non-distributable ADL/NVML headers.

### LINKED ISSUE #120: support building without ADL (and without CUDA)

magnumripper:
The needed glue is copied into our respective local headers. Should close #120.

Please review the changes to Makefile and deps.sh, not sure they are good as-is. I left some stuff in there, but commented out, in case someone temporarily wants to use the real headers during devel.

Tested on Linux 64-bit and OSX. The statically linked NV stuff for Windows was left as-is.

### DIFF
--- a/docs/BUILD.md
+++ b/docs/BUILD.md
@@ -18,11 +18,9 @@ To compile **oclHashcat** the following third party libraries are required:
 The next thing to do is download all the third party libraries listed above and put these files into the *deps/tmp* directory.
 
 The following files are needed inside the *deps/tmp* directory:
-    
-    ADL_SDK9.zip
-    R352-developer.zip
-    gdk_linux_amd64_352_55_release.run
-    
+
+    R352-developer.zip  (only needed when cross-compiling for Windows)
+
 # Building oclHashcat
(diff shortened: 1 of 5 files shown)


jsteube:
We should also use those headers for Windows so that we can totally get rid of the #ifneq stuff in Makefile. Can you do that?

magnumripper:
We need that test for nvapi.h anyway, because it also (sort of) tells us that nvapi.lib (or nvapi64.lib) is there. Until we dynamically load that lib, there's no point in changing that stuff.

PR #177, opened 2016-01-29
Acquire an exclusive lock before writing to any file.

### LINKED ISSUE #172: Consider using file locking
Some time ago I had concurrent sessions of oclHashcat completely thrash a potfile because they were both using the same one - and both assumed they had exclusive use to it. Can't remember the details but I probably used the --potfile option pointing to one same file assuming file locking was in place. Perhaps that was stupid but nothing stopped me from doing it and no docs made me any wiser.

Now that I get to see the code, I see you are using O_APPEND (actually fopen "a") but no locking. This is a decent first step but not safe enough.

This is simple to fix though. Any write to the pot file should be wrapped with a blocking write lock. I'm willing to implement this and hand you a PR unless you have some opinion against it for whatever reason?

magnumripper:
Note that in some cases we never explicitly unlock a file because fclose will do it implicitly. Closes #172.

Locking code will be silently skipped for Windows builds.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -28,6 +28,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <search.h>
+#include <fcntl.h>
 
 #ifdef _POSIX
 #include <sys/time.h>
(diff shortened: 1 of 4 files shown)


magnumripper:
Build verified on Linux, Windows and OSX. Actual functionality is verified on Linux. If the file is already locked by another process, we'll simply wait until we get the lock.

jsteube:
Excellent, thanks!

PR #192, opened 2016-01-31
Update ADL/OpenCL Library Loader

matrix:
Rewrite ADL loader like OpenCL loader.

If someone with OSX and AMD Gpu could test if ADL work is great :)


jsteube:
Great work!

PR #197, opened 2016-02-03
Get rid of nvidia NvAPI static linking

matrix:
From [nvidia nvapi concepts](http://docs.nvidia.com/gameworks/content/gameworkslibrary/coresdk/nvapi/concepts.html)

> Use a Static Link with Applications
> 
> NvAPI cannot be dynamically linked to applications. You must create a static link to the library and then call NvAPI_Initialize(), which loads nvapi.dll dynamically. 
> If the NVIDIA drivers are not installed on the system or nvapi.dll is not present when the application calls NvAPI_Initialize(), the call just returns an error. The application will still load. 

We can do it, and it seems to work.

nvapi.lib is basically a wrapper for nvapi.dll
Some of nvapi functions are not exported but addresses ... are online :)
By using the exported function "nvapi_QueryInterface" we can call internal dll functions => game over men ;)

No more NvAPI static linking.
The nvapi.dll or nvapi64.dll are shipped with the nvidia driver,
it is enough for what interests us.

Time to new beta ;)

~(=^‥^)ﾉ◎～

### DIFF
--- a/include/ext_nvapi.h
+++ b/include/ext_nvapi.h
@@ -295,6 +295,7 @@ typedef struct
 // Macro for constructing the version field of NV_GPU_DYNAMIC_PSTATES_INFO_EX
 #define NV_GPU_DYNAMIC_PSTATES_INFO_EX_VER MAKE_NVAPI_VERSION(NV_GPU_DYNAMIC_PSTATES_INFO_EX,1)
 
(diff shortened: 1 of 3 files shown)


Xanadrel:
Ok for merging.

PR #200, opened 2016-02-04
Bugfix: putenv(3) was used with a stack variable, that's a no-no.

magnumripper:
From my man page:  _The string pointed to by string becomes part of the environment.  A program should not alter or free the string, and should not use stack or other transient string variables as arguments to putenv().  The setenv() function is strongly preferred to putenv()._

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -5072,7 +5072,7 @@ int main (int argc, char **argv)
 
   if (compute)
   {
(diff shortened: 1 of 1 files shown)


magnumripper:
New, uglier but MinGW-compatible, version using putenv and a static.

jsteube:
Thanks! And I agree, I wish we could switch to setenv()

Issue #191, opened 2016-01-31
Display driver crashes using Intel HD Graphics 4400


Fist0urs:
Using oclhashcat64.exe on a Windows 8.1 Professionnal and latest Intel drivers with OpenCL gpu runtime:

```
oclHashcat64.exe -b
```

display driver crashes. Looking at the eventlog, I get:

```
Display driver igfx stopped responding and has successfully recovered.
EventId: 4101
```

I know it will be difficult to debug with such few details, so if I can add some more details, do not hesitate to ask

epixoip:
We probably need a timeout patch for Intel like we do with Nvidia.

Zyntax3rror:
That patch changes registry TdrLevel from 3 to 0, so it should work on any card(did work on my hd4000). (If your computer hangs after the patch (after you have started oclHashcat64) then you have to use the reset button/power-button)

TdrLevel
0 - Detection disabled
1 - Bug check on detected timeout, for example, no recovery.
2 - Recover to VGA (not implemented).
3 - Recover on timeout. This is the default value.

there is plenty of guides (just search for TdrLevel). This is one way of doing it:
patch (create a file called patch.reg (not patch.reg.txt) and paste the following. Then save and double click the file(and answer yes  and then yes again) then reboot your computer)
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\GraphicsDrivers]
"TdrLevel"=dword:00000000

to go back to default:
(default.reg   copy/paste + save + double click +yes+yes+reboot)
Windows Registry Editor Version 5.00

[HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\GraphicsDrivers]
"TdrLevel"=dword:00000003

epixoip:
Yes, this is already covered on our wiki: http://hashcat.net/wiki/doku.php?id=timeout_patch

jsteube:
@Fist0urs can you please try again with the registry patch and tell us if it worked?

Fist0urs:
I tried and my screen freezes. I tried supplying just a single hash of known plain to see if it actually worked but it doesn't

magnumripper:
Does it crash on MD4 already or what format is causing the hang? My HD4000 works just fine until an `Abort trap: 6` in SHA-3(Keccak) (and that may just as well be an Apple driver bug).

matrix:
[...]

PR #210, opened 2016-02-06
Fix SHA-3(Keccak) issue with HD4000 (issue #191) and reorder macro

matrix:
@magnumripper could you test it with your HD4000 if there're other "trap 6" error ?

Thank you,
matrix


magnumripper:
I can now complete a benchmark of all formats with the HD4000, but it takes 45 minutes. And Bitcoin/Litecoin reports 0 H/s.

matrix:
@magnumripper, great news :+1: 

Thank you for testing

Issue #150, opened 2016-01-23
Scrypt bug?


Zyntax3rror:
Using a GTX 750 Ti (no overclocking);
a hash created with parameters SCRYPT:512:16:1
oclHashcat(2.10beta) fails finding the correct password (is in dictionary), 
hashcat (v2) does find the password (same hash used) 

(only 3 words was used in my test dictionary, one of them is the correct password)

Zyntax3rror:
#2
Could be related to the above problem
Forcing the CPU to do the work fails too(used version 2.10b64)
(I used a litecoin hash for the CPU (but this time I used my other computer))
(hashcat is one of the words in the dictionary)

oclHashcat64.exe -m 8900 test1024c.hash example.dict --opencl-device-types 1

Device #1:        Intel(R) Core(TM) i7-3770S CPU @ 3.10GHz, 8113/32452 MB alloca
table, 3100Mhz, 8MCU
Device #2: Intel(R) HD Graphics 4000, skipped

Session.Name...: oclHashcat
Status.........: Exhausted
Input.Mode.....: File (example.dict)
Hash.Target....: SCRYPT:1024:1:1:MDIwMzMwNTQwNDQyNQ==:5FW+...
Hash.Type......: scrypt
Time.Started...: 0 secs
Time.Estimated.: 0 secs
Speed.Dev.#1...:        0 H/s
Recovered......: 0/1 (0.00%) Digests, 0/1 (0.00%) Salts
Progress.......: 3/3 (100.00%)
Rejected.......: 0/3 (0.00%)
HWMon.GPU.#1...: N/A Util, N/A Temp, N/A Fan

Started: Sun Jan 24 14:11:51 2016
Stopped: Sun Jan 24 14:11:54 2016

Zyntax3rror:
Done some further analysis (hash is hascat)

problem is that when r > 8 it fails!

OK
SCRYPT:64:1:1:dGVhbWhhc2hjYXQ=====:SDYBLsXIefUhSOacrgoZE2vEdduSyl+2uU+OTD8eUD8=
SCRYPT:64:2:1:dGVhbWhhc2hjYXQ=====:5b9cH8RssHn8NQ4fA+40seHscB18Smy1SCgkkUDob0E=
SCRYPT:64:3:1:dGVhbWhhc2hjYXQ=====:upfDMzSsnbPyA2wlogTBrIQPJe22ZJaDShPe+yyCt9I=
[...]

PR #211, opened 2016-02-06
Fixed "VENDOR_ID" format (probably relate with issue #150)

matrix:
The bug is

`-DVENDOR_ID=%d`

if vendor_id (unsigned int) is bigger than INT_MAX, overflow.

Example:

`-DVENDOR_ID=-1 -DCUDA_ARCH=0 -DVECT_SIZE=4 -DDEVICE_TYPE=2
`

Now is

`-DVENDOR_ID=4294967295 -DCUDA_ARCH=0 -DVECT_SIZE=4 -DDEVICE_TYPE=2`

I think issue #150 is also fixed

(＾_･.･_＾)

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -13227,12 +13227,12 @@ int main (int argc, char **argv)
      */
 
     #ifdef HAVE_HWMON
-    int *temp_retain_fanspeed_value = (int *) mycalloc (devices_cnt, sizeof (int));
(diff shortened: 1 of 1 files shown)


jsteube:
I'm wondering if 4294967295 is really the VENDOR_Id. I'd say it's probably a -1 return code and they want to say something like "Parameter unknown/not supported".

matrix:
All the clinfo output show 4294967295 as vendor id for Apple cpu, [like here](https://groups.google.com/forum/#!topic/aparapi-discuss/ryA8IvjI0AU)

PR #214, opened 2016-02-09
Use per-device timer resolution in total_time calculations

matrix:
Reference: [Intel OpenCL Performance Debugging Intro](https://software.intel.com/en-us/articles/performance-debugging-intro/)

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -797,7 +797,7 @@ void status_display_automat ()
     hc_clGetEventProfilingInfo (data.ocl, device_param->event, CL_PROFILING_COMMAND_START, sizeof (time_start), &time_start, NULL);
     hc_clGetEventProfilingInfo (data.ocl, device_param->event, CL_PROFILING_COMMAND_END,   sizeof (time_end),   &time_end,   NULL);
 
-    const double total_time = (time_end - time_start) / 1000000.0;
+    const double total_time = ((double)(time_end - time_start) * (double)(device_param->device_timer_resolution)) / 1000000.0;
 
     fprintf (out, "%f\t", total_time);
   }
@@ -12578,6 +12578,14 @@ int main (int argc, char **argv)
         hc_clGetDeviceInfo (data.ocl, device_param->device, CL_DEVICE_MAX_CLOCK_FREQUENCY, sizeof (device_maxclock_frequency), &device_maxclock_frequency, NULL);
 
         device_param->device_maxclock_frequency = device_maxclock_frequency;
+
+        // device_timer_resolution
+
+        cl_ulong device_timer_resolution;
+
+        hc_clGetDeviceInfo (data.ocl, device_param->device, CL_DEVICE_PROFILING_TIMER_RESOLUTION, sizeof (device_timer_resolution), &device_timer_resolution, NULL);
+
+        device_param->device_timer_resolution = device_timer_resolution;
 
         // skipped
 
--- a/include/types.h
+++ b/include/types.h
@@ -847,6 +847,7 @@ struct __hc_device_param
   u64     device_maxmem_alloc;
   u64     device_global_mem;
   u32     device_maxclock_frequency;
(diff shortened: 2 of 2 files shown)


src/oclHashcat.c
```diff
@@ -797,7 +797,7 @@ void status_display_automat ()
     hc_clGetEventProfilingInfo (data.ocl, device_param->event, CL_PROFILING_COMMAND_START, sizeof (time_start), &time_start, NULL);
     hc_clGetEventProfilingInfo (data.ocl, device_param->event, CL_PROFILING_COMMAND_END,   sizeof (time_end),   &time_end,   NULL);
 
-    const double total_time = (time_end - time_start) / 1000000.0;
+    const double total_time = ((double)(time_end - time_start) * (double)(device_param->device_timer_resolution)) / 1000000.0;
 
     fprintf (out, "%f\t", total_time);
```
magnumripper, line 802: Sorry if I'm totally confused here, but is this really correct? I think it isn't.

Before this patch:
- If a device with 1 ns resolution gave a `time_end - time_start` resulting in 8000, we know it took 8 microseconds +/- one nanosecond.
- If a device with 80 ns resolution gave `time_end - time_start` resulting in 8000, we know it took 8 microseconds +/- 80 nanoseconds.

But after this patch, the latter will show up like... 8000*80 == 640000 meaning 640 microseconds? That's just not right.

matrix:
@magnumripper everything is possible, look the reference link and propose a change :)
Bye,
matrix

src/oclHashcat.c
magnumripper, line 802: Revisiting eg. https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/clGetEventProfilingInfo.html I see it's a bit ambiguous but as far as I'm aware it's like I said above: The result is _already_ nanoseconds and the _resolution is only needed to know the accuracy_.

http://stackoverflow.com/questions/34276371/timer-resolution-in-opencl-profiling

matrix:
@magnumripper from your link:

> OpenCL devices are required to correctly track time across changes in device frequency and power states. The CL_DEVICE_PROFILING_TIMER_RESOLUTION specifies the resolution of the timer i.e. the number of nanoseconds elapsed before the timer is incremented.

magnumripper:
> @magnumripper everything is possible, look the reference link and propose a change :)

My proposition is to revert this commit. It was correct before, it's busted now.

magnumripper:
> OpenCL devices are required to correctly track time across changes in device frequency and power states. The CL_DEVICE_PROFILING_TIMER_RESOLUTION specifies the resolution of the timer i.e. the number of nanoseconds elapsed before the timer is incremented.

You read it as "after 80 ns, counter will increase by 1". I am pretty sure the correct interpretetion is "after 80 ns, counter will increase by 80".

A timer resolution of 80 merely says you can't measure smaller things - like 40 ns. It will go from 0 to 80 to 160.

matrix:
You're probably right, but what is the correct usage of CL_DEVICE_PROFILING_TIMER_RESOLUTION in profiling?

magnumripper:
We don't need it at all for our purposes. It's only there so you know what accuracy you are dealing with. My Macbook's GT650M has a resolution as bad as 1000 ns, so I simply can't measure anything smaller than a microsecond. My CPU device has the full 1 ns resolution.

matrix:
Now I agree :)
For the current usage it's not useful.
Make happy @jsteube and do a new PR when you have time please :)

magnumripper:
@gm4tr1x you should have a "revert" button right here below. It will create a PR for revertion of this one.

magnumripper:
And don't be sad, you made like a hundred **damn good** commits before this one ;-)

matrix:
Why sad ? :)
I'm happy, I just learned something new, one in "OpenCL" and the revert function of github, first time use :)

Ty,
matrix

PR #217, opened 2016-02-10
Use device's preferred vector width, not its native one.

magnumripper:
For example, the Intel MIC announces a native width of 16 but a preferred width of 1. That's because it will auto-vectorize. If you give it a vectorized kernel anyway, the driver will unconditionally auto-**de**-vectorize it and then auto-vectorize the result of _that_. Yes it's very stupid but that is what will happen. Build time will increase and the result will likely end up worse than if you give it scalar code (which it's politely asking for). Plain CPU drivers are the same. I've done a lot of tests with this.

Also, don't assume vector width for 'long' is half of that for 'int'. Usually it is, but there's no reason to make assumptions when we can get it with a proper query.

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -12478,11 +12478,11 @@ int main (int argc, char **argv)
         {
           if (tuningdb_entry->vector_width == -1)
           {
-            hc_clGetDeviceInfo (data.ocl, device_param->device, CL_DEVICE_NATIVE_VECTOR_WIDTH_INT, sizeof (vector_width), &vector_width, NULL);
-
             if (opti_type & OPTI_TYPE_USES_BITS_64)
             {
-              if (vector_width > 1) vector_width /= 2;
+              hc_clGetDeviceInfo (data.ocl, device_param->device, CL_DEVICE_PREFERRED_VECTOR_WIDTH_LONG, sizeof (vector_width), &vector_width, NULL);
+            } else {
+              hc_clGetDeviceInfo (data.ocl, device_param->device, CL_DEVICE_PREFERRED_VECTOR_WIDTH_INT, sizeof (vector_width), &vector_width, NULL);
             }
           }
           else


jsteube:
It seems we have a bigger problem related to this PR. All Intel CPU using Intels OpenCL runtime now return 1 as their preferred vector width. The only way of selecting the best vector width is to use the native one. Any Ideas?

cc @philsmd

epixoip:
Maybe if this has changed for CPU then it's also changed for MIC? If not, can we put something in the .hctab to fix this?

magnumripper:
Like I said, if they return 1 as preferred width, they are virtually always performing best if you honor that. And like I said, they will auto-vectorize anyway.

This is the whole reason for the "preferred" query to exist.

magnumripper:
Here's JtR with an AMD CPU driver, which has preferred width of 4 and is given that (note the `4x` in the algorithm name within brackets):

```
$ ../run/john -test -form:wpapsk-opencl -dev=3
Device 3: Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
Benchmarking: wpapsk-opencl, WPA/WPA2 PSK [PBKDF2-SHA1 OpenCL 4x]... DONE
Raw:    21557 c/s real, 689 c/s virtual
```

Here's an Intel driver instead. It has preferred width of 1 so is given scalar code:

```
$ ../run/john -test -form:wpapsk-opencl -dev=4
Device 4: Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
Benchmarking: wpapsk-opencl, WPA/WPA2 PSK [PBKDF2-SHA1 OpenCL]... Build log: Compilation started
Compilation done
Linking started
Linking done
Kernel <wpapsk_init> was successfully vectorized
Kernel <wpapsk_loop> was successfully vectorized
Kernel <wpapsk_pass2> was successfully vectorized
Kernel <wpapsk_final_md5> was successfully vectorized
Kernel <wpapsk_final_sha1> was successfully vectorized
Done.
DONE
Raw:    21557 c/s real, 672 c/s virtual
```

Curiously enough they end up exactly the same speed.

Here's the result if disobeying the preferred vector size and using the native 4x anyway:

```
$ ../run/john -test -form:wpapsk-opencl -dev=4 --force-vector=4
Device 4: Intel(R) Xeon(R) CPU E5-2670 0 @ 2.60GHz
Benchmarking: wpapsk-opencl, WPA/WPA2 PSK [PBKDF2-SHA1 OpenCL 4x]... Build log: Compilation started
Compilation done
Linking started
Linking done
Kernel <wpapsk_init> was not vectorized
Kernel <wpapsk_loop> was not vectorized
Kernel <wpapsk_pass2> was not vectorized
Kernel <wpapsk_final_md5> was not vectorized
Kernel <wpapsk_final_sha1> was not vectorized
Done.
DONE
Raw:    21277 c/s real, 684 c/s virtual
```

jsteube:
I think for hashcat that doesn't work, as we don't do any manual interlacing. OTOH, the PHI performance using CL_DEVICE_NATIVE_VECTOR_WIDTH_INT was pretty good. I'll switch back

magnumripper:
JtR was just an example, it's very basic code with no interleaving (we do that only in non-OpenCL CPU code).

The thing I wrote above about "auto-de-vectorizing and then auto-vectorizing" was not something I made up, it's what Intel says will happen when you pass a vectorized kernel even though the device tells you not to. End result: longer build time, worse performance. In some odd cases the performance may end up slightly better by "coincidence" but that should be rare. I'm pretty sure I never saw it happen.

Anyway, the best arbitrator is empirical data from testing oclHashcat of course. Is there a way to force scalar, or a certain width, from command line?

epixoip:
Yes, this can be overridden on the command line with --opencl-vector-width and also in the new .hctab file.

magnumripper:
No matter what width I use or what accel/loops I try, oclHashcat is only doing 25% of JtR speed for WPA-PSK (eg. 5K vs. 21K) on same Intel CPU as the benchmarks mentioned above. This must be some unrelated issue. @jsteube that's the gear Solar gave you an account on btw.

jsteube:
WPA (-m 2500) is not converted to the new simd code yet. Actually none of the slow hashes is converted.  You can find out by looking on the kernel header for NEW_SIMD_CODE. I'll do that later. In the meanwhile you can test with all the -a 3 kernel, ideally -m 900. This one is converted and should beat JtR's speed on all device type.

magnumripper:
OK that explains it. I will hold back my testing until we've got a really slow format supporting vector code, otherwise we'll just see interleaving effects that make the results very hard to draw conclusions from.

Meanwhile I did find a new-ish whitepaper from Intel that contradicts things I've read and heard before. Perhaps the CPU drivers (as opposed to MIC/Phi ones) will auto-vectorize _only if given scalar code_ and use already vectorized code as-is. I always thought that would be a better way to go even for the MIC. I really hate it when I give them pre-vectorized code and their compiler just ruins it. Anyway I can't imagine why the heck they would announce 1 if 4 or 8 is better.

PR #223, opened 2016-02-12
Get rid of hc_clEnqueueNDRangeKernel double call

matrix:
With Apple CPU if local_work_size if wrong _clEnqueueNDRangeKernel_ fail with _CL_INVALID_WORK_GROUP_SIZE_

I found a way for check the value before call _clEnqueueNDRangeKernel_

```
size_t workgroup_size = 0;
hc_clGetKernelWorkGroupInfo (data.ocl, kernel, device_param->device, CL_KERNEL_WORK_GROUP_SIZE, sizeof(size_t), &workgroup_size, NULL);
if (kernel_threads > workgroup_size) kernel_threads = workgroup_size;
```

Bye,
matrix


magnumripper:
This is good measure for generic device support. Perhaps it should be queried right after building kernel and cached, instead of querying for every call.

You should probably use it for the case of `if ((data.opts_type & OPTS_TYPE_PT_BITSLICE) && (data.attack_mode == ATTACK_MODE_BF))` in line 2473 also, you just have to divide the result by 32 (which is done anyway).

PR #230, opened 2016-02-17
Support vector width 16

magnumripper:
Please test (test.sh will support `-V 16` now), I could only test it to 90% due to various unrelated bugs with OSX.


jsteube:
Excellent, many thanks!

PR #237, opened 2016-02-23
Fairly strict stamping of WPA salt components

gpuhash:
While experimenting with synthetic handcrafted multi-handshake HCCAP files we discovered that oclHashcat does not recognize unique salts in WPA (-m 2500) mode even if all WPA salt components are the same (but the MIC fields are different due to different known passwords used). Despite this situation is practically impossible in real handshakes due to ANonce and SNonce randomness, it will be quite useful to have strict stamping of WPA salt components.
Possible use of this feature is some kind of WPA proof-of-work system then untrusted worker gets several synthetic HCCAP handshakes with the same salt components but with different MIC fields.
He should find all possible password candidates not knowing who is the real one. If oclHashcat recognises these handshake as non-unique salts the performance of cracking drops significantly.
We tested attached patch with few synthetic multi-handshake HCCAP files and found that now it reliable detects unique salts in WPA mode.

### DIFF
--- a/src/shared.c
+++ b/src/shared.c
@@ -10073,6 +10073,7 @@ int wpa_parse_hash (char *input_buf, uint input_len, hash_t *hash_buf)
 
   uint32_t *p0 = (uint32_t *) in.essid;  
   uint32_t c0 = 0; 
+  uint32_t c1 = 0; 
   
   for (unsigned int i = 0; i < sizeof(in.essid)/sizeof(uint32_t); i++) c0 ^= *p0++;  
   for (unsigned int i = 0; i < sizeof(wpa->pke)/sizeof(wpa->pke[0]); i++) c1 ^= wpa->pke[i];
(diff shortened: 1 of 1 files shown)


gpuhash:
OMG just realised we lost one line while commiting.

https://github.com/gpuhash/oclHashcat/commit/541f231c23dc4f35f266600d8cfcec205a51beaf

philsmd:
so long story short, we should pull this fix because it could happen that in very, very, very rare cases it could happen that:
**non-unique "salts"** were falsely considered as unique and therefore only 1 of those hashes were successfully loaded and cracked

Can you please provide an (anonymized) sample (at least 2 *.hccap files?) such that we can verify this issue and keep those files as a kind of documentation here (to justify this change)?
Thank you very much

---

Now that I'm thinking about it, also the opposite could be true... i.e. that unique salts were not recognized as unique (but they were considered as 2 distinct "salts"), even if they should be marked as unique... Still some examples would be good (for maybe both of these cases).

PR #249, opened 2016-03-01
New format -m 13200 AxCrypt

Fist0urs:
This implements the AxCrypt format (check details here: https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/doc/AxCrypt-Auditing-HOWTO.md )

At the moment, only "passphrase" ciphering is supported (provision of "keyfile" is possible using this software).

I've implemented this format using the same input hash format that in my John The Ripper implementation.

test.pl uses random iteration rounds (AxCrypt default is 10k and minimum is 6) between 6 and 100k

AxCrypt also stores secrets in memory (look at link above), I don't know if it worths implementing it as a separate format (cc @jsteube)

Hash format to add to the wiki would be: 

$axcrypt$*1*6*385c8570bc26c39c96ef9cc39f30b367*cf9814d37fbc72a397420fb1c10271a2104d6a6663aa3bc3

Finally, the tool to extract hashes from AxCrypt encrypted file is "axcrypt2john.py" provided with John The Ripper

Cheers!


epixoip:
Fantastic job, ok to merge!

Issue #245, opened 2016-02-28
make DEBUG=1 doesn't allow debugging at all


philsmd:
This is an issue for developers only.

The current Makefile with the check if the DEBUG variable is set, i.e. if the user runs:

```
make DEBUG=1
```

does not allow debugging at all.

There are actually 2 subproblems:
1. -O2 is used even if DEBUG=1 enables -g
2. -s (removing symbol table and relocation information from the binary) is used even if DEBUG=1 is set

Furthermore, currently the address sanitizer (-fsanitize=address) will be enabled whenever debugging is on. I'm not totally convinced that this makes sense in general. Maybe it should be enabled only whenever DEBUG=2 or something like that was specified on the command line. For instance, for "just" debugging a segmentation fault, it doesn't really make sense to enable compiler debug flags that do more than what e.g. -g does.

Thx

jsteube:
I agree we should have 2 debugging levels.

The first one should "just" enable -g, without optimizations and stripping. 

The second one should be like the first, but additionally enable the sanitizer

Btw, what about that "-ggdb" flag? Does that do any help?

PR #254, opened 2016-03-06
#245: Makefile DEBUG fix

philsmd:
This fix should be enough to make the "debugging" version of oclHashcat to work again under a debugger (like gdb).
3 important changes are included:
- only use -s (stripping) flag for the compiler if debugging was not enabled
- 2 levels of debugging (DEBUG=1 and DEBUG=2), where only DEBUG=2 does sanitize-address checks
- do not use -O2 while debugging (if either DEBUG=1 or DEBUG=2 was enabled)

Please close issue #245 if you agree that this fix solves all problems mentioned within the "make DEBUG=1 doesn't allow debugging at all" issue.

Thank you very much

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -86,10 +86,17 @@ VERSION_SUM              := $(shell git describe --tags --dirty=+ | cut -d- -f3)
 ## Compiler flags
 ##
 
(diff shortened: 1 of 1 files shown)


jsteube:
Many thanks!

PR #258, opened 2016-03-07
fixed overwrite_at_le_4x4 () function (if offset is 32)

philsmd:
There was a small problem (a typo probably) in the OpenCL overwrite_at_le_4x4 () function for offset of exactly 32. The problem could end up in missing cracks, because the wrong buffer was overriden.
Thanks


epixoip:
Ouch, yeah, please merge!

PR #264, opened 2016-03-11
added some checks to the -m 111 parser

philsmd:
There was an important sanity check missing in the -m 111 = nsldaps, SSHA-1(Base64), Netscape LDAP SSHA parser.
This is an example to verify it:

```
{SSHA}AZKja2fbuuB9SpRlHqaoXxbTc3==
```

The problem occurred when the length of the base64 decoded input was less than 20, in this case we got a negative salt->salt_len = [number_less_than_20] - 20 < 0 and this salt->salt_len was afterwards used in the memcpy () function.

This missing sanity check (length check) could lead to a segmentation fault/crash, this pull request should reject such hashes with invalid decoded lenght(s).

Thank you very much

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -151,6 +151,10 @@ Type.: Bug
 File.: Host
 Desc.: Fixed a bug in line counter: Conditional jump or move depends on uninitialised value
 
+Type.: Bug
+File.: Host
+Desc.: Fixed some checks in the parser of -m 111 = nsldaps, SSHA-1(Base64), Netscape LDAP SSHA
+
 * changes v2.00 -> v2.01:
 
 Type.: Bug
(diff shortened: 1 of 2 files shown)


philsmd:
Note: since the format is "{SSHA}" . base64 ($hash . $salt) we cannot make the assumption that base64_decode ($hash . $salt) == 20.
We only can make the check that it must be **at least** of length 20, but since there could be non-zero length salt, we can't test if != 20.

Btw: With the above patch, I introduced both checks PARSER_HASH_LENGTH and PARSER_SALT_LENGTH even if one would be for sure enough. Why the first one is enough ? The answer is, whenever the decoded length is at least 20, the salt is either 0 or longer (which is perfectly fine!). But to make it more clear that both conditions should be satisfied, I added both checks to be safe and easier/clearer to understand (the compiler, especially with -O2 flag, will notice this "dependency" of these two checks and do its best to make only one check and therefore optimize it for us).

PR #270, opened 2016-03-13
added check for hash_len/hash_buf after calls to hlfmt_hash ()

philsmd:
calls to hlfmt_hash () should check if the resulting hash_buf is not NULL and the length (hash_len) is at least 1.
I discovered that sometimes it happens that the format was e.g. detected as SHADOW, but parsing it failed, therefore the len was 0 and the buf was NULL. oclHashcat didn't check for such situations and sometimes it resulted in a segfault/crash.

Thank you very much

### DIFF
--- a/src/oclHashcat.c
+++ b/src/oclHashcat.c
@@ -11144,6 +11144,8 @@ int main (int argc, char **argv)
 
           if (hash_len < 1) continue;
 
+          if (hash_buf == NULL) continue;
+
           if (username)
           {
             char *user_buf = NULL;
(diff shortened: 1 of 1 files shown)


philsmd:
Hmm, maybe we shouldn't just ignore it. But instead print a warning about what happened!
Let me know if this makes sense and you think that would be much better.

philsmd:
https://github.com/philsmd/oclHashcat/commit/169905751649b48cc5c47e4b6d2f81ac15567f0f adds the suggested warning messages. Check the entire changes here https://github.com/hashcat/oclHashcat/pull/270/files

jsteube:
That's clever, I like it

PR #276, opened 2016-03-21
fixed -m 5500 parser, avoid strange crashes

philsmd:
There was a (probably copy-pasted) duplicated check for srvchall_pos, but no check if hash_pos is NULL.

This missing check for hash_pos == NULL was the culprit for several crashes with corrupted or invalid NetNTMLv1 hashes.

Thank you very much


jsteube:
Cool!

PR #280, opened 2016-03-25
use a GCC equivalent for __stdcall where applicable

MestreLion:
GCC has **attribute**((stdcall)) as an equivalent to Windows __stdcall, so use it conditionally instead of leaving it blank for all non-Windows platforms. This may benefit all platforms where GCC or compatible compilers are used, including (but not limited to) Linux.


jsteube:
Thanks!

PR #281, opened 2016-03-29
fixed base64_decode () input length validation for -m 8900

philsmd:
There was a missing check within the scrypt (-m 8900) parser function which should prevent a buffer overflow of "tmp_buf". The (decoded) salt length should be limited to 32 bytes (max 45 base64 encoded).

This patch adds the missing check to prevent crashes/memory corruption with too long base64-encoded salts.

Thank you very much


jsteube:
many thanks!

PR #285, opened 2016-04-02
fixed some parser checks for new keepass format (-m 13400)

philsmd:
for this _new_ format, -m 13400 = keepass 1/2, there were some incorrect checks within the parser function.

The problem is that we need to check for NULL pointers _immediately_ after calling strchr (), calling it afterwards might already be too late (and could result in crashes/segmentation faults).

Thank you very much


Fist0urs:
Seems good to me

PR #289, opened 2016-04-06
Added support of keyfiles within Keepass 1.x and Keepass 2.x

Fist0urs:
Hi guys,

this PR provides support of keyfile usage within Keepass 1.x and Keepass 2.x. You have to use keepass2john for John The Ripper suite to get right input format (actually the PR that makes things work within keepass2john is done on jtr).

Happy cracking!

Fist0urs


jsteube:
Awesome! Many thanks!

PR #291, opened 2016-04-09
Added new hash mode -m 125 = ArubaOS

hops:
This adds support for ArubaOS hashes used in the Aruba Wireless Controllers.
I have verified it with hashes from the following two devices.
7010 Cloud Services Controller running ArubaOS 6.4.2.12
3200 Mobility Controller running ArubaOS 6.1.1.0

They are stored in the configuration like this:
mgmt-user admin root d325047b01c23ad649a0a8c99b4c73ff6c675925c157c7ff76
enable secret "64dcc0c80168e6c41f279e59f9545ed9a9af980c702a713aa4"
The first line is the user (admin) and the max length of the password is 32.
The second line is used to get in the privileged mode and the max length of the password is 15.

The first 10 hex chars form the salt (in binary) while the first four bytes are random generated and the fifth is the signature (always 01).
What comes after is the SHA1 hash of $salt.$pass

You can add one of the above hashes to the wiki as example (password is "hashcat")


jsteube:
Excellent! Many thanks

jsteube:
From all the documents I've found so far, the maximum password length for ArubaOS seem to be 64, not 32. This is confusing, what's the right value? I'm currently refactoring password length related kernels and therefore it's important ot have the correct value

hops:
I'm pretty sure on the 3200 Mobility Controller running ArubaOS 6.1.1.0 (quite old) the limit is 32.
I currently don't have access to that device, but can test it tomorrow evening.
Also I'll ask someone I know who has more up to date devices to test the limit.

hops:
@jsteube The maximum password lenght is definitely 32 characters.
It has been tested on a 3200 mobility controller running ArubaOS version 6.4.4.14 (compiled on 22th of June 2017) and a 7210 mobility controller running ArubaOS running 6.4.4.8 and 6.5.1.2.
In which documentation did you see the 64 characters limit? Maybe that was for different products?

PR #292, opened 2016-04-10
Many rules touched so they will work out of the box with john-jumbo

jfoug:
I have made changes to john-jumbo, so that it can handle all HashCat rules, and logic changes.  That change is still waiting some discussion, but is rock solid, and end results are lining up between the 2 rules engines just fine.

https://github.com/magnumripper/JohnTheRipper/issues/2095

The comment for the commit (next message in this PR) explains a lot of what is going on, but simply put, for this to work, I had to change all rules which start out with a . or a [    Those have special meaning within john .conf file loading code, and thus, were causing the file to not be able to be loaded.  I simply changed all of those lines to start with a : no op code.


jsteube:
I'll reject this as I don't see any advantage for hashcat. Also, in case of PasswordsPro updates it's default ruleset, who's going to update it then?

There's 2 things to consider:
- Change JtR parsing/handling code for [ in case "!! HashCat logic ON" is used.
- Change your default loading behavior for rules to what you call "HashCat logic" and create "JtR-old logic" you can use for your old default rulesets. That would make all rule handling of all tools compatible to each other. Background for this is that hashcat logic isn't hashcat logic, it's based on PasswordsPro.

jfoug:
This is not rules 'loading'.  It is the config (.conf file) loading.  They are 2 totally separate items.  The config file loading loads the whole john.conf, and in jumbos case, anything that is .included.   It is this logic where the interpretation of the leading [ and . characters is the problem.

I may be able to make that change, but it will be very hackish at best.

The advantage is for the users.  It allows usage of already made rulesets between more tools.  There is little or no downside to having some of the rules files have a few no-op commands added to them.

jfoug:
This was made to work without too much difficulty modifying the config.c file that parses a line, so the changes requested here, are not required to have john able to use these rulez files.

PR #303, opened 2016-04-19
Fix LLVM compilation for OpenCL kernels

neheb:
They otherwise fail with "static type not supported". LLVM only does OpenCL 1.1.

### DIFF
--- a/include/kernel_vendor.h
+++ b/include/kernel_vendor.h
@@ -4,6 +4,7 @@
  */
 
 #pragma OPENCL EXTENSION cl_khr_byte_addressable_store : enable
+#pragma OPENCL EXTENSION cl_clang_storage_class_specifiers : enable
 
 /**
  * vendor specific


jsteube:
This looks almost good. Can you please try to put it into include/kernel_vendor.h instead? That's where the extensions usually are called from. If that's not possible, for whatever reason, please tell me and we'll merge it as it is

neheb:
ah that file looks to be a much better fit! I got a bit confused on where to place the statement. I believe LLVM's OpenCL support works on both AMD and Nvidia cards. I'm only testing AMD at the moment though. In any case the change has been force pushed.

jsteube:
Thanks!

jsteube:
@neheb Can you please test if it's still working on LLVM

neheb:
yeah it's working. still doesn't compile right but that's a different issue.

PR #304, opened 2016-04-20
dive.rule: removed simple duplicates

jfoug:
Very simple dupe removal. Removed the 2nd identical rule, where one rule has spaces between each operation, and one does not.

There were so many that git browser view does not show the diff file.


jsteube:
Cool! Thanks!

PR #305, opened 2016-04-21
Disable AMD optimizations for LLVM

neheb:
libclc does not implement any 3rd party OpenCL extensions which happens to include all of the amd_ ones. As such, disabling them is needed. I have no idea what the proper solution to this is. Implementing them using asm?

Anyway, not tested on fglrx.

### DIFF
--- a/OpenCL/types_ocl.c
+++ b/OpenCL/types_ocl.c
@@ -199,11 +199,9 @@ static inline u32 rotl32_S (const u32 a, const u32 n)
 
 static inline u64 rotr64_S (const u64 a, const u32 n)
 {
-  #if DEVICE_TYPE == DEVICE_TYPE_CPU
+  #if (DEVICE_TYPE == DEVICE_TYPE_GPU)
 
-  const u64 r = rotate (a, (u64) 64 - n);
-
-  #else
+  #ifdef amd_bitalign
 
   const u32 a0 = h32_from_64_S (a);
   const u32 a1 = l32_from_64_S (a);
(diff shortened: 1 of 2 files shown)


jsteube:
This seem not to be correct. For example, If you don't want to use amd_bfe(), why do you check for "cl_clang_storage_class_specifiers" and not simply "amd_bfe" ?

The or of cl_clang_storage_class_specifiers in combination with DEVICE_TYPE may work but is hard to understand. I think we have to find a cleaner way to handle this.

Can you tell me how I can make use of LLVM like you do use it here? Maybe I can help finding a proper way.

neheb:
Good point. I guess I was thinking of LLVM too much. I'll redo the commit when I get home.

As for LLVM, you mean to use it with hashcat? That requires ditching fglrx and installing mesa with its opencl package. Also commenting out the wavefront call since it will fail(unimplemented).

Oh also a 7000 or newer GPU is required. 6000 technically works but global atomics are not implemented.

jsteube:
Cool I really have to try that. How's the performance compared to fglrx?

neheb:
Can't say since my card is a 6k series. The kernels compile now but they crash upon use of atomic_add or atomic_or.

neheb:
Fixed up, Should be good now.

jsteube:
I'm still not 100% happy. 

Note that rotate() is not part of cl_amd_media_ops, it's a basic OpenCL 1.0 standard: https://www.khronos.org/registry/cl/sdk/1.0/docs/man/xhtml/rotate.html

It's weird that an OpenCL runtime does not include it, it totally should!

Another thing is:

```
#ifdef cl_amd_media_ops
    return amd_bfe (a, b, c);
```

It should be:

```
#ifdef amd_bfe 
    return amd_bfe (a, b, c);
```

Then there's:

```
(DEVICE_TYPE == DEVICE_TYPE_CPU) || !defined (cl_amd_media_ops)
```

First thing is that we try to avoid NOT cases, so we try to avoid '!'. Another thing is that we try to avoid && and || in branches. This makes exception handling easier. Please rewrite it to have two seperate branches for this (use 2x if()).

neheb:
Another attempt. Not sure if the if statements are quite what you were looking for.

jsteube:
All right yes I think that's good, many thanks! I should totally try to get the same setup running as you have so that we can make it compile by default.

PR #307, opened 2016-04-21
2nd group of dupe removal. Also contains some rules put back into dive.rule that were over agressively removed on first PR

jfoug:
**_DO NOT MERGE_** yet.  I have to update my script, and fix the 2nd commit.


jfoug:
HOLD UP!

I think I see some problems.  Let me validate.  I see some that I think should not be there:

```
69544 : $1 $2 d ] ]
68589 : $8 d ]
```

I am pretty sure those were removed because of the `d` rule, and that is not valid.

Yup, validated, and that is wrong.

```
$ cat > x.rule
d
$a d ]
$a$b d ]]

$ ./a x
Total lines read 3 Unique lines written 1
1 : $a d ]
2 : $a$b d ]]
```

```
$ echo test | run/john -pipe -stdout -rule=:'d;$ad];$a$bd]]'
testtest
testatest
testabtest
3p 0:00:00:00 0.00% 14.15p/s testabtest
```

jfoug:
Here is another dupe type (but hard to 'spot', and likely not used much)

```
$x d ]   is dupe to M $x 6
$x$y d ]] is dupe to M $x$y 6  (2nd is less steps)  Also, in JtR, the M can/should be skipped.
$x$y$z d ]]] dupe to M$x$y$z6 or ^z^y^xd[[[
```

Not sure this would show up often, I doubt it.  But it could in random rules.  But this is the type of duplicate finder I am 'trying' to work to find, in a rules emulator that would emulate the work being done, holistically view what was trying to be accomplished, and then see if there is a canonical more optimal way to do something.  So in the $xd] what we are trying to do is insert x into the middle of the word duplicated. So if M$x6 is emulated, it should also show that we are trying to insert x into the middle of the word duplicated, thus showing that $xd] and M$x6 are identical.  However, my script is nowhere to this level of intelligence yet, lol.

jsteube:
Wait, we can't really replace 'd' with 'M' because all memory store operations like 'M', '4' and '6' . They are not support in oclHashcat

jfoug:
I had not started on any of the dupes listed in the last post, so do not worry.

So, if I did see a dupe of `$x d ]` and `M$x6` I should remove the 2nd one, leaving the first, because we want to make sure that anything left works on more systems?

jfoug:
Please simply reject this PR and I will start over with new ones, that can be validated.

PR #308, opened 2016-04-21
Simple dupe removal

jfoug:
These are 'pure' duplicates, and superfluous white space duplicates (which create 'pure' duplicates). These should be merged as safe, prior to any other logic found duplicates are looked at.

I have included ALL script files used to find these duplicates, and auto-remove them.


jsteube:
Thanks, good work!

Issue #306, opened 2016-04-21
Support for PeopleSoft's PS_TOKEN SHA1


fgaudreault:
Would love to see a new mode added for that. There is one for the application password, but if we could get a sha1($salt.utf16le($pass)) to break the SHA1 signature of the token, that would be great. The salt length is very long, 130-150bytes, so we can't really use any existing mode.

jsteube:
Wait, they have a 130 byte long salt on start? hihihi I guess they didn't read my post about how not to salt a hash :) https://hashcat.net/forum/thread-4429.html

Please give us an exact number for the salt length range

fgaudreault:
It will be hard to give you an exact bracket or a fixed number, the salt length will vary depending of variables such as the username, the node name, the date, etc. :/ The other thing is that the salt will contain null bytes.

jsteube:
The null byte shouldn't be a problem. You can encode the salt in hex and use --hex-salt or we're creating a special parser for the mode. I'd recommend the 2nd in case the different salts are not concatinated into a single long string.

fgaudreault:
Ok, that make sense. I actually created a new parser yesterday on a local copy, but now I have to figure out how to change the algorithm (which is almost identical to -m 140), except that the salt is bigger and the password is encoded in utf16le.

jsteube:
you need to create a new kernel to make and because of this you can't use the -m 140 address space for it. use the next free -m

jsteube:
i can help you with the kernel details once you pushed the parser and kernel template

fgaudreault:
Ok, so I will prepare a pull for the parser code (which is basically a copy of 140, but with higher min/max salt length constants), and a copy of the 140 kernels as a base. My local test picked the salt, but the core was not able to crack the password. I believe that's normal nice the kernels are not able to handle that data yet. I used -m 134 since the PeopleSoft application passwords were -m 133.

jsteube:
Please don't use * here

    #define DISPLAY_LEN_MIN_13500 40 + 1 + 16 * 2
    #define DISPLAY_LEN_MAX_13500 40 + 1 + 512 * 2

Simply do:

    #define DISPLAY_LEN_MIN_13500 40 + 1 + 32
    #define DISPLAY_LEN_MAX_13500 40 + 1 + 1024

Sorting of this:

    13400,
    13500,

Should be:

      133,
    13500,

You're still using an invalid option for the salt which you need to remove:

[...]

PR #311, opened 2016-04-22
Adding parser and basic kernels for -m 13500

fgaudreault:
Adding the parser for PS_TOKEN long salt, and copy of the -m 140 kernels to start with.

See Issue #306


jsteube:
It's a good start, but we have to do some changes:
- First I'd like to move the hashmode from 134 to 13500 (I'll reserver that for your hash-mode).
- Then I've seen that you've used parse_and_store_salt() to store the salt. This is invalid in your case, as the salt is so long. Also note that the salt_buf in salt_t is limited to 64 byte, so you've propably already overwrote it.
- To make this correctly, you first need to define a new esalt (external salt) and give it enough space to handle it. Like 512 byte. Note that you have to work with 32 bit integers so you need to create something like u32 buf[128].
- After that you need to create the space for the salt, see other algorithms like 13400 in how to do that. 
- Next you need to store the salt into this esalt manually and without parse_and_store_salt. There's also plenty of examples in other parser functions. 
- Finally I'd suggest to copy at least 16 byte of that salt into the original salt_t salt_buf because based on that it will find if a hash is unique in some internal sorting routine

After this is done we can start with the kernel code

fgaudreault:
We changed the mode to 13500, and added an esalt structure. I believe the buffers are populating the data according to your recommendations. So what are the next steps? :)

jsteube:
That looks much better. Following things are missing:
- Code in ascii_digest(). This function is responsible for generating the output string. It's very important to generate 1:1 the same output string as the input string is. 
- Add unit testing code in test.pl and test.sh.
- You use memcpy for the salt buffer, but you'd actually have to decode the hex data first. something like this:
  for (i = 0; i < salt_len / 2; i++)
  {
    out[i] = hex_to_u8 ((const u8 *) &salt_pos[i \* 2]);
  }
- You marked the hash with the option OPTS_TYPE_ST_ADD80 but using OPTS_TYPE_ST_ADD80  makes sense only for schemes like sha1($pass.$salt) where the $salt comes after the $pass. I think what you want to use is OPTS_TYPE_PT_ADD80. But note this will work on -a3 kernel only automatically. For -a0 and -a1 you have to do it in the kernel.

fgaudreault:
The tests are still missing, as well as the appropriate kernel code. But for the rest, we should have the base in place. More tests will be required afterward.

jsteube:
If you want I can merge and finish the kernel

fgaudreault:
I think that would be great. I'll just make sure we have the tests in place, and remove the warning on compiling. Should be done in the next commit. Meanwhile, I will also fix the test.pl file for the new Digest::ECB versions, the constants are no longer exported.

jsteube:
OK, waiting for your "go"

fgaudreault:
I think we are good now. The test.pl results have been cross-checked using JtR, and it cracks properly. Unfortunately, I had to use a constant salt, but that should do the trick.

jsteube:
Please don't use \* here

```
#define DISPLAY_LEN_MIN_13500 40 + 1 + 16 * 2
#define DISPLAY_LEN_MAX_13500 40 + 1 + 512 * 2
```

Simply do:

```
#define DISPLAY_LEN_MIN_13500 40 + 1 + 32
#define DISPLAY_LEN_MAX_13500 40 + 1 + 1024
```

Sorting of this:

```
13400,
13500,
```

Should be:

```
  133,
13500,
```

You're still using an invalid option for the salt which you need to remove:

```
| OPTS_TYPE_ST_ADDBITS15;
```

The following optimization is not possible since our data buffer is > length 55, you need to remove it:

```
| OPTI_TYPE_PRECOMPUTE_MERKLE
```

This code:

```
memset(pstoken_tmp, 0, mysalt_len + 1);
```

Is wrong in two ways. First you should use sizeof():

```
memset (pstoken_tmp, 0, sizeof (pstoken_tmp));
```

Another problem is that you're using a value which is unknown at compile time to declare a number of elements in an array.

Instead of:

```
u8 pstoken_tmp[mysalt_len + 1];
```

You should do:

```
u8 pstoken_tmp[512 + 1];
```

More problems with snprintf():

```
snprintf((char *)(pstoken_tmp + i), (size_t)2, "%02x", pstoken->salt_buf[i]);
```

Problem 1, the "%02x" does not match salt_buf[] which is 32 bit, not 8 bit. You propably need to something like this befoire you can use it:

```
u8 *salt_buf_ptr = (u8 *) pstoken->salt_buf:

... and then use salt_buf_ptr[i] in the snprintf()
```

Problem 2, the casting of pstoken_tmp from u8 \* -> char \* is not necessary. 

```
Just declare pstoken_tmp as char *.
```

Suggestion:

```
The use of snprintf() is slow, take a look bin_to_hex_lower (). This is not always useful but sometimes it is.
```

Inside the parser you do:

```
u8 pstoken_tmp[DISPLAY_LEN_MAX_13500 - 40 - 1];

memset(pstoken_tmp, 0, DISPLAY_LEN_MAX_13500 - 40 - 1);
```

I'd recommend to use the same declaration as in ascii_digest instead:

```
u8 pstoken_tmp[512 + 1];

memset (pstoken_tmp, 0, sizeof (pstoken_tmp));
```

This optimization does not work here, because we need a 2nd call to sha1_transform because of the length > 55. You need to remove it:

```
digest[0] -= SHA1M_A;
digest[1] -= SHA1M_B;
digest[2] -= SHA1M_C;
digest[3] -= SHA1M_D;
digest[4] -= SHA1M_E;
```

That was easy stuff so fow, but now in test.pl it's getting real problematic. From your original issue you said:

> Would love to see a new mode added for that. There is one for the application password, but if we could get a sha1($salt.utf16le($pass)) to break the SHA1 signature of the token, that would be great. The salt length is very long, 130-150bytes, so we can't really use any existing mode.

What you totaly did not mention is the use of DES inside the algorithm. Now I'm very confused. Please explain in words how DES is involved in the process. It's important to write a proper kernel.

fgaudreault:
There is no DES in the algorithm. I fixed the use Crypt::ECB module only because the newer version is no longer using PADDING_NONE and PADDING_AUTO, you have to push the padding in the encrypt sub. That being said, we will fix the problems.

fgaudreault:
Aight. So at this point, we think we would need the kernels in order to test the full loop on our own and fix the remaining issues. Because it's the first time we touch the code base, having working kernels would help, and if you can help us with that, it would go much faster. Thanks for your patience btw ;)

jsteube:
I'll now rewrite the kernel and inform you when I've pushed a working version

PR #312, opened 2016-04-22
Remove duplicate rules (s chains, s sorting and : removal)

jfoug:
Pretty simple dupe finds.  All easily validated.


jsteube:
Another ~250 removed. Great work!

PR #313, opened 2016-04-23
Duplicates removed. Based on $ and ^ reorder only

jfoug:
There are still dupes based upon `$` and `^` reorder.  We will get more of them as I add other reorders (especially when I get to `[` and `]`  and `{` `}` groups

I am doing this a step at a time, to make validation easier.  We have caught times where over aggressive rule modifications have cause non-duplicate rules to be removed as duplicate. I am trying to avoid that, or at least catch the cases where it happens, to add additional logic to work around specific issues.

In this commit, I changed my script to work in a one-off manner.  No more using unique, diff and going back to original file.  Now the first script detects the duplicate, outputs original line, and provides a log file of what line was removed, what the rule was, and what the original line number (and rule) which were already written, but now detected as a duplicate.  Each commit has a comment that is this 'log' of data. It is very useful in validating that ONLY duplicate rules are removed.   It takes a bit of time to look through it, but we now should easily be able to check that nothing that should have stayed in the files was removed.


jfoug:
There is one more block of duplicates after this one.  But now the duplicates are getting fewer, and 'harder' to find and know they are really duplicates.

After the next block, I think I am gonna call it quits for now.  What we have now is a great number of them removed.  Yes, there still are duplicates, but finding enough of them to make the ROI worthwhile is hard.

I will however, do some rule 'auto re-write', once all of the dupes are in. I want to remove the true no-ops from the file. So far, I have preserved original data (with a couple of exceptions).  The no-ops are things like:     `:` `^x[` `$x]` and others.  But I will wait until all duplicate stuff is in, before doing those.  Those changes simply make the rules shorter (fewer steps and faster), by removing commands that play no part in the final word.

jfoug:
Well it looks like all recent comments were dumped into this PR, so I think it contains all of the work I have done so far.   That being the case, this is all of the duplicates I plan on right now, unless I spot something else which has a good ROI for the work it takes to find them.

jfoug:
This task is not being written off as completed!

jfoug:
Here is the entire script used for finding dupes (and finally to remove the no-op data).  NOTE, lots of lines were commented out to start, and then each step uncommented more and more blocks, so that only 1 'type' of duplicate finding was done at each step.  This script was made to run with either HC or john logic.  Also in the john side, we read data from the .conf files with multiple sections.

If run with -- hc as the 2 command line params, then only HC logic is used, also neither of the external perl module files will be required.  Those do nothing useful in HC mode.  In john mode, they load and process the .conf files, and perform the preprocessing logic to convert johns input rule (with rpp data in it), into the multiple rules that are used.  I can think of no way to script find duplicates within john's pre-processor extension.  It simply adds too much ambiguity to the work. Duplicates can easily be found with the pp, but it will have to be done by hand. This script works fine at showing the duplicates, BUT whoever is doing it will have to figure out where/how to hook it back into the pp code.  HC has no pp helper logic, so simplistic rule duplication logic is all that is required.

``` Perl
#!/usr/bin/perl

# script to canonize (or try to) john/hc/crack rules
#
#  1 move all validation to as early as possible.
#     length validations, move to start, or to just after prior length changing command
#     is char validations.  Move to start (right after lenght validations), or right after or character modifiers.
#  s rule, drop any 2nd usage of original letter (the first ate it).
#  s rule, if target (2nd letter) of an s is later the first letter of another s, then merge the 2 rules.
#  sort s rules by original letter (and pack back to back).
#  s logic groupings are bounded by things like case or other changers.
#  place all adjacent [ before ] so [][][ becomes [[[]] and ][ becomes []
#  if we find a [ and there was a prior prepend, then drop the most recent prepend and the [
#  if we find a ] and there was a prior append, then drop most recent append and the ]
#    Note for prior 2, seeing many rule objects has to remove knowledge of the prior prepend, append
#    or both.  Things like d } { and many others make removing that prior  append/prepend not be
#    something that can be done as a no-op any more.
#  [ and ] are grouped with {} and possibly o O x v
#  remove all : ops, UNLESS : is the entire rule.
#  TODO: { in middle of } block can be removed (and one })  { in middle of } block likewise.

use strict;

my @ar = ();
my $hc = 0;
my %map = ();
my $lncnt = 0;

# to do raw hashcat rules, just do:    cat xxx.rule | ./rulecanonize -- hc
if ($ARGV[0] eq '--') {
    if (defined $ARGV[1] && $ARGV[1] eq 'hc') { $hc = 1; }
    @ar = (<STDIN>);
} else {
    # we may need to control 'what' gets loaded.
    use File::Basename;
    use lib dirname(__FILE__).'/';
    use lib dirname(__FILE__).'/../run';
    require jtrconf;
    jtrconf::load();
    @ar = jtrconf::getsection('List.Rules:'.$ARGV[0]);
}

chomp @ar;
my $force_output_noop_removed = 0;

foreach my $rule (@ar) {
    my $cnt=20000;
    ++$lncnt;
    if (substr ($rule, 0, 1) eq '#' || length($rule) == 0) { print "$rule\n"; next; }
    my $rulepp = $rule;
    if ($hc == 0) {
        use File::Basename;
        use lib dirname(__FILE__).'/';
        use lib dirname(__FILE__).'/../run';
        require jtr_rulez;
        jtr_rulez::jtr_rule_pp_init($rule, 125, $cnt, $hc);
    }
    my $rule_cnt = 0;
    while (defined ($rulepp) && length($rulepp)>0) {
        my $canon = canonize($rulepp, $hc, 1);
        if (!defined $map{$canon}) {
            $map{$canon} = $lncnt-1;
            if ($force_output_noop_removed == 1) {
                print "$canon\n";
                print STDERR "ln-$lncnt:  $rule  reduced to $canon\n";
            } else {
                print "$rule\n";
            }
        } else {
            print STDERR "ln-$lncnt:  $rule  dup of ln-$map{$canon}: $ar[$map{$canon}]\n";
        }

        if ($hc == 0) {
            use File::Basename;
            use lib dirname(__FILE__).'/';
            use lib dirname(__FILE__).'/../run';
            require jtr_rulez;
            $rulepp = jtr_rulez::jtr_rule_pp_next();
        } else {
            $rulepp = "";
        }
    }
}

sub ch { my ($r, $i) = @_;
    return substr($r,$i,1);
}
sub canonize { my ($rule, $hc, $sp) = @_;
    my $rules = tokenize($rule, $hc);
    my $mod = 1;
    if (!defined $hc) { $hc = 0; }
    $force_output_noop_removed = 0;
    while ($mod == 1) {
        $mod = 0;
        my @ar = @$rules;
        # remove all : and -: (unless it is the ONLY rule)
        for (my $i = 0; $i < scalar @ar; ++$i) {
            if (scalar @ar > 1 && $ar[$i] eq ':') {
                splice(@ar, $i, 1);
                $force_output_noop_removed = 1;
                --$i;
            }
            if ($ar[$i] eq '-:') {
                splice(@ar, $i, 1);
                --$i;
            }
        }
        # remove all identical - rejection rules (keeping the first only.
        if ($hc == 0) {
        for (my $i = 0; $i < scalar @ar-1; ++$i) {
            if (length($ar[$i]) > 1 && ch($ar[$i],0) eq '-') {
                for (my $j = $i+1; $j < scalar @ar; ++$j) {
                    if ($ar[$i] eq $ar[$j]) {
                        splice(@ar, $j, 1);
                        --$j;
                    }
                }
            }
        }}
        # put all rejection rules (-) first.
        if ($hc == 0) {
        my $rej1 = 0;
        for (my $i = 0; $i < scalar @ar-1; ++$i) {
            if (length($ar[$i]) > 1 && ch($ar[$i],0) eq '-' && $i != $rej1) {
                my $s = $ar[$i];
                splice(@ar, $i, 1);
                splice(@ar, $rej1, 0, $s);
                ++$rej1;
            }
        }}

#       # remove [ and ] if there were $ or ^ (or A[zl] or A0) Likely we should first convert
#       # A[lz] into $ and A0 into ^ and then later they get converted back into Az/0
#       # NOTE, { } will cause this to fail without adjustments, D may, as could i. as could most anything (such as d)
#       # ITW, I did see d and $ad] $a$bd]] which are obviously 3 differnt rules.
        my $h=-1; my $t=-1;
        for (my $i = 0; $i < scalar @ar; ++$i) {
            my $c = ch($ar[$i],0);
            if ($c eq '$') { $t = $i; next; }
            if ($c eq '^') { $h = $i; next; }
            if ($c eq '[' && $h != -1) {
                splice(@ar, $i, 1);
                splice(@ar, $h, 1);     # need to handle Az here, and possibly in next line.
                $force_output_noop_removed = 1;
                $h = $t = $i = -1;
                next;
            }
            if ($c eq ']' && $t != -1) {
                splice(@ar, $i, 1);
                splice(@ar, $t, 1);     # need to handle A0 here, and possibly in next line.
                $force_output_noop_removed = 1;
                $h = $t = $i = -1;
                next;
            }
            if (index('dfpODPIxq{}', $c) != -1) {
                $h = $t = -1;
            } elsif ($c eq 'r') {
                my $tmp = $h; $h = $t; $t = $tmp;
            } elsif (index('kzy4,', $c) != -1) {
                $h = -1;
            } elsif (index('KZY6.\'', $c) != -1) {
                $t = -1;
            }
        }
        # drop 'dupe' s values, or s where an output later becomes an s input.
        for (my $i = 0; $i < scalar @ar - 1; ++$i) {
            if (ch($ar[$i],0) eq 's') {
                for (my $j = $i+1; $j < scalar @ar; ++$j) {
                    if (ch($ar[$j],0) eq 's') {
                        if (ch($ar[$i],1) eq ch($ar[$j],1)) {
                            $mod = 1;
                            splice(@ar, $j, 1);
                            $force_output_noop_removed = 1;
                            # before proceeding, remove ANY others like this
                            for (my $k = $j; $k < scalar @ar; ++$k) {
                                if (ch($ar[$i],1) eq ch($ar[$k],1)) {
                                    splice(@ar, $k, 1);
                                    --$k;
                                    next;
                                }
                            }
                            --$j;
                            next;
                        }
                        if (ch($ar[$i],2) eq ch($ar[$j],1)) {
                            $mod = 1;
                            substr($ar[$i],2) = substr($ar[$j],2);
                            my $c = substr($ar[$j],1,1);
                            splice(@ar, $j, 1);
                            $force_output_noop_removed = 1;
                            # before proceeding, remove ANY others like this
                            for (my $k = $j; $k < scalar @ar; ++$k) {
                                if ($c eq ch($ar[$k],1)) {
                                    splice(@ar, $k, 1);
                                    --$k;
                                    next;
                                }
                            }
                            --$j;
                            next;
                        }
                    }
                }
            }
        }
        # now that sCC has been 'reduced' removing dupes and cycles, canonize it by outputting groupings in sorted order
        for(my $i = 0; $i < scalar @ar; ++$i) {
            if (ch($ar[$i],0) eq 's' && ch($ar[$i+1],0) eq 's' && ord(ch($ar[$i],1)) > ord(ch($ar[$i+1],1))) {
                my $s = $ar[$i];
                $ar[$i] = $ar[$i+1];
                $ar[$i+1] = $s;
                $mod = 1;
                next;
            }
        }

        # convert $ ^ blocks into A0'bbb' and Az'xxx' with A0 packed first.
        for(my $i = 0; $i < scalar @ar; ++$i) {
            if (ch($ar[$i],0) eq '$' && ch($ar[$i+1],0) eq '$') {
                $ar[$i] .= substr($ar[$i+1],1);
                splice(@ar, $i+1, 1);
                --$i;
                $mod = 1;
                next;
            }
            if (ch($ar[$i],0) eq '$' && ch($ar[$i+1],0) eq '^') {
                my $s = $ar[$i];
                $ar[$i] = $ar[$i+1];
                $ar[$i+1] = $s;
                $mod = 1;
                next;
            }
            if (ch($ar[$i],0) eq '^' && ch($ar[$i+1],0) eq '^') {
                $ar[$i] = '^' . substr($ar[$i+1],1) . substr($ar[$i],1);
                splice(@ar, $i+1, 1);
                --$i;
                $mod = 1;
                next;
            }
            if ($ar[$i] eq ']' && $ar[$i+1] eq '[') {
                $ar[$i] = '[';
                $ar[$i+1] = ']';
                $mod = 1;
                next;
            }
            if (index('cClut', $ar[$i]) != -1 && $ar[$i+1] eq ']') {
                $ar[$i+1] = $ar[$i];
                $ar[$i] = ']';
                $mod = 1;
                next;
            }
            if (index('lut', $ar[$i]) != -1  && $ar[$i+1] eq '[') {
                $ar[$i+1] = $ar[$i];
                $ar[$i] = '[';
                $mod = 1;
                next;
            }
            # not sure these are right, so left out of code.
            #if (substr($ar[$i],0,1) eq '$' && $ar[$i+1] eq '[') {
            #   $ar[$i+1] = $ar[$i];
            #   $ar[$i] = '[';
            #   $mod = 1;
            #   next;
            #}
            #if (substr($ar[$i],0,1) eq '^' && $ar[$i+1] eq ']') {
            #   $ar[$i+1] = $ar[$i];
            #   $ar[$i] = ']';
            #   $mod = 1;
            #   next;
            #}
        }

        @$rules = @ar;
    }
    my $s = "";
    foreach my $c (@$rules) {
        if (length($c)>2 && ch($c,0) eq '$') {
            if ($hc > 0) {
                my $s = "";
                for (my $j = 1; $j < length($c); ++$j) {
                    if (defined $sp && $sp != 0 && length ($s)) { $s .= ' '; }
                    $s .= '$'.substr($c,$j,1);
                }
                $c = $s;
            } else {
                $c = "Az`".substr($c,1).'`';
            }
        }
        if (length($c)>2 && ch($c,0) eq '^') {
            if ($hc > 0) {
                my $s = "";
                for (my $j = length($c)-1; $j > 0; --$j) {
                    if (defined $sp && $sp != 0 && length ($s)) { $s .= ' '; }
                    $s .= '^'.substr($c,$j,1);
                }
                $c = $s;
            } else {
                $c = "A0`".substr($c,1).'`';
            }
        }
        if (defined $sp && $sp != 0 && length ($s)) { $s .= ' '; }
        $s .= $c;
    }
    return $s;
}

sub tokenize { my ($rule, $hc)=@_;
    my @ar = split //, $rule;
    my @rules = ();
    for (my $i = 0; $i < scalar @ar; ++$i) {
        my $c = $ar[$i];
        next if ($c eq ' ');
        my $esc = 0;
        # 1 byte rules.
        if (index('E:lucCtrd{}f12[]PISVMQUkK46q', $c) != -1) { # 1 char rules
            push @rules, $c;
            next;
        }
        if ($hc == 0) {
            if ($c eq '-') {  # jtr rejections
                my $s = $c.$ar[++$i];
                if ($ar[$i] eq '<' || $ar[$i] eq '>') {
                    $s .= $ar[++$i];
                }
                push @rules, $s;
                next;
            }
            if (index('pRL', $c) != -1) { # 1 char rules
                # if next char is digit, then use HC pRL rules.
                if (ord($ar[$i+1]) < ord('0') || ord($ar[$i+1]) > ord('9')) {
                    push @rules, $c;
                    next;
                }
            }
        }
        if (index('T$^<>_\'DpRL+-zZ.,yY', $c) != -1) { # 2 char rules (pRL here is pN 2 char rule)
            my $s = $c.$ar[++$i];
            push @rules, $s;
            next;
        }
        if (index('xioO*', $c) != -1) { # 3 char rules
            my $s = $c.$ar[++$i]; $s .= $ar[++$i];
            push @rules, $s;
            next;
        }
        if (index('Xv', $c) != -1) { # 4 char rules
            my $s = $c.$ar[++$i]; $s .= $ar[++$i]; $s .= $ar[++$i];
            push @rules, $s;
            next;
        }
        if ($c eq 'A') {
            my $s = $c.$ar[++$i];
            $c = $ar[++$i];
            $s .= $c;
            while ($ar[$i+1] ne $c) {
                $s .= $ar[++$i];
            }
            $s .= $ar[++$i];
            push @rules, $s;
            next;
        }
        # handle rules that can use classes in JtR mode
        if (index('@!/()e', $c) != -1) { # 2-3 char rules
            my $s = $c.$ar[++$i];
            if ($hc == 0 && $ar[$i] eq '?') { $s .= $ar[++$i]; }
            push @rules, $s;
            next;
        }
        if (index('s', $c) != -1) { # 3-4 char rule with ? 2nd char
            my $s = $c.$ar[++$i];
            if ($hc == 0 && $ar[$i] eq '?') { $s .= $ar[++$i]; }
            $s .= $ar[++$i];
            push @rules, $s;
            next;
        }
        if (index('=%', $c) != -1) { # 3-4 char rule with ? 3rd char
            my $s = $c.$ar[++$i]; $s .= $ar[++$i];
            if ($hc == 0 && $ar[$i] eq '?') { $s .= $ar[++$i]; }
            push @rules, $s;
            next;
        }
        print "Error, unknown rule token at $c next char $ar[$i+1]  Orig: $rule\nRules found so far : ";
        foreach my $s (@rules) {
            print " $s ";
        }
        print ("\n");
        exit 1;
    }
    return \@rules;
}
```

jsteube:
This is crazy. Can I merge? How many rules did you remove in total?

jfoug:
Yes, this can be merged. I validated every rule was a duplicate.  There are about 1100 more removed in this PR.

NOTE, I wrote code to find the rest.  There are a couple of types that I can auto-remove.  The others may take hand looking.  There are roughly 12k more duplicates in dive, than have been removed as of now.

The code to find the rest simply built a dictionary file with some random character lines. I started with john's password.lst, then did Rock-You top 50k (which was slow), but in the end, I found the best wordlist simply to be the random character lines, where i have lines of varying lengths (3 to 32), and 50 lines of each length.    The method, is I run each rule over every word. I then hash all the outputs which are different than the input word (simply add them using MD5_Update).  Then when done, do an MD5_Final to a buffer FOR THAT RULE.  Then move on to the next rule against exact same wordlist.   When complete, I simply do a O(n^2) braindead search for rules that have the same hash value (i.e. rules which produced EXACTLY the same end result).  This is not 100% safe, the duplicates DO have to be looked at by hand.  BUT there can be no other rules in the entire set which do produce duplicates. The list I have now is ALL of them.

But I am going to wait for this one to be merged (this one has been validated).  If you want, you can look at each commit. There is a comment of exactly which rules were found to be duplicated with which other rules, to see that everything WAS done properly.  There were 2 rules where my script errantly removed a rule. I put them back in by hand.  It had to do with 2 parts of the script fighting with each other (the sorting of the s commands, and the cycle/noop remove code of the s commands)

jfoug:
Here is the first 'taste' of the duplicates found by running data:

```
duplicate rules:
(0)  c
(1304)  T0tc
(1761)  uc
(2372)  lc
(85293)  Ct
(88696)  Ec
duplicate rules:
(1)  l
(10231)  cl
(18932)  l*B5*B5
(33011)  ut
(47389)  lsMF
(108387)  ll
duplicate rules:
(2)  u
(148)  uu
(1138)  cu
(14801)  Cu
(27600)  lu
(63926)  usa3
```

All of those are 100% duplicate.  There are some things that could be scripted for, but I think this will simply have to be looked at by hand, and then a script written to read in this log file and properly remove the lines from the rule file.

I knew there were duplicates like this (in all, I think dive will end up with 25% or more of it removed, and still have 100% identical finding capability).

However, before I proceed with this, we need to make a call on whether we should change all of the xNM commands to ONM   I think so.  Looking at the data, it really appears that when these rules were generated, that the logic of the ONM is what was wanted for these.  But I will leave that call up to you.  It does greatly impact the duplication finding logic. There are about 2500 fewer duplicates found of we switch from x to O, but most of all, I think there are about 5000 more rules that actually produce hits if we switch from x to O

That being the case, it is very likely that all rule files produced prior to when you made that switch should be changed. You might want to keep both flavors, in case there are people running older versions of HC, where the x still means omit.   But if the rules have not been updated since that logic change was done, then x almost certainly is wrong when run on the newer HC versions/

jsteube:
> However, before I proceed with this, we need to make a call on whether we should change all of the xNM commands to ONM I think so. 

I think changing the xNM to ONM is the right way as the dive.rule was created back when xNM what is today ONM. Go for it!

> You might want to keep both flavors, in case there are people running older versions of HC, where the x still means omit.

I'm not so much about supporting old rules. We don't need to care about that case.

jfoug:
> I think changing the xNM to ONM is the right way as the dive.rule was created back when xNM what is today ONM. Go for it!

I will do that as a stand alone commit, so it does not get 'confused' with anything else.

> I'm not so much about supporting old rules.

Ok, ONM it is.   A very easy change.  Were there other 'older' .rule files that would have this same issue?  Getting the rules working with proper logic will only help to improve their find rates.  If a rule was supposed to do the logic of Onm, but instead is now doing the extract logic, then it will match many fewer candidates. Actually, it becomes a sore point for duplicates, because often we strip the word down to just 1 or 2 bytes, instead of removing those 1 or 2 bytes.

PR #320, opened 2016-04-28
dive.rule: removed all the rest of the duplicate rules

jfoug:
This should be the remainder of the duplication


jsteube:
See Issue discussion

PR #321, opened 2016-04-28
Adding final code to test.pl

fgaudreault:
Missing blurb to complete the test.pl script.


jsteube:
I can see a print in there, guess it's some debugging print. Will remove it afterwards. Thanks!

PR #324, opened 2016-05-02
Remove clang extension

neheb:
It's only needed for using the static keyword. Since you removed those, clover builds it fine without it. Also removed the check since Mesa is considered IS_GENERIC now.


jsteube:
Thanks!

PR #325, opened 2016-05-04
dive.rule: duplicates found by running data have been removed.

jfoug:
/hc2/rules/dupe/z $ md5sum after b4
f1af3c6e84632dc7961f390a400e5376 *after
f1af3c6e84632dc7961f390a400e5376 *b4

This test was jtr password.lst and RockYou top 50k, run against dive.rule before the change (b4), and then run after this change.

The command was run using --stdout, then put through unique and then sort.  The resultant files (b4 and after), then compare 100% identical, as shown by md5sum

I will send you some data offlist if you want to look it it.  But I am pretty sure you would not be able to find any false duplicates in there.

Jim.


jfoug:
There are some rules in the final dive.rule that are 'almost' duplicates, but have been kept. They may have only produced a couple of unique words on the 55k input file run, but they were kept intact none the less.  I do not see us being able to remove much more than what is removed, without starting to have 'some' affect on the end generated words.  Right now, it is 100% the 'same' as where we started, only 20-25% fewer duplicate words generated.

This ruleset still does generated a fairly large number of duplicates, that is mostly due to not using any rejection code.  Running the rules against the input file I have done, produces about 70% or so duplicates.  But this is almost all where the word is left unchanged most of the time.  I am not 100% sure for oclHashcat on the usage of rejection logic.  I now in john for most any salted hash, getting the rejection logic right helps things out. It can cause the speed to be 'less', BUT the overall time required to complete the work is less.  For slower hashes, rejection logic is critical.  That is also with the helper logic in john, that does a uniq logic for generated passwords, and only uses them if the word being generated is different that the last one generated.  That helps, but does not fully eliminated duplicates.   The 70% duplicates is USING the john logic. I bet the percentage is quite a bit higher than that when running the rules through HC.

jsteube:
> I am not 100% sure for oclHashcat on the usage of rejection logic. 

For GPU, there's no rejection rules. That's mostly because of the traditional goal to crack fast hashes instead of slow ones. The rule engine was simply faster not to create branches for rejections rather than to check them.

Please continue to send PR for the other rules.

PR #328, opened 2016-05-07
d3adone.rule: removed duplicates.

jfoug:
Ready for merge. This is all the remaining duplicates removed.


jsteube:
That's something :)

jfoug:
It was a big difference waiting until the xNM were changed to ONM.  The differences were quite a few more before this, and they did not 'look right'.  But after the change, things lined up properly.  I did check the entire list, and all were easy to spot as being duplicates.

It looks like generated.rules (and generated2) also need to have the xNM to ONM logic done on them.  They also have quite a few dupes, that can be found using the tools I have.  I will get the x->O checked in first with only that change.  It makes it easier to later look over the duplicate finds, when you have only 1 thing to look at.  Helps to keep my mind focused.

PR #341, opened 2016-05-17
add version header and time to --status-automat

roycewilliams:
For tracking how, where and when automated output was generated.

```
# hashcat (v3.00-beta-29-g18f05e9+) Tue May 17 06:29:28 2016

1:1500:354400613
2:1500:154215960

```

First pull request!

### DIFF
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -5876,6 +5876,10 @@ int main (int argc, char **argv)
         log_info ("%s (%s) starting in benchmark-mode...", PROGNAME, VERSION_TAG);
         log_info ("");
       }
+      else
+      {
+        log_info ("# %s (%s) %s", PROGNAME, VERSION_TAG, ctime (&proc_start));
+      }
     }
     else if (restore == 1)
     {


jsteube:
I just removed this stuff with a previous commit: https://github.com/hashcat/oclHashcat/commit/c6e5ff2a68b21572d54c2577e415e74768465313

Why bring it back? I thought it's important to have the data output as plain as possible so that there's no need to cut away the header....

roycewilliams:
It's a compromise. Same reasoning that I used when I requested the commit hash in the verbose version output -- historical comparison value. A set of stats, even automated stats, is much more useful when you have the context. If people are working with automated stats over time, they will need to know how they were generated. Better IMO to have that as part of the new automated file "format", and pay the simple price of cutting away the header (which is very easy, because it's prepended with the comment character).

roycewilliams:
It's the same reason why the performance comparison Google Docs spreadsheet lists which hashcat versions are being compared. I would even suggest expanding the automated format to include the device type as well as the device ID (which could be stripped if not needed) but that's probably too ambitious. :)

PR #354, opened 2016-05-27
Change --status-automate to --machine-readable.

ghost:
The former didn't really illustrate the point of the command though the functionality is great to have.  Updated ASCII tables around options output.


jsteube:
If you rename --status-automat to --machine-readable you'd rename the variable (and macro) from status_automat to machine_readable, but not status_machine_readable. Please fix.

ghost:
I did this because on the code side, it's easier for one to see this has something to do with the status output, and that it is transforming it into machine readable code.

jsteube:
Thanks!

PR #364, opened 2016-06-05
Add cygwin support

neheb:
Doesn't actually build but "make win64" or "make win32" still works.

Actually, why is the operating system even checked in the makefile?

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -14,7 +14,7 @@ PROG_NAME                := hashcat
 
 UNAME                    := $(shell uname -s)
 
-ifeq (,$(filter $(UNAME),Linux Darwin))
+ifeq (,$(filter $(UNAME),Linux Darwin CYGWIN_NT-10.0))
 $(error "! Your Operating System ($(UNAME)) is not supported by $(PROG_NAME) Makefile")
 endif
 
--- a/src/Makefile
+++ b/src/Makefile
@@ -13,8 +13,9 @@ PROG_NAME                := hashcat
 ##
 
 UNAME                    := $(shell uname -s)
+UNAME			 := $(patsubst CYGWIN_NT-%,CYGWIN_NT-,$(UNAME))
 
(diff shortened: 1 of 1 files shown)


jsteube:
I'm confused. What is this doing and why. Are you saying you tried to build on windows with mingw and it worked?

neheb:
yeah. basically in cygwin you have "gcc" but you can also cross compile it with "mingw64". the difference is a cygwin build(does not currently work) uses cygwin.dll to emulate POSIX stuff while mingw does not.

philsmd:
I'm not yet convinced if we should really add it like this "CYGWIN_NT-10.0".

What about other windows versions where the uname changes (slightly), CYGWIN_NT-5.0, CYGWIN_NT-5.1, CYGWIN_NT-5.2, CYGWIN_NT-6.0, CYGWIN_NT-6.1, CYGWIN_NT-6.2, CYGWIN_NT-6.3, CYGWIN_NT-6.4, ... and all the WOW/WOW64 variants ?

I think it should be at least as flexible as CYGWIN_NT-6\* and CYGWIN_NT-10\* such that it accepts all windows versions from vista/windows7 up to windows 10.
There shouldn't be any reason to exclude windows 7, 8, 8.1... because if it compiles correctly on windows 10, it also should compile without errors on those older windows operating system versions.

Do you agree?

neheb:
yeah. i only did it like that since i have no idea how bash works. So can it be replaced with CYGWIN-NT-*?

philsmd:
I guess something like this would work (i.e. just remove the windows version number from the UNAME variable):

```
UNAME := $(shell uname -s)

UNAME := $(patsubst CYGWIN_NT-%,CYGWIN_NT-,$(UNAME))

ifeq (,$(filter $(UNAME),Linux Darwin CYGWIN_NT-))
...
```

neheb:
i tested it, seems to work.

jsteube:
Thanks!

PR #368, opened 2016-06-10
Removed obsolete entries from install: target.

hlein:
These files have been renamed+moved under OpenCL, and are already
installed from there elsewhere in the install: target.


jsteube:
Thanks!

PR #369, opened 2016-06-10
Added DESTDIR support

hlein:
This is a common thing for distros or build environments that want to install to a sandbox location as part of testing or package-building. It should be a harmless/no-op change when make is called without DESTDIR defined.  Also a related tweak to make installing binaries to a path that does not exist yet, work.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -36,6 +36,7 @@ endif
 ## Installation paths (Linux only)
 ##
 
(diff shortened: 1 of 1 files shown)


jsteube:
I always wondered what DESTDIR is used for, saw it couple of time. Makes sense, thanks!

hlein:
Sweet, thanks!

PR #379, opened 2016-06-19
Removed duplicate rule 'cl' & 'l'

mhasbini:
'cl' (line 1568) will have the same effect as 'l' (line 1322), so I removed 'cl'.


jsteube:
Thanks!

PR #394, opened 2016-06-29
Make linux-build POSIX compatible

ghost:
Compilation on musl-libc linux systems fails, as the termio.h header has been obsoleted and succeeded by the POSIX termios.h header.

Including termios.h and sys/ioctl.h as well as changing the struct names allows hashcat to build and run on gentoo with musl-libc as well as debian stable with glibc.

Please note that this is just a quick compilation fix, not sure if it introduces weird behaviour.

From glibc's termio.h:
/\* Compatible <termio.h> for old `struct termio' ioctl interface.
   This is obsolete; use the POSIX.1`struct termios' interface
   defined in <termios.h> instead.  */

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -41,7 +41,8 @@
 #include <limits.h>
 
 #ifdef LINUX
-#include <termio.h>
+#include <termios.h>
+#include <sys/ioctl.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


jsteube:
This is cool, many thanks!

PR #401, opened 2016-06-30
test: fixed verify for -m 13100, -m 13200, -m 13300 and -m 13400

philsmd:
First of all, I'm sorry that this is a fix of an amalgamation of different problems. But I felt that the problems here are way too similar to justify different pull requests.

(Note: this pull request only affects 1 single file: test.pl and mostly only the "verify" invocation.)

Indeed, the common problem here for these algorithms:
1. -m 13100 = Kerberos 5 TGS-REP etype 23 
2. -m 13200 = AxCrypt
3. -m 13300 = AxCrypt in memory SHA1
4. -m 13400 =Keepass 1/2 (with and without keyfile)

was basically that verify did not do the reverse step (e.g. while the "hashing" task (passthrough or single) was encrypting, verify has to decrypt the encrypted data etc).

In addition to that this commit includes some small fomatting fixes.

Thank you very much.

### DIFF
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -2211,7 +2211,7 @@ sub verify
 
       next unless (exists ($db->{$hash_line}) and (! defined ($db->{$hash_line})));
     }
-    elsif ($mode == 13100 )
+    elsif ($mode == 13100)
     {
       ($hash_in, $word) = split ":", $line;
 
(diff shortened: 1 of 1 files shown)


jsteube:
Great work! Thanks! @Fist0urs Please check this out

Fist0urs:
@philsmd told me about that. Actually I knew there were some problems (not all them looking at the commits though) but I did not have time to fix stuff :/

I'll take a closer look for the next algorithm, sorry about that

Great work!

PR #407, opened 2016-07-01
Fix makefile on OSX

jdek:
First commit just defaults to GNU sed as opposed to the built-in sed which doesn't have the features which are being used.
Second commit fixes the file extension. The .app file extension is strictly for app folder bundles which consist of more than a binary, and aren't actually files. OSX behaves the same as linux in regards to executables on the command line, so the .app is not needed.


jsteube:
OK, cool. Thanks for the PR and the explanation

PR #409, opened 2016-07-01
hashcat.c: change osx to OSX use shared_dir

### LINKED ISSUE #408: hashcat looking in binary folder for resources
Compiling and installing master from source (using `make install`) and then trying to run ends up with:

```
$ hashcat -m 100 -a 3 -1 abcdef ~/hashes.txt "?1?1?1?1?1?1?1?1?1?1"
hashcat (v3.00-16-g86cf95e) starting...



/usr/local/bin/hashcat.hctune: No such file or directory
```

jdek:
This fixes #408. 

_Note: I originally created the issue because I thought I didn't have enough time to look into a fix, but turns out I needed a fix enough to make time._

### DIFF
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -6171,7 +6171,7 @@ int main (int argc, char **argv)
 
   char *exec_path = get_exec_path ();
 
-  #ifdef LINUX
+  #if defined(LINUX) || defined(__APPLE__)
 
   char *resolved_install_folder = realpath (INSTALL_FOLDER, NULL);
   char *resolved_exec_path      = realpath (exec_path, NULL);


jsteube:
Two things to mention here:
- I never expected this install system to be run on OSX. There could be problems
- The macro _APPLE_ is kind of official? Reason is we're currently using DARWIN on the most spots. What the correct/better form?

magnumripper:
On my machine (latest OSX) both the native and Homebrew's gcc define `__APPLE__` by default, but not `__DARWIN__`

```
$ gcc -dM -E -x c /dev/null | egrep "(APPLE|DARWIN)"
#define __APPLE_CC__ 1
#define __APPLE__ 1
$ /usr/bin/gcc -dM -E -x c /dev/null | egrep "(APPLE|DARWIN)"
#define __APPLE_CC__ 6000
#define __APPLE__ 1
```

jsteube:
Excellent. I think we should switch to that. /cc @unix-ninja

unix-ninja:
Sure. If we wanted to remove the custom declaration then I am fine with using **APPLE**

unix-ninja:
That was supposed to be 

```
__APPLE__
```

XeniaGabriela:
Dear hashcat team,

I just downloaded the latest version and installed it in an Ubuntu 14.04 virtual machine on Windows7 by 
1. unzipping it to ~/.hashcat with p7zip
2. copying hashcat64.bin to /usr/bin/
3. making a symblic link hashcat to hashcat64.bin

'hashcat --help' works fine ...
But I am getting the ERROR '/usr/bin/hashcat.hctune: No such file or directory' when I run the program.

Did I do something wrong with the installation?

Regards,

Xenia

PR #413, opened 2016-07-04
special case handling for commas within masks of .hcmask file

philsmd:
There was a slight problem with how masks within a .hcmask file were parsed when they contained a comma, an example:

> ?l?d,?1?1?1?1?a?a\,?a?a

This, according to the docs, would mean that the "\," will be interpreted as a single "," (note: we need to use "\," because otherwise without the slash everything before the comma would be interpreted as --custom-charset2).

This means, that the final mask read from .hcmask files need to be parsed and every "\," needs to be replaced by a single "," (in addition to that, we need to allow the special case "\\," which means a literal "\" followed by a ",").

The commit attached does implement exactly this very special case handling.

Thank you very much

### DIFF
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -17499,6 +17499,36 @@ int main (int argc, char **argv)
 
             mask = mask + str_pos + 1;
           }
+
+          /**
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks!

PR #415, opened 2016-07-04
extend global CFLAGS and LDFLAGS to aid distro packaging

anthraxx:
This preserves globally defined CFLAGS and LDFLAGS and simply
extends those variables to aid distro based packaging toolchains
and predefined distro wide defaults like SSP, relro etc.


jsteube:
How could we have missed that, thanks!

PR #417, opened 2016-07-05
this fixes #400: --loopback was not working

### LINKED ISSUE #400: --loopback not working
Induct directories get filled with found entries but aren't used and deleted afterwards.

philsmd:
There was a new problem with --loopback (mentioned in issue #400), which made the --loopback command line option do nothing.

The fix I propose here is to allow --loopback (and the induction folder in general) work again by just allowing additional inner loop run (variable dictpos).

Of course, this is a little hack. But the comment I added and the otherwise very complicated way to properly fix it, should be enough to justify a "little hack" in this particular case.

Thank you very much


jsteube:
Thanks for fix!

PR #419, opened 2016-07-06
fixed problem where hashcat did not remove WPA/WPA2 hashes found in potfile

philsmd:
The removing of hashes found in potfiles was not working for hash mode WPA/WPA2 because the comparison of the hashes/salts within the potfile versus the ones in the current hash list was not correct.

The comparison did compare too many bytes and therefore failed.

Thank you very much


jsteube:
Nice find

PR #421, opened 2016-07-06
FreeBSD support

lifanov:
This adds FreeBSD platform support to the new CPU+GPU hashcat.
Changes are restricted to FreeBSD fixes only, not general improvements.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -52,6 +52,11 @@
 #include <mach/mach.h>
 #endif
 
+#ifdef __FreeBSD__
+#include <termios.h>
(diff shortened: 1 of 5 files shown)


jsteube:
This is great, thanks! Too bad travis doesn't support FreeBSD, so I can't check if it's working.

lifanov:
No problem! It works, according to me.

lifanov:
FreeBSD port for new hashcat landed. I renamed the previous port to hashcat-legacy as per upstream, so both are available.

Issue #425, opened 2016-07-07
Can't build with MinGW anymore


neheb:
make win64 errors with

make: **\* No rule to make target '/usr/x86_64-w64-mingw32/lib/CRT_glob.o', needed by 'hashcat64.exe'.  Stop.

it does the same on both cygwin and fedora 24.

edit: the fix was to change CRT_GLOB_64 to /usr/x86_64-w64-mingw32/sys-root/mingw/lib/CRT_glob.o

jsteube:
Yeah I was going to say you maybe just need to adjust the path. Problem is I don't know if it's possible to not use a path and let MinGW find the object itself.

philsmd:
I think the correct path could be programmatically found (w/ the Makefile) like this:

linux - 32 bit cross-compilation:
`$ i686-w64-mingw32-ld --verbose | grep -m 1 SEARCH_DIR | sed 's/.*"\([^"]*\)".*/\1/'                                                                         
/usr/i686-w64-mingw32/lib`

linux - 64 bit cross-compilation:
`$ x86_64-w64-mingw32-ld --verbose | grep -m 1 SEARCH_DIR | sed 's/.*"\([^"]*\)".*/\1/'
/usr/x86_64-w64-mingw32/lib`

It should work the same way under mingw too, we need just someone to test this. @neheb, can you try this ? The "ld --verbose ..." command should output the correct lib path even under mingw @ windows

btw: the grep ... | sed ... could also be replaced by:
`grep -Po -m 1 'SEARCH_DIR\("\K.*?(?=")'`
but I'm not sure if mingw (and all other supported operating systems used to build from source) supports this

neheb:
both commands return /usr/lib on fedora 24. not sure on cygwin. replacing the grep | sed commands with the last one results in:

=/usr/i686-w64-mingw32/lib
=/usr/local/lib
=/lib
=/usr/lib

and

=/usr/x86_64-w64-mingw32/lib
=/usr/local/lib
=/lib
=/usr/lib

respectively.

philsmd:
Are you sure that you are trying with the correct "mingw32-ld" command (i686-w64-mingw32-ld or x86_64-w64-mingw32-ld) ?
Using a different ld version (e.g. not the mingw32 provided one), also gives me different results. But this is ofc to be expected. We are trying to get the correct path from mingw32-ld after all.

neheb:
i am

philsmd:
maybe something more flexible would be this:
[...]

PR #433, opened 2016-07-14
fix for #425: make check for CRT_glob.o more flexible and show warnings in case of failure

philsmd:
This commit should fix the problem mentioned in #425.

The changes were successfully tested by the OP and seem to work flawlessly.

The main changes are that we make the CRT_glob.o search/check more flexible, only do them if we build windows binaries and show a warning in case the globbing can't be enabled.

Thank you very much

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -31,7 +31,9 @@ endif
 
 MAKEFLAGS                += -l -j -rR --no-print-directory
 
-ifneq ($(findstring clean,$(MAKECMDGOALS)),)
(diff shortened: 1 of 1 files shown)


jsteube:
It took me a while to make the Makefile as clean and easy to read as it is now, but with those changes it becomes hard to read again. Can we move that code section out to an external file and include it from within the Makefile?

philsmd:
k, I will close this pull request for now ... and maybe reopen a new one when I have a (better) solution.
Thanks

Issue #129, opened 2016-01-14
Request for basic DES-ECB cracking


mattames:
Hi Guys,

Can you add a basic DES-ECB cracker? Since Hashcat contains several algorithms (-m 1500 and 3000) that use DES-ECB as a core component, I figure most of the work is done already. This mode would simply take a known 64 bit DES input and output plaintext/ciphertext pair and then brute force the key used.

This would be greatly advantageous to myself as part of some other work I am doing on another open source project, and it would enable some cool comparisons between GPU and FPGA solutions in terms of keys per second/cost/complexity/power consumption.

jsteube:
For me this sounds like a bad idea to compare FPGA/CPU with GPU using DES since DES is so extremely FPGA/CPU friendly. A comparison based on MD5 makes much more sense since this algorithms is friendly for both platforms.

Anyway I think such an algorithm can make sense for known-plaintext based attacks since so many crypto schemes still base on raw DES but it should be mode-independant. You can anyway simply reverse the ECB from it before putting it to oclHashcat.

mattames:
Yes, once the plaintext/ciphertext pair is extracted (easy to do) they can then be fed into oclHashcat. FPGA implementations are hard to come by, require experience with FPGAs, expensive software tools to be able to create the bitstreams on the larger devices ($5000 for one vendor), and can be time consuming to create an image and troubleshoot it.

By comparison, oclHashcat on GPU is fast, off the shelf and if you collaborate with a few friends with GPUs they can 'lend' you their spare cycles and it would be feasible to split the keyspace up into slices for a zero-cost solution using the GPUs that people already own...

STUCK12:
Seconded. Is there any public software that can actually support 64-bit DES (CPU or GPU based)?

mattames:
Hi Guys,

Is there any chance someone might be able to implement this one? 

Cheers,
Matt

roycewilliams:
https://github.com/hashcat/hashcat/commit/71a8459d851d246945343ea59effa1d46b965bf8

PR #452, opened 2016-08-02
Added functionality to crack DES-ECB for known-plaintext. The "hashty…

fransla:
…pe" is 1510, and the format to crack is:

plaintext:ciphertext, where plaintext is a single block of plaintext, and ciphertext is the output of standard
DES encryption of the plaintext in ECB mode, no padding, in hex.
For example, encrypting the plaintext PlainTXT with password S3cr3ts1 will result in the ciphertext
a1ff88c804540b1f. The format for the input file should be:
PlainTXT:a1ff88c804540b1f
And cracking it with:
hashcat -m 1510 -a 3 S?dcr?ats?d might result in the following:
PlainTXT:a1ff88c804540b1f:S3cr2ts0
Note that DES only uses the MSB 7 bits out of each 8 of the password, so the resultant password might not quite match
the orginal used, but will still decrypt the ciphertext.
OpenSSL command used to generate the above example:
echo -n "PlainTXT" | openssl enc -des-ecb -K "5333637233747331" -nopad | xxd

As per https://github.com/hashcat/hashcat/issues/129

### DIFF
--- a/OpenCL/m01510_a0.cl
+++ b/OpenCL/m01510_a0.cl
@@ -1,0 +1,741 @@
+/**
+ * Authors.....: Jens Steube <jens.steube@gmail.com>
+ *               Gabriele Gristina <matrix@hashcat.net>
+ *               magnum <john.magnum@hushmail.com>
+ *               Frans Lategan <frans.lategan+hashcat@gmail.com>
+ *
+ * License.....: MIT
+ */
+
+#define _DES_
+
+#define NEW_SIMD_CODE
+
+#include "inc_vendor.cl"
+#include "inc_hash_constants.h"
+#include "inc_hash_functions.cl"
+#include "inc_types.cl"
+#include "inc_common.cl"
+#include "inc_rp.h"
+#include "inc_rp.cl"
+#include "inc_simd.cl"
+
+#define PERM_OP(a,b,tt,n,m) \
+{                           \
+  tt = a >> n;              \
+  tt = tt ^ b;              \
+  tt = tt & m;              \
+  b = b ^ tt;               \
+  tt = tt << n;             \
+  a = a ^ tt;               \
+}
+
+#define HPERM_OP(a,tt,n,m)  \
+{                           \
+  tt = a << (16 + n);       \
+  tt = tt ^ a;              \
+  tt = tt & m;              \
+  a  = a ^ tt;              \
+  tt = tt >> (16 + n);      \
+  a  = a ^ tt;              \
+}
+
+#define IP(l,r,tt)                     \
+{                                      \
+  PERM_OP (r, l, tt,  4, 0x0f0f0f0f);  \
+  PERM_OP (l, r, tt, 16, 0x0000ffff);  \
+  PERM_OP (r, l, tt,  2, 0x33333333);  \
+  PERM_OP (l, r, tt,  8, 0x00ff00ff);  \
+  PERM_OP (r, l, tt,  1, 0x55555555);  \
+}
+
+#define FP(l,r,tt)                     \
+{                                      \
+  PERM_OP (l, r, tt,  1, 0x55555555);  \
+  PERM_OP (r, l, tt,  8, 0x00ff00ff);  \
+  PERM_OP (l, r, tt,  2, 0x33333333);  \
+  PERM_OP (r, l, tt, 16, 0x0000ffff);  \
+  PERM_OP (l, r, tt,  4, 0x0f0f0f0f);  \
+}
+
+__constant u32 c_SPtrans[8][64] =
+{
+  {
+    /* nibble 0 */
+    0x02080800, 0x00080000, 0x02000002, 0x02080802,
+    0x02000000, 0x00080802, 0x00080002, 0x02000002,
+    0x00080802, 0x02080800, 0x02080000, 0x00000802,
+    0x02000802, 0x02000000, 0x00000000, 0x00080002,
+    0x00080000, 0x00000002, 0x02000800, 0x00080800,
+    0x02080802, 0x02080000, 0x00000802, 0x02000800,
+    0x00000002, 0x00000800, 0x00080800, 0x02080002,
+    0x00000800, 0x02000802, 0x02080002, 0x00000000,
+    0x00000000, 0x02080802, 0x02000800, 0x00080002,
+    0x02080800, 0x00080000, 0x00000802, 0x02000800,
+    0x02080002, 0x00000800, 0x00080800, 0x02000002,
+    0x00080802, 0x00000002, 0x02000002, 0x02080000,
+    0x02080802, 0x00080800, 0x02080000, 0x02000802,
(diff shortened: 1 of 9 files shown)


halexan:
How about the benchmark of DES-ECB on your own computer? @fransla 
I want to know the performance of GPU's DES-ECB and compare it with FPGA.

fransla:
./hashcat -m 1510 -b -D 1,2 -w 4
hashcat (v3.00-70-gdbe33d8) starting in benchmark-mode...

# OpenCL Platform #1: Apple
- Device #1: Intel(R) Core(TM) i5-4690 CPU @ 3.50GHz, 2048/8192 MB allocatable, 4MCU
- Device #2: AMD Radeon R9 M290X Compute Engine, 512/2048 MB allocatable, 20MCU

Hashtype: DES

Speed.Dev.#1.: 33435.1 kH/s (125.36ms)
Speed.Dev.#2.:   743.5 MH/s (477.53ms)
Speed.Dev.#*.:   777.0 MH/s

Started: Tue Aug  2 16:48:51 2016
Stopped: Tue Aug  2 16:48:57 2016

halexan:
A german company http://www.sciengines.com/which are doing some cracking with FPGA

This is their benchmark http://prntscr.com/bm5net

Maybe GPU really doesn't good at cracking DES.

But I think cracking DES-ECB should be added to hashcat,

jsteube:
That's nice. Can you please create dedicated hash-modes for them, too? Please use 1510 -> 14000 and 1530 -> 14100. 

Note that I think you can do bitsliced DES for -a 3 kernel, which will increase the speed by a very large factor. In theory, benchmark speed for your DES kernel should match the LM kernel.

Also please add test.pl and test.sh target

halexan:
As we know, DES only uses the MSB 7 bits out of each 8 of the password, so the key length is just 56 bits, the other 8 bits is useless. Hashcat don't need to consider the 8 bits and just set the password "56 bits hex" (like "1A8D938F83920D").

Using char password (like "S3cr3ts1") is not good, because different char password may equal to same key (like "S3cr2ts0" = "S3cr3ts1"), this may cause redundant calculation, so I think it's better to use hex password.

jsteube:
@fransla please respond

fransla:
Hi Jens,
Sorry for the slow response, I am a bit busy at work but will make most of the changes. The bitslicing will probably not be done yet though

Frans

jsteube:
Hey Frans, no worries, I'll add bitsliced stuff afterwards.

mattames:
Also, since it would be great to run many instances of this across several different PCs (I have access to a few at work with beefy GPUs) why not add a key start range..

For example, one PC can start at key 0x00001000000000, the next could be told to start searching at say 0x000F0000000000 and so on...

Since password to DES input key is application specific, it really doesn't relate to DES-ECB at all, so as halexan commented, it would be best to just work with the raw DES keys in 56-bit hex format.

Thanks so much for this addition, it is really great to see this implemented!!!

jsteube:
@mattames that's already supported since a long time. see here: https://hashcat.net/forum/thread-3047.html

mattames:
@jsteube - my apologies, I should have looked harder. Thanks for the info!!

Cheers,
Matt.

halexan:
Thank you @jsteube ! Hastcat is a really huge software, it has so many functions, it's hard for everyone to master it.

This thread (https://hashcat.net/forum/thread-3047.html) does not support reply. Can I ask a question here?

If DES-ECB key is in 0x00000000000000-0x00000000000FFF, is the following command right?

``` xml
./hashcat.bin DES-ECB.txt -m 14000 -a 3 ?b?b?b?b?b?b?b?b?b?b?b?b?b?b -s 0 -l 4096
```

jsteube:
@halexan No, hashcat abstracts this for you. Please the read forum post again.

jsteube:
I didn't want to wait anymore and rewrote it myself, fixed some smaller bugs, added unit-tests and added bitsliced kernel which makes it around 10 times faster. Here's the full commit: https://github.com/hashcat/hashcat/commit/71a8459d851d246945343ea59effa1d46b965bf8
Thanks for your help

mattames:
Awesome work jsteube!! Can you give an indication of how fast the bitsliced DES-ECB kernel runs on your machine?

Thanks again!

jsteube:
Yes, a friends 6x1080 machine is doing 120GK/s. IOW: Cracks every DES key in 80 hours average

jark22:
Can anyone show an example that works with DES-ECB?  I was using the following command using that custom character set but I could be doing it wrong.

 hashcat -m 14000 --custom-charset1=charsets/DES_full.charset -a 3 0123456789abcdef:8a5ae1f81ab8f2dd     ?1?1?1?1?1?1?1?1  

where the key was all 1's.  It takes a couple of minutes to run but obviously does not run through the whole keyspace and fails to find a solution

jsteube:
I think you just forgot to add --hex-charset

Forts117:
Hi guys... great work you are doing here with this!  I'm hoping someone can point me in the right direction though.  I've run multiple sample numbers and I never seem to find the key.  For example jark22s post above, I don't find the key.  I've also used DA93B01EF7F98852:1B5B3EDF6511A554 which should be key: 0101010101012626 with no success (these numbers were confirmed as good with this tool - http://des.online-domain-tools.com/).  

Here is the command line I'm using: 
hashcat64.exe -m 14000 hashes.txt -o cracked.txt -a 3 -1 charsets/DES_full.charset --hex-charset ?1?1?1?1?1?1?1?1 -w 3

Does anyone have any suggestions as to what I might be doing wrong?  Or have a known working set of numbers I can try?

Many many thanks!

fransla:
You should put the CipherText first, then the known PlainText, so try 1B5B3EDF6511A554:DA93B01EF7F98852

Forts117:
That's where it gets confusing, the very first message in this thread says PT:CT, yet I've seen other comments like yours suggesting CT:PT.  Regardless, I've tried both ways... No joy. Does it work for you? (If you happen to have a minute to give those numbers a try that is....)

Thanks!!

fransla:
jsteube changed the order around when merging back into hashcat; the pull request had PT:CT, but this code now requires CT:PT. The bitslicing for -a 3 for DES might affect your results, so try using your example as a single key 3DES, which is equivalent, so this works for me:
hashcat -m 14100 1b5b3edf6511a554:da93b01ef7f98852 -a 3 -1 ./charsets/DES_full.charset --hex-charset 01010101010126?101010101010126?1010101010101?1?1
I am not quite sure why it seems to give trouble in -m 14000 -a 3

bigfella237:
Even if this 3DES work-around works as above for a known key, that does not help when searching for an unknown key since we have nothing with which to populate the first two variables?

bigfella237:
@fransla can you confirm that -m 14000 does not find the test key with the verified CT:PT?

Is this a bug in the DES code and if so, how do we report it?

fransla:
I can confirm that -m 14000 did not work for me

PR #459, opened 2016-08-17
some kernels used COMPARE_M_SIMD instead of COMPARE_S_SIMD in singlehash mode

hops:
The following kernels used COMPARE_M_SIMD in singlehash mode:
m00900_a1, m01710_a0, m06900_a3, m13300_a0, m13300_a1, m13300_a3


jsteube:
Nice find actually, thanks! I think it's safe that there's no opposite case for this, like having COMPARE_S_SIMD in the multihash kernel, because this would require to have the search[] array declared, so it would cause a compile error.

PR #460, opened 2016-08-18
new feature: allow -o to be used in stdout mode

philsmd:
This is basically a new feature request and diff that implements the following:
when using --stdout together with --outfile nothing happened so far, i.e. the -o option was just ignored.

I found this a little bit disturbing, either it should allow it or it should give an error (we can decide whatever is best for hashcat, but ignoring it is not a good option IMO).

Of course hashcat didn't support it so far, therefore this is a _new feature_ and not really a bug fix. But still I think we should either use this variant (i.e. supporting -o in --stdout) or be very strict and give an error ("-o is not allowed in stdout mode").

This PR implements the option using the outfile. Feel free to argument against or for it.
Thank you so much.
~ Phil


jsteube:
It's good, -o should be used incase --stdout is used, also saves the redirect

PR #462, opened 2016-08-18
add sanity check to disallow --show and --outfile-autohex-disable

philsmd:
These few lines of code change prohibit the user of using --outfile-autohex-disable together with --show (or better said ... if the hashcat user specifies --outfile-autohex-disable with --show he will now see an error).

The best fix would be to allow outfile-autohex-disable also within --show, meaning that the pot file output would be converted and "unhexed" if it is a $HEX[xx...] converted string.

Since we currently do not manipulate the output of the .pot file at all when using --show, --left ... this would be a major change and maybe not worth the time/effort. (at least I don't know of any user requesting this in the past).

Therefore, I would suggest that (like this commit is trying to accomplish) it's best to filter such command lines and disallow --show together with --outfile-autohex-disable for the time being.

Thank you very much indeed


jsteube:
We should setup an issue for that otherwise it's forgotten

Issue #457, opened 2016-08-13
Display Warning or Error when using --skip together with a .hcmask file


franky436:
When using --skip together with a (multi-line) .hcmask file, hashcat printed this error:

"ERROR: Restore value greater keyspace"

which is quite irritating/misleading. Possibly happens only if skip is higher than the keyspace of the first mask in the .hcmask file, haven't checked sources for that yet.

Suggestion: check if --skip was used together with a .hcmask file and print a warning/error that this combination is not supported.

jsteube:
Yes, that check sounds useful

PR #463, opened 2016-08-18
sanity check for #457: do not allow mask files or increment w/ --skip/--limit

philsmd:
This should fix the issue mentioned in #457.

The problem here is that --skip and --limit are not supported with increment and mask files and therefore an error should be shown.

The current version applies skip and limit to the first mask and therefore some strange behavior or error messages could appear.

I suggest we prohibit the use of --skip and/or --limit altogether for the time being - together with mask file and/or --increment - (until we support --skip and/or --limit in those very special cases, if we decide to support it at all in the future).

Thank you very much
~ phil


jsteube:
We should close that issue then. Thanks!

PR #464, opened 2016-08-18
fixes #425: flexible way of searching for CRT_glob.o on linux and cygwin

### LINKED ISSUE #425: Can't build with MinGW anymore

philsmd:
This should fix the problem mentioned in issue #425.

The problem was that CRT_glob.o was not successfully found on cygwin/windows machines (if you tried to build the source code within the cygwin shell).

This adds a new file to do the searching in case it detects that windows binaries need to be compiled.

Thank you very much


jsteube:
Very good, thanks!! I'll close the #425 Issue as well

PR #482, opened 2016-08-31
Fix some more warnings.

neheb:
clang's -Weverything seems to catch a lot of stuff. Here's some of it.

### DIFF
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -769,7 +769,7 @@ static double get_avg_exec_time (hc_device_param_t *device_param, const int last
   {
     double exec_ms = device_param->exec_ms[(exec_pos + i) % EXEC_CACHE];
 
-    if (exec_ms)
+    if (exec_ms > 0)
     {
       exec_ms_sum += exec_ms;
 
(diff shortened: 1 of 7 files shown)


src/hashcat.c
```diff
@@ -2895,7 +2896,7 @@ static int run_kernel (const uint kern_run, hc_device_param_t *device_param, con
     {
       if (data.opti_type & OPTI_TYPE_SLOW_HASH_SIMD)
       {
-        num_elements = CEIL ((float) num_elements / device_param->vector_width);
+        num_elements = CEIL (num_elements / device_param->vector_width);
```
magnumripper, line 2899: Are you sure about this?

neheb: Briefly looking at it, both variables seem to be uints.

magnumripper: Hence the need for the typecast you removed. I'm not saying you are wrong, I'm just asking if you actually understand what you are doing here.

neheb: Not really. I added a printf to see if anything changed after I removed the cast. Numbers were still the same.

The only issue I can potentially see is that dividing could cause a remainder to get thrown away. num_elements seems to end up being a multiple of 8 though.

jsteube: should be fine, thanks for cross check @magnumripper

PR #485, opened 2016-09-04
Cmake build

KOLANICH:


jsteube:
I'm trying to process the single commits historically the same order you did and comment each commit how I handled them.

https://github.com/hashcat/hashcat/pull/485/commits/d288608b85b4ded6010128c7ed40e9532a3ccc02

This commit is unacceptable for multiple reason:

I just hope it's really only about formating to that I'm able to merge other changes i like
- OpenCL/inc_hash_constants.h was changed in the meantime from myself into enum types
- Removed comments, no need to reformat
- "int log_out_nn (...)" is the correct format, "int log_out_nn(...)" is wrong, no need to change those
- Code like this is easier to read if it's correctly aligned, no need to change those:

```
     temp = (te2[(temp >> 16) & 0xff] & 0xff000000)
          ^ (te3[(temp >>  8) & 0xff] & 0x00ff0000)
```

```
     temp = (te2[(temp >> 16) & 0xff] & 0xff000000)
      ^ (te3[(temp >>  8) & 0xff] & 0x00ff0000)
```
- Same here, code like this is easier to read if it's correctly aligned, no need to change those:

```
       td0[te1[(rdk[j + 1] >> 24) & 0xff] & 0xff] ^
       td1[te1[(rdk[j + 1] >> 16) & 0xff] & 0xff] ^
       td2[te1[(rdk[j + 1] >>  8) & 0xff] & 0xff] ^
       td3[te1[(rdk[j + 1] >>  0) & 0xff] & 0xff];
```

```
       td0[te1[(rdk[j + 1] >> 24) & 0xff] & 0xff] ^
       td1[te1[(rdk[j + 1] >> 16) & 0xff] & 0xff] ^
       td2[te1[(rdk[j + 1] >> 8) & 0xff] & 0xff] ^
       td3[te1[(rdk[j + 1] >> 0) & 0xff] & 0xff];
```
- This change looks like a bug in your automated replacement:

```
-  char *block_ptr = (char *) block;
-  char *plain_ptr = (char *) plain;
```

```
+  char *block_ptr = (char *)block;
+  char *plain_ptr = (char *)plain;
```
- Not sure why people think #ifdef should not be aligned, it just makes it hard to read, will not change:

```
-  #if defined (_WIN32) || defined (_WIN64)
```

```
+#if defined (_WIN32) || defined (_WIN64)
```
- Those are copies from ADK SDK, we better don't do changes if we want to maintain them over time:

```
-  ADLODPerformanceLevel aLevels [2];
```

```
+  ADLODPerformanceLevel aLevels[2];
```
- Same goes for NVAPI and NVML
- This just looks ugly:

```
-static inline void CPU_ZERO  (cpu_set_t *cs)          { cs->count = 0; }
-static inline void CPU_SET   (int num, cpu_set_t *cs) { cs->count |= (1 << num); }
-static inline int  CPU_ISSET (int num, cpu_set_t *cs) { return (cs->count & (1 << num)); }
```

```
+static inline void CPU_ZERO(cpu_set_t *cs) { cs->count = 0; }
+static inline void CPU_SET(int num, cpu_set_t *cs) { cs->count |= (1 << num); }
+static inline int  CPU_ISSET(int num, cpu_set_t *cs) { return (cs->count & (1 << num)); }
```

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/d26eef45ab95b414ff3bff19199f871ba4ea666c

This commit simply creates a folder "hwmon" and moves the code from the main src folder into that (same for include).

It's not like that would create any better view on the folder, because in the main src folder is 10 files anyway:

```
-rw-r--r-- 1 root root  18024 Sep  5 10:10 ext_ADL.c
-rw-r--r-- 1 root root   5626 Sep  5 10:10 ext_nvapi.c
-rw-r--r-- 1 root root  13041 Sep  5 10:10 ext_nvml.c
-rw-r--r-- 1 root root  13650 Sep  4 20:11 ext_OpenCL.c
-rw-r--r-- 1 root root   4777 Sep  5 10:10 ext_xnvctrl.c
-rw-r--r-- 1 root root 695854 Sep  4 21:53 hashcat.c
-rw-r--r-- 1 root root  11151 Sep  4 20:11 Makefile
-rw-r--r-- 1 root root  96535 Sep  4 20:11 rp_kernel_on_cpu.c
-rw-r--r-- 1 root root 602378 Sep  4 21:53 shared.c
-rw-r--r-- 1 root root   2015 Sep  4 20:11 win_file_globbing.mk
```

It's not like this creates alot of confusion. OTOH, it creates lots of problems with the Makefile. We'd need to rewrite it in a way so that it is able to handle the new folder structure. 

I don't see the benefit yet. Maybe this changes when I continue going through the other commits.

KOLANICH:
1 I used the tool in VS to reindent the code. Reindentation was needed to be able to reorder commits. It crashed the handmade alignment and reindented using own standards. I (or you) can filter the branch with AStyle, which can solve some issues, will it be enough?
2

> It's not like that would create any better view on the folder, because in the main src folder is 10 files anyway

The count is increased in later commits.
3

> It's not like this creates alot of confusion. OTOH, it creates lots of problems with the Makefile. We'd need to rewrite it in a way so that it is able to handle the new folder structure. 

The makefile is to be replaced with cmake script, which uses globbing to get list of all c and h files in the project.

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/58b728230b666cabf7f5dee445dc3ece34e1068e

I merged this, or better I took the idea and applied it to the oder cpu-*.c files as well. While doing this I find there's a lot of duplicated macros if you compare it with  OpenCL/inc_hash_functions.cl and  OpenCL/inc_hash_constants.cl we maybe want to change that at a later time.

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/75a3ab9d47bd7d929e08f1f93de4399909743ef7

I merged this, or better I took the idea and applied it to constants that apply to this. I'm sure there will be more files that can be improved like this, whenever I see them I will fix them accordingly.

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/4323b4a70ad879b1a441449ff96143701647b087

I merged this, or better I took the idea and applied it to constants that apply to this. I'm sure there will be more files that can be improved like this, whenever I see them I will fix them accordingly.

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/30df3c66c2d14b17ba314c71aa527b592f873145

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/555675f955eeef601ebed27399ca7cebf99b0b75

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/11f58ee7959b84ecbcbfed0c7d1a28ee3390eae4

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/5481c882931dd2e967ac81d693dbcbd3d3384a72

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/2351e8ec8f6d0872c1e1183d3bd5352e3c05e5da

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/f440897d442006705f9b28cc9f30fa6b03fc5168

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/7d4ab2ff331f78bab129e146de04b8d040a2e482

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/2a7723bab3af5ada61a2b726e2283d620cf470ec

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/8fc2437e951535c5a00a7ddc0783acc291fcf139

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/96ea03db72de1f1022de3ae11e0952ad8f9d83e6

Partially merged:
- Fixed unused typedefs for windows "uint" and "uint64_t" (double)

To fully merge, please explain:
- What is _BASETSD_H_ ?
- Is _BASETSD_H_ a pre-existing macro from vc or set from Makefile?

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/a55052cc66029a3a5ef76ae4bc1a216f812464f2

merged

KOLANICH:
> What is BASETSD_H ?
> Is BASETSD_H a pre-existing macro from vc or set from Makefile?

A macro defined in a header supplied with compiler where the types you define are defined. Don't know if it exists in gcc and clang, but in VS it is.

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/96ea03db72de1f1022de3ae11e0952ad8f9d83e6

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/135555f4572e4d6def1150eaf77800b665d7ed63

Can not merge. 

Datatype switch to static is not possible because variable is used from different object via extern

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/d57d9245dbaaa2dca9135fde1a674548fb36b52e

merged

jsteube:
https://github.com/hashcat/hashcat/pull/485/commits/1aafa13e09973568ae58e4a539ce1d7772bd7e0b

merged, even created a new file for that, but since I'm this as an object, we can no longer inline it.

KOLANICH:
> we can no longer inline it.

That's sad. There are dedicated instructions for this in cpus, so those header is just compiler abstraction layer, they are really meant to be inlined. In later commit I've implemented them as compiler intrinsics (which means their calls would be converted into a single instruction, and they really are, as disasm says).

jsteube:
> There are dedicated instructions for this in cpus, so those functions are just compiler abstraction layer, they are really meant to be inlined.
> The use of them is not time intensive, so it's fine

KOLANICH:
> The use of them is not time intensive, so it's fine

It's true, but it is ugly. We still need headers, so I see no reason to move the implementations to a separate binary library. BTW, why do we keep cpu implementations of algorithms at all? GPUs perform much better for computations and there are opencl runtimes for cpus for the cases we need to do the operation on a cpu (and we don't need to optimize those kernels for cpus because as you mentioned performance is not critical in that case)?

jsteube:
There's two reasons:
- It's only faster in large quantities. But when we use those cpu algorithms, it's mostly just one call
- The OpenCL devices are not initialized at this point and it makes no sense to move them

jsteube:
Just in case you're wonding. Currently doing pretty intensive refactoring so that it's easier to merge your changes. Takes some time

KOLANICH:
OK

KOLANICH:
How is the progress?

jsteube:
You can see it from the commits I do, there's a lot of work being done already. But I'm not finished yet. I think it will take a few more weeks to get to a point to merge this stuff. Note that alot of your changes are a result of the same refactoring I'm doing, like moving the code into seperate files. When I'm done with that, I'll go through your PR and compare it with my changes to find out what I'm missing.

KOLANICH:
OK.

jsteube:
Refactoring almost done; Integrated hand selected changes; Continue discussion on separate issue https://github.com/hashcat/hashcat/issues/520

PR #488, opened 2016-09-06
Fix multiple build issues with OSX

matrix:
Current OSX building fail without that,
I test the patch also in linux, but only build + example0
Bye :)


jsteube:
This helped alot, thanks m8!

matrix:
✌

PR #533, opened 2016-10-09
Change Makefile to strip symbols in the linker instead of the compile…

coolbry95:
…r to make compatible with clang


jsteube:
Thanks!

PR #550, opened 2016-10-27
Add --restore-file-path option

Etienne-gh:
This option set the restore file path.
It's useful when we need all session files in specific directory.

Example:
hashcat -m 0 --session "Test" --potfile-path "/path/to/my/potfile.potfile"
--restore-file-path "/path/to/my/restore_file.restore" hashs.txt -a 3 ?a?a?a?a
Restore session with :
hashcat --sesion "Test" --restore --restore-file-path
"/path/to/my/restore_file.restore"


jsteube:
Nice, thanks!

PR #552, opened 2016-10-28
Fix OSX segfault on check_hash()

matrix:
See #508 for details

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -18,7 +18,7 @@
 
 #if defined (_WIN)
 
-#define hc_thread_create(t,f,a)     t = CreateThread (NULL, 0, (LPTHREAD_START_ROUTINE) &f, a, 0, NULL)
+#define hc_thread_create(t,T,f,a)   t = CreateThread (T, 0, (LPTHREAD_START_ROUTINE) &f, a, 0, NULL)
 #define hc_thread_wait(n,a)         for (u32 i = 0; i < n; i++) WaitForSingleObject ((a)[i], INFINITE)
 #define hc_thread_exit(t)           ExitThread (t)
 
@@ -29,7 +29,7 @@
 
 #elif defined (_POSIX)
 
-#define hc_thread_create(t,f,a)     pthread_create (&t, NULL, f, a)
+#define hc_thread_create(t,T,f,a)   pthread_create (&t, T, f, a)
 #define hc_thread_wait(n,a)         for (u32 i = 0; i < n; i++) pthread_join ((a)[i], NULL)
 #define hc_thread_exit(t)           pthread_exit (&t)
 
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -179,7 +179,7 @@ static int inner2_loop (hashcat_ctx_t *hashcat_ctx)
     thread_param->hashcat_ctx = hashcat_ctx;
     thread_param->tid         = device_id;
 
-    hc_thread_create (c_threads[device_id], thread_autotune, thread_param);
+    hc_thread_create (c_threads[device_id], NULL, thread_autotune, thread_param);
   }
 
   hc_thread_wait (opencl_ctx->devices_cnt, c_threads);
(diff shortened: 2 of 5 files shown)


jsteube:
That solution slows down cracking massive, because allocating an setting memory to zero would be done for each cracked hashes. If you have a large unsalted hashlist, cracking speed will drop 1000th of times.

A better way to handle this would be to preallocate that buffer somewhere and move it with the context. Can you please change?

magnumripper:
Also, what is the actual problem here? Why would you need to alloc/free at all? Why doesn't the stack variable do the job just fine?

mouse07410:
> That solution slows down cracking massive...

As compared to crashing with SEGV...?

jsteube:
@magnumripper That's a good question actually. When I saw this I thought about this is some OSX speciality I don't know about. This works fine for Linux/Windows usually.

matrix:
Hi,
I try to debug this issue with lldb, reproducible by exec example0.sh,
and at day 3 when I switch static allocation with dynamic hashcat stopped crash.

> include/common.h:#define HCBUFSIZ_LARGE      0x50000

There're a lot of static allocation with that size inside hashcat (called also "scratch buffer" in some comments), probably the stack is full.
@magnumripper try example0.sh (without apply my patch) and check if you got a segmentation fault.

@jsteube probably the best solution is not preallocate that buffer in hashcat_ctx struct, I think we should make better use of memory allocations.

@mouse07410 if my patch slow down the cracking speed we can't use it, also if without got a segfault. If you need hashcat for some particular reason you can use my repo temporarily until we solve the bug in the best way possible or switch to linux solution.

Sorry all for later response but yesterday I woke up with fever :( I'm sick ... grrr

magnumripper:
> @magnumripper try example0.sh (without apply my patch) and check if you got a segmentation fault.

Yeah I do. I'll debug it a little.

magnumripper:
Debug data is very strange. I think the real problem is something earlier, like a stack smash. But ASan doesn't help me. Did anyone try running with Valgrind (on Linux)?.

```
Process 7438 stopped
* thread #5: tid = 0x65c6c, 0x00000001000485af hashcat`check_hash(hashcat_ctx=<unavailable>, device_param=<unavailable>, plain=<unavailable>) + 14 at hashes.c:225, stop reason = EXC_BAD_ACCESS (code=1, address=0x700000eb2718)
    frame #0: 0x00000001000485af hashcat`check_hash(hashcat_ctx=<unavailable>, device_param=<unavailable>, plain=<unavailable>) + 14 at hashes.c:225
   222  }
   223  
   224  void check_hash (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, plain_t *plain)
-> 225  {
   226    debugfile_ctx_t *debugfile_ctx = hashcat_ctx->debugfile_ctx;
   227    loopback_ctx_t  *loopback_ctx  = hashcat_ctx->loopback_ctx;
   228  
(lldb) print hashcat_ctx
error: Couldn't materialize: couldn't get the value of variable hashcat_ctx: read memory from 0x700000eb2718 failed (0 of 8 bytes read)
error: errored out in DoExecute, couldn't PrepareToExecuteJITExpression
```

It does happen on **second** call to `check_hash()`.

matrix:
@magnumripper for print "hashcat_ctx" you need go back in previous frame (try "bt" command) and then print it with \* like:

`print *hashcat_ctx
`
I do 3 days of debugging for understand the issue :)

magnumripper:
Oh, right. But anyway I think the alloc/free just hides a problem still present. That stack size theory might be correct, is there a way to bump it?

matrix:
You can try comment out some code in "check_hash" function, you will be redirect to another function call, "potfile_write_append". But as I say before the only way I found to mitigate the segfault has been switch the memory allocation.
You can also try the @jsteube hint about prealloc that buffer somewhere for check if the segfault still here or not.

matrix:
Hi,
With DEBUG=2 we can confirm the stack overflow bug in OSX version.

look here http://pastebin.com/huHCVSKp

I do some mods on thread handling, in details I get the thread stack size with **pthread_attr_getstacksize**, double it and set with **pthread_attr_setstacksize**.

I think that could be one of the best mitigation :)

Happy Halloween,
~(=^‥^)ノ☆

mouse07410:
> > That solution slows down cracking massive...
> 
> if my patch slow down the cracking speed we can't use it, also if without got a segfault

As my timing shows, that patch does slow things up, but by no means "massive". The difference appeared to be be between 19 minutes and 15 seconds, and 19 minutes and 6 seconds. Not overwhelming.

Now trying your v2 (currently running :).

> If you need hashcat for some particular reason you can use my repo temporarily until we solve the bug in the best way possible

Both are fine (waiting for the best solution, and using your repo for now). I am not in any rush. Initially I stumbled upon this project as a good way to exercise OpenCL.

> or switch to linux solution

Sorry, impractical for me. And I'm not that eager to run Linux in a virtual machine.

> Sorry all for later response but yesterday I woke up with fever :( I'm sick ... grrr

Wish you to get healthy soon!

Also, I hope my findings (like, where the first Mac OS X-related failure came about) can be of use:
https://github.com/hashcat/hashcat/issues/508#issuecomment-257064018

**Update**
The current hashcut master (not oclHashcat) runs `example0.sh` in 19 minutes 9 seconds - only three seconds longer than the oclHashcat (when it doesn't crash :).

matrix:
I'm wrong with word "mitigation", I meant "alternative" :)

Another could be reduct the HCBUFSIZ_LARGE value, but I have absolutely no desire to test all the algorithms implemented to see if it crashes or not, i'm very lazy :)
but coding make me happy also in sick mode :D 

Thanks for your time

matrix:
@jsteube, version with preallocated buffer here

https://github.com/matrix/hashcat/commit/1df6ca2a47307b8b8c15fcf4a0cce0bb7580fc5b

mouse07410:
This is a change to accommodate Mac OS X OpenCL framework:

```
diff --git a/src/Makefile b/src/Makefile
index 8c9f2d6..8b7cdd5 100644
--- a/src/Makefile
+++ b/src/Makefile
@@ -94,7 +94,7 @@ VERSION_TAG              := $(shell test -d .git && git describe --tags --dirty=
 ## Compiler flags
 ##

-INCLUDE_PATHS            := -Iinclude/ -IOpenCL/ -I$(OPENCL_HEADERS_KHRONOS)/
+INCLUDE_PATHS            := -Iinclude/ -IOpenCL/

 CFLAGS                   += -pipe -W -Wall -std=c99

@@ -123,10 +123,12 @@ BINARY_NATIVE            := $(PROG_NAME)

 ifeq ($(UNAME),Darwin)
 export MACOSX_DEPLOYMENT_TARGET=10.9
-CFLAGS_NATIVE            :=
+CFLAGS_NATIVE            := -framework OpenCL
 CFLAGS_NATIVE            += $(CFLAGS)
-LFLAGS_NATIVE            := -lpthread
+LFLAGS_NATIVE            := -lpthread -framework OpenCL
 LFLAGS_NATIVE            += $(LDFLAGS)
+else
+INCLUDE_PATHS            += -I$(OPENCL_HEADERS_KHRONOS)/
 endif # darwin

 ifeq ($(UNAME),Linux)
```

I'd appreciate if you could merge it. I took care to make sure that neither Linux nor FreeBSD would be impacted by this change.

Also, 1df6ca2 works (just tested). Will provide timing later.

jsteube:
@mouse07410 I've pushed https://github.com/hashcat/hashcat/commit/2a330122b4095506aa16fce1d2d83707e528c2cf which includes the changes you suggested

jsteube:
@matrix Good work. Looks like you found the root of the problem (stack size). The solution using the buffer in context is exactly what I was looking for. Big Thanks!

mouse07410:
@matrix and @jsteube thank you - it works now with the latest/current master oclHashcat.

jsteube:
please close the issue then

PR #561, opened 2016-10-30
Fix a bunch of warnings. 

neheb:
Also added more warnings to the Makefile.

Note: -Wundef errors with the OpenCL Headers. That's been fixed upstream but 'git submodule update' sets up an old version. git submodule update --remote pulls the latest.


jsteube:
@neheb Can you fix those please: cc1: error: unrecognized command line option ‘-Wfloat-conversion’

PR #563, opened 2016-10-31
More cleanups

neheb:
Since GCC now sees a few functions as format string functions, new warnings have popped up. I fixed several but many more remain.

### DIFF
--- a/include/bitops.h
+++ b/include/bitops.h
@@ -6,12 +6,12 @@
 #ifndef _BITOPS_H
 #define _BITOPS_H
 
-u32 rotl32 (const u32 a, const u32 n);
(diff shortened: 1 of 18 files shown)


jsteube:
Please don't use **attribute** ((const)) and **attribute**((pure)) it makes the code kind of hard to read.
But I liked the **attribute** ((format (printf, 2, 3))) and hand-merged it.
I also hand-merged your changes related to fix the issue that are visible due to this.
I had to use -Wformat-zero-length because the empty string in event_log_*() has a special meaning.
Thanks!

neheb:
I remember GCC warning me about another function that needs the printf but the call was to vsnprintf so I was unsure how to apply it.

PR #570, opened 2016-11-04
Fix file locking (again).

magnumripper:
I added file locking in #177 but unfortunately it's been completely ruined since. First, lock_file() isn't even called anymore for pot file writing (which is what really **needs** locking).

Second (and much less of a problem), the lock_file() function was modified in 27bec8b so it's no longer signal safe. Why? The while loop in my original code wasn't something I made up just for lulz. It has to be there.

This fixes both issues.

### DIFF
--- a/src/debugfile.c
+++ b/src/debugfile.c
@@ -8,6 +8,7 @@
 #include "memory.h"
 #include "event.h"
 #include "debugfile.h"
+#include "locking.h"
 
 static void debugfile_format_plain (hashcat_ctx_t *hashcat_ctx, const u8 *plain_ptr, const u32 plain_len)
 {
@@ -109,6 +110,13 @@ int debugfile_init (hashcat_ctx_t *hashcat_ctx)
     if (debugfile_ctx->fp == NULL)
     {
       event_log_error (hashcat_ctx, "Could not open debug-file for writing");
+
+      return -1;
+    }
+
+    if (lock_file (debugfile_ctx->fp) == -1)
+    {
+      event_log_error (hashcat_ctx, "%s: %s", debugfile_ctx->filename, strerror (errno));
 
       return -1;
     }
--- a/src/dictstat.c
+++ b/src/dictstat.c
@@ -8,6 +8,7 @@
 #include "memory.h"
 #include "event.h"
 #include "dictstat.h"
+#include "locking.h"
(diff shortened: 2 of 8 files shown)


jsteube:
I guess that must have been going under while I was refactoring everything, sorry. It didn't happen intentionally. 

I'm wondering, you're locking the files, but never unlock them. Do you think it's ok?

magnumripper:
Sure, closing a file implicitly unlocks it. In fact, unlocking before close would be a race condition unless you flush before unlocking.

magnumripper:
In fact I think there's more room for improvement: You should ideally open your session file at start of run, then lock it and never close it (I think that's not the case right now) until actually exiting. This way there's no chance you accidentally start another session with same name (or default name) and end up with some **one** of them actually saved (the one that happened to write last). This is how JtR was made some 20 years ago and it has saved many many hours of cut/paste mistakes...

magnumripper:
I'll produce a PR for you to review and test.

magnumripper:
I realized I had missed one instance, added now.

jsteube:
> You should ideally open your session file at start of run, then lock it and never close it (I think that's not the case right now) until actually exiting

If it's not the case, I will rewrite it to do that because that's the expected behaviour. I'll do that after the merge.

## 

Are you planing to add something to this or can I merge already?

magnumripper:
> Are you planing to add something to this or can I merge already?

I'm pretty sure it's fine right now. I haven't tested it as much as I did #177 though.

jsteube:
OK I merged, the compiler complains about the following when compiling for windows:

```
In file included from src/logfile.c:11:0:
src/logfile.c: In function ‘logfile_append’:
include/locking.h:23:28: warning: statement with no effect [-Wunused-value]
 #define lock_file(dummy)   0
                            ^
src/logfile.c:54:3: note: in expansion of macro ‘lock_file’
   lock_file (fp);
   ^
```

Can you try to find a different solution?

magnumripper:
I don't do windows at all ever. What if you change it to

``` c
#define lock_file(dummy) void dummy
```

magnumripper:
...no that will probably not be better. Perhaps we should #ifdef all calls to (un)lock_file instead. Or just that call.

Issue #568, opened 2016-11-04
Don't use $HEX[deadbabe] for UTF-8


magnumripper:
Virtually every user of Hashcat is running a UTF-8 capable terminal, right? Personally I'd much appreciate not getting $HEX[deadcafe] output for perfectly fine UTF-8 strings. This goes for pot files as well as screen output. I can make a PR but I thought we'd discuss details first:

I propose we add a UTF-8 test but keep the "printable ASCII" test, and add some option for which of them to use. I'd personally vote for using the UTF-8 test by default. BTW in some cases where we *know* hashes can't be made from UTF-8 (eg. LM), we'd auto-disable it and fall back to the "printable ASCII" test.

See also #567.

jsteube:
I think I can agree _if_ we find a way to tell if a string is a valid UTF-8 string. For example $HEX[00003030] surely is not. In such a case we should stick to $HEX[...]. Is there any easy way to check this?

magnumripper:
Sure, that's the base of the idea. I already have that code in use elsewhere and it's proven to work like a champ. For extremely short (eg. two bytes) passwords consisting of random binary data there's a slight chance of mis-interpreting - but actually there's not much harm there either. Once you realise this is the case you can always re-encode it. No data is lost! For lengths of 3-4 bytes already, the risk of mis-interpreting is really slim. For real-life passwords, it practically never happens.

Having said that, I think it's a good idea to "blacklist" LM from this default behaviour since we know it's never ever made of UTF-8 but from an "OEM codepage" like CP850. Same goes for 40-bit RC4 brute-force as in eg. oldoffice.

jsteube:
merged, thanks!

PR #571, opened 2016-11-05
Change need_hexify so it optionally tests for printable UTF-8. See #568

magnumripper:
Here's code for #568. Currently it falls back to ASCII for the LM format only (possibly a few other formats could be added, and maybe even a command line option although I see no need for one).

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -7,17 +7,84 @@
 #include "types.h"
 #include "convert.h"
 
-bool need_hexify (const u8 *buf, const int len)
+static bool printable_utf8 (const u8 *buf, const int len)
+{
+  u8 a;
+  int length;
+  const u8 *buf_end = buf + len;
+  const u8 *srcptr;
+  const char trailingBytesUTF8[64] = {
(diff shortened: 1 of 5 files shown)


src/convert.c
```diff
@@ -7,17 +7,84 @@
...
+static bool printable_ascii (const u8 *buf, const int len)
 {
   for (int i = 0; i < len; i++)
   {
     const u8 c = buf[i];
 
-    if (c < 0x20) return true;
-    if (c > 0x7f) return true;
+    if (c < 0x20) return false;
+    if (c > 0x7e) return false;
```
magnumripper, line 72: Note that 0x7f is not printable!

jsteube:
I'll merge this. I'll maybe change the check for LM and move it to here:

https://github.com/hashcat/hashcat/blob/master/src/interface.c#L17428-L17443

and add an option like OPTS_TYPE_PT_NO_UTF8 to mark those algorithms. Thanks for that check, saves me alot of headache checking for it :)

magnumripper:
Yeah a format flag is the way to go, that'll be perfect.

Issue #572, opened 2016-11-06
Support skip and limit options for mask file


Ins1ghtLabs:
It would be great if hashcat would support using skip and limit options per mask for mask file.
For example:

line 1 : ?l?l?l?l?l?d?l?l?l?l,100000,10000000
line 2 : ?u?u?l?l?l?l?l?l?l?l,1000000,500000
...

This would allow better distribution to be carried by splitting masks with large keyspace into smaller chunks and evenly distribute the work load using mask files.

Ins1ghtLabs:
Seems I need to somehow add additional parser in mpsp.c and change status_ctx->words_off and 
status_ctx->words_base when the job is starting.

Ins1ghtLabs:
Pull request:
https://github.com/hashcat/hashcat/pull/574

magnumripper:
What's the reason you ended up using `|` instead of `,`? Just curious.

Ins1ghtLabs:
@magnumripper because it's easier to use '|' than ',' , so I won't have to write code to figure out whether that part is a custom charset or mask or skip/limit number, or touch the mask parsing logic.

jsteube:
I don't like this alot, because it creates some inbalance. It would enable to set -s and -l for an iterated attack, but only for one special. For example, this would not work to set -s and -l for a folder full of wordlists, which hashcat supports and which is also an iterated attack.

OTOH I don't see a reason how this helps in a distributed solution. In such a scenario, there should be no iterated attack at all, because it becomes hard to control.

Ins1ghtLabs:
"For example, this would not work to set -s and -l for a folder full of wordlists, which hashcat supports and which is also an iterated attack."

Which attack mode is this? can you elaborate?

jsteube:
Like, every attack that support wordlist also supports wordlist folders, except -a1. 

See here:

```
root@sf:~/hashcat# cat hash
7f306bca1a462127e59b9154599c443a
root@sf:~/hashcat# cat wordlist1.txt 
word1
root@sf:~/hashcat# cat wordlist2.txt 
word2
root@sf:~/hashcat# ./hashcat hash *.txt --quiet                       
7f306bca1a462127e59b9154599c443a:word2 
root@sf:~/hashcat# 
```

That means if we change support for -s and -l we should change it for all attack modes, otherwise this creates confusion. 

[...]

PR #574, opened 2016-11-08
Add support for skip and limit in mask file.

Ins1ghtLabs:
For example:

line1: ?u?d?d?d?u?d?d?d|30000|20000

30000 is the skip value and 20000 is the limit value. You may also only specify the skip value such as ?u?d?d?d?u?d?d?d|30000. No behavior will change if you leave the mask as before.

### DIFF
--- a/src/mpsp.c
+++ b/src/mpsp.c
@@ -1410,7 +1410,8 @@ int mask_ctx_parse_maskfile (hashcat_ctx_t *hashcat_ctx)
 {
   mask_ctx_t     *mask_ctx     = hashcat_ctx->mask_ctx;
   user_options_t *user_options = hashcat_ctx->user_options;
-
+  status_ctx_t  *status_ctx=hashcat_ctx->status_ctx;
+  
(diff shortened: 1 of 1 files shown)


Ins1ghtLabs:
Well, the delimiter '|' can't be escaped, therefore it can't be used in any inline charsets or masks. Although '|' is very rarely used anywhere, which is the reason why use it as the delimiter.

jsteube:
Can you please explain why do you need this feature at all?

Ins1ghtLabs:
Without this feature, hashcat can only distribute the workload one mask at a time, if the hash file is large, it takes a long time to restart because it has to reload the hash file into memory and map them  again.  By pre-calculating keyspaces and ECDs of all masks in a mask file before hand, we can divide heavier masks into smaller workloads and generate mask files with equal workload mask files  then assign these files to the worker nodes. We can also tailor workloads to different sizes to suit worker nodes that have different performance.

jsteube:
See https://github.com/hashcat/hashcat/issues/572 for discussion

PR #576, opened 2016-11-10
Add hex charset (?h) to mpsp

unix-ninja:
Just making it easier to do hex masks since they can be often used.


jsteube:
Nice, thanks!

Issue #412, opened 2016-07-03
`Abort trap: 6` when benchmarking `SHA-3(Keccak)` on Iris Pro


leonklingele:
```
$ ./hashcat.app -d2 -b -m 5000
hashcat (v3.00-14-gb58f7a4) starting in benchmark-mode...

OpenCL Platform #1: Apple
=========================
- Device #1: Intel(R) Core(TM) i7-4870HQ CPU @ 2.50GHz, skipped
- Device #2: Iris Pro, 384/1536 MB allocatable, 40MCU
- Device #3: GeForce GT 750M, skipped

Hashtype: SHA-3(Keccak)

Abort trap: 6
```

Works fine with -d3 though.

Using latest master (b58f7a4) on OS X 10.11.5 (15F34)

kardipapa:
I have the same problem with Iris Pro graphics, i use mid 2015 MacBook Pro with OS X 10.11.6, hashcat version: v3.00-69-g804ee28, compiled from git source. But when i use dictionary attack with 10GB dict. files, just works fine with Iris Pro

zrayburn:
I am unsure entirely but it may be the amount of VRAM the card can allocate. For instance in mining Ethereum you must have a card that has at least 1.5 gb of VRAM in the GPU to load the full operations buffer.

mouse07410:
This is not Iris Pro, but maybe it would be helpful (from MacBook Air with Intel HD Graphics 5000):

```
CL_DEVICE_TYPE_GPU[0]
    CL_DEVICE_NAME: HD Graphics 5000
    CL_DEVICE_AVAILABLE: 1
    CL_DEVICE_VENDOR: Intel
    CL_DEVICE_PROFILE: FULL_PROFILE
    CL_DEVICE_VERSION: OpenCL 1.2 
    CL_DRIVER_VERSION: 1.2(Aug 29 2016 22:20:39)
    CL_DEVICE_OPENCL_C_VERSION: OpenCL C 1.2 
    CL_DEVICE_MAX_COMPUTE_UNITS: 40
    CL_DEVICE_MAX_CLOCK_FREQUENCY: 1100
    CL_DEVICE_MAX_WORK_GROUP_SIZE: 512
    CL_DEVICE_ADDRESS_BITS: 64
    CL_DEVICE_MEM_BASE_ADDR_ALIGN: 1024
    CL_DEVICE_MAX_MEM_ALLOC_SIZE: 402653184
[...]

PR #584, opened 2016-11-14
Fix CL_BUILD_PROGRAM_FAILURE on OSX (#412)

matrix:
ω(=＾・＾=)ω


magnumripper:
This is the silliest driver (or run-time) bug ever. It was present already in OS X Mavericks three years ago - we had a similar problem in JtR back then.

jsteube:
Unbelieveable! Thanks :)

Issue #603, opened 2016-11-20
Clang analyzer take 2


neheb:
https://dl.dropboxusercontent.com/u/102011983/hashcat.tar.xz

Interestingly enough, VERIFY_PTR also fools clang into thinking it's a file descriptor.

jsteube:
How can I reproduce this?

neheb:
My full command line is,

scan-build -enable-checker alpha.core.BoolAssignment -enable-checker alpha.core.CallAndMessageUnInitRefArg -enable-checker alpha.core.CastSize -enable-checker alpha.core.CastToStruct -enable-checker alpha.core.DynamicTypeChecker -enable-checker alpha.core.FixedAddr -enable-checker alpha.core.IdenticalExpr -enable-checker alpha.core.PointerArithm -enable-checker alpha.core.SizeofPtr -enable-checker alpha.core.TestAfterDivZero -enable-checker alpha.core.TestAfterDivZero -enable-checker alpha.deadcode.UnreachableCode -enable-checker alpha.security.ArrayBound -enable-checker alpha.security.ArrayBoundV2 -enable-checker alpha.security.MallocOverflow -enable-checker alpha.security.ReturnPtrRange -enable-checker alpha.security.taint.TaintPropagation -enable-checker alpha.unix.cstring.BufferOverlap -enable-checker alpha.unix.cstring.NotNullTerminated -enable-checker alpha.unix.cstring.OutOfBounds -enable-checker alpha.unix.Stream -enable-checker alpha.unix.SimpleStream make

As of now, 55 warnings show up. The last two checkers are responsible for the VERIFY_PTR issue. Interestingly enough, if those get disabled, other warnings such as unreachable code ones also disappear.

jsteube:
Yeah I kind of run into the same problems here. For example, I wanted to
examine the divide by zero problem reported (which in the html looks
more like a bug in scan-build itself) but when I'm selecting only
"alpha.core.TestAfterDivZero" the problem disappears. We have to be very
careful with some of this stuff.

jsteube:
I've reduced what scan-build calls "bugs" to a short number. The remaining ones are files handles that need to stay open during the session, so this is not an error.

Please close the issue if that's fine for you.

Thanks for the input anyway, this is good stuff.

neheb:
I actually left out a checker in that command line: alpha.core.PointerSub

[...]

PR #604, opened 2016-11-20
Some fixes from #603

neheb:
I am not completely sure about sizeof(*binary). Googling indicates the result should be the same.

### DIFF
--- a/src/folder.c
+++ b/src/folder.c
@@ -148,7 +148,7 @@ char *first_file_in_directory (const char *path)
 
     #endif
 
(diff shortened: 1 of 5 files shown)


matrix:
@neheb, "sizeof(*binary)" it's not good for me and not only ...
Have you tried running hashcat after apply these patches?

jsteube:
Be careful of this trap, it's actually not a bug. This is some weird AMD OpenCL runtime error, they actually want the size of the pointer address. Let's continue on https://github.com/hashcat/hashcat/issues/603

PR #609, opened 2016-11-21
add missing includes for FreeBSD

lifanov:
These were lost during refactoring and found again
in preparation for port update to upcoming 3.20.


jsteube:
Thanks!

PR #618, opened 2016-11-27
Fixed event_log_warning() format in debug builds

matrix:
Fixed:

```
src/opencl.c:3480:149: warning: format specifies type 'unsigned long long' but the argument has type 'size_t' (aka 'unsigned long') [-Wformat]
      if (user_options->quiet == false) event_log_warning (hashcat_ctx, "SCRYPT tmto optimizer value set to: %u, mem: %" PRIu64, scrypt_tmto_final, size_scrypt);
                                                                                                                      ~~~                           ^~~~~~~~~~~
src/opencl.c:3808:144: warning: format specifies type 'long' but the argument has type 'off_t' (aka 'long long') [-Wformat]
          if (user_options->quiet == false) event_log_warning (hashcat_ctx, "* Device #%u: Kernel %s (%ld bytes)", device_id + 1, cached_file, cst.st_size);
                                                                                                      ~~~                                      ^~~~~~~~~~~
                                                                                                      %lld
src/opencl.c:3827:142: warning: format specifies type 'long' but the argument has type 'off_t' (aka 'long long') [-Wformat]
        if (user_options->quiet == false) event_log_warning (hashcat_ctx, "* Device #%u: Kernel %s (%ld bytes)", device_id + 1, source_file, sst.st_size);
                                                                                                    ~~~                                      ^~~~~~~~~~~
                                                                                                    %lld
src/opencl.c:4012:142: warning: format specifies type 'long' but the argument has type 'off_t' (aka 'long long') [-Wformat]
        if (user_options->quiet == false) event_log_warning (hashcat_ctx, "* Device #%u: Kernel %s (%ld bytes)", device_id + 1, cached_file, cst.st_size);
                                                                                                    ~~~                                      ^~~~~~~~~~~
                                                                                                    %lld
src/opencl.c:4154:142: warning: format specifies type 'long' but the argument has type 'off_t' (aka 'long long') [-Wformat]
        if (user_options->quiet == false) event_log_warning (hashcat_ctx, "* Device #%u: Kernel %s (%ld bytes)", device_id + 1, cached_file, cst.st_size);
                                                                                                    ~~~                                      ^~~~~~~~~~~
                                                                                                    %lld
5 warnings generated.
```

### DIFF
--- a/src/opencl.c
+++ b/src/opencl.c
@@ -3479,7 +3479,7 @@ int opencl_session_begin (hashcat_ctx_t *hashcat_ctx)
       }
 
       #if defined (DEBUG)
-      if (user_options->quiet == false) event_log_warning (hashcat_ctx, "SCRYPT tmto optimizer value set to: %u, mem: %" PRIu64, scrypt_tmto_final, size_scrypt);
+      if (user_options->quiet == false) event_log_warning (hashcat_ctx, "SCRYPT tmto optimizer value set to: %u, mem: %ld", scrypt_tmto_final, size_scrypt);
       if (user_options->quiet == false) event_log_warning (hashcat_ctx, "");
       #endif
     }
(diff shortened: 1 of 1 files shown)


jsteube:
Using this PR, it produces warnings on both linux and windows:

```
src/opencl.c:4156:75: warning: format ‘%lld’ expects argument of type ‘long long int’, but argument 5 has type ‘__off_t {aka long int}’ [-Wformat=]
```

I think, for the PRIu64, the correct format specifier would be to use %zu not %ld, since we use -std=c99 and size_t and not ssize_t. For the off_t, using lld is a problem on windows 32 bit machines, We should use %zd here.

But in both cases, mingw (i think windows in general because it uses printf() from msvcrt), doesn't like z* types.

So the only way out is to cast the variable to something. I've done that with commit: https://github.com/hashcat/hashcat/commit/e77c029dadd82d6f572ab2d94a1d8de1f182eed7

This seems to work pretty good on Linux and Windows. In theory, it should work on OSX as well. If not please open an issue for it.

neheb:
Is there an issue with using MinGW's ANSI stdio feature? It removes the PRI formats but adds ℅z

jsteube:
I've tried exactly that, but didn't work. Warning still appears

PR #620, opened 2016-11-27
Request build log only if size > 1

matrix:
If build_log_size is 1 will be printed an empty line for each hc_clGetProgramBuildInfo call


jsteube:
This PR is invalid, because this: 

> if (CL_rc == -1 && build_log_size > 1)

Some OpenCL runtime, when they have error, do not set CL_rc to -1, they only have build_log_size > 1

PR #887, opened 2016-11-29
Change fseek and ftell to o variants

neheb:
Standard way to do 64-bit file sizes

I do wonder if o-less functions also work on 32-bit binaries. (64-bit filesizes).


jsteube:
I don't know either. Let's hope 32 bit OS will die soon :)

PR #891, opened 2016-11-29
Fix compilation for native Cygwin

neheb:
set_cpu_affinity was if'd out since cygwin has no cpu_set_t or pthreads_setaffinity_np support.

Still doesn't quite work(OpenCL kernels don't even run) but at least it gets to that point.

### DIFF
--- a/include/affinity.h
+++ b/include/affinity.h
@@ -11,7 +11,9 @@
 
 #if defined (_POSIX)
 #include <pthread.h>
+#if defined (__linux__)
 #include <sys/sysctl.h>
+#endif // __linux__
 #endif // _POSIX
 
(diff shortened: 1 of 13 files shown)


jsteube:
Great going forward fix. Thanks!

But tell me more about cygOpenCL-1.dll

* Why was it needed?

* Is OpenCL v1.2 compatible?

* Which OpenCL runtimes did you test it with?

neheb:
It's the same thing as libOpenCL.so, just the cygwin flavor. pocl is the only supported runtime.

edit: more directly,

1: all of the other dlopen calls fail. Only this succeeds.

2: I assume pocl is. yes.

3: pocl.

PR #892, opened 2016-11-30
Fixed a few warnings.

neheb:
Used -Wstrict-aliasing=5 and -Wformat=2 for these. No idea why they don't normally get picked up.

Oh and removed _LARGEFILE64_SOURCE. _GNU_SOURCE already defines it but not the other one.


jsteube:
Good fixes, thanks!

PR #895, opened 2016-12-02
Only lock loopback and pot files while actually writing to them.

### LINKED ISSUE #893: hashcat hangs when called with --left option
It is not possible to get uncracked hashes with the --left option while cracking instance is still running.

**How to reproduce**
try to crack a hash that is not cracked immediately
> 098f6bcd4621d373cade4e832627b4f6:test
> 5fa0fd04d89bbf2f09c891add76ac48c:hashcat-uncrackable$_-_

content of md5.hash
> 098f6bcd4621d373cade4e832627b4f6
> 5fa0fd04d89bbf2f09c891add76ac48c

run hashcat
> hashcat64.bin -m 0 -a 3 md5.hash ?a?a?a?a?a?a?a?a?a?a --potfile-path=result.pot -i --increment-max=10

while hashcat is cracking the hashes run following command
> hashcat64.bin md5.hash -m 0 --potfile-path=result.pot --left

the command doesn't return (hangs) until the cracking instance has finished or is aborted.

**Affected versions**
tested with hashcat-3.20-rc5 (30-Nov-2016 16:53)
hashcat v3.10 is not affected, the --left command returns immediately

magnumripper:
Closes #893.

### DIFF
--- a/src/loopback.c
+++ b/src/loopback.c
@@ -112,15 +112,6 @@ int loopback_write_open (hashcat_ctx_t *hashcat_ctx)
     return -1;
   }
 
-  if (lock_file (fp) == -1)
-  {
-    fclose (fp);
-
-    event_log_error (hashcat_ctx, "%s: %s", loopback_ctx->filename, strerror (errno));
-
-    return -1;
-  }
-
   loopback_ctx->fp = fp;
 
   loopback_ctx->unused = true;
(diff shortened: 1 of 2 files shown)


jsteube:
Many thanks! There's other files that do locking, too. For example

* debugfile: filename must be given be the user, but would lock in case same file used: not sure if we need a fix at all in such a case, but I don't think so..
* dictstat: only uses locking for writing and instantly closes afterwards: no fix required
* hashfile: like dictstat, also requires --remove
* logfile: like fixed loopback
* kernelfiles (cache): like dictstat
* outfile: like debugfile
* stdout: like debugfile

In general, should be fine as-is

magnumripper:
Yes, I too reviewed all of them and hopefully it's all OK now.

PR #896, opened 2016-12-03
Few more fixes.

neheb:
It compiles so I assume it works fine. Interestingly cppcheck complains about ctime_r but not ctime when it should complain about both.

I tried avoiding defining _POSIX_THREAD_SAFE_FUNCTIONS directly but the only other way is to define _POSIX_C_SOURCE which defines ANSI stdio as well.


neheb:
Rebasing

PR #897, opened 2016-12-03
FreeBSD: fix refactoring some more - sysctl produces a string

lifanov:
Sorry, I only found this post-release once I got cracking :)


jsteube:
OK, so the sysctl () call in FreeBSD returns the real path, not a link, right?

PR #898, opened 2016-12-03
Thread safety fixes.

neheb:
First commit removes strerror, which is not thread safe. At least not under linux. %m is a GNU extension and as such, the attribute had to be changed to gnu_printf. Otherwise MinGW threw warnings.

I left two functions as normal printf because of MinGW stupidity( PRIu64 getting evaluated to %u and whatnot).

Second commit changed ctime to ctime_r. Pretty straight forward. The define also exposes gmtime_r in MinGW.

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -17,6 +17,7 @@
 #elif defined (_WIN32) || defined (_WIN64)
 #define _WIN 1
 #define WIN 1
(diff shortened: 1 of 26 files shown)


jsteube:
Nice, didn't know that. About the printf() leftover, I think there's no messages that are neither warnings nor errors that use strerror(). So it should be ok. Thanks!

PR #903, opened 2016-12-03
Add custom tuning settings for Iris Pro / SHA-3(Keccak)

matrix:
#412


leonklingele:
This fixes #412, yay!

matrix:
@leonklingele, could you do a full benchmark only with Iris Pro device to identify eventually other problems?
Thanks

leonklingele:
@matrix I can't :/ It still crashes when benchmarking Whirlpool: https://github.com/hashcat/hashcat/issues/412#issuecomment-264703313

jsteube:
Just a suggestion, why don't you just pair the Iris and Iris Pro into a group so that the tuning will work on both the same way? hashcat.hctune supports that.

matrix:
@jsteube I will see why Apple named it as "Pro" before set an alias in hctune :)

PR #907, opened 2016-12-04
Only apply gnu_printf to MinGW

neheb:
Needs verification before being pulled. No problems with MinGW or Linux but I cannot test OS X or FreeBSD so I have no idea if %m is supported by them or not.

Ideally test by invoking one of the functions. For example

./hashcat -a 2 -m ?? hashfile.txt filenotfound.dict

@matrix ?


neheb:
Never mind

PR #908, opened 2016-12-04
Revert "Remove strerror in favor of thread safe %m"

neheb:
Not compatible with MinGW. Just prints an m on error.


neheb:
I really need to stop this...

PR #912, opened 2016-12-06
Fix %m with MinGW and potentially OS X.

neheb:
This needs testing on the latter. If it doesn't work, don't bother pulling.


jsteube:
Is there a special reason to use "__MINGW32__" (32 bit only) mingw macro / was that done intentional?

neheb:
Yeah I looked through the defines for 32 and 64 compilers and found they both share this one.

PR #913, opened 2016-12-06
extend global CFLAGS and LDFLAGS to aid distro packaging

anthraxx:
This preserves globally defined CFLAGS and LDFLAGS and simply
extends those variables to aid distro based packaging toolchains
and predefined distro wide defaults like SSP, relro etc.

this re-fixes 7f8aaf74302816d03fbff62dab5c987d498acdde after it was
somehow undone ;)

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -101,7 +101,7 @@ BINARY_NATIVE           := $(PROG_NAME)
 ## General compiler and linker options
 ##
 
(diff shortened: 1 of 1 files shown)


jsteube:
I'm sorry I think this was lost during refactoring.

anthraxx:
@jsteube no problem, i was just wondering when observing it again :yum:

PR #918, opened 2016-12-15
Zero padding rule

dellastreet:
Rule to make a fixed length password by padding with zero's

### DIFF
--- a/OpenCL/inc_rp.cl
+++ b/OpenCL/inc_rp.cl
@@ -1185,11 +1185,11 @@ inline u32 rule_op_mangle_reflect (const u32 p0, const u32 p1, u32 buf0[4], u32 
   return out_len;
 }
 
-/** 
+/**
 * Pad with zero's or truncate to padding length
 * p0 is length of padded password
 */
-inline u32 rule_op_fixedlenpad (const u32 p0, const u32 p1, u32 buf0[4], u32 buf1[4], const u32 in_len)
+inline u32 rule_op_fixpad (const u32 p0, const u32 p1, u32 buf0[4], u32 buf1[4], const u32 in_len)
 {
   if (in_len > p0) return (p0);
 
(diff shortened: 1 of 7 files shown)


jsteube:
Nice, the first PR which adds a rule. The question is just why we would need such a rule. I know there's hash-modes that need zero byte padding, but in this case it's better to have a dedicated hash-mode added. Can you please explain what's the reason you put this into a rule?

dellastreet:
For cracking a password with zero padding :). It seemed to me that adding a hash mode for this purpose was more complicated then adding a rule. So the main reasons are prob. Lack of knowledge if the code + lazyness.

dellastreet:
I guess i also thought it would be more flexible. For bruteforce modes it would be possible to add the zeros to the search pattern.. So its mostly relevant for dictionary attacks anyway on which i want to do rule expansion anyway.

magnumripper:
JtR has 'P' for english grammar (eg. crack -> cracked). We've been talking about trying to unify rules commands but in this case it's probably not a problem since HC will likely never get the JtR version of 'P' anyway and on the other hand, JtR can't currently implement *this* version of 'P' since it can not currently handle nulls within plains.

So, in short: No problem there.

jsteube:
I'd prefer to not have any collisions between hashcat and JtR rule
commands. Can you make a suggestion for an alternative name that is not
in use by JtR currently?

magnumripper:
There already are many collisions. We have https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/doc/RULES#L104-L303.
Recently Jim added a "hashcat rules mode" where we behave differently and try to mimic hashcat (hashcat legacy even, I think) rules. https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/doc/RULES-hashcat

@solardiz @jfoug

jsteube:
I'm really happy to see a new rule function added via PR, but in this case I think the correct way to implement this logic into the hash mode kernel. Otherwise you can't do mask attacks and combinator attacks against such a hash mode because rules are not supported in those. If you need help in how to add a new hash-mode, contact me on IRC.

dellastreet:
Ok, will look into that method. Probably won't look into before mid january.

PR #925, opened 2016-12-20
Replace some snprintf calls with asprintf.

neheb:
Simplifies the code.

Runtime tested. Seems to work fine.

### DIFF
--- a/src/dictstat.c
+++ b/src/dictstat.c
@@ -43,11 +43,10 @@ int dictstat_init (hashcat_ctx_t *hashcat_ctx)
   if (user_options->attack_mode == ATTACK_MODE_BF) return 0;
 
   dictstat_ctx->enabled  = true;
-  dictstat_ctx->filename = (char *)       hcmalloc (HCBUFSIZ_TINY);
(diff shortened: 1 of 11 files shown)


jsteube:
Nice! Thanks

jsteube:
I'm getting tons of those now:

warning: ignoring return value of ‘asprintf’, declared with attribute warn_unused_result [-Wunused-result]

neheb:
Is this on OS X?

jsteube:
No, Ubuntu 16.04

neheb:
agh this also seems to be a problem in 14.04. And only in Ubuntu.

So far I have been unable to reproduce in Fedora, MinGW, or Cygwin. This sounds like Ubuntu's version of glibc has a very specific patch to add warn_unused_result to asprintf. Hillarious since neither malloc nor snprintf do.

I'd suggest adding -Wno-unused-result in the Makefile to silence the warning.

jsteube:
Added a wrapping function: https://github.com/hashcat/hashcat/commit/6ece83760d3501ee34a8ee2076751edb8945dd81

PR #937, opened 2016-12-29
Update docs to use a light clone for building

bluehallu:


jsteube:
While you're absolutely right I prefer to let it as-is, only because it's what people (which doesn't have any experience with git) expect to see.

PR #940, opened 2016-12-30
respect ports tree CFLAGS and LFLAGS

lifanov:
FreeBSD ports need to respect flags passed down by the build system.
This enables FreeBSD hashcat port to use upstream hashcat source code
patch-free.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -177,10 +177,12 @@ LFLAGS_NATIVE           += -lpthread -ldl
 endif # Linux
 
 ifeq ($(UNAME),FreeBSD)
+ifndef PORTNAME
 CFLAGS_NATIVE           := $(CFLAGS)
-CFLAGS_NATIVE           += -I$(OPENCL_HEADERS_KHRONOS)/
-CFLAGS_NATIVE           += -march=native
 LFLAGS_NATIVE           := $(LFLAGS)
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks! So does that mean that the split into hashcat library and hashcat frontend works on FreeBSD?

lifanov:
Yes, it works. It has been working since release. I'm just trying to de-uglify the port.

jsteube:
No, I meant the yesterday change. There's two files produced by "make"
now. That is "hashcat" and "libhashcat.so".

lifanov:
Hmm, let me re-test

lifanov:
Yep, it works:
```
$ file hashcat
hashcat: ELF 64-bit LSB executable, x86-64, version 1 (FreeBSD), dynamically linked, interpreter /libexec/ld-elf.so.1, for FreeBSD 12.0 (1200019), FreeBSD-style, stripped
$ ldd hashcat
hashcat:
	libhashcat.so => ./libhashcat.so (0x800827000)
	libthr.so.3 => /lib/libthr.so.3 (0x800b07000)
	libc.so.7 => /lib/libc.so.7 (0x800d2e000)
```

lifanov:
Runtime works too

PR #941, opened 2016-12-30
respect ports tree CFLAGS and LFLAGS  step 2

lifanov:
It's actually less confusing to let the port set
the other two options as well, so just move the whole
block under ifdef PORTNAME.

Sorry about the back-to-back thing. The port works 100% patch-free now.


lifanov:
This is all that's needed for hashcat port to be patch-free in the next release:
https://github.com/freebsd/freebsd-ports/commit/19bf8a22ad805fbf22ce7b5dabe4bd6a693a4e48

PR #961, opened 2017-01-11
Filter out OpenCL Intel Runtime broken check on Apple

matrix:


jsteube:
Can you please rewrite so that it does not use a NOT. We try to stick to positive cases only for easier reading.

matrix:
Ok, is broken only in windows?

jsteube:
All systems (just not apple?)

matrix:
If only Apple is not affected the right code has already been written :)

PR #981, opened 2017-01-21
log topid/subid should use u32 everywhere 

philsmd:
For the ids we do not need signed integers.
This changes should make sure that we use unsigneds in general when dealing with ids (both top and sub ids for the logfile).

Thank you


jsteube:
Nice find, thanks!

Issue #965, opened 2017-01-12
-m 11600 (7z): Support LZMA/LZMA2 decompression of decrypted data


philsmd:
For -m 11600 = 7-Zip the archived data sometimes needs to be decompressed via LZMA/LZMA2 to calculate the correct CRC32 checksum. This problem was first discussed on hashcat's forum: https://hashcat.net/forum/thread-4328-post-32963.html#pid32963

Hashcat currently does not support lzma_decode () or lzma2_decode (), but as mentioned on the forum post it could probably simple be implemented via the 7z SDK and a hook (this system is already supported by hashcat).

Recently 7z2hashcat.pl was updated to produce $7z$0$ (uncompressed), $7z$1$ (LZMA1 compressed) or $7z$2$ (LZMA2 compressed) "hashes" (because 7z2hashcat knows this info).
That means that the hash format was also updated and I already made a pull request (https://github.com/hashcat/hashcat/pull/964) to make hashcat aware of those new signatures.

This issue here just indicates that this is work-in-progress and the LZMA/LZMA2 decompression will be a new feature (that still needs to be implemented).

Thank you (and feel free to discuss what you think about this and how it should be implemented below)

philsmd:
I just wrote a proof of concept and realized that we miss some further information in the hash format:
1. the length of the uncompressed data (needed to allocate the memory and tell LZMA how long the output will be)
2. the length of the first "stream" (i.e. how many bytes of the start of the decompressed data will be needed to generate the CRC32). Note: if there are more files in the .7z file, the data of all files will be "concatenated" but the CRC32 checksum we need to check was generated by just the bytes of the first file.
3. the attributes (lc, pb, lp and the dicSize) especially needed for LZMA1

(all of this info should be easily available/accessible with 7z2hashcat.pl, but we probably need to adapt the hash format again!!!)
@magnumripper 

Here is a proof of concept/test (feel free to use it): https://gist.github.com/philsmd/99aebdc722472b5322a4e0b0408d743a
Thank you very much

magnumripper:
Yes I just realised this too. So perhaps we should just add new fields in the end (those fields shall be invalid for `$7z$0$`).

philsmd:
Yeah, maybe something like this:
whenever we have $7z$0$ hashes we use:
...$[data]
whenever we have $7z$1$ or $7z$2$ hashes we use:
...$[data]$[uncompressed_data_length]$[first_file_size]$[coder_attributes]

[...]

PR #986, opened 2017-01-24
#965: new hook function to support LZMA1/LZMA2 decompression for 7z

philsmd:
This should solve the problem mentioned in #965 (i.e. the feature that was missing to decompress the data stream if it was LZMA1/LZMA2 compressed, to be able to validate the CRC32 "hash").

This PR commit includes the hook function seven_zip_hook_func () which is able to do the decompression on the host (CPU) and notify the kernel about the results.

The additional files are mainly from the LZMA SDK (public domain license).

The consequence of these changes should be to have much less false positives + be able to crack much more files (which have compressed data). It also implements the new parsing "rules" to accept the newly formatted output by 7z2hashcat.pl.

Thanks


jsteube:
Great work, thanks!

PR #1020, opened 2017-02-04
Restore strerror as %m is unsupported by the BSDs

neheb:
The BSDs just print m instead of the actual error message. Benefit
is not worth the cost.

Most of the usages of %m are reporting an error, not a warning. Thread safety doesn't really matter at that point. Some Linux and Windows specific usages were kept.


jsteube:
Thanks :)

PR #1024, opened 2017-02-06
Adding windows appveyor builds

stephengroat:
Currently having an issue on Windows x86_64, was able to re-create in `appveyor`

included the `.appveyor.yml` file for reference. you can also integrate appveyor badges
[Build status](https://ci.appveyor.com/project/stephengroat/hashcat)

Problem builds on [MSYS x86_64](https://ci.appveyor.com/project/stephengroat/hashcat/build/job/pm4vna3xg20gq233) and [cygwin x86_64](https://ci.appveyor.com/project/stephengroat/hashcat/build/job/asxqgt7o5rmnaqbo)

PR #1025, opened 2017-02-07
Disable glob file under Cygwin

neheb:
Cygwin is fully capable of generating MinGW or native binaries. No
need to restrict it. Also removed a bug where the glob file would
be applied to native cygwin builds, causing a failed compile.

All of that should also apply to MSYS2 but I'll leave it since I
have not tested it.

### DIFF
--- a/src/win_file_globbing.mk
+++ b/src/win_file_globbing.mk
@@ -98,7 +98,7 @@ endif
 ## win native
 ##
 
-IS_WIN_BUILD_NATIVE := $(filter CYGWIN,$(UNAME))$(filter MSYS2,$(UNAME))
+IS_WIN_BUILD_NATIVE := $(filter MSYS2,$(UNAME))
 
 ifneq (,$(IS_WIN_BUILD_NATIVE))
 


stephengroat:
@neheb is there any test case I can use to test the functionality of the glob?

neheb:
make win32 or win64 makes use of it.

philsmd:
@neheb 
I'm just questioning if this is the right way to approach this problem, my concerns (I'm not trying to discourage you, I just want to fix this correctly or at least discuss this a little bit more):
1. may this be related to this "issue" https://github.com/hashcat/hashcat/pull/1024 ? I see no clear reference to the issue but it seems to be related and I'm wondering if the chance to have a combination of an issue and PR at almost the same time about the very same problem should be too low such that it doesn't seem very likely that these are 2 distinct "problems". Maybe you just forgot to mention it, dunno... or we are very (un)lucky and it was all just a coincidence (anyway, if you fix an issue, you should reference it in the future)
2. I don't think we should just disable file globbing like this
3. There was a commit (by @atom, this one https://github.com/hashcat/hashcat/commit/cd3fae958dbf28b79ce190f5759c8e5ce3e46006 ?) that explicity made the changes to support file globbing with the native target. Essentially, the changes within your PR, kind of silently revert this. I think atom added the support on purpose (not by mistake)... we should be a little bit careful here what happened within previous commits, just saying (no offense! this is something that I find myself struggling with, e.g. understanding why somebody added something and to forget to check the file history etc)
3. There is a whole separate section within win_file_globbing.mk labeled "win native" ... and a variable called IS_WIN_BUILD_NATIVE ... your changes kind of make this section useless and also the variable name now becomes kind of misleading (the result is no more "all native windows builds" as the variable name might suggest and it was previously).
4. I think the right way to fix this is to detect the architecture of the native build and always enable file globbing on windows (no matter if it was native or a "release"/binary build)... whenever the correct CRT_glob.o (for the correct architecture) can be found
5. btw: I also find this CRT_GLOB := $(shell $(FIND) / -name $(CRT_GLOB_FILE_NAME) -print -quit) a little bit too much ... at least when we look at where it tries to search i.e. the whole (mysys/cygwin/) file system which could be huge  (this has nothing to do with your PR, I just try to suggest/discuss some changes)... we should limit the search to the libs!

Therefore my suggestion would be to try to fix the root of the problem (not just disable file globbing in general for all native builds just because it sometimes uses the wrong CRT_glob.o since that file might be of a different/incorrect architecture).
I'm not saying that I fully grasp how all this cygwin/mysys differences work... but I suggest that we do not disable file globbing in general for native builds, because, as said, I think this was added on purpose.

I would be happy to discuss this with all of you... maybe I'm just totally wrong about my conclusions and if so please explain it here why we need to disable file globbing like this. (yeah, I admit I'm more used to work with linux so that might limit my understand here a lot).

As said, would be great if we can somehow eliminate my concerns and find a solution that does not just work around the build problems, but instead fix the root of the problems...
Please discuss it with me... I'm sure we can find a good solution that is acceptable by all of us (and hopefully that "just works" - preferably with file globbing enabled ;) -).

Thanks

---

BTW: I forgot to tell you that I really like your (past) contributions to hashcat and really appreciate them... I just think that here we need some "room" for discussion (at least to understand the main problem)

neheb:
Not related to #1024 .

The real issue is that the CRT glob is being used for a non-windows platform (that includes Cygwin) even though it is a MinGW specific thing. Cygwin and MSYS2 (by default) do not generate windows binaries that are usable without Cygwin.dll.

The main use case of compiling hashcat with Cygwin is to get proper SSH support (MinGW binaries don't work correctly with Cygwin's sshd due to a difference in input handling).

Oh I also just remembered. I made this after I noticed 100% CPU usage in Cygwin's find command after running make. Although FWIW I've been commenting it out for a while.

jsteube:
@neheb The fix for IS_WIN_BUILD_NATIVE is ok, but the fix for cross-compilation is not. The cross-compilation targets are ment to be called from linux only and to generate the binary package. If we remove it its too confusing. Can you please modify the PR ?

philsmd:
Thanks @neheb for explaining this. I didn't know this.
I suggest that we document that (with some more comments in win_file_globbing.mk) such that we keep this information safe for the future.

So, if I'm not totally mistaken, file globbing should work with both cross-compiling (mingw) and non-cross-compiling, MYSYS and cygwin. For cygwin it's a little bit different and we should make this clear also in win_file_globbing.mk (by changing the code and/or adding comments).
Could you do these changes, @neheb?

Thanks again

neheb:
Last I remember, MSYS2 creates several shortcuts. The regular MSYS2 is basically equivalent to Cygwin. The MinGW ones compile with MinGW and also change the uname.

File globbing only works with MinGW targets. Will do some testing on Windows and update the PR.

edit: at least MSYS2 fails properly: 

find: failed to read file names from file system at or below ‘/’: No such file or directory
src//win_file_globbing.mk:117:

! The MinGW CRT GLOB library was not found on your system. Please make sure that CRT_glob.o exists
! ATTENTION: File globbing will be disabled

Actually this makes sense. /cygdrive is where regular drives get mounted. Must be trying to read the whole file system.

neheb:
Going to resend. Something got messed up.

PR #1031, opened 2017-02-08
Add NVML and NVAPI support to Cygwin. Disable file globbing.

neheb:
Runtime tested. The NVML/NVAPI stuff ehaves the same as a MinGW build.

Typedef was changed as long is 64-bits under Cygwin.

### DIFF
--- a/include/ext_nvapi.h
+++ b/include/ext_nvapi.h
@@ -12,8 +12,8 @@
 
 #define NVAPI_INTERFACE extern NvAPI_Status
 
-typedef unsigned long NvU32;
-typedef   signed long NvS32;
+typedef unsigned int NvU32;
+typedef   signed int NvS32;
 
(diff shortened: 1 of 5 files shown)


philsmd:
I just made some tests with appveyor (I didn't know this even existed before @stephengroat made his PR). See: https://ci.appveyor.com/project/philsmd/hashcat/build/1.0.2/job/0wcvyc7ncdh84o85#L597
There still are some errors with your latest code @neheb.
Since I can't guarantee that my appveyor_test branch is always up-to-date, could I kindly ask you that you also test with appveyor and link your appveyor test builds here (with "neheb" in the URL, you need to sign in, e.g. with oauth to appveyor)?

Do you know how these remaining errors can be correctly fixed?

Thank you very much for your contributions!

jsteube:
Both changes look good to me. Not sure what's going on with AppVeyor but this seems to be a different issue. We should open an issue for it, but that's not related to this PR

PR #1036, opened 2017-02-09
make install error on MacOS Sierra

gum0x:
macOS Sierra does not uses gsed, but sed. Makefile set gsed for all Darwin based OSes.

1. Added OS Version on Makefile validation for Darwin platform.
    If OS type is Darwin and OS Version is 16, sed is command is used instead of gsed. 
2. Added SEDOPS variable in order to construct the correct options syntax depending on the system who executes the sed command. 
    On macOS's sed command, -i option needs to explicitly set at least an empty string as a parametre.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -16,6 +16,11 @@ UNAME                   := $(shell uname -s)
 # we need to strip the windows version number to be able to build hashcat on cygwin hosts
 UNAME                   := $(patsubst CYGWIN_NT-%,CYGWIN,$(UNAME))
 
+##
+## Detect Operating System version
+## 
+OSVERSION		:= $(shell uname -r|cut -d. -f1)
+
 # same for msys
 UNAME                   := $(patsubst MSYS_NT-%,MSYS2,$(UNAME))
 UNAME                   := $(patsubst MINGW32_NT-%,MSYS2,$(UNAME))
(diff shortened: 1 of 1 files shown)


philsmd:
Hey @gum0x, thx for your contribution.
I think the formatting (indentation, tab vs spaces etc) got mixed up a little bit within these changes (it seems that you used tabs in places where we always use spaces etc).
Could you please try to fix this?

BTW: I think this variable name "OSVERSION" is a little bit confusing, because under linux the result seems to be just the first digit of the linux kernel number. Maybe we should add this variable to the Darwin section (because it is something specific to OSX?) and call it something like OSXVERSION ? (As you see, I'm not totally sure if my assumptions of this osxversion vs kernel version numbering are correct... maybe you can find this out too ? Thank you so much)

---

Update: maybe we should make it even more generic and replace the check with a check for versions >= 16 ?

magnumripper:
This would work
```
OSDARWIN		:= $(shell uname -s | grep -c Darwin)

(...)

ifeq ($(OSDARWIN),1)
    (...)
```

magnumripper:
...but I see now there already is some `ifeq ($(UNAME),Darwin)` in there. So just moving the stuff around should do. But why limit it to 16 at all? Darwin <16 didn't have gsed either, yes?

gum0x:
Hi, 

Sorry for the delay on my answer. I have been disconnected these days. The fact is as I saw a special condition for Darwin OSes, I supposed all versions except MacOS Sierra used gsed instead of sed. I did not tested in previous versions of OSX. 

Thanks for the fix and the corrections on the identation. Let me know if you need something from my side. 

Regards, 
Josep.

Issue #1032, opened 2017-02-08
Fix file globbing with win (e.g mingw64) and add test for appveyor


philsmd:
We recently saw here (https://github.com/hashcat/hashcat/pull/1024, thanks @stephengroat)  that tests with appveyor would be a nice test addition.
Furthermore, we recently found out that appveyor does in fact report some problems even with the latest hashcat version (https://github.com/hashcat/hashcat/pull/1031#issuecomment-278268781).

Therefore I suggest:

1. we add the .appveyor.yml file (similar to the one here: https://github.com/hashcat/hashcat/pull/1024) and sign up for appveyor test builds
2. we test the appveyor buids and fix the file globbing problems
3. we fix all additional problems that appveyor might report
4. add the appveyor badges to README.md
5. we should add some hashcat-run tests (preferably to the appveyor tests, if possible) to check if the globbing really works (like hashcat -m 0 --show *ple0.hash, it should expand the name to example0.hash and if globbing didn't work the exit code should be different from 0)

Maybe you, @neheb, could have a look at this issue? You seem to grasp the problem with file globbing and the different environments that we need to test (cygwin/mingw/mysys etc) more than anyone else here.
Thanks

---

I would suggest that the person that tries to implement all of this,  #signs up (with his/her own account) to appveyor and links his/her test results here. If everything seems good and after merging, this github user ("hashcat") should sign up too and link the master branch + add the badges to the README.md

neheb:
I'll need to test this again but I believe the MinGW shortcuts provided by MSYS2 change environment variables to help out configure scripts find the right binaries.

Now, that's important to mention since the hashcat makefile just selects standard GCC under the MinGW environment and compiles a standard POSIX binary.

I can't say I want to work on this since I'm not an MSYS2 user. They both have different use cases.

stephengroat:
i was still modifying my appveyor script, i can start a new pull request and allow both maintainers and anyone interested to edit

the only problem that i currently see if that msys 64bit builds currently fail because they find the i686 `CRT_glob.o`.

I think adding this to main removes the need to find the `.o`:
````c
#if defined(__MINGW64__) || defined(__MINGW32__)
int _dowildcard = -1; 
#endif
````

Quick question: how do you run tests? it's not just running tools/test.sh, right?

philsmd:
[...]

PR #1042, opened 2017-02-10
cleaned up the .appveyor.yml file and implemented the test_script(s) #1032

philsmd:
This implements the test_script which run on cygwin32/64 and mingw32/64.
The tests for now are very simple: just test for file globbing.
(other tests might be usefull, but it seems we can't really run hashcat.exe without a working OpenCL driver environment).

Thanks


jsteube:
Thanks

PR #1043, opened 2017-02-10
Cleanup zero-length printf format string warnings

stephengroat:


jsteube:
The use of NULL for the format variable is something gcc doesn't like much. Given the code:

```
printf(NULL);
```

gcc returns with:

```
warning: null argument where non-null required (argument 1) [-Wnonnull]
```

So this seems to be the wrong approach to me. But I can see the need for it. There's another way to do it. Instead of 

```
event_log_info (hashcat_ctx, NULL);
```

one could do

```
event_log_info (hashcat_ctx, "%s", "");
```

This makes every compiler happy but not us coders, because that looks kind of strange.

So as another alternative I'll try to use the patch as you suggested and add a check for null pointer and then let's see how the compiler react. There's many different compilers and versions and it seems they all react different to this.

PR #1044, opened 2017-02-11
Fix more cppcheck warnings

neheb:
Please verify the changes. I'm a little unsure about the variable++ removals. cppcheck complains that the variable's assigned value goes unused.


jsteube:
It's fine, the num_files++ can be remove. The only reason it's there is because of this:

```
    files[num_files] = hcstrdup (path);

    num_files++;
```

By using a code like this:

```
  files[num_files] = NULL;

  num_files++;
```

It's easier for the coders' eye to identify that there's some "pattern" in which two variables are involved, but it's not that important.

PR #1048, opened 2017-02-11
travis test: workaround undefined _WIN32 with a sed fix

philsmd:
This should fix some warnings that only seem to appear with older gcc/clang compilers.
But the root of the problem is that the #if #elsif etc do miss the defined () check.

Thank you


jsteube:
Very nice, thanks!

PR #1053, opened 2017-02-12
Remove defined macros in favor of preprocessor defaults

neheb:
Totally untested on the BSDs but should work. Compile and runtime tested on MinGW[32/64] and Cygwin (hashcat -b). 

A few notes: __WIN32__ is defined on both 32 and 64 bit versions of MinGW (for stupid Microsoft reasons) so it can be used for both.

ADL support was added to Cygwin. Totally untested since I have no hardware but should work.

Miscellaneous cleanups here and there.

### DIFF
--- a/include/affinity.h
+++ b/include/affinity.h
@@ -9,12 +9,12 @@
 #include <stdlib.h>
 #include <stdarg.h>
 
-#if defined (_POSIX)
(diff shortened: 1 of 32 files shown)


neheb:
Aaaaaannnd the BSDs are where this fails. Closing for now.

PR #1100, opened 2017-02-15
Remove defined macros in favor of preprocessor defaults

neheb:
2nd attempt.

Kept _POSIX since __unix__ is not defined on BSDs. Also added support for ADL on Cygwin. Totally untested(no hardware) but should work.

### DIFF
--- a/include/affinity.h
+++ b/include/affinity.h
@@ -22,9 +22,9 @@
 #include <mach/thread_policy.h>
 #endif // __APPLE__
 
-#if defined (_WIN)
+#if defined (__WIN32__)
 #include <windows.h>
-#endif // _WIN
+#endif // __WIN32__
 
 int set_cpu_affinity (hashcat_ctx_t *hashcat_ctx);
 
--- a/include/common.h
+++ b/include/common.h
@@ -8,22 +8,17 @@
 
 #define PROGNAME "hashcat"
 
-#if   defined (__linux__) || defined (__CYGWIN__)
+#if defined (__unix__) || defined (__APPLE__) || defined (__FreeBSD___) 
 #define _POSIX
-#elif defined (__APPLE__)
-#define _POSIX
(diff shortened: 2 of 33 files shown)


magnumripper:
Strange, `__unix__` should be defined on all BSDs except MacOS and `_POSIX` should be less safe to use. I think you updated your patch though.
```
-bash-4.4$ uname -mprsv
OpenBSD 6.0 GENERIC.MP#116 amd64 amd64
-bash-4.4$ cc -dM -E -x c /dev/null | grep -iE 'unix|posix'
#define __unix__ 1
-bash-4.4$ gcc -dM -E -x c /dev/null | grep -iE 'unix|posix'
#define __unix__ 1
```

jsteube:
I'm not sure if I like the changes to _WIN. It feels much better to read "WIN" instead of "WIN32" because "WIN32" could be misunderstand that a 64 bit compiler will not use this macro (even if it does, i've just verified). I think we should stick to the current _WIN.

Also what's confusing is this:
```
+#if defined (__unix__) || defined (__APPLE__) || defined (__FreeBSD___) 
```
but according to @magnumripper the __FreeBSD__ should also use __unix__ ?! since I don't have a BSD system I can not verify myself, I need a good recommendation.

The ADL on Cygwin sounds good.

Another thing I saw is that you've changed copy/paste headers from 3rd party sources, like ADL. I'd prefer to have them non-modified, even if they do not match our format settings. I know it happend before but I try to keep them as they are.

jsteube:
I'm closing this as it's already not longer auto-merging. Can you please resend in the ADL HW-Monitor support for cygwin? About the other stuff (Macros) we should discuss separate.

neheb:
Using __WIN32__ was done for consistency purposes. The other relevant ones also have two __ on both sides. Anyway, I'll resend the ADL stuff.

PR #1114, opened 2017-02-18
Add ADL support to Cygwin

neheb:
Untested but should work.

### DIFF
--- a/include/ext_ADL.h
+++ b/include/ext_ADL.h
@@ -45,13 +45,13 @@ typedef struct AdapterInfo
   char strDisplayName[ADL_MAX_PATH];
   int  iPresent;
 
-  #if defined (_WIN32) || defined (_WIN64)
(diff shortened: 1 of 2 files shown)


philsmd:
It seems that appveyor reports some errors related to these changes: https://ci.appveyor.com/project/jsteube/hashcat/build/1.0.211/job/po6xvvd8ootrkas0#L336 and https://ci.appveyor.com/project/jsteube/hashcat/build/1.0.211/job/6kay6311dhde3iie#L333

I'm also wondering if we should change this "_WIN64" checks like this... I would prefer just adding the CYWIN support for now and discuss the macro problems (WIN32, WIN64, WIN etc) within a new issue/PR, or is my assumption wrong that the cygwin changes are/should be unrelated?

neheb:
fixed.

PR #1120, opened 2017-02-21
Fix typo in README file (from Paul Wise)

LocutusOfBorg:
see https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=855687


philsmd:
Thank you @LocutusOfBorg.
As you may have noticed, we did some further changes at the project's description, see: https://github.com/hashcat/hashcat/commit/d200b8493e30f25eaae441107f8d059c0f009a59

Maybe you can do the same changes for the package description?
Thx

update: hmm, the 200+ algo change might be only correct with the newest version... but the grammar correction should be applied already now to both (packages and README.md)

LocutusOfBorg:
fixing shortly!

PR #1126, opened 2017-02-22
add autoconf/automake supports

yhfudev:
The Makefiles are handled by autoconf/automake, so

* the software can be cross compiled without specify 32 bit or 64 bit version.
* the release file can be generated with command
    make dist-gzip
* tested in Ubuntu

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,21 +8,29 @@ environment:
       CYG_SETUP: setup-x86_64.exe
       BASH: C:\cygwin64\bin\bash
       CC: gcc
+      DOCUMENT_FOLDER: /usr/local/share/doc/hashcat/
+      SHARED_FOLDER: /usr/local/share/hashcat/
     - CYG_ROOT: C:\cygwin
       CYG_CACHE: C:\cygwin\var\cache\setup
       CYG_SETUP: setup-x86.exe
       BASH: C:\cygwin\bin\bash
       CC: gcc
+      DOCUMENT_FOLDER: /usr/local/share/doc/hashcat/
+      SHARED_FOLDER: /usr/local/share/hashcat/
     - MSYSTEM: MINGW64
       MSYS_CACHE: C:\msys64\var\cache\pacman\pkg
       KERNEL_CACHE: C:\msys64\usr\local\bin\OpenCL\
       BASH: C:\msys64\usr\bin\bash
       CC: gcc
+      DOCUMENT_FOLDER: /bin/
+      SHARED_FOLDER: /bin/
     - MSYSTEM: MINGW32
       MSYS_CACHE: C:\msys64\var\cache\pacman\pkg
       KERNEL_CACHE: C:\msys64\usr\local\bin\OpenCL\
       BASH: C:\msys64\usr\bin\bash
       CC: gcc
+      DOCUMENT_FOLDER: /bin/
+      SHARED_FOLDER: /bin/
 
 # if we have too many commits at the same time, we might need to download more than just the last commit for appveyor to succeed
 # otherwise we get the error: "fatal: reference is not a tree <commit>"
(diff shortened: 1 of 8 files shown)


jsteube:
I'll probably merge this after v3.40 release which is already in RC phase (so that it will go into v3.50). Great PR!

stephengroat:
@yhfudev you could also update the `.travis.yml` and `.appveyor.yml` files

for `.travis.yml`:
````
script:
  - ./autogen.sh && ./configure && make
````

for `.appveyor.yml`:
````
build_script:
  - if defined BASH (%BASH% -lc "cd $(cygpath ${APPVEYOR_BUILD_FOLDER}) && git submodule update --init && ./autogen.sh && ./configure && make install")
````

also, should you remove the existing `Makefile`s?

**Comment by yhfudev** [comment]
@stephengroat

Thanks! I updated the CI scripts so it can be compiled in cygwin, mingw, linux, and osx now.

**Comment by stephengroat** [comment]
@yhfudev looks nice!

since you're overriding `CPPFLAGS=-I$(pwd)/deps/OpenCL-Headers/` for every build, do you just want to add it to the `Makefile.am` to make it the default?

**Comment by jsteube** [comment]
I've taken some time to look into this. I must admit I'm still not convinced using autoconf/automake.

First of all when I run ./autogen.sh I get tons of warnings. From technical perspective this is usually not a problem, but from a user perspective it gives the impression that the developers did not even test their own tools and the first impression counts for software, too. The expectation of the entire toolset is lowered from the beginning.


```
chmod: cannot access 'config.guess': No such file or directory
chmod: cannot access 'config.rpath': No such file or directory
chmod: cannot access 'config.status': No such file or directory
chmod: cannot access 'config.sub': No such file or directory
chmod: cannot access 'configure': No such file or directory
chmod: cannot access 'depcomp': No such file or directory
chmod: cannot access 'install-sh': No such file or directory
chmod: cannot access 'missing': No such file or directory
```

```
src/Makefile.am:137: warning: shell date +%s: non-POSIX variable name
src/Makefile.am:137: (probably a GNU make extension)
src/Makefile.am:139: warning: shell cd "$(top_srcdir: non-POSIX variable name
src/Makefile.am:139: (probably a GNU make extension)
```

Why is this GPL now? We're on MIT


```
Makefile.am: installing './COPYING' using GNU General Public License v3 file
Makefile.am:     Consider adding the COPYING file to the version control system
Makefile.am:     for your code, to avoid questions about which license your project uses
```

It then completely failed because on my system there was no package "libtool" installed. So it called "libtoolize" (whatever that is)  but still tried to continue after it failed to locate it. Of course this did not work but for a user with no development experience he would never find out how to solve this. 

After that it automatically ran ./configure with tons of options we don't care at all. For example:

```
checking how to run the C++ preprocessor... g++ -E
```

This is a C source, not C++. Why is this check done at all? Can't we disable it?

Finally it finished and I'm ending up with a mess created by tons of files that I have no clue what they are doing. None of these files existed before:


```
	AUTHORS
	COPYING
	ChangeLog
	INSTALL
	Makefile.am
	Makefile.in
	NEWS
	aclocal.m4
	ar-lib
	autoclean.sh
	autogen.sh
	autom4te.cache/
	compile
	config.guess
	config.h
	config.h.in
	config.h.in~
	config.status
	config.sub
	configure
	configure.ac
	depcomp
	hashcat-3.3.0.tar.gz
	install-sh
	libtool
	ltmain.sh
	missing
	src/.deps/
	src/Makefile.am
	src/Makefile.in
	src/libhashcat.pc
	src/libhashcat.pc.in
	src/lzma_sdk/.deps/
	stamp-h1
```

This is ugly and confusing. Whenever there's a problem with it, I wouldn't even know where to start looking.

I see that you tried to generate some sort of auto-documentation, but I'm really not a fan of this. Auto-generated documentation out of git's log can not be trusted because we can not force a good documentation of the changes verified from a technical system (git). Therefore changes should always be hand-written.

I think there's also a misunderstanding of the package distribution. The script tools/package.sh is the one script which creates the files for distribution of Linux and Windows binaries. Especially Windows binaries. But there's no windows binary in the final package hashcat-3.3.0.tar.gz. With our hand writte Makefile we call "make binaries" to compile them using mingw GCC. But I do not see a single use of the substring "mingw" in autogenerated src/Makefile so there's defenitly a major feature missing. Does autoconf stuff even support cross compiler?

I see some files/notices that references hashcat v3.3.0. We call that v3.30 but mean the same. We maybe change this to v3.3.0 notation in a later version.

Finally I still don't see the need of all this. With our hand-written Makefile all the feature we want are working fine. There's no warnings, no errors and it works on multiple systems. Maybe not all but at least on such where the device drivers come with a working OpenCL runtime. All the others are not useable for us anyway?!

yhfudev:
@stephengroat

The user should use their own OpenCL-Headers, not this one, and we use it here is for compiling test.

yhfudev:
@jsteube

From your post above, my only fault is not create the files AUTHORS, COPYING for you. The other questions, please google by yourself, here's the hints,

* most of the files are generated by autoscan/autoconf/automake
* the start point for user is 'configure', not 'autogen.sh'
* it can be compiled in expected platforms (updated .travis.yml and appveyor.yml)
* 'configure' is a 'API' for distributions.
* do some experience of packing binaries for various distributions(rpm,apt,pacman) if not sure.

magnumripper:
A bunch of files (eg. autom4te.cache, config*.cache and config.status) should be added to .gitignore

stephengroat:
@magnumripper you could add the https://github.com/github/gitignore/blob/master/Autotools.gitignore

magnumripper:
Yeah, and it should be added to this PR

jsteube:
With my comment I hoped you maybe can convince me why it makes sense to replace our really good working hand-written Makefile with autoconf. The argument "do some experience of packing binaries for various distributions(rpm,apt,pacman) if not sure." reads like "because everyone is doing it" which is something I don't care much about. Plus the Makefile as it is right now works pretty well for them. There's packages for many Linux distributions, even FreeBSD. 

At this time I don't see why we should switch to autoconf and therefore I will close this PR. I'm not saying we'll never switch, but at this time I'm simply not convinced.

@yhfudev @stephengroat I would like to thanks for your work involved. I just hope that rejecting this PR will not stop you from contributing to hashcat.

PR #1130, opened 2017-02-23
Prefer _WIN to _POSIX for defines.

neheb:
Also miscellaneous fixes here and there.

mainly a consistency update.

defined( > defined (

WIN is completely unused in the code.


jsteube:
Thanks for PR! From what I've seen I think it's no longer required to do this:

```
#include <sys/types.h>
#include <sys/stat.h>
```

in both pidfile.h and restore.h. I think we can remove them. What you think?

PR #1145, opened 2017-03-01
fix rule 'i' in host mode in case pos = length

### LINKED ISSUE #1144: Possible wrong behaviour for `i` rule with -j option
Similar to #969 but with `i` rule.

```
$ ./hashcat --stdout -j 'i1u' /tmp/1
1
$ ./hashcat --stdout -r /tmp/d /tmp/1
1u
$ cat /tmp/d
i1u
$ cat /tmp/1
1
$ ./hashcat --version
v3.30-382-g50b46d5d
```

mhasbini:
close #1144 

to test:

```
$ echo -n 1 | ./hashcat --stdout -j 'i1u'
1u
```


jsteube:
Thanks!

PR #1195, opened 2017-03-20
Update outfile.c

### LINKED ISSUE #1194: outfile_write method does not output crack_pos.
When using --outfile-format=11, hashcat is suppose to include the guess number as the last item.  However, this was the output:

a2593e6a5ae8ba54b95ff7b182d738a8:jblaze:

Looking at outfile_write, I tracked down the problem.  Variable 'tmp_len' needs to be increased in size to accommodate the crackpos text.  Here is my rudimentary fix on line 439:

435:  if (outfile_ctx->outfile_format & OUTFILE_FMT_CRACKPOS)
436:  {
437:    sprintf (tmp_buf + tmp_len, "%" PRIu64, crackpos);
438:  }
439:  tmp_len += 20; // Need to move end of the buffer to include crackpos.
440:  tmp_buf[tmp_len] = 0;

After adding line 439, the output is now:
a2593e6a5ae8ba54b95ff7b182d738a8:jblaze:9612916440

I'm using version 3.40 of hashcat.

syntaxmonkey:
Fix #1194.
crackpos was not being output properly when "--outfile-format 11" was used.

### DIFF
--- a/src/outfile.c
+++ b/src/outfile.c
@@ -436,6 +436,7 @@ int outfile_write (hashcat_ctx_t *hashcat_ctx, const char *out_buf, const unsign
     sprintf (tmp_buf + tmp_len, "%" PRIu64, crackpos);
   }
 
+  tmp_len += 20; // Required to include the text generated for crackpos.
   tmp_buf[tmp_len] = 0;
 
   if (outfile_ctx->fp != NULL)
--- a/src/outfile.c
+++ b/src/outfile.c
@@ -343,6 +343,17 @@ void outfile_write_close (hashcat_ctx_t *hashcat_ctx)
   fclose (outfile_ctx->fp);
 }
 
+int numDigits(u64 number)
+{
+    int digits = 0;
+    while (number) {
+        number /= 10;
+        digits++;
+    }
+    return digits;
+}
+
+
 int outfile_write (hashcat_ctx_t *hashcat_ctx, const char *out_buf, const unsigned char *plain_ptr, const u32 plain_len, const u64 crackpos, const unsigned char *username, const u32 user_len, char tmp_buf[HCBUFSIZ_LARGE])
 {
   const hashconfig_t   *hashconfig   = hashcat_ctx->hashconfig;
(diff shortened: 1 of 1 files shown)


philsmd:
I see 3 problems with this PR:
1. why should we use the value 20 (hardcoded)? this doesn't make any sense to me
2. why do you not increase the length within the  OUTFILE_FMT_CRACKPOS branch? The value should only be increased if that specific condition holds and therefore only if that branch was taken
3. we should use snprintf () instead of sprintf () and check that we do not use more bytes than HCBUFSIZ_LARGE (this was not a problem with your PR, but we should fix this anyway)

Can you please try to update your PR?
thx

---

update: forgot one important problem:
4. we need to update docs/changes.txt too to mention the problem that versions <= 3.40 didn't work well with crackpos

syntaxmonkey:
@philsmd - changes made as suggested.

The changes.txt appears to be committed changes based on version.  I'm not entirely clear how this change should be added.

philsmd:
For what regards docs/changes.txt, you just need to add something like this:
"- Fixed --outfile-format 11 up to --outfile-format 15: the crack position was not correctly outputted"

within the "Bugs" section:
```
##
## Bugs
##
```

Furthermore, I'm not sure if we really need this new numDigits () function (also: it should respect the coding standards, see README.md, e.g. num_digits () etc...), because snprintf already is able to tell us the number of bytes that are needed for the digits. That means, it is kind of useless to ignore the return value, but instead introduce a further function that does basically the same.

There is some further room of improvement that comes to my mind (but doesn't really affect only your PR): we shoud also make sure that whenever we do:
`tmp_buf[tmp_len] = 0;`

that tmp_len is never greater or equal to HCBUFSIZ_LARGE, e.g something like this code should be used:

```
tmp_len = MIN (tmp_len, HCBUFSIZ_LARGE - 1);
tmp_buf[tmp_len] = 0;
```

What do you think?

BTW: thank you very much for your contributions, these are some great finds and fixes!

PR #1216, opened 2017-04-03
standardize usage output

roycewilliams:
Updates names of some software to match official capitalization or punctuation.


philsmd:
I think we should avoid having 2+ labels/names for the same algorithms. Therefore we should also change the names in https://github.com/hashcat/hashcat/blob/master/src/interface.c#L79-L302
Do you agree?

roycewilliams:
@philsmd - Concur 100%. I should have caught that. I think that I have them synced now. It was relatively easy to sync usage.c and interface.c (because I could automate comparing them), but docs/readme.txt is a whole other animal (significantly different names, multiple hash types rolled up into single entries, different sort order), so I just manually fixed the obvious things like case in readme.txt, but otherwise had to defer for time/consensus. 

I tried to research the conflicts -- such like exact version numbers in CMSes, etc where hashing methods changed -- but I may have made some errors, so a sanity check would be great.

roycewilliams:
Once this PR is stable and accepted, I'll also sync the names as shown in the example hashes page on the hashcat.net wiki.

roycewilliams:
Wiki is synced now, except for some differences due to specific hash examples being different even though the hash mode is the same (MD5 phpBB vs Joomla, etc.) or other such purpose-specific differences.

Issue #1218, opened 2017-04-05
make clean is dangerous


magnumripper:
I'm writing this as an issue instead of a patch because we might disagree.

In my humble opinion, `make clean` should **absolutely not** destroy any pot files or session files. Instead, they should be listed in .gitignore so they don't turn up in eg. `git status` or accidentally get added in a commit.

If you really want to, you could have an alternative target `make distclean` that actually does destroy such files, but in the world of git it's not really needed. For example, you'd better create archives (eg. zip files) using `git archive` instead of a "manual" zip command. The former will never include anything not actually in the git tree.

jsteube:
OK, we should do that

PR #1219, opened 2017-04-06
Ensure 'make clean' is a safe operation

magnumripper:
See #1218.


magnumripper:
Not sure about .induct and .outfiles, should they too be moved to distclean?

jsteube:
The .induct and .outfiles are temporary files which are no longer valid after the session terminates. We should clean them.

PR #1223, opened 2017-04-10
fix strings

stephengroat:
enable and fix const string warnings


roycewilliams:
Are the whitespace differences a correction to follow the existing standard, or were they inadvertent?

stephengroat:
on purpose, attempts to follow the existing alignment

jsteube:
thanks!

PR #1226, opened 2017-04-16
Blake2b raw hash implementation 

DoZ10:
Hi,

I'm pretty new to OpenCL and did try a shot to implement Blake2b raw hash function. My work is based on hashcat's SHA512 implementation and adapted it to the Official Blake2b reference https://github.com/BLAKE2/BLAKE2.

Unfortunately, I'm kinda stuck with the kernel due to my lack of experience with OpenCL, and debugging gets frustrating with all those bswaps and architecture ninjutsu. Interface, should be okay tho. Target hash is the one I am hunting for in my test files: b2test.sh & b2test.hash

If anyone experienced with hashcat could give me directions, I would really appreciate to learn more about OpenCL impl. for hashcat. 

Thanks,


DoZ10:
Ok might have found a way to tshoot this. All GPU debuggers failed on my Ubuntu system, but I could probably return values in buffers of mark() call. Will look at this tomorrow.

What you guys use for debugging ?

jsteube:
You can debug a kernel using printf() in case you have a nvidia GPU.
There's a few things you should do additionally:

1. Set DEBUG mode to 1 in Makefile

2. Run you test with -u1 -n1 --force (this is a special condition that
will disable the autotuner which otherwise would confuse printf output)

3. Additionally use -d 1 in case you have multiple GPU

4. Use --weak-hash-check 0 (also to reduce confusion of printf output)

In case you have an AMD GPU you need need to force hashcat to use the
JiT kernel without caching, otherwise printf() fails. Find the line that
says:

opencl_ctx->force_jit_compilation = -1;

In the opencl.c source and replace the -1 with your kernel number (600).

DoZ10:
Just peed on myself. Never been so happy to have printf() working.

Thx rly much... will fix my crap.

DoZ10:
Alright, debugged the kernel code for single hash function s04.

Can you please confirm coding style is okay, before I move on to other functions and kernel files ?

Also, I am using the right params for the right kernel purpose or did I hack something in a useless way ? (I am not talking about optimizing my own code, but rather if I am using the right kernel parameters/macros/structs/functions.).

Finally, a quick description of m04, m08 and m16 funcs ? Do they relate to different input dimensions or kernel threads dimensions ? 

Thanks for helping me out !

jsteube:
I don't know much about the feature of blake2b, but it seems it has some sort of built-in salting support. If that's the case, then the hash format need to have each salt somehow stored along with each hash. From your example hash, it doesn't seem to do so. 

OTOH, hashcats hash formats/encoding traditionally tries to stay compatible to jtrs. Since jtr already supports blake2b we should make the format 1:1 to the the one from jtr. That will also give a good inspiration about how to configure the different blake2b features (per hash).

Here are some more things that looks suspicious in your code:


1. This can not be true if the hash is salted: 

```
hashconfig->salt_type      = SALT_TYPE_NONE;
```

2. This function seem to be just byte swaps:

```
u64 load64(const void *src)
u64 load64_reverse(const void *src)
```

There's swap64() in OpenCL/inc_common.h maybe you find a way to simply replace them with our functions.


3. Try to avoid void * in the kernel, that is compatible to all OpenCL runtimes:

```
u64 load64(const void *src)
```

4. This function seem is simple bit rotate:

```
+u64 rotr64_w(const u64x w, const u32 c)
```

We already have rotr64() in the types.h. Please try to replace it, too.

About your questions:

> Finally, a quick description of m04, m08 and m16 funcs ? Do they relate to different input dimensions or kernel threads dimensions ?

the "m" stands for multiple hashes, where "s" stands for single hash. With a single hash, we can do a single integer comparison at the end ofthe kernel, while with "m" the kernel has to run through a bloom filter and some binary tree. For you it's not so important, it basically the same code, but because of performance reasons it's not implemented in form of a branch. Just take a look at any other fast hash kernel, for example md5. The only different is that you do not have a search[] array and the COMPARE_S_SIMD() is replaced with COMPARE_M_SIMD().

> Also, I am using the right params for the right kernel purpose or did I hack something in a useless way ? (I am not talking about optimizing my own code, but rather if I am using the right kernel parameters/macros/structs/functions.).

I can't say too much about that as long as you do not full make use of the blake2 features because this changes some stuff in the code, you will find out when you do.

Also feel free to join #hashcat on freenode in case you need direct help

DoZ10:
Understood. 

Yes there is a built-in salt in Blake2b and also a variable size output, but I was planning to add it at the very end to avoid too much confusion since this is my first kernel implementation... you know what they say... Romantic Dinner before backseat P0rn. But, I'll comply with JtR format and apply the changes during the weekend.

Thanks.

DoZ10:
Alright.. so looking at JtR 
https://github.com/magnumripper/JohnTheRipper/blob/8bcaddef35f703c47367fad612e473061459b156/src/rawBLAKE2_512_fmt_plug.c 

it looks like optional salt is not implemented for this RAW 512 bit implementation also with fixed size output. I guess this would make more sense in the Argon2 implementation.

It should not be a hurdle to implement this (variable hash length + salt), but is it preferable to do it with another -m number such as 610 ?

For now, all 3 kernels are working and were tested. Hash format works against JtR. Please let me know the final steps, if any, to get this work merged.

Thanks,

DoZ10:
Okay there is a problem with test.pl... will have to fix this first.

If anything else to look at comes to mind, please let me know.

jsteube:
Please get test.sh and test.pl working first, it's important for tests

jsteube:
I can't check the test.pl unless the conflict is resolved. However, here's at least one odd thing i saw by watching the source:

blake2b_compress(pw, pw_len, digest);

pw_len should be out_len for -a 0 kernel.

DoZ10:
I can see what is wrong, but I don't know the best approach to the problem. I am now testing only with Attack Mode 0. VECT_SIZE will define some kind of array to every u??x types, so I need to process this in "parallel" in the algorithm.

Any thoughts on the best way to achieve this, I still feel there is one piece missing when I look at other implementations...

jsteube:
I've tested all modes (0, 1, 3, 6, 7) and they seem to work fine on the tested systems:

- nv 750ti
- amd rx 480
- amd r9 290x

Some notes:

1. It would be nice to have a blake2b_transform function for _a0 and _a1 instead of having the code embedded into the most inner loop. reason here is that we need it in case we need to transport it to a new hash mode (one that is using blake2 interal, ex: pbkdf2-hmac-blake2b) -- Ignore this it already exist!
2. For this reason (1) it would be nice to have the salt and pepper supported by the function. Even for mode 600 we should use this function without suffering for performance drop, because the compiler are clever enough to recognize that the "faked" initial buffer we are all set to zero.
3. I've noted that you use the vector-datatypes for the blake2b_state structure only in hash mode 0, but not in mode 1 and 3, example: --> u64x h[8]; in a0, u64 h[8]; in a1
4. The same for the blake2b_param struct, but here it makes sense, because we always handle the same salt within a single call of the kernel and therefore3 the parameters must be the same for all password candidates. therefore you should switch those to the scalar types, not the vector types
5. For performance reason you can try to move the blake2b_sigma[] array into the inner kernel as fixed data block, maybe it gives you a performance increase
6. please unroll the following section and do the assigns by hand:

```
    S->h[i] ^= *((u64x*)(p + sizeof(S->h[i]) * i));
```

7. P->digest_length seems to be unused. I'm not familar with blake2b, is this some kind of feature like it is in pbkdf2 where you can specify a desired output length? However it seems not to be used, maybe you want to create blake2b_param_optimized structure instead and remove all the unneccessary elements
8. If you do not use a structure to hold the state but only the elements from it as single integer arrays you help the compiler because it does not require to hold the data in that particular order, its faster
9. For our source code convention, please replace code like this:

```
  u32 i;
  ...
  for (i = 0; i < 8; ++i) 
```

with this:

```
  for (int i = 0; i < 8; i++) 
```

It's easier to read

10. This one looks suspicious:

```
    m[i] = swap32(pw[i * 2]);
    m[i] <<= 32;
    m[i] |= swap32(pw[i * 2 + 1]); 
```

I guess what you want to use instead is this:

hl32_to_64() see OpenCL/inc_types.h

11. Please unroll this section by hand:

```
  for (i = 0; i < 8; ++i)
    v[i] = S->h[i];
```

Please do the above, then it will be easier for me to understand the code and to start debugging

magnumripper:
> 11. Please unroll this section by hand:
> ```
>   for (i = 0; i < 8; ++i)
>     v[i] = S->h[i];
> ```

Why would you want things like that manually unrolled? Just curious.

jsteube:
If you take a look below, v[8] .. v[15] is initially set, too.

```

 v[ 8] = blake2b_IV[0];
 v[ 9] = blake2b_IV[1];
 v[10] = blake2b_IV[2];
 v[11] = blake2b_IV[3];
 v[12] = blake2b_IV[4] ^ S->t[0];
 v[13] = blake2b_IV[5] ^ S->t[1];
 v[14] = blake2b_IV[6] ^ S->f[0];
 v[15] = blake2b_IV[7] ^ S->f[1];
```

If everything is aligned in a picture-like format we have a clear visual
representation of how the algorithm organizes its data and how its
structured. This helps to better understand the algorithm in case we
want to optimize it afterwards

DoZ10:
Thanks for the great feedback !

I pretty much understand all your points, I'll push all fixes soon. 

I'll include support for blake2b options in kernels.  The big question here is since we need to test them, do we include directly these tests in a 610 mode in the current PR or not ?

DoZ10:
Okay, spent some hours looking at this and I think I am doing it wrong by initializing the state and parameters in the kernel. Parameter block (p_) is used only once for state initialization and state (s_) is used and updated for each hash update/final.

Perhaps, it would be better to handle those initialization variables in CPU and then just relay the state to the GPU kernels. That way, I would not be stuck with all conditional statements/execution branches and would improve speed.

I saw that there were special cases such as SCRYPT that relays data from CPU directly to kernel function arguments. That would be convenient for this purpose, but I feel this is may become a major change in the current project. In the end, it would really simplify kernel code especially for KEYED mode where pre-hashing is required.

What do you think ?

PS: Btw only attack mode 0 received some love this weekend, As per dicussion above, next commits will be applied to all other modes.

DoZ10:
Gave it a shot !

Params and State are initiated in interface.c and relayed through "esalt" buffer down to GPU kernel. This should be suitable for transforms and KEYED-mode pre-computing.
 
Now everything looks fine for attack modes 0, 1 & 3. VECT_SIZE > 1 also fixed for vector-types.

I guess we can call this a day !

Cheers,

jsteube:
Some (not so important) things left:

1. Replace:

```
for (int i = 0; i < 8; ++i) 
```

With

```
for (int i = 0; i < 8; i++) 
```

2. In OpenCL, you can replace:

```
u64 p_key[16] = { 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```

With

jsteube:
Some (not so important) leftovers:

1. There's still some leftovers of "blake2b_compress", and "blake2b_transform" - Please make it "transform" everyhwere
2. Is the vector cracking working now?
3. Have you tried moving the blake2b_sigma to the inside of the kernel and compared if this gives a speed boost?
4. You can also try copying esalt_bufs->h[] to a local variable outside the inner loop, this way the innerloop does not need to try to access the global memory for each iteration. compare performance afterwards  but it's possible the compiler already auto-optimized it for you. either way you should do it. Same for t[] and f[]
5. What's the reason for having blake2 as cpu function? I mean it's good to have it, just wondering what's the background
6. In case we want to have Blake2-256, what do we need to change? Simply truncate the output or is it more complex?
7. This section in the parser:

```
  for (u8 i = 0; i < sizeof(blake2_params_t); i++)
    ((u8 *)S)[i] ^= ((u8 *)P)[i];
```

Is it really required to use a cast here? I'd prefer to see each attribute xor in a single for () loop. This greatly helps understand why we do this because it propably shows which values are set to zero at this point.

8. As you can see here:

```
    case   600: hashconfig->esalt_size = sizeof (blake2_state_t);   break;
    case  2500: hashconfig->esalt_size = sizeof (wpa_t);            break;
```

The convention of the type is always XXX_t. Can you please rename blake2_state_t to blake2_t

jsteube:
9. One more thing. When running the benchmarks it shows:

```
Hashtype: Blake2b
```

But shouldn't it be Blake2b-512 or so?

jsteube:
Sometime, in the kernel, you use:

```
for (u32 il_pos = 0; il_pos < il_cnt; il_pos += VECT_SIZE) 
{
```

```
for (u32 il_pos = 0; il_pos < il_cnt; il_pos += VECT_SIZE) {
```

The latter is correct.

DoZ10:
Okay. For the fixes I should be able to commit something tonight. For the questions:

2) Vector cracking was working yesterday for VECT_SIZE 2 for attack mode 0, 1 & 3 using tools/test.sh. Will take a look to see if any bug was induced, but it should work now. Initial problem was a wrong usage of vector-types uXXx.

3) Not yet but will do. I needed to have a fully working kernel before doing benchmark tests.

5) Intent was for KEYED mode in Blake2b performs a first hash with the KEY. This could become handy to pre-hash this in CPU and relay the output through esalt buffer for computation without too much overhead on kernel code. Unless you believe there is a neat performance gain of putting all this in kernel, I could do so. For now, cpu_blake2.h only holds types used in interface.c.

6) According to the official reference on github, the output is simply truncated. (see. blake2b_final(...) in blake2b-ref.c). So should be a piece of pie to implement in interface.c

9) I'll comply with JtR convention.

Thanks

DoZ10:
Alright. Pushed all the fixes.

Great perf gain for the blake2b_sigma integration. Over 100% in gain.
Not so much for the esalt_bufs out of the loop tho.

jsteube:
I think we're almost done. Here's some missing pieces left:

1. Please add mode 600 to extra/tab_completion/hashcat.sh to the HASH_MODES variable
2. Please add note about addition to docs/changes.txt
3. Fix following formating in the parser:

```
  blake2_t  *S = (blake2_t *) hash_buf->esalt;

  memset(S,  0, sizeof(blake2_t));
```

to:


```
  blake2_t *S = (blake2_t *) hash_buf->esalt;

  memset (S, 0, sizeof (blake2_t));
```

4. Fix following formating in the parser:

```
  S->h[0] ^= 0x0000000001010040; // digest_lenght = 0x40, depth = 0x01, fanout = 0x01
```

to:

```
  S->h[0] ^= 0x40 << 0; // digest_lenght = 0x40
  S->h[0] ^= 0x01 << 4; // depth = 0x01
  S->h[0] ^= 0x01 << 6; // fanout = 0x01
```

Beware! Please verify I understood the positioning and comment correctly

5. Please do a global search on the substring "Blake2" - I think we should name it Blake2b. For example here in usage.c it says:

```
"    600 | Blake2-512                                       | Raw Hash",
```

But maybe this is not neccessary, I just don't know what's the difference between Blake2 and Blake2b. You have to decide.

6. Some notes:

I've tested with test.sh and it worked for me as well. I've tested with AMD, NV and Intel. 
I've also added a default hash to wiki example hashes.

This is really great work!

DoZ10:
Blake2 comes in four flavors. The "b" one is the most used right now so that is why it is probably referred to only "Blake2" in JtR. But the true name is "Blake2b" and I don't think this will confuse users since perl and python libraries refer to "Blake2b".

Thanks for your help and guidance on this one. Was fun to do, more to come ;)

PR #1235, opened 2017-04-30
Add support for rule: eX

### LINKED ISSUE #1231: new rule: title case, with alternate word separators (like JtR 'e')
John the Ripper has an 'e' rule, which works like 'E' but allows the user to specify what separator should be used. Briefly described here by @jfoug:

http://www.openwall.com/lists/john-users/2016/04/17/2

> I have also added a '2nd' title case command.  Hashcat added the 'E' 
> rule, which is title case based solely upon space character separators. 
> The eC and e?C rules are extensions to this. They allow any character 
> (or class of characters) to be used as the separator for title casing.

(This issue was inspired by https://hashcat.net/forum/thread-6489-post-34578.html)

mhasbini:
```
$ cat /tmp/in
ab-bcd-cef
abc
ab cd ef
-ab
 ab

$ ./hashcat --stdout -j 'e-' /tmp/in
Ab-Bcd-Cef
Abc
Ab cd ef
-Ab
 ab
```

I've tested:

1- using 'e-' with -j switch:  `./hashcat --stdout -j 'e-' /tmp/in`.
2- using 'e-' with -r switch: `./hashcat --stdout -r <(echo 'e-') /tmp/in`
3- using 'e-' to crack hash: 
```
$ ./hashcat --quiet -r <(echo 'e-') /tmp/h /tmp/in
ff46c8192da24c9d1b30fb90b7461ec2:Ab-Bcd-Cef
```
Closes #1231

### DIFF
--- a/OpenCL/inc_rp.cl
+++ b/OpenCL/inc_rp.cl
@@ -2454,7 +2454,7 @@ inline u32 rule_op_mangle_dupeblock_last (const u32 p0, const u32 p1, u32 buf0[4
   return out_len;
 }
 
-inline u32 rule_op_mangle_title (const u32 p0, const u32 p1, u32 buf0[4], u32 buf1[4], const u32 in_len)
+inline u32 rule_op_mangle_title_sep (const u32 p0, const u32 p1, u32 buf0[4], u32 buf1[4], const u32 in_len)
 {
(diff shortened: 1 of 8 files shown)


jsteube:
Nice, IIRC, the first contribution that adds a new rule function. Thanks!

PR #1237, opened 2017-05-07
New algorithm: Chacha20 

DoZ10:
New stream cipher algorithm Chacha20 !

Implementation is based on Perl and Python3 implementations and was tested on all 3 kernels for attack modes 0, 1 and 3. Vector-types were tested as well. 

Proposed hash format (all params are hex encoded):
$SIGNATURE$\*POSITION\*IV\*PLAINTEXT\*CIPHERTEXT
ex.
$Chacha20$\*0000000000000001\*0000000000000000\*4141414141414141\*7c4099cb07618083

Only 1 bug to figure out here that is not directly related to the current work, but could be nice to look at. When using 32 byte "passwords" from a dictionnary the password is discarded and not processed. But 31 byte passwords work fine, indicating a trailing null or carriage return bug in core CPU code for line parsing. This bug is NOT from the kernel and other hashing algorithms seem to be affected too such as ArubaOS.

Cheers.


DoZ10:
Okay I found what is the problematic code, but I'm not sure if this is some "desired" behavior.

The pw_max is overriden to 31 in _straight.c_ at function _straight_ctx_init_ see below:

```
[...]

const bool has_noop = kernel_rules_has_noop (straight_ctx->kernel_rules_buf, straight_ctx->kernel_rules_cnt);

  if (has_noop == false)
  {
    switch (user_options_extra->attack_kern)
    {
      case ATTACK_KERN_STRAIGHT:  if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                  break;
      case ATTACK_KERN_COMBI:     if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                  break;
    }
  }
  else
  {
    if (hashconfig->attack_exec == ATTACK_EXEC_INSIDE_KERNEL)
    {
      switch (user_options_extra->attack_kern)
      {
        case ATTACK_KERN_STRAIGHT:  if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                    break;
        case ATTACK_KERN_COMBI:     if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                    break;
      }
    }
    else
    {
      // in this case we can process > 32
    }
  }
```
 ATTACK_EXEC_INSIDE_KERNEL is set for my hash mode 670 and has_noop is true.

What was the original intent from this code ? Since it now seem to conflict with modes that permits pw > 31, I am not sure if we should use a workaround by selecting different configuration options or figure out a patch.

jsteube:
Again, great contributions! Many thanks so far! Here are some comments:

1. kernel code quality is very good, but maybe add some "salt" section to the initialization code by simply commenting it

```
+  u32 iv[2]    = { 0 };
+  u32 plain[2] = { 0 };
+  u32 position[2] = { 0 };
+ 
+  position[0] = esalt_bufs->position[0];
+  position[1] = esalt_bufs->position[1];
+
+  u32 plain_length = esalt_bufs->plain_length;
+
+  iv[0] = esalt_bufs->iv[0];
+  iv[1] = esalt_bufs->iv[1];
+
+  plain[0] = esalt_bufs->plain[0];
+  plain[1] = esalt_bufs->plain[1];
```

2. What's the reason to have a "plain_length" in the esalt? Isn't this a block cipher?

3. In the chacha20_parse_hash(), can you add some length verification checks to ensure iv length is actually 16. The best way to accomplish that is by using the code as we do it, for example see the itunes_backup_parse_hash() function. You will find out pretty easily the pattern and you will see how we do the tests for length.

4. The algorithm makes use of an esalt, but misses to set an artificial salt. While this is not a technical requirement, it's more a requirement of hashcat itself. In order to help hashcat to identify duplicate salts we have to create a unique (artificial) salt for each unique esalt. See pstoken_parse_hash() for example. I think it would be fine simply to copy all data from the esalt data into salt_buf[] array and set the salt_length to 32 or so-

5. Most important. Please do not use mode 670. Please use mode 15400 instead.

DoZ10:
1) Sure.
2) Early implementation leftover.. Thanks for catching it I'll remove it. This is a stream cipher and originally I wanted to allow variable length, but revised such design since it is totally useless for cracking purposes.
3) I'll implement the is_valid_hex_string() since the hash format is fixed size. Thanks for pointing this out.
4) Wasn't aware of this. Will fix.
5) Sure.

DoZ10:
One cool thing to add would be position + offset values to be more flexible for the cipher position in use.

I could commit something to reflect that change, but it requires a new format parameter.

Such as $SIGNATURE$\*POSITION\*OFFSET\*IV\*PLAINTEXT\*CIPHERTEXT

DoZ10:
Alright..

Combining __position__ and __offset__ parameters, you can now reach any consecutive 8 bytes in any keystream. This new addition also enables to reach an 8 byte cipher across two keystream blocks !

It makes more sense to allow users to reach any part of the keystream.

Will wait for your comments !

jsteube:
1. Please replace this:

```
for (u8 i = 0; i < 10; ++i)
```

with this:

```
for (u8 i = 0; i < 10; i++)
```

2. Prepend for() loops with a fixed iteration count with #pragma unroll

```
  for (u8 i = 0; i < 10; i++)
```

to:

```
  #pragma unroll
  for (u8 i = 0; i < 10; i++)

```

3. Drop w2[] and w3[] in the inner loop, they are not accessed

4. Load the base word outside the inner loop, that saves accessing the slow global memory, ex:

outside:

```
  u32 w0[4];
  u32 w1[4];

  w0[0] = pws[gid].i[0];
  w0[1] = pws[gid].i[1];
  w0[2] = pws[gid].i[2];
  w0[3] = pws[gid].i[3];
  w1[0] = pws[gid].i[4];
  w1[1] = pws[gid].i[5];
  w1[2] = pws[gid].i[6];
  w1[3] = pws[gid].i[7];
```

inside (replace):

```
    u32x w0_t[4];
    u32x w1_t[4];

    w0_t[0] = w0x;
    w0_t[1] = w0[1];
    w0_t[2] = w0[2];
    w0_t[3] = w0[3];
    w1_t[0] = w1[0];
    w1_t[1] = w1[1];
    w1_t[2] = w1[2];
    w1_t[3] = w1[3];
    ...
    chacha20_transform (w0_t, w1_t, position, offset, iv, plain, digest);

```

jsteube:
1. There's still remainings of this in the single hash kernels. 

```
  position[0] = esalt_bufs->position[0];
  position[1] = esalt_bufs->position[1];
```

Even if this is always just "0" please rewrite it, too. That makes it easier to copy/paste. It will have no impact on performance.

```
  position[0] = esalt_bufs[digests_offset].position[0];
  position[1] = esalt_bufs[digests_offset].position[1];
```

2. Hash mode 15400 is missing in extra/tab_completion/hashcat.sh in HASH_MODES variable

3. Try to use more "const" keywords whenever possible, for example:

```
  u32 index  = offset / 4;
  u32 remain = offset % 4;
```

to:

```
  const u32 index  = offset / 4;
  const u32 remain = offset % 4;
```

4. Please fix alignment for spaces:

```
  u32 iv[2]    = { 0 };
  u32 plain[2] = { 0 };
  u32 position[2] = { 0 };
  u32 offset = 0;
```

5. Those do not need to be u32x in the outer loop, u32 is fine and saves some register:

```
  u32x w0[4];
  u32x w1[4];
  u32x w2[4];
  u32x w3[4];
```

6. The tmp variable can be u32, no need to u32x here:

```
    u32x tmp[3];
```

jsteube:
Great work again! Many thanks!

PR #1245, opened 2017-05-12
Add support for rule position 'p'

### LINKED ISSUE #1035: New rule: replace Nth character
I would like to propose a new feature, which would be this rule to replace the Nth instance of a character with a different character. The classic 's' rule replaces _all_ instances of the given character; this new rule could perhaps be 'S'. The idea is to be able to generate candidates such as 'pa$swords', based solely on premises e.g. that some users like to use leetspeak for the first 'a' in the word but not the rest - or things of that nature.

New rule could be of the form SNXY: replace Nth instance of character X with Y

So, to illustrate:

Rule | Input | Output
--- | --- | ---
ss$ | passwords | pa$$word$
S1s$ | passwords | pa$swords
S2s$ | passwords | pas$words
S1s$S1s$ | passwords | pa$$words

**EDIT:**
After about a month since opening this issue, I just wanted to reiterate how useful I think it will be when it is implemented.

The idea came to me when I couldn't crack the password 'Odes$a77'. I spent some decent time with fairly good hardware, my favorite wordlists and rule files, some good masks and combinations... nothing worked, even though I knew (due to the nature of the challenge) that it had to be a somewhat simple password ('Odessa' is in every wordlist, obviously). Luckily I had the opportunity to ask the challenge owner for the correct answer and that's when I realized that none of the existing rules would have matched it, except maybe some very random rule by chance - delete s @ N then insert $ @ N, etc.

When this gets implemented, a think a new, more selective/fine-grained leetspeak rule file will be due, and I can't wait to see what it will catch when run against large bases :)

mhasbini:
The loop condition in `RULE_OP_REJECT_CONTAINS` was changed to be compatible with JtR. 

```
echo 'remember' | ./hashcat --stdout -j '/e op3'
r3member

echo 'remember' |  ./hashcat --stdout -j '%2m ip!'
reme!mber

echo 'remember' |  ./hashcat --stdout -j '%3e op3 %2e op3 /e op3'
r3m3mb3r
```

Using `p` without using `/` or `%` before it, would result in an invalid rule.
E.g. `echo 'remember' |  ./hashcat --stdout -j 'op3'` will return empty response with exit status 1.

This may not close #1035. I thought it's nice to have.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -14,6 +14,7 @@
 - Wordlist encoding: Support added for internal convert from and to user-defined encoding during runtime
 - Wordlist encoding: Added parameters --encoding-from and --encoding-to to configure wordlist encoding handling
 - Rules: Support added for rule 'eX'
+- Rules: Support added for position 'p' in host mode (using -j or -k)
 
 ##
 ## Improvements
--- a/docs/rules.txt
+++ b/docs/rules.txt
@@ -56,4 +56,4 @@
 #define RULE_OP_REJECT_EQUAL_AT         '=' // reject plains that do not contain char X at position N
 #define RULE_OP_REJECT_CONTAINS         '%' // reject plains that contain char X less than N times
 #define RULE_OP_REJECT_MEMORY           'Q' // reject plains that match the plain saved (see M), i.e. if unchanged
-
+#define RULE_LAST_REJECTED_SAVED_POS    'p' // position of the character last found with '/' or '%'
--- a/include/types.h
+++ b/include/types.h
@@ -293,6 +293,7 @@ typedef enum rule_functions
   RULE_OP_REJECT_EQUAL_AT        = '=',
   RULE_OP_REJECT_CONTAINS        = '%',
   RULE_OP_REJECT_MEMORY          = 'Q',
+  RULE_LAST_REJECTED_SAVED_POS   = 'p',
 
   RULE_OP_MANGLE_SWITCH_FIRST    = 'k',
   RULE_OP_MANGLE_SWITCH_LAST     = 'K',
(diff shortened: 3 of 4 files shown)


jsteube:
Thanks for your contributions. Please see my comments below:

1. In the changes.txt:

```
+- Rules: Support added for position 'p'
```

This should be modified to reflect this change will work only for the CPU rule enging by using -j and -k, but not on GPU

2. The macro name convention:

```
 #define RULE_OP_REJECT_EQUAL_AT         '=' // reject plains that do not contain char X at position N
 #define RULE_OP_REJECT_CONTAINS         '%' // reject plains that contain char X less than N times
 #define RULE_OP_REJECT_MEMORY           'Q' // reject plains that match the plain saved (see M), i.e. if unchanged
-
+#define RULE_REJECT_SAVED_POS           'p' // position of the character last found with '/' or '%'
```

I'm wondering if the "REJECT" substring is correct, or is it really working only for reject cases?

mhasbini:
@jsteube Thanks for checking!

> 1. In the changes.txt:

I'll update the changelog entry to be:

> - Rules: Support added for position 'p' in host mode (using -j or -k)

What do you think?

> 2. The macro name convention

I wasn't sure about the name.
`p` will only work if it's preceded by `/` or `%`, and will have the position of the last matched character as value.

Maybe `RULE_LAST_REJECTED_SAVED_POS` or `RULE_CONTAINS_SAVED_POS`?

jsteube:
1. OK
2. OK. RULE_LAST_REJECTED_SAVED_POS sounds good to me.

roycewilliams:
Examples are now documented here: https://hashcat.net/wiki/doku.php?id=rule_based_attack#using_p_nth_instance_of_a_character_with_positional_rules

Issue #1246, opened 2017-05-14
Different behaviour for some rules with input of length 32 between host & GPU mode 


mhasbini:
Examples:

- rule `@`:

```
$ echo '123456790abcdefghijklmnopqrstuvw' | ./hashcat --stdout -j '@a'
123456790bcdefghijklmnopqrstuvw
$ echo '123456790abcdefghijklmnopqrstuvw' | ./hashcat --stdout -r <(echo '@a')
$
```

- rule '4

```
$ echo '123456790abcdefghijklmnopqrstuvw' | ./hashcat --stdout -j "'4"
1234
$ echo '123456790abcdefghijklmnopqrstuvw' | ./hashcat --stdout -r <(echo "'4")
$
```

Same behaviour for `D`, `O`, `Y`, `[`, `]`, `x` and `y` rules:

host mode: rule executed correctly.
GPU mode: blank response.

Other shared rules have consistent behaviour between both modes: blank response.

Is this an expected behaviour ?

DoZ10:
I noticed a similar problem for some algorithms in GPU mode with input legnth 32 bytes. Problem was mentionned in my current PR, but not addressed yet... it might be related... here's a copy.

```
Okay I found what is the problematic code, but I'm not sure if this is some "desired" behavior.

The pw_max is overriden to 31 in straight.c at function straight_ctx_init see below:

[...]

const bool has_noop = kernel_rules_has_noop (straight_ctx->kernel_rules_buf, straight_ctx->kernel_rules_cnt);

  if (has_noop == false)
  {
    switch (user_options_extra->attack_kern)
    {
      case ATTACK_KERN_STRAIGHT:  if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                  break;
      case ATTACK_KERN_COMBI:     if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                  break;
    }
  }
  else
  {
    if (hashconfig->attack_exec == ATTACK_EXEC_INSIDE_KERNEL)
    {
      switch (user_options_extra->attack_kern)
      {
        case ATTACK_KERN_STRAIGHT:  if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                    break;
        case ATTACK_KERN_COMBI:     if (pw_max > PW_DICTMAX) pw_max = PW_DICTMAX;
                                    break;
      }
    }
    else
    {
      // in this case we can process > 32
    }
  }

[...]

PR #1248, opened 2017-05-16
Add tools/test_rules.pl script

mhasbini:
Add `tools/test_rules.pl` script to run some tests on available rules.

There are 2 modes for tests:
1- host mode (`-j $RULE`)
2- GPU mode (`-r $RULE_FILE`).

Cases are added to `tools/rules-test-cases.yaml` with this format:

```
rule:
    mode: 1 (or 2) # optional, if available the test will only run for the specified mode.
    input:
    ....
    expected_output:
    ....
```

```
$ perl tools/test_rules.pl --help
usage: tools/test_rules.pl [rule] [mode]
       [mode]: 1 for host mode, 2 for GPU mode
       If [rule] was not specified, run all test cases
       If [mode] was not specified, run test for all modes

examples:
run i3! case on modes 1 & 2  : perl tools/test_rules.pl i3!
run O04 case on mode 1       : perl tools/test_rules.pl O04 1
run sab case on mode 2       : perl tools/test_rules.pl sab 2
```

These tests will save some -manual- testing time while refactoring/adding new rules.

I've tried to test some common cases:
- string with length 31, 32 (rejection limit).
- limit positions (e.g. string with length 2 with `'1` rule).
- the general application of the rule.

The tests doesn't cover all edge cases for limit positions.

There are 10 tests failing currently related to #1246 (whether host mode and GPU mode should behave similarly when string hid rejection limit for some rules).

I'll modify the script to adhere with #1246 conclusion.

### DIFF
--- a/tools/rules-test-cases.yaml
+++ b/tools/rules-test-cases.yaml
@@ -1,0 +1,796 @@
+---
+::
+  input: |
+    AbCd
+    abcd
+    ABCD
+    123456790abcdefghijklmnopqrstuvw
+    123456790abcdefghijklmnopqrstuvwx
+
+  expected_output: |
+    AbCd
+    abcd
+    ABCD
+
+l:
+  input: |
+    AbCd
+    abcd
+    ABCD
+    123456790abcdefghijklmnopqrstuvw
+    123456790abcdefghijklmnopqrstuvwx
+
+  expected_output: |
+    abcd
+    abcd
+    abcd
+
+u:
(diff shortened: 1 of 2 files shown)


jsteube:
Nice tool, will try to find out if the reported errors are real errors or if we have to patch test_rules.pl script.

mhasbini:
FTR current test run: http://termbin.com/euv7

PR #1255, opened 2017-05-18
Fix for Periodic Status Leak

### LINKED ISSUE #1254: Leaking memory while periodically printing stats
Some Background: I've been working on a task management system in Go that links against libhashcat (similar to what Rich5 is doing [here](https://github.com/Rich5/pyHashcat)) to start & manage hashcat jobs and have been noticing a significant memory leak over time. 

The leak can be triggered whenever a user runs hashcat with the arguments`--status --status-timer=1` which I think is caused by not free'ing the results of the strdup calls from some of the `char* status_get_XXX` functions.

I have a coroutine that is running during an active cracking session that periodically pulls the status via `hashcat_get_status` and sends that upstream via a websocket to provide real-time updates to the browser. Disabling this and the leak is gone. 

Here's the output from the latest version of hashcat (v3.5.0) compiled with `DEBUG=1` with me periodically streaming the status out: https://gist.github.com/tankbusta/8ad06e98bed01ca7a3399dd76945ec72

tankbusta:
This fixes #1254 by freeing some of the contents within `hashcat_status_t`.

Valgrind Before: https://gist.github.com/tankbusta/8ad06e98bed01ca7a3399dd76945ec72
Valgrind After: https://gist.github.com/tankbusta/02640f63705626a22549d8688e94a5b4

### DIFF
--- a/include/status.h
+++ b/include/status.h
@@ -91,4 +91,6 @@ void        status_progress_reset                 (hashcat_ctx_t *hashcat_ctx);
 int         status_ctx_init                       (hashcat_ctx_t *hashcat_ctx);
 void        status_ctx_destroy                    (hashcat_ctx_t *hashcat_ctx);
 
+void        status_status_destroy                 (hashcat_status_t *status_ctx);
+
 #endif // _STATUS_H
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -37,6 +37,7 @@
 
 - Fixed a missing type specifier in a function declaration of the RACF kernel
 - Fixed a condition that caused a hybrid attack using a maskfile to not select all wordlists from a wordlist folder
+- Fixed a memory leak that is present when a user is periodically printing hashcat status (using --status-timer)
 
 ##
 ## Technical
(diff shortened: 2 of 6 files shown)


include/status.h
```diff
@@ -91,4 +91,6 @@ void        status_progress_reset                 (hashcat_ctx_t *hashcat_ctx);
 int         status_ctx_init                       (hashcat_ctx_t *hashcat_ctx);
 void        status_ctx_destroy                    (hashcat_ctx_t *hashcat_ctx);
 
+void        status_status_destroy                 (hashcat_status_t *status_ctx);
```
tankbusta, line 94: I'm not a fan of this name but willing to change it on suggestion

jsteube:
Thanks for the contribtion, looks like a nice fix! Some (minor important) comments:

1. The function name status_status_destroy() isn't perfect but it's good enough for me now. We can rename it later if we see a need to do so.
2. Please add a note to docs/changes about your fix
3. Please fix format:

```
hcfree(status_ctx->session);
```

to 


```
hcfree (status_ctx->session);
```

and 

```
 if (NULL == status_ctx)
    return;
```

to

```
 if (NULL == status_ctx) return;
```

tankbusta:
Awesome thanks! Just committed the changes you requested.

PR #1256, opened 2017-05-18
Add support for rejection rule _N

mhasbini:
Ported from JtR.

Example: `_2`

input:

```
a
ab
abc
```
output:

```
ab
```


roycewilliams:
Nice!

For testers, here is a bash one-liner commandline test:

```
$ (for item in a ab abc; do echo $item; done) | hashcat --stdout -j '_2'
ab
```

Issue #1264, opened 2017-05-30
compiling v3.5.0-147-g2dcad3f0 causes gcc 7.1.1 warnings


ZerBea:
After updating gcc to 7.1.1, I got the following warnings when compiling 
v3.5.0-147-g2dcad3f0:

src/convert.c: In Funktion »printable_utf8«:
src/convert.c:41:36: Warnung: diese Anweisung könnte durchfallen
[-Wimplicit-fallthrough=]
       if ((a = (*--srcptr)) < 0x80 || a > 0xbf) return false;
           ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
src/convert.c:42:5: Anmerkung: hier
     case 3:
     ^~~~
src/convert.c:43:36: Warnung: diese Anweisung könnte durchfallen
[-Wimplicit-fallthrough=]
       if ((a = (*--srcptr)) < 0x80 || a > 0xbf) return false;
           ~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~
src/convert.c:44:5: Anmerkung: hier
     case 2:
     ^~~~
src/convert.c:51:21: Warnung: diese Anweisung könnte durchfallen
[-Wimplicit-fallthrough=]
       case 0xF4: if (a > 0x8f) return false;
                     ^
src/convert.c:54:5: Anmerkung: hier
     case 1:
     ^~~~

Best regards
Mike

jsteube:
Thanks fort reporting. This is an code snippet from @magnumripper - can you take a look please?

magnumripper:
The code is OK. We are deliberately using the fall-through feature of switch/case and apparently the `-Wall` option now comments on that. Just add `-Wno-implicit-fallthrough` to get rid of the warning.

ZerBea:
Thank you for the info. I know that the code is working well. It was just an information that there are some gcc warnings using the default options of gcc 7.1.1.
Best regards
Mike

magnumripper:
Will still might want to add `-Wno-implicit-fallthrough` before most dists has a gcc that new.

neheb:
Adding __attribute__((fallthrough)) seems to be the recommended way to silence the warning.

ZerBea:
Well, it's a good idea to silence the warning, before gcc 7.1.1 reaches UBUNTU.

magnumripper:
[...]

PR #1266, opened 2017-05-31
Mute gcc 7.1.1 warnings about intentional fall-throughs. See #1264.

magnumripper:
OK, this didn't turn out so bad. This will not fail on older compilers (or clang) and it actually makes the code a little more clear about what's happening.

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -7,6 +7,12 @@
 #include "types.h"
 #include "convert.h"
 
+#if (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__) >= 70101
+#define FALLTHROUGH __attribute__ ((fallthrough))
+#else
+#define FALLTHROUGH
+#endif
+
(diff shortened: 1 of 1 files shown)


src/convert.c
```diff
@@ -38,17 +44,17 @@ static bool printable_utf8 (const u8 *buf, const int len)
...
+      if ((a = (*--srcptr)) < 0x80 || a > 0xbf) return false; FALLTHROUGH;
     case 2:
       if ((a = (*--srcptr)) < 0x80 || a > 0xbf) return false;
 
       switch (*buf) {
       case 0xE0: if (a < 0xa0) return false; break;
       case 0xED: if (a > 0x9f) return false; break;
       case 0xF0: if (a < 0x90) return false; break;
-      case 0xF4: if (a > 0x8f) return false;
+      case 0xF4: if (a > 0x8f) return false; FALLTHROUGH;
```
Empact, line 57: This may as well be break, for symmetry.

jsteube:
Thanks!

PR #1268, opened 2017-06-04
tidy changes.txt and name normalizations

roycewilliams:
Strictly speaking, 'Chacha20' should be 'ChaCha20'. If hashcat recently created the $Chacha20$ format, if it's not too late to change it to '$ChaCha20$', it should be changed.


magnumripper:
IMHO, while the format name might better be ChaCha20, the tag (or signature) should be `$chacha20$`in all lower. Some tags are all upper (I think) but few ones should have a mix.

roycewilliams:
That's fair. Of the tags from the hashcat examples, there's a definite trend:

```
1
2a
5
6
7z
8
9
B
BLAKE2
ChaCha20
DCC2
DPAPImk
H
P
PHPS
RAR3
S
apr1
axcrypt
axcrypt_sha1
bitcoin
blockchain
cram_md5
ecryptfs
episerver
ethereum
itunes_backup
keepass
krb5pa
krb5tgs
md5
ml
mysqlna
office
oldoffice
pdf
postgres
racf
rar5
sha1
sip
zip2
```

@magnumripper, I'm not attached to either approach strongly, but just out of curiosity, other than historical trends, are there any practical or programmatic reasons why mixed case would be bad?

jsteube:
I don't mind how it's written, as long everyone works with it afterwards. In case you want to change it again, please send in a PR, but do it fast because of the release on friday

magnumripper:
> @magnumripper, I'm not attached to either approach strongly, but just out of curiosity, other than historical trends, are there any practical or programmatic reasons why mixed case would be bad?

It's not the end of the world but I know for sure that mixed-case tags will lead to confused users. Tags aren't labels and they aren't clear-text. They are just... tags, and in the end of the day they are more like "machine readable" than "human readable". They should definitely be simple: Either all upper or all lower IMO.

magnumripper:
Also, the real arbitrator is compatibility. If hashcat would have ended up having it as `$cHaCha20$` before we implemented anyting else, we would have used that in JtR too, fugly or not.

roycewilliams:
Indeed. I like the unified machine-readable and human-readable lower-case simplicity. As you can see, committed as such. Thanks for speaking up.

PR #1271, opened 2017-06-06
Chacha20: Fixed error in format parser separator validation

DoZ10:
An unmatched separator will return a NULL + 1 value.

Fixed accordingly.


jsteube:
I can see the problem, but tbh I don't like the fix. Can you please rewrite the code so that it matches the other parser functions? If you do, the problem will be fixed, too. See here for an example: https://github.com/hashcat/hashcat/blob/master/src/interface.c#L14965-L15043

PR #1284, opened 2017-06-24
Fix signed overflow warnings

neheb:
This is most definitely a false positive. Might as well fix it anyways.


jsteube:
I'm sure there's many such false positives, but merging it, it doesn't hurt. Thanks!

PR #1288, opened 2017-07-04
Add additional DES charsets with corrected parity

mattames:
Add additional charsets with corrected parity and other charsets to split up key searches by the most significant key byte.


jsteube:
Thanks! I'll merge and afterwards move some of the files to a new folder, to avoid confusion.

PR #1292, opened 2017-07-09
Added new rules

evilmog:
expanded-cutb-clean - compilation of a large number of rules largely auto generated, also credit to toxic as I used his tmesis utils to generate some of them

rule file is massive so I compressed it and put it in a new folder called archive so that it doesn't break workflow

only use on really fast hashes - evilmog


jsteube:
File is too large, GitHub doesn't like it. It will also increase package size. Anyway, this is a great piece of resource. You should create some dedicated repository for it.

PR #1294, opened 2017-07-12
Add example_hashes.csv

ghost:
We don't always have access to the wiki, so add the example hashes page as a CSV file in the docs/ folder.


roycewilliams:
Interesting - could be handy.

@usernamestaken, a related side note: 

Multiple search engines seem to suppress the example-hashes wiki page sometimes. We are not sure why. The page is still there, but is removed entirely from search results, even when using very specific search strings.

roycewilliams:
It may be useful to separate with tabs or pipes instead, to support various CSV-parsing systems. This will also sidestep any potential commas-in-the-hash-name parsing issues.

roycewilliams:
@usernamestaken - stray wrapping with leftover quoting around line 17:

```
400|"phpass, WordPress (MD5),
Joomla (MD5)"|$P$984478476IagS59wHZvyQMArzfx58u.
```

ghost:
Should be good now, libreoffice doing weird stuff with newlines in the cells

roycewilliams:
Closer ... but the GitHub CSV parser still sees an error (though I'm honestly not sure why yet):

https://github.com/usernamestaken/hashcat/blob/f38fd0f47f2141fab97cfc265b086b4a543d98b0/docs/example_hashes.csv

Maybe because it's the first record that actually contains commas. Maybe tab-separated would be more universally applicable. Kinda flying by the seat of my pants here. :)

ghost:
Row 17 and row 19 are similar but no warning is thrown for it.  Maybe something with the parser, either way, the file is fine for reference regardless of whether they think its pretty or not.  Fact is, this isn't meant to be viewed online anyway, but grep'd, cat'd, less'd, more'd, (insert viewing method here).

roycewilliams:
Fair. My only reason for citing the online viewer was that if its parser is happy, most other tools will probably also have no problem. Glad to have it, regardless. I hope it's accepted.

jsteube:
There's already the same data in here:

https://github.com/hashcat/hashcat/blob/master/src/interface.c#L25-L266

It would make more sense to just print this data on user request (--example-hashes or so) otherwise there'd be two separate places to maintain the same data.

roycewilliams:
Ah, didn't realize that was there! A cmdline option to show them could be handy.

jsteube:
Implemented with https://github.com/hashcat/hashcat/commit/12d95fd22c7cab80e0e466946d5a3129eeb15b52

Issue #138, opened 2016-01-18
feature: expanded version information


roycewilliams:
In the new world of open source and publicly available betas, it might be handy to have the git commit hash in both the output of -h/--help/-V/--version, and the general "oclHashcat vX.YY starting" messages.

This allows precise version identification in pasted/posted output, when bugs are reported/fixed, etc. 

It might be useful to leave it out for official binary releases, so that they can be clearly differentiated from any and all self-compiled/test/unofficial/beta versions. Or a "-release" suffix might be used instead, to explicitly disambiguate.

One possible method:

```
$ ./oclHashcat -V
2.10 (30293a0)

$ ./oclHashcat -h | head -2
oclHashcat, advanced password recovery
v2.10 (30293a0)

$ ./oclHashcat -b -m 1500
oclHashcat v2.10 (30293a0) starting in benchmark-mode...
[snip]
```

... or:

```
$ ./oclHashcat -V
2.10-30293a0

$ ./oclHashcat -h | head -2
oclHashcat, advanced password recovery
v2.10-30293a0

$ ./oclHashcat -b -m 1500
oclHashcat v2.10-30293a0 starting in benchmark-mode...
[snip]
```

... with official binary releases either using the current basic version number only, or a simple tag:

```
$ ./oclHashcat -V
2.10

$ ./oclHashcat -V
2.10-release

$ ./oclHashcat -V
2.10 (release)
```

philsmd:
would it be okay if we just append the compile time to the version number information ?
There is a variable COMPTIME used within the Makefile for such purposes.

Why not git commit hash? Because somtimes (well, not that often, but rarely) it happens that betas will be compiled just for testing (even before every piece of new/changed code was committed) and therefore we do not know any git commit SHA1 hash at that time.
On the other hand, adding COMPTIME to the version number should be fine (in my opinion).
What do you think ? @roycewilliams would that be okay?

roycewilliams:
Thanks for the opportunity to weigh in. This may seem like a bit of trouble -- but once the effort is invested, it will benefit many future developers, testers, troubleshooters and researchers forever. :)

I hadn't thought about the "compile from uncommitted changes and publish that" corner case -- good point.

[...]

PR #1304, opened 2017-07-20
visual indicator of active options when benchmarking

roycewilliams:
To disambiguate benchmark output, add indicator of:

* whether -O was used
* what level of -w was used

Much like adding the git commit to the version information (#138), this can help to compare benchmarks over time in a variety of contexts.

Adding this to both interactive output and --machine-readable output is warranted.

This commit implements one method of doing this:

```
$ ./hashcat -b -m 0 -O
hashcat (v3.6.0-244-g434af9f+) starting in benchmark mode...
Active options: -O, -w 2
[...]

$ ./hashcat -b -m 0
hashcat (v3.6.0-244-g434af9f+) starting in benchmark mode...
Active options: -w 2
[...]

$ ./hashcat -b -m 0 -O -w 4
hashcat (v3.6.0-244-g434af9f+) starting in benchmark mode...
Active options: -O, -w 4
[...]

$ ./hashcat -b -m 0 -w 4
hashcat (v3.6.0-244-g434af9f+) starting in benchmark mode...
Active options: -w 4
[...]

$ ./hashcat -b -m 0 -O --machine-readable
# hashcat (v3.6.0-244-g434af9f+)
# Active options: -O, -w 2
[...]

$ ./hashcat -b -m 0 --machine-readable
# hashcat (v3.6.0-244-g434af9f+)
# Active options: -w 2
[...]

$ ./hashcat -b -m 0 -O -w 4 --machine-readable
# hashcat (v3.6.0-244-g434af9f+)
# Active options: -O, -w 4
[...]

$ ./hashcat -b -m 0 -w 4 --machine-readable
# hashcat (v3.6.0-244-g434af9f+)
# Active options: -w 4
[...]
```

Another option would be something like:
```
# Optimized kernel: no; workload profile: 2
```
```
# Optimized kernel: yes; workload profile: 4
```
... etc.

But there may be other flags that affect benchmark performance, so being brief may be warranted, to conserve space and to keep the output to a single line.


jsteube:
Please change the order of the branch so that the test is testing a true case, not a false case. That makes it faster to read. 
There's more options that can influence benchmark performance: -n, -u, --opencl-vector-width  
If not set, they are handled by the tuning database

roycewilliams:
Will do. May be a few days before I can get to it.

PR #1305, opened 2017-07-20
tests: fix tests for -m 14000, 14100, 14900 and 15400

philsmd:
This PR fixes some test cases where the password length is fixed and only has a specific allowed length.
In these special cases we need to apply some extra logic to make sure that the correct mask lengths and/or splitted dict lengths are used.

Thanks


jsteube:
Great! Many thanks

PR #1309, opened 2017-07-28
get_random_num: Simplify

neheb:
random() is available everywhere except Windows and DOS.

edit: Switched Windows to rand_s as well.


neheb:
Fixed MinGW compile. Had no good solution to it since I would have had to modify lzma_sdk/Alloc.c

jsteube:
Appveyor still showing build failed. Is this safe to merge?

magnumripper:
Appveyor MINGW32 as well as MINGW64 (Cygwin 32/64 are fine):
```
src/shared.c: In function 'get_random_num':
src/shared.c:379:3: error: implicit declaration of function 'rand_s'; did you mean 'rand_r'? [-Werror=implicit-function-declaration]
   rand_s(&r);
   ^~~~~~
```
This function wasn't prototyped by MINGW headers some time ago - is it now? If not, just add this in some sensible place.
```
extern errno_t rand_s (unsigned int *randomValue);
```
...or maybe it is prototyped nowadays although Appveyor runs an older MINGW version where it isn't. Then you'll need a few gears to sort that out...

neheb:
I tested the compile on Fedora 26. MinGW that is. It works fine(except for the iconv stuff). So it does work.

I think I'll just revert to rand(). rand_s is nice since it returns an unsigned 32 bit integer but it looks like it's too much trouble.

jsteube:
Thanks!

PR #1312, opened 2017-08-05
implement sha256(sha256(password))

ckuethe:
Doesn't yet pass self-test on my machine; here's a PR so someone can explain what I'm doing wrong in the opencl...


philsmd:
Oh! Ha!

Unfortunately, one does not simply write optimized kernel code by just running something like sed 's/sha1/sha256/g'

Hint: there are also different input/output lengths e.g. for the hex conversation etc. To make it very clear, sha256 hashes/digests "are longer" compared to sha1 hashes... this is just an example what is completely wrong in your kernels.
You are also missing a lot of host code changes, like entries in src/usage.c etc, please look at previous commits which did add a new hash type to hashcat to see what is required.

... and before I forget it (and actually the most important thing!): thank you very much for your contribution!!!

ckuethe:
Thanks for the useful feedback - that's exactly what I was looking for.

docs/changes.txt
```diff
@@ -5,6 +5,7 @@
 ##
 
 - Added hash-mode  2501 = WPA/WPA2 PMK
+- Added hash-mode  1470 = sha256(sha256(password))
```
philsmd: Please use the same format that we used for other hash modes i.e.:
- Added hash-mode  1470 = sha256(sha256($pass))

jsteube:
The mode number 1470 doesn't exactly fit into the numbering system (as bad as it is). Can you please use mode 15900 and rewrite the code.

docs/changes.txt
ckuethe: ACK

ckuethe: ACK

jsteube:
I think there's still the mode 1470 used in the kernel sources. Did you remove the cached kernels before retrying ?

ckuethe:
I haven't had much time to work on the opencl bits this week. Other stuff at work...

jsteube:
Any news here?

jsteube:
PR is a bit outdated (lots of conflicts with latest version). Please reopen when fixed, thanks!

ckuethe:
ACK. I still haven't got the OCL working, but I haven't forgotten.

PR #1331, opened 2017-08-16
use an overflow check instead of a hard coded max. value for ETA

philsmd:
There was a recent bug report on the hashcat forum (https://hashcat.net/forum/thread-6774.html) that made me wonder if there is some kind of upper limit of the date displayed...

and indeed there was a fixed value
```
if (sec_etc > 100000000)
```
(this value converted from seconds to years is about 3 years).

This pull request tries to fix this hardcoded value with calls of the overflow_check_u64_add () function.
Furthermore, I've refactored some code into a new function called status_get_sec_etc ().

This commit also introduces a new hashcat specific type called hc_time_t and some time-specific function which use this new type (the idea is to convert all time_t into hc_time_t soon).
Thanks


jsteube:
Great! Thanks!

PR #1358, opened 2017-09-16
Fix overflow in mangle_dupechar_last function

mhasbini:
The problematic line is here: https://github.com/hashcat/hashcat/blob/master/src/rp_kernel_on_cpu.c#L533
when `len` is zero (e.g. rule "]Z1" applied on "a").

To check set DEBUG to 2 and run:

```
$ cat /tmp/2
a

$ ./hashcat --stdout -r <(echo ]Z1) /tmp/2
=================================================================
==99299==ERROR: AddressSanitizer: stack-buffer-overflow on address 0x70000052c42f at pc 0x00010bdcf1e9 bp 0x70000052bde0 sp 0x70000052bdd8
READ of size 1 at 0x70000052c42f thread T15
```

### DIFF
--- a/OpenCL/inc_rp.cl
+++ b/OpenCL/inc_rp.cl
@@ -503,6 +503,7 @@ static int mangle_dupechar_last (MAYBE_UNUSED const u8 p0, MAYBE_UNUSED const u8
 {
   const int out_len = len + p0;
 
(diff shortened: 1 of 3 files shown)


jsteube:
Nice find, thanks!

PR #1373, opened 2017-09-27
Use stat64 for all platforms except cygwin.

neheb:
MinGW as well as the other major platforms support using stat64 directly. Not cygwin though.

Tested on Cygwin, Linux, and MinGW. No idea about FreeBSD or OS X. A quick google search shows that they work.


krytarowski:
stat64 breaks NetBSD

neheb:
Forget it then

PR #1379, opened 2017-10-02
Adjust hcmalloc to be the same as calloc

neheb:
calloc is almost equivalent to malloc + memset(0) except that it's faster with big allocations because of OS trickery. It also protects against integer overflow and throws a null pointer on overflow whereas malloc does not.

### DIFF
--- a/src/memory.c
+++ b/src/memory.c
@@ -23,16 +23,8 @@ void *hccalloc (const size_t nmemb, const size_t sz)
 
 void *hcmalloc (const size_t sz)
 {
-  void *p = malloc (sz);
-
-  if (p == NULL)
-  {
-    fprintf (stderr, "%s\n", MSG_ENOMEM);
-
-    return (NULL);
-  }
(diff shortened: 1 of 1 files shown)


philsmd:
This is confusing. We already have a hccalloc () function.

jsteube:
I like the Idea, but can you please replace the calloc() with hccalloc() and add a note why we use calloc() inside the malloc() wrapper? This could be confusing otherwise.

neheb:
done

edit: One concern I have is that hcmalloc = malloc + memset. The memset zeroes the memory allocated, but I'm sure there's code that does not need this.

Which begs the question, is this patch correct or should the memset be removed? I don't think integer overflow is an issue unless we're talking about 32-bit systems. There's also existing code of the form "void *p = hccalloc(foo, 1).

jsteube:
It's not needed, that's right, but it also doesn't hurt.

PR #1382, opened 2017-10-03
replace more snprintf calls with hc_asprintf.

neheb:
from my fairly basic testing, nothing has blown up. I'd still wait after the release to merge.


philsmd:
could it be that this patch introduces some memory leaks?
Whenever we use asprintf () we also should free () the buffer afterwards.
The problem only happens when there were buffers on the *stack* beforehand (where the buffers automatically wen't out of scope and therefore deleted) and now we use asprintf which uses the heap (where we do need to make sure to call free ()).

jsteube:
I'll fix the missing free()

PR #1400, opened 2017-10-20
Replace PRIu64 with ju. Less verbose.

neheb:
Tested with MinGW (the poster child for compiler issues). cppcheck will argue the format is wrong, (uint64_t vs. uintmax_t) but they're equivalent, even on 32-bit.


jsteube:
Yeah and even if uintmax_t somewhere in the future becomes 128 bit, it would still work as expected. It's fine for me, thanks!

jsteube:
Somehow this breaks mingw cross compiled binaries on windows, result looks like this:

```
* Device #2: GeForce GTX 980, ju/ju MB allocatable, 1024MCU
```

Therefore have to revert the changes, sorry!

PR #1419, opened 2017-10-31
partially fix compile for musl

neheb:
last issue is that musl has no qsort_r. not sure how to fix that.

### DIFF
--- a/include/affinity.h
+++ b/include/affinity.h
@@ -11,9 +11,6 @@
 
 #if defined (_POSIX)
 #include <pthread.h>
-#if defined (__linux__)
-#include <sys/sysctl.h>
-#endif // __linux__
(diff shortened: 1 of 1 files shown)


Duncaen:
I fixed it by using the fallback `sort_r_simple` function.
https://github.com/voidlinux/void-packages/blob/a481cae1c61f2122919c372b8b7c92d8f80c58be/srcpkgs/hashcat/patches/musl.patch

Kochise:
Fyi, `qsort_r` has been added to Alpine on 2021/09/22 (musl-1.2.2-r7) :

https://gitlab.alpinelinux.org/alpine/aports/-/commit/bf5bbfdbf780092f387b7abe401fbfceda90c84d

Profit.

PR #1438, opened 2017-11-11
use fallback qsort implementation on linux systems without glibc

Duncaen:


Kochise:
Fyi, `qsort_r` has been added to Alpine on 2021/09/22 (musl-1.2.2-r7) :

https://gitlab.alpinelinux.org/alpine/aports/-/commit/bf5bbfdbf780092f387b7abe401fbfceda90c84d

Profit.

PR #1444, opened 2017-11-13
Replace char[] with *char

neheb:
The latter is more correct. The side effect is that filesize becomes smaller.

gcc    clang

before

O2
756112 762120

Os
674192 680216

after

O2
748456 760728

Os
666536 674728

There are a bunch of entries remaining in include files. I'm not exactly sure what to do with them. Changing them to const char* causes unused variable warnings.


jsteube:
I think we replaced the * to [] because of some recommendation of the coverity system. Can you make sure this change is ok? What are the variables creating the unused variable warnings?

neheb:
The variables are in header files like opencl.h. Once converted they become unused variables since they are not used within the header file.

That coverity warning sounds bogus. Most if not all uses of the strings expect a char pointer. Currently those variables rely on array decay, which is an implicit(and unnecessary) conversion.

Quick hashcat -b shows no problems, speed or otherwise.

I also have another patch that fixes a bunch of cast-qual warnings, some a result of this patch.

jsteube:
OK, let's do the change and see if coverity complains again. For the other variables let's create a separate PR.

PR #1451, opened 2017-11-16
always include the error message even if hash is long

philsmd:
This problem was first reported here: https://hashcat.net/forum/thread-7013-post-37438.html#pid37438

Whenever the hash is very long and there was a parsing error for this specific hash, we should make sure that we do not truncate the line before the actual error message within event_log_warning ().

We should always make sure that the error messages is guaranteed to show up within the error line. We could safely truncate the hash instead. This is what this patch is implementing. It shortens the hash if needed.

Thanks

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -15,6 +15,12 @@
 * changes v4.0.0 -> v4.0.1:
 
 ##
+## Improvements
+##
+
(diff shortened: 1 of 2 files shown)


philsmd:
@neheb thank you very much for the suggestion

PR #1480, opened 2017-12-26
Osx clock gettime

matrix:
Hi,
clock_gettime() in not defined in OSX <= 10.11

> $ make
clang -c -pipe -std=gnu99 -Iinclude/ -Iinclude/lzma_sdk/ -IOpenCL/ -W -Wall -Wextra -Wfloat-equal -Wundef -Wshadow -Wmissing-declarations -Wmissing-prototypes -Wpointer-arith -Wstrict-prototypes -Waggregate-return -Wswitch-enum -Winit-self -Werror-implicit-function-declaration -Wformat -ftrapv -Wwrite-strings -Wno-cast-align -Wno-cast-qual -Wno-conversion -Wno-padded -Wno-pedantic -Wno-sizeof-pointer-memaccess -O2 src/timer.c -o obj/timer.NATIVE.STATIC.o
src/timer.c:36:3: error: implicit declaration of function 'clock_gettime' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
  clock_gettime (CLOCK_MONOTONIC, a);
  ^
src/timer.c:36:18: error: use of undeclared identifier 'CLOCK_MONOTONIC'
  clock_gettime (CLOCK_MONOTONIC, a);
                 ^
2 errors generated.
make: *** [obj/timer.NATIVE.STATIC.o] Error 1

I fixed that by
- checking the Product Version in Makefile
- add a clock_gettime() alternative by patching the hc_timer_set() function

Merry Christmas,
Matrix
(★ゝз･)ﾉ~*

### DIFF
--- a/src/timer.c
+++ b/src/timer.c
@@ -33,7 +33,7 @@ inline double hc_timer_get (hc_timer_t a)
 
 inline void hc_timer_set (hc_timer_t* a)
 {
-  #if defined (__APPLE__)
+  #if defined(__APPLE__) && defined(MISSING_CLOCK_GETTIME)
   // taken from proxmark3/client/util_posix
   static uint64_t clock_start_time = 0;
   static mach_timebase_info_data_t timebase_info = {0, 0};
--- a/src/timer.c
+++ b/src/timer.c
@@ -33,7 +33,25 @@ inline double hc_timer_get (hc_timer_t a)
 
 inline void hc_timer_set (hc_timer_t* a)
 {
+  #if defined (__APPLE__)
+  // taken from proxmark3/client/util_posix
+  static uint64_t clock_start_time = 0;
+  static mach_timebase_info_data_t timebase_info = {0, 0};
+  uint64_t now = mach_absolute_time();
+
+  if (clock_start_time == 0)
+  {
+    mach_timebase_info(&timebase_info);
+    clock_start_time = now;
+  }
+
+  now = (uint64_t)((double)(now - clock_start_time) * (double)timebase_info.numer / (double)timebase_info.denom);
+
+  a->tv_sec = now / 1000000000;
+  a->tv_nsec = now % 1000000000;
+  #else
   clock_gettime (CLOCK_MONOTONIC, a);
+  #endif
 }
 
 inline double hc_timer_get (hc_timer_t a)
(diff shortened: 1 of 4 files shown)


jsteube:
That's not how I'd have done that in src/timer.c (The other stuff in Makefile is ok). The use of static variables gives me some headache. Can you rewrite it do use this kind of code as used in v4.0.1?

types.h:

```
typedef struct timeval    hc_timer_t;
```

timer.c:

```
inline void hc_timer_set (hc_timer_t* a)
{
  gettimeofday (a, NULL);
}

inline double hc_timer_get (hc_timer_t a)
{
  hc_timer_t hr_tmp;

  hc_timer_set (&hr_tmp);

  return (double) (((hr_tmp.tv_sec - (a).tv_sec) * 1000) + ((double) (hr_tmp.tv_usec - (a).tv_usec) / 1000));
}
```

src/timer.c
```diff
@@ -42,6 +46,9 @@ inline double hc_timer_get (hc_timer_t a)
 
   hc_timer_set (&hr_tmp);
 
+  #if defined(__APPLE__) && defined(MISSING_CLOCK_GETTIME)
+  return (double) (((hr_tmp.tv_sec - (a).tv_sec) * 1000) + ((double) (hr_tmp.tv_usec - (a).tv_usec) / 1000));
```
philsmd, line 50: hmm, could/should this be improved with something like this:
```
   hc_timer_t s;
 
   s.tv_sec  = hr_tmp.tv_sec  - a.tv_sec;
   s.tv_usec = hr_tmp.tv_usec - a.tv_usec;
 
   if (s.tv_usec < 0)
   {
     s.tv_sec  -= 1;
     s.tv_usec += 1000;
   }
 
   double r = ((double) s.tv_sec * 1000) + ((double) s.tv_usec / 1000);

   return r;
```
???
( I didn't test it, but if I look at the other branch - the #else - it seems to be more consistent)
we could also perform the return after the branch (because it is common code to both branches) and the variable definition of hc_timer_t s before the branches (but this isn't too important, it just would emphasize where the main differences between the 2 cases are, i.e. the nsec vs usec etc)

jsteube: I've merged the original version from @matrix but it would be nice if you can send in a PR with the easier to read version you've posted in the comment, too.

PR #1493, opened 2018-01-15
added -m 16400 = CRAM-MD5 Dovecot

mohemiv:
This implements the CRAM-MD5 hash for dovecot server ( command `doveadm pw`).

Thanks!


jsteube:
Cool stuff, thanks for the contribution! Can you please change the hash-mode from 10201 to 16400?

mohemiv:
Done

jsteube:
OK, thanks so far!

The next thing that's missing is the test.pl implementation. Usually we need it to do the unit tests but it's also used to create hashes with just a given password (for example to create the self-test hash). Please add the test.pl implementation.

jsteube:
Excellent contribution, thank you!

PR #1498, opened 2018-01-24
Modified makefile to add soname for libhashcat

### LINKED ISSUE #1496: Missing soname of libhashcat.so
Hello,
According to the documentation i've read for hashcat, libhashcat.so isn't private or plugin but public interface, so i think it should have soname. Could you please add soname to this library?

TomasKorbar:
closes #1496

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -8,6 +8,55 @@ DEBUG                   := 0
 PRODUCTION              := 0
 
 ##
+## Native compiler paths
+##
+
+CC                      := gcc
+FIND                    := find
+INSTALL                 := install
+RM                      := rm
+SED                     := sed
+SED_IN_PLACE            := -i
+
+ifeq ($(UNAME),Darwin)
+CC                      := clang
+# the sed -i option of macOS requires a parameter for the backup file (we just use "")
+SED_IN_PLACE            := -i ""
+PROD_VERS               := $(shell sw_vers -productVersion | cut -d. -f2)
+endif
+
+ifeq ($(UNAME),FreeBSD)
+CC                      := cc
+SED                     := gsed
+endif
+
+##
+## Cross compiler paths
+##
+
+CC_LINUX_32             := gcc
+CC_LINUX_64             := gcc
+
+CC_WIN_32               := i686-w64-mingw32-gcc
+CC_WIN_64               := x86_64-w64-mingw32-gcc
+
+##
+## Misc stuff
+##
+
+COMPTIME                := $(shell date +%s)
+
+# the following variable value will be automatically replaced by the "git archive" command
+# (which is automatically run for every github release)
+# the value will be something like this: "tag: vX.Y.Z, refs/pull/K/head" or "HEAD -> master, tag: vX.Y.Z"
+
+VERSION_EXPORT          := $Format:%D$
+VERSION_TAG             := $(shell test -d .git && git describe --tags --dirty=+ || echo "$(VERSION_EXPORT)" | $(SED) 's/.*: v\([\.0-9]*\).*/v\1/')
+
+SONAME_VERSION          := $(shell echo "$(VERSION_TAG)" | $(SED) 's/^v\([\.0-9]\+\)-.*/\1/')
+SONAME_VERSION_MAJOR    := $(shell echo "$(SONAME_VERSION)" | $(SED) 's/^\([0-9]\+\)\..*/\1/')
+
+##
 ## Detect Operating System
 ##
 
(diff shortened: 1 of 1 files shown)


yarda:
AppVeyor build failures seems unrelated to the patch for me.

jsteube:
Why is there

```
SONAME                  := 0.0
```

Is that really what we want?

jsteube:
Bug: The target "make clean" does not remove the file:

libhashcat.so.0.0

yarda:
Just proposal, we need API version. Usually in new projects we start from 0.0 (major/minor), but there can be any sane API numbering.

yarda:
@TomasKorbar please update the PR to address the bug, i.e. edit it to have (you can do git push -f to force push it):
```
-$(RM) -f obj/*.o obj/lzma_sdk/*.o *.bin *.exe *.so *.dll *.pid hashcat core
+$(RM) -f obj/*.o obj/lzma_sdk/*.o *.bin *.exe *.so *.so.* *.dll *.pid hashcat core
```

jsteube:
What about this as an alternative to SONAME?

```
SONAME_VERSION := $(shell git describe --tags --dirty=+ | sed 's/^v\([\.0-9]*\).*/\1/')
```

jsteube:
This is also wrong I think

```
	cd $(DESTDIR)$(LIBRARY_FOLDER) && ln -s $(HASHCAT_LIBRARY) $(HASHCAT_LIBNAME)
```
Instead of doing the ln manually, ldconfig should be used:

```
root@ht:~/hashcat# ls -l libhashcat.so*
-rwxr-xr-x 1 root root 817368 Jan 25 22:17 libhashcat.so.4.0.1
root@ht:~/hashcat# ldconfig -n .
root@ht:~/hashcat# ls -l libhashcat.so*
lrwxrwxrwx 1 root root     19 Jan 25 22:30 libhashcat.so.4 -> libhashcat.so.4.0.1
-rwxr-xr-x 1 root root 817368 Jan 25 22:17 libhashcat.so.4.0.1
```

And then we should link to libhashcat.so.4

yarda:
No problem, if the e.g. 4.0.1 API is not compatible with the 4.0.0. E.g. all consumers of the API needs to be recompiled upon bump to 4.0.1, 4.0.2, etc. Most projects bump the stable API version after while only when doing API changes.

yarda:
@jsteube  my previous comment is wrong, yours proposal is better.

jsteube:
Not sure if they are really better, I'm just playing around with this. I have no real experience with it, therefore I was asking you to provide the PR.

yarda:
Usually auto tools do all the magic :) I will re-think it and provide updated PR soon.

yarda:
I think libhashcat.so.4.0.1 is OK for the filename of the library, but on Linux we should probably set SONAME to libhashcat.so.4, i.e. use only the MAJOR number for it when linking, ie.: call ld with '-Wl,-soname,libhashcat.so.4', this way the consumers of the API will have to be recompiled only if the MAJOR version of the API will be bumped taking into account the hashcat version means MAJOR.MINOR.PATCH. Well, it's probably non-portable (it seems e.g. FreeBSD use different approach than Linux). It's usually better to keep the magic on the libtool / auto-tools than doing it the hard way :)

I will update the patch.

yarda:
Also the SONAME_VERSION will have to be set by some bootstrap script before making the upstream release, because git history is not available in the tarball. Or maybe we could parse it from the VERSION_EXPORT variable.

yarda:
it seems VERSION_TAG already does it.

jsteube:
Please send in an update PR, I'll review

yarda:
Tomas will update the PR later today - I am supervising/sponsoring him in Fedora.

yarda:
@TomasKorbar this is wrong patch.

yarda:
@jsteube sorry for delay, please wait for next PR.

yarda:
FYI for Tomas this is the first contribution to real upstream/github project.

jsteube:
The current PR does not apply cleanly and looks like something went terrible wrong. Please fix!

yarda:
@TomasKorbar mostly. Why is there hunk on lines 342 - 354? Also you shouldn't change the whitespaces .e.g lines 347/365 and many others. Please provide updated PR.

yarda:
@TomasKorbar you probably didn't update your repo and rebased the commit by:
git pull
git rebase

yarda:
@jsteube the basic idea should be clear even from this PR. We do not call now ldconfig in the install, because I think it should be handled by admin, i.e.:
```
# make install
# ldconfig
```
e.g. with RPM packages the ldconfig is called in the RPM %postinst phase to make all the symlinks for new libraries. But we can add it to the install, if you think it's good to have.

yarda:
@TomasKorbar great, well done, it looks much better now.

yarda:
@jsteube it's ready for review :)

jsteube:
Looks good, I'll merge it. I'll do some small changes afterwards.

jsteube:
OK, I've pushed my changes. Btw I'm planing to release 4.1.0 soon. That should make it easier to use the changed code from your linux distribution.

yarda:
OK, great, thanks.

PR #1511, opened 2018-02-06
Update rule syntax in best64.rule

Chick3nman:
Updated syntax for xMN to OMN.


magnumripper:
I must have missed the original change in rules syntax. Where did that happen? Was it good or bad in terms of compatibility?

Chick3nman:
I'm actually not sure when it happened @magnumripper, I need to dig through the versions and find out. It doesnt seem bad at all, considering all it does is add another rule for 'x' (Extract range) and move the original function (delete range) to 'O'.

jsteube:
@magnumripper We changes the x rule to be the O rule a while back to be JtR compatible. We just forgot to update the rule files, too. That's what this PR is doing.

PR #1524, opened 2018-02-13
added hcstat2 files

evilmog:
Added generated hcstat2 files


jsteube:
That's some nice work, but please create a smaller set (max. 10 files) and name it so that a user who has no Idea what this is get's an idea of your idea.

evilmog:
Modified as requested

hcstat2s/readme.txt
```diff
@@ -0,0 +1,15 @@
...
+built in hashcat hcstat file which is trained on rockyou.txt
+
+These adjust the marvok model and per position generators. Use this if you want to adjust
+how a3 does searches.
+
+More of these can be found at https://github.com/evilmog/hashcat-hcstat
+
+To use these run ./hashcat64.bin -a 3 -markov-hcstat all.hcstat2 -m [insert mode] example0.hash [insert mask here]
+
+These can also be used with statsprocessor ./sp64.bin [options] hcstat-file [filtermask]
```
jsteube, line 11: not true

hcstat2s/readme.txt
```diff
@@ -0,0 +1,15 @@
...
+
+These files may work out better than the built in or they may not, YMMV
```
jsteube, line 13: What is YMMV

hcstat2s/readme.txt
```diff
@@ -0,0 +1,15 @@
+This is a collection of hcstat2 files for Hashcat. These are used in -a 3 to replace the
+built in hashcat hcstat file which is trained on rockyou.txt
+
+These adjust the marvok model and per position generators. Use this if you want to adjust
+how a3 does searches.
+
+More of these can be found at https://github.com/evilmog/hashcat-hcstat
+
+To use these run ./hashcat64.bin -a 3 -markov-hcstat all.hcstat2 -m [insert mode] example0.hash [insert mask here]
```
jsteube, line 9: If you do an example for example0.hash, the -m mode is always 0

hcstat2s/readme.txt
```diff
@@ -0,0 +1,15 @@
+This is a collection of hcstat2 files for Hashcat. These are used in -a 3 to replace the
+built in hashcat hcstat file which is trained on rockyou.txt
+
+These adjust the marvok model and per position generators. Use this if you want to adjust
```
jsteube, line 4: Removing word-wrap please

jsteube (commented):
There's still 74 files in the PR, but should be 10. There's also a line in the docs/changes.txt missing. Please fix.

philsmd:
some further problems: marvok is probably a typo for markov
the command line parameter is --markov-hcstat (2 dashes at the beginning)

Some of the .hcstat2 files (as atom already mentioned) are not within the hcstat2s folder.

Another question is, should we add a note also within the docs/ folder? Or only use the docs/ folder for the documentation and use the hcstat2s folder for the compressed hcstat2 files only ? I'm not sure about this. Maybe we should use a considtent way for adding these extra files. This doesn't really depend on this PR, but maybe to make it more consistent we should also add README files to the rules, charset and masks folders too?

Another observation I made is that the files are about 14 MB (already compressed)... is this an acceptable grow in total file size of the release, @jsteube ? If needed we could maybe only pick the best ones !?

evilmog:
moving to 4.2

PR #1536, opened 2018-03-01
POC for cracking SAP "half hashes" CODVN B+F/G 

erpscanteam:
Hi,
you can find in this PR an implementation of the SAP hashes 7701 and 7801 that are half 7700 and 7800 and very often the only one you will get when pen-testing SAP ABAP servers via their RFC interfaces.
We enjoy using hashcat a lot and think this little patch may interest not only us so here it is!

Cheers,

-- Mathieu


jsteube:
Thanks for the PR, looks very nice!

To merge there are some modifications needed:

1. Add the missing kernels. That is the -a1 and -a3 kernel.
2. Add the test.pl and test.sh entries so we can do unit tests.
3. Remove the example77xx and 78xx files. Those are obsolte after (2) is implemented.

PR #1604, opened 2018-06-27
Mute a macOS -Wpointer-sign warning.

magnumripper:
Mutes this:
```
src/rp_cpu.c:503:31: warning: passing 'char *' to parameter of type 'const u8 *'
      (aka 'const unsigned char *') converts between pointers to integer types with different sign
      [-Wpointer-sign]
      const u8 c = hex_to_u8 (rule_new + rule_pos + 2);
                              ^~~~~~~~~~~~~~~~~~~~~~~
include/convert.h:26:26: note: passing argument to parameter 'hex' here
u8  hex_to_u8  (const u8 hex[2]);
                         ^
1 warning generated.
```


jsteube:
Thanks!

PR #1629, opened 2018-07-22
Fix shared install on macOS

RAN1:
Fix name flag when building the shared library on macOS. clang doesn't support soname.

Change the macOS library name to match the dylib convention.

Minor fix for subdirectory installs so files are placed in the correct paths (i.e. include/lzma_sdk/).

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -50,6 +50,7 @@ INSTALL                 := install
 RM                      := rm
 SED                     := sed
 SED_IN_PLACE            := -i
+CD                      := cd
 
 ifeq ($(UNAME),Darwin)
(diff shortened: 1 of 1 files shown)


jsteube:
I'm fine with the .dylib change, the other ones I don't understand. Please explain what's going on there and why it is needed. Please send multiple PR for multiple changes.

jsteube:
Note that with commit https://github.com/hashcat/hashcat/commit/2530f83029da1f657ae54bea62f2a1d90a806b99 I've moved out the LZMA files from the include folder, therefore the change in the installation routines in your PR should no longer be neccessary

PR #1633, opened 2018-07-23
Fix shared install on macOS

RAN1:
Remove -soname on Darwin builds, macOS doesn't use SONAME and its ld throws an unknown option error when -soname is passed. The equivalent options for versioning on macOS, -current_version and -compatibility_version, have been added.

By default on macOS, library searches for -lx look for libx.dylib. Filenames with .so aren't supported, so fixed library name and updated .gitignore accordingly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@
 hashcat
 hashcat.exe
 libhashcat.so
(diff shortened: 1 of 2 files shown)


jsteube:
Thanks!

PR #1634, opened 2018-07-24
Fix a couple of gcc-6 -Wpointer-sign build warnings.

magnumripper:
Fixes this:
```
src/interface.c:4953:20: warning: initializing 'const u8 *' (aka 'const unsigned char *') with an
      expression of type 'char [1]' converts between pointers to integer types with different sign
      [-Wpointer-sign]
  static const u8 *zero = "";
                   ^      ~~
src/interface.c:13895:11: warning: initializing 'char *' with an expression of type 'u8 *'
      (aka 'unsigned char *') converts between pointers to integer types with different sign
      [-Wpointer-sign]
    char *account_info_start = input_buf + 12; // we want the * char included
          ^                    ~~~~~~~~~~~~~~
2 warnings generated.
```

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -4950,7 +4950,7 @@ int md5md5_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNU
    * This way we can save a special md5md5 kernel and reuse the one from vbull.
    */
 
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks!

Issue #956, opened 2017-01-08
insecure relative RPATH handling


anthraxx:
hashcat adds a relative RPATH to the current directory, which has security implications and may cross privilege boundaries.
This issue can possibly be exploited by a local attacker to run arbitrary code as some other user if victim user can be convinced to run hashcat in an attacker controlled directory with specially crafted content.
RPATH should never be used for relative or world-writable paths.

jsteube:
This is important to help the not so advanced users to be able to compile a native hashcat binary and run it without using the install target or using LD_LIBRARY_PATH. The only practical way I see to handle this is to remove the RPATH when using the make install target. What do you think?

anthraxx:
well advanced user here or there but this is a security issue and especially non-advanced users are even less aware of the security implication :smile: 
yes, right now i simply strip out the whole rpath to mitigate this threat. stripping it on install would work for those, but the thing is, technically the users not using the install target will still be vulnerable to this problem. in my opinion people need to learn to make things work without the relative rpath to not expose anyone to a potential threat.

how about some kind of simply shell script with a readme block that aids non-advanced users to run it locally without installing it into the system?

jsteube:
Teaching people how to use LD_LIBRARY_PATH is far too big of a goal for the hashcat project. The only solution is that we remove the rpath when the user uses make install target. How exactly did you do it? I see the tools "patchelf" and "chrpath" can do it. Which one would you suggest?

anthraxx:
as said, you don't need to teach anyone about LD_LIBRARY_PATH if you have a wrapper script that does so and uses explicitly the directory where the shared lib was built in. Otherwise its still vulnerable for all people who don't use the install target.
The only solution to this issue is to not use relative rpath at all and add a wrapper script to start hashcat when not being in a 'installed' target, otherwise the users are always vulnerable to unintended code execution.

jsteube:
I kind of agree, but such an wrapper script would confuse new users.
It's not very intuitive, so it's not a good solution. Do you have any
other ideas?

anthraxx:
I don't see the difference, wherever you put you hashcat executable could be the place to have the hashcat wrapper script. Nobody would ever notice the difference if you adjust the install target accordingly and distribute the non-wrapper version instead. I honestly don't see where the problem is.

jsteube:
Well as already discussed, the problem would occur only with the
self-compiled version which then is not used with the make install
target, because this target will have the rpath removed. That means that
[...]

PR #1635, opened 2018-07-24
Add macOS library install pathnames

RAN1:
macOS uses run-time install pathnames for shared libraries, set by an -install_name flag on clang and also using install_name_tool. The following changes set up install names in macOS.

Normally the install name is set on the shared library with the -o path as it was passed to clang. and this path is copied to the executable during linking. The -o path is copied verbatim and the path uses the shell's working directory at run-time, so testing shared hashcat in its own directory causes an error if the shell isn't working in the same directory.

I've set the $(HASHCAT_LIBRARY) Darwin target to use the executable path instead (via token) to prevent this.

During install, install_name_tool is used to change the path on both the install destination library and executable to use the absolute destination path since otherwise the bin-installed hashcat and anything linked against the library will use executable paths first.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -57,6 +57,7 @@ CC                      := clang
 SED                     := /usr/bin/sed
 SED_IN_PLACE            := -i ""
 PROD_VERS               := $(shell sw_vers -productVersion | cut -d. -f2)
+INSTALL_NAME_TOOL       := install_name_tool
 endif
 
 ifeq ($(UNAME),FreeBSD)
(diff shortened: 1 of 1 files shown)


jsteube:
Please explain why we need to do this and what's happening in case we do not merge this PR.

RAN1:
Currently the first library pathname checked is the executing shell's $PWD/libhashcat. Because that pathname is copied from the library to executables, this is true of all executables linked, whether the library's in hashcat/ or the library directory. If there isn't a libhashcat in $PWD, then the default library paths are checked. This can cause misbehavior, so the PR changes it to the executable path for the initial build (in case the library is used in place) and again after install to the library directory.

jsteube:
> Currently the first library pathname checked is the executing shell's $PWD/libhashcat.

This sounds really strange to me. In Linux, there's no library search path for "." because of the security implication this creates. Here's an older issue that dealt with the same problem: https://github.com/hashcat/hashcat/issues/956

>  If there isn't a libhashcat in $PWD, then the default library paths are checked. This can cause misbehavior

Wait here, what kind of misbehavior are you talking about?

> so the PR changes it to the executable path for the initial build (in case the library is used in place)

OK, but wouldn't this create a relative path which is what we need to avoid (see the issue linked above)?

RAN1:
> This sounds really strange to me. In Linux, there's no library search path for "." because of the security implication this creates. Here's an older issue that dealt with the same problem: #956 

As far as I can tell this is default behavior with macOS libraries if at build time the output isn't to an absolute path and the install path isn't specified.

> Wait here, what kind of misbehavior are you talking about?

I meant it is misbehavior for more or less clear reasons.

> OK, but wouldn't this create a relative path which is what we need to avoid (see the issue linked above)?

It's effectively the same as using $ORIGIN. The alternative is to just pass the library install path to -install_name and remove the changes to install_library and install_hashcat.

jsteube:
Why is this so different to Linux or Unix in general? Usually if you have a shared library you copy it to /usr/lib or whatever folder your distribution uses for this case. Done.

If you don't want to copy the shared library you set LD_LIBRARY_PATH to the folder where you did installed the library. This can be ".", too. Is there no LD_LIBRARY_PATH on macOS?

RAN1:
On macOS, library pathnames are hardcoded in the executable's load commands and they are checked first at runtime. If the path is "relative/libfoo" then it will look in "./relative/libfoo" first. This will cause problems similar to the ones RPATH causes.

<s>Yes, there is the equivalent DYLD_LIBRARY_PATH on macOS, but it has lower precedence than the hardcoded paths.</s>

EDIT: I tested this badly, DYLD_LIBRARY_PATH is checked before everything else, the hardcoded paths only have precedence over the regular library search paths.

RAN1:
I amended my PR to use only the library install path.

jsteube:
OK, thanks!

PR #1643, opened 2018-07-31
Added hash-mode 16900 = Ansible Vault

### LINKED ISSUE #1532: Support for Ansible Vault
Hello,

It would be nice to add the support for Ansible Vault. Further information are added below.

Vault algorithm:
- https://github.com/ansible/ansible/blob/devel/lib/ansible/parsing/vault/__init__.py

Decryption examples:
- https://github.com/ansible/ansible/blob/060001b08d98969217156b2b696613ade44eb8c5/test/units/parsing/vault/test_vault.py

Example of vault password files: 
- https://raw.githubusercontent.com/ansible/ansible/e238ae999b90b771f662ac15e6cceff8aa0b721f/test/integration/targets/vault/format_1_1_AES.yml
- https://raw.githubusercontent.com/stevenharradine/ansible-vault-brute-force/master/vault.yml

Thanks in advance!

hops:
This closes #1532.
You can use the ansible2john script found here: https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/run/ansible2john.py
Some background about the algorithm:
First a PBKDF2-HMAC-SHA256 hash with dklen = 80 is generated using a 32byte salt and 10000 iterations.
The bytes 32-64 are used as key to generate a HMAC-SHA256 hash over the (AES256-CTR encrypted) ciphertext.
That's how this attack is implemented. I've limited the size of the ciphertext to 16kb which should be more than enough in most cases.
More details about the vault format can be found here: https://docs.ansible.com/ansible/2.4/vault.html (right at the bottom)


jsteube:
This is great! Can you add test.sh and test.pl cases please?

roycewilliams:
What category/family should this be? It's not really a password manager per se, but it sort of is.

hops:
@jsteube Yes I'm working on it. Unfortunately I can't finish this today, I'll continue tomorrow.
@roycewilliams That's what I was asking myself and password manager was the category that fitted the most. I'm open to other suggestions.

hops:
I amended the commit including the changes to test.pl and test.sh.
My perl foo isn't the best so please double check my changes.
However the tests pass on a GTX 1080 and a i7-6700K.

jsteube:
Almost perfect, I'll merge this. There's three minor things missing for perfection:

- Missed entry in read docs/readme.txt
- Missed entry in read extra/tab_completion/hashcat.sh 
- Instead of OPTS_TYPE_HASH_COPY add the missing/unused attributes to an extra attribute in esalt so you can rebuild the original string using sprintf in ascii_digest()

The perl stuff looks good, too.

PR #1666, opened 2018-08-23
hcstat2: allow to use option --markov-hcstat2 instead of just --markov-hcstat

philsmd:
I think it makes sense to show in the --help output very clearly that the new hcstat2 (lzma compressed) markov file format should be used with new versions of hashcat.
I think it therefore also makes sense to allow to specify --markov-hcstat2 (but because of the getopts shortcut continue to allow --markov-hcstat too... BUT the file itself always needs to be in the new format of course since version 4.0 of hashcat!).

Thanks

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -8,6 +8,7 @@
 - Allow bitcoin master key length not be exactly 96 byte a multiple of 16
 - Getting rid of OPTS_TYPE_HASH_COPY for Ansible Vault
 - Add a tracker for salts, amplifier and iterations to status screen
+- Add option --markov-hcstat2 to make it clear that the new hcstat2 format (compressed hcstat2gen output) must be used
 
 ##
 ## Bugs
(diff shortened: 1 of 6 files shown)


jsteube:
Thanks!

JensTimmerman:
Could the help also explain how to generate these new stats?
I tried with hcstatgen.bin and hcstat2gen.bin, but both of these commands write out files that hashcat 4.2.1 then fails with: `Could not uncompress data.`

roycewilliams:
It would be good to have such an example. In the meantime, see: 

https://hashcat.net/wiki/doku.php?id=hashcat_utils#hcstat2gen

This wiki entry is very new, so it isn't showing up yet in Google search results for this problem.

JensTimmerman:
Thanks for that link, doesn't seem to work though, my lzma in fedora refuses to write to stdout for compressing, had to use 
`lzma --compress --format=raw wpa --suffix=hcstat2` and this generated a wpa.hcstat2 file that could be used by hashcat.

roycewilliams:
Good alternative - added to the wiki. Thanks!

PR #1710, opened 2018-10-02
Add support for TOTP (RFC 6238)

unix-ninja:
This algorithm needs supplementary processing to be practical. We need to reasonably reduce the keyspace while still traversing most of it in order to find the duplicate secrets (there are going to be a *lot* of collisions, but all of the collisions should be unique.) Nonetheless, the framework has been provided here, and we leave it up to the user to figure out the nuances of a practical attack.


jsteube:
This line:

> otp_code = ((ctx.opad.h[otp_offset/4] & 0xffffff) << 8) | ((ctx.opad.h[otp_offset/4+1] % 0xffffff00) >> 24);

You do:

> (ctx.opad.h[otp_offset/4] & 0xffffff) << 8

This is fine, but then you do:

> (ctx.opad.h[otp_offset/4+1] % 0xffffff00) >> 24

This looks like some bytewise align, IF you'd have used & 0xffffff00 and not % 0xffffff00. 

The outcome with both operators should be the same, but I think you can replace it with & and make it more logical than arithmetic.

This line:

> s[idx] = swap32_S (salt_bufs[salt_pos].salt_buf[idx]);

You can, if you want, do the swapping in the parser code. 

This has the advantage that hashcat don't has to call swap32_S() again and again in the kernel. 

I doubt it will give alot of performance increase on GPU, but maybe on a slow CPU you can see the difference. Note that you will have to do another swap() in the ascii_digest() to synchronize it with the original salt.

> switch(otp_offset%4)

On some systems, &3 is faster than %4 while the result is the same

If you write a mask like this 

> 0xffffff

and also use a mask like this:

> 0xffffff00

It's better to write 0x00ffffff instead of 0xffffff. This keeps the code aligned in width. While it's the same for the compiler, it's much easier for the human brain to figure out the idea if it does have to deal with alignment.

One more thing:

We can assume this mode to produce tons of collisions. Since this is the expected behavior, you can tell hashcat to enable --keep-guessing by setting it in the hash-mode switch case:

opts_type |= OPTS_TYPE_PT_NEVERCRACK;

This will tell hashcat about to continue cracking

roycewilliams:
Also, don't forget to update changes.txt - this is a cool add to be able to brag about. :)

OpenCL/m18100_a0-pure.cl
```diff
@@ -0,0 +1,190 @@
...
+
+    // calculate the offset using the least 4 bits of the last byte of our hash
+    const int otp_offset = ctx.opad.h[4] & 0xf;
+
+    // initialize a buffer for the otp code
+    unsigned int otp_code = 0;
+
+    // grab 4 consecutive bytes of the hash, starting at offset
+    // on some systems, &3 is faster than %4, so we will use it in our switch()
+    switch(otp_offset&3)
```
philsmd: the format we normally use is
```
switch (otp_offset & 3)
```
(spaces before the opening bracket and between the operators)

OpenCL/m18100_a0-pure.cl
```diff
@@ -0,0 +1,190 @@
...
```
philsmd: idem (formatting)

OpenCL/m18100_a3-pure.cl
```diff
@@ -0,0 +1,210 @@
...
+
+    // calculate the offset using the least 4 bits of the last byte of our hash
+    const int otp_offset = ctx.opad.h[4] & 0xf;
+
+    // initialize a buffer for the otp code
+    unsigned int otp_code = 0;
+
+    // grab 4 consecutive bytes of the hash, starting at offset
+    // on some systems, &3 is faster than %4, so we will use it in our switch()
+    switch(otp_offset&3)
```
philsmd: idem (add spaces)

OpenCL/m18100_a3-pure.cl
```diff
@@ -0,0 +1,210 @@
...
```
philsmd: idem (missing spaces)

src/hashes.c
```diff
@@ -275,6 +275,21 @@ void check_hash (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, pl
...
+  // TOTP should be base32 encoded
+  if (hashcat_ctx->hashconfig->hash_mode == KERN_TYPE_TOTP_HMACSHA1)
+  {
+    // we need a temp buffer for the base32 encoding
+    u32 temp_buf[64] = { 0 };
+    u8 *temp_ptr = (u8 *) temp_buf;
+
+    // encode our plain
+    base32_encode (int_to_base32, (const u8 *) plain_ptr, plain_len, (u8 *) temp_buf);
+    plain_len = strlen((const char *) temp_buf);
```
philsmd: formatting:
```
plain_len = strlen ((const char *) temp_buf);
```

src/hashes.c
```diff
@@ -275,6 +275,21 @@ void check_hash (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, pl
...
+
+    // copy the base32 content into our plain buffer
+    plain_ptr = (u8 *) strdup((const char *) temp_ptr);
```
philsmd: should be:
```
plain_ptr = (u8 *) strdup ((const char *) temp_ptr);
```
(with added space)

src/interface.c
```diff
@@ -5156,6 +5158,68 @@ int sha1s_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNUS
...
+    token.attr[1] |= TOKEN_ATTR_VERIFY_HEX;
+  }
+
+  const int rc_tokenizer = input_tokenizer (input_buf, input_len, &token);
+
+  // now we need to reduce our hash into a token
+  int otp_code = 0;
+  for (int i = 0; i < 6; i++)
+  {
+    otp_code = otp_code  * 10 + itoa32_to_int(input_buf[i]);
```
philsmd, line 5196: idem (space before the opening bracket)

src/interface.c
```diff
@@ -5156,6 +5158,68 @@ int sha1s_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNUS
...
+  }
+  if (rc_tokenizer != PARSER_OK) return (rc_tokenizer);
+
+  digest[1] = otp_code;
+
+  u8 *salt_pos = token.buf[1];
+
+  // convert ascii timestamp to ulong timestamp
+  unsigned long timestamp = 0;
+  timestamp = hc_strtoul((const char *) salt_pos, NULL, 10);
```
philsmd, line 5206:
```
timestamp = hc_strtoul ((const char *) salt_pos, NULL, 10);
```

src/interface.c
```diff
@@ -5156,6 +5158,68 @@ int sha1s_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNUS
...
+
+  // divide our timestamp by our step. We will use the RFC 6238 default of 30 for now
+  timestamp /= 30;
+
+  // convert counter to 8-byte salt
+  for (int i = 2; i--; timestamp >>= 32)
+  {
+    salt->salt_buf[i] = byte_swap_32(timestamp);
```
philsmd, line 5214:
```
salt->salt_buf[i] = byte_swap_32 (timestamp);
```

src/interface.c
```diff
@@ -22266,6 +22331,14 @@ int ascii_digest (hashcat_ctx_t *hashcat_ctx, char *out_buf, const size_t out_le
         ptr[15], ptr[14]
       );
   }
+  else if (hash_mode == 18100)
+  {
+      // salt_buf[1] holds our 32 bit value. salt_buf[0] and salt_buf[1] would be 64 bits.
+      // for now, we only need to worry about 32 bit counters.
+      // we also need to multiply salt by our step to see the floor of our original timestamp range.
+      // again, we will use the default RFC 6238 step of 30.
+      snprintf (out_buf, out_len - 1, "%06d:%d", digest_buf[1], byte_swap_32(salt.salt_buf[1]) * 30);
```
philsmd, line 22340:
```
snprintf (out_buf, out_len - 1, "%06d:%d", digest_buf[1], byte_swap_32 (salt.salt_buf[1]) * 30);
```

tools/test.pl
```diff
@@ -9971,6 +9972,23 @@ sub gen_hash
 
     $tmp_hash = sprintf ("%s", $hash_buf);
   }
+  elsif ($mode == 18100)
+  {
+    my $paddedTime = sprintf("%016x", int(int($salt_buf) / 30));
```
philsmd:
```
my $paddedTime = sprintf ("%016x", int (int ($salt_buf) / 30));
```

tools/test.pl
```diff
@@ -9971,6 +9972,23 @@ sub gen_hash
...
+    my $data = pack('H*', $paddedTime);
```
philsmd:
```
my $data = pack ('H*', $paddedTime);
```

tools/test.pl
```diff
@@ -9971,6 +9972,23 @@ sub gen_hash
...
+    $hash_buf = hmac_hex ($data, $key, \&sha1, 64);
+
+    my $offset =  hex (substr ($hash_buf, -8)) & 0xf;
+    $offset *= 2;
+    my $token = hex (substr ($hash_buf, $offset, 8));
+    $token &= 0x7fffffff;
+    $token %= 1000000;
+
+    ## token must be leading zero padded, and salt leading zero stripped
+    $tmp_hash = sprintf ("%06d:%d", $token, int($salt_buf));
```
philsmd:
```
$tmp_hash = sprintf ("%06d:%d", $token, int ($salt_buf));
```

src/interface.c
```diff
@@ -5156,6 +5158,68 @@ int sha1s_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNUS
...
+
+  // convert ascii timestamp to ulong timestamp
+  unsigned long timestamp = 0;
+  timestamp = hc_strtoul((const char *) salt_pos, NULL, 10);
+
+  // divide our timestamp by our step. We will use the RFC 6238 default of 30 for now
+  timestamp /= 30;
+
+  // convert counter to 8-byte salt
+  for (int i = 2; i--; timestamp >>= 32)
```
philsmd, line 5212: I think it doesn't make much sense to use a loop here. It's a little bit distracting/confusing.
I would suggest to unroll it and therefore make it clearer how the single salt_buf[x] will look like (2 assignments).
```
salt->salt_buf[0] = byte_swap_32 (timestamp >> 32);
salt->salt_buf[1] = byte_swap_32 (timestamp >>  0);
```

philsmd (commented):
good job.
I think the m18100_a1-pure.cl kernel is still missing and the extra/tab_completion/hashcat.sh must also be updated (currently missing).

Should we "only" name it TOTP or add "One-Time Password" such that it can be found more easily ?

unix-ninja:
@philsmd I made the changes you requested.

As for using the abbreviated form or expanding it, I don't feel strongly either way. People likely to care about this mode are probably going to recognize it, but if you really want to expand it that should be a trivial change.

roycewilliams:
If there are more than one OTP formats that could be eventually supported, giving them their own category could neatly solve the short-vs-long issue.

unix-ninja:
Right now I list it as `TOTP (HMAC-SHA1)` to address the specific OTP algorithm. Do you think that covers it?

roycewilliams:
That seems perfect for the official algo. name. If the last column of the --help output (which currently has things like "Operating Systems", etc.) could use the category name of "One-Time Passwords", I think it would meet the rest of the"findability" goal.

unix-ninja:
ok, I will add that and push in a little while.

OpenCL/m18100_a0-pure.cl
```diff
@@ -0,0 +1,190 @@
...
+
+    // calculate the offset using the least 4 bits of the last byte of our hash
+    const int otp_offset = ctx.opad.h[4] & 0xf;
+
+    // initialize a buffer for the otp code
+    unsigned int otp_code = 0;
+
+    // grab 4 consecutive bytes of the hash, starting at offset
+    // on some systems, &3 is faster than %4, so we will use it in our switch()
+    switch(otp_offset & 3)
```
philsmd: ideally we would use this format:
```
switch (otp_offset & 3)
```
add a further space after "switch" on all instances of this switch()

OpenCL/m18100_a3-pure.cl
```diff
@@ -0,0 +1,210 @@
...
+    const int otp_offset = ctx.opad.h[4] & 0xf;
+
+    // initialize a buffer for the otp code
+    unsigned int otp_code = 0;
+
+    // grab 4 consecutive bytes of the hash, starting at offset
+    // on some systems, &3 is faster than %4, so we will use it in our switch()
+    switch(otp_offset & 3)
+    {
+		case 1:
```
philsmd: minor formatting problem: there is a tab character here

OpenCL/m18100_a3-pure.cl
```diff
@@ -0,0 +1,210 @@
...
+    sha1_hmac_ctx_vector_t ctx;
+
+    sha1_hmac_init_vector (&ctx, w, pw_len);
+
+    sha1_hmac_update_vector (&ctx, s, salt_len);
+
+    sha1_hmac_final_vector (&ctx);
+
+    // calculate the offset using the least 4 bits of the last byte of our hash
+    const int otp_offset = ctx.opad.h[4] & 0xf;
```
philsmd: compiler complains about the type incompatibility (int vs u32x), we probably need to use u32x types here and modify the switch () code accordingly

philsmd:
The base32 output is also a little bit special/confusing. I guess the format expects this base32 "passwords".
It's just confusing that the password candidates (e.g. "Candidates.#1") are different compared to the final output (which is base32 encoded).
Furthermore, the -m 18100 --example output is "hashcat", but the final output is base32 ("hashcat") etc....
I can live with it, but i guess if it is not really required to encode the password with base32, it could also make sense to skip this step to avoid confusion (does the user know that the output was not the raw output but the encoded output ? I guess, that experienced users can notice that)

jsteube:
> Added hash-mode 18100 = TOTP (HMAC-SHA1)

and this:

> static const char *HT_18100 = "TOTP (HMAC-SHA1)";

and this:

> 18100 | TOTP (HMAC-SHA1)  ...

Didn't you want to change this to Onetime Token or something?

>  unsigned long timestamp = 0;
>  timestamp = hc_strtoul ((const char *) salt_pos, NULL, 10);

Two things here:

1. Do we really want to use unsigned long? Using long datatypes is bad coding practise because it changes it's size depending on the 32 or 64 bit compilation. It's better to use u32 or u64 instead.
2. If both line should be put in the same line to make it easier to read:

  unsigned long timestamp = hc_strtoul ((const char *) salt_pos, NULL, 10);

>  if (hashconfig->opts_type & OPTS_TYPE_ST_HEX)
>  {
>    token.len_min[1] *= 2;
>    token.len_max[1] *= 2;
>    token.attr[1] |= TOKEN_ATTR_VERIFY_HEX;
>  }

Is there a reason to allow hex encoded salts? From what I've seen, the salt is already encoded.

>  int otp_code = 0;
>  for (int i = 0; i < 6; i++)

Formatting:

Please put in a blank line here

> otp_code = otp_code * 10 + itoa32_to_int (input_buf[i]);

This line is the same as with the hc_strtoul. Please don't do two steps at once, it makes it hard to read.

Please convert it first into a dedicated fixed size buffer and then use hc_strtoul to convert it to a number.

> timestamp >> 32

By seeing this it means you probably want to use u64 for timestamp, not u32.

> salt->salt_buf[1] = byte_swap_32 (timestamp);
> salt->salt_buf[0] = byte_swap_32 (timestamp >> 32);

Please think on the easy read code again. In this case, we do this instead:

salt->salt_buf[1] = byte_swap_32 (timestamp >>  0);
salt->salt_buf[0] = byte_swap_32 (timestamp >> 32);

I know it's kind of OCD but it actually has some meaning.

However, you're about to assing a u64 to an u32 input of that function. 

It's better to convert it down to 2 times u32 variables first. That makes it easier to understand.

> @@ -275,6 +275,21 @@ void check_hash (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, pl

Can you please explain why you need to add code to check_hash() function? That's kind of unusual. I'm not saying it's wrong, I just want to understand the reason behind/

OpenCL/m18100_a0-pure.cl
unix-ninja: agh! I can't believe I missed that. sorry. Will fix it this morning.

OpenCL/m18100_a3-pure.cl
unix-ninja: wow- I am not sure how the tab got in there, because I have tab expansion set in vim, but it's fixed now. pushing soon.

unix-ninja:
@jsteube I changed the category in usage.c to "One-Time Password", which is what I thought Royce was talking about, to solve the find-ability issue. I didn't touch the name of the algorithm, though. This should actually be the proper way to address it, do you agree?

I guess you have a point there. I will switch timestamp to u64.

TOKEN_ATTR_VERIFY_HEX was boilerplate straight from m150. For this algorithm we prob want TOKEN_ATTR_VERIFY_DIGIT since this should be a timestamp. I will swap that.

The code in check_hash() was mostly off-loading base32 encoding of the founds to CPU. I didn't want to slow down the kernel with this. The result needs to display encoded, even though the keyspace we are searching is not.

unix-ninja:
@philsmd so, I struggled with that point myself. By the RFC, the secrets must be base32 encoded (this is the way you will input it to something like pyotp.) It seems weird to do that extra work inside the Hashcat kernel though, which would just needlessly slow down performance when you can just encode it at the end. So this way is faster, but I agree that it's confusing unless you know what's going on. I am writing a blog post to explain this attack, which I hope helps. In the end, I thought having that extra performance was probably worth it.

OpenCL/m18100_a3-pure.cl
unix-ninja: I used an explicit cast for this. I think that should be good.

jsteube:
> const int otp_offset = (int) ctx.opad.h[4] & 0xf;

This case is kind of strange. Since you commented out the vector macro I'd suggest to replace it with this:

const u32x otp_offset = ctx.opad.h[4] & 0xf;

> unsigned int otp_code = 0;

We never use "unsigned int" in the kernels. We either use just "int" or "u32" to hold some 32 bit data.

In this case I'd suggest replacing this to:

u32 otp_code = 0;

> plain_ptr = (u8 *) strdup ((const char *) temp_ptr);

This one creates a new buffer but i can not see any free() for it. memleak?

Since you're overwriting the original pointer value in plain_ptr which is not allocated memory you can not use free() on this variable unless you do the same branch again. However that's not very readable. I'd suggest to use a second fixed size plain_ptr2 the same way plain_ptr works and then you don't need to care anymore. you can then memcpy the new result into plain_ptr before leaving the block

> if (hashcat_ctx->hashconfig->hash_mode == KERN_TYPE_TOTP_HMACSHA1)

We're not using -> twice in the same variable. Please add something like this

  hashconfig_t   *hashconfig   = hashcat_ctx->hashconfig;

to the check_hash() function and then use hashconfig->hash_mode in the branch line

>  // now we need to reduce our hash into a token
>  int otp_code = hc_strtoul ((const char *) input_buf, NULL, 10);

This section should come after the PARSER_OK check. Basically same like other parsers do.

> digest[1] = otp_code;

Why are you using digest[1] here? that's kind of unnatural. I'd suggest to replace with:

digest[0] = otp_code;
digest[1] = 0;
digest[2] = 0;
digest[3] = 0;

The 1,2 and 3 helps when hashcat is sorting the hashes in memory, since you specified DGST_SIZE_4_5 in the switch case

Note to make this work, you have to modify the switch case constants too:

hashconfig->dgst_pos0      = 0;
hashconfig->dgst_pos1      = 1;
hashconfig->dgst_pos2      = 2;
hashconfig->dgst_pos3      = 3;

> It seems there's a major error here in the follow scenario. First i created a random hash (I guess there's not error so far, but please take a close look and verify)

$ echo -n hashcat | tools/test.pl passthrough 18100
504267:85457442135216

> Then I tried to crack it, but it did not work:

./hashcat -m 18100 504267:85457442135216 hashca?1 -1 t -a 3 

However, with smaller salts (less than 2^32) it works. So there seems to be a problem with larger salt values.

src/interface.c
```diff
@@ -5156,6 +5158,62 @@ int sha1s_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, MAYBE_UNUS
...
+  int otp_code = hc_strtoul ((const char *) input_buf, NULL, 10);
+
+  if (rc_tokenizer != PARSER_OK) return (rc_tokenizer);
+
+  digest[1] = otp_code;
+
+  u8 *salt_pos = token.buf[1];
+
+  // convert ascii timestamp to ulong timestamp
+  u64 timestamp = hc_strtoul ((const char *) salt_pos, NULL, 10);
```
philsmd, line 5202: there is a bug here, must be:
```
   u64 timestamp = hc_strtoull ((const char *) salt_pos, NULL, 10);
```
(i.e. use hc_strtoull () for 64 bit, instead of hc_strtoul () which would only allow 32 bit)

src/interface.c
```diff
@@ -22266,6 +22325,14 @@ int ascii_digest (hashcat_ctx_t *hashcat_ctx, char *out_buf, const size_t out_le
         ptr[15], ptr[14]
       );
   }
+  else if (hash_mode == 18100)
+  {
+      // salt_buf[1] holds our 32 bit value. salt_buf[0] and salt_buf[1] would be 64 bits.
+      // for now, we only need to worry about 32 bit counters.
+      // we also need to multiply salt by our step to see the floor of our original timestamp range.
+      // again, we will use the default RFC 6238 step of 30.
+      snprintf (out_buf, out_len - 1, "%06d:%d", digest_buf[1], byte_swap_32 (salt.salt_buf[1]) * 30);
```
philsmd, line 22334: I found that the hashes can't be cracked multiple times this way (e.g. the output of hashcat can't be cracked again for certain hashes, e.g. "hashcat" -> 504267:85457442135216). My suggestion here is something like this:
```
      u64 tmp = (((u64) byte_swap_32 (salt.salt_buf[0])) << 32) | ((u64) byte_swap_32 (salt.salt_buf[1]));

      tmp *= 30;

      snprintf (out_buf, out_len - 1, "%06d:%lu", digest_buf[1], tmp);
```
(and remove the comment about the 32 bit is enough)
Note: we need 64 bit because the * 30 could also overflow etc

philsmd (commented):
I troubleshooted the problems mentioned by @jsteube a little bit (I problably saved you a headache ;) ).
... see my code review below.

The solution with the ascii_digest () / hash output, on the other hand, is still not 100% clean, because we somehow lose the original salt value because of the integer calculation: (x / 30) * 30 . Maybe we could just add a further salt_buf[2] where we store the original value (it's a bit of cheating - a hack - but the output hash would then always be the same as the input hash)
maybe somthing like this:
totp_parse_hash ():
```
  // keep/store the original salt value:

  salt->salt_buf[3] = (u32) (timestamp >>  0);
  salt->salt_buf[2] = (u32) (timestamp >> 32);

  // divide our timestamp by our step. We will use the RFC 6238 default of 30 for now

  timestamp /= 30;

  salt->salt_buf[1] = byte_swap_32 ((u32) (timestamp >>  0));
  salt->salt_buf[0] = byte_swap_32 ((u32) (timestamp >> 32));

  salt->salt_len = 16; // only 8 are used (the 2nd half is to store the original "salt")
```

ascii_digest:
```
      u64 tmp = (((u64) salt.salt_buf[2]) << 32) | ((u64) salt.salt_buf[3]);

      snprintf (out_buf, out_len - 1, "%06d:%lu", digest_buf[1], tmp);

```

and within the kernels (all of them!) we need to set the salt length to hardcoded 8 (which could sometimes be even faster, because the compiler could optimize the fixed length if we are lucky)
```
const u32 salt_len = 8;
```


What do you think?

I think the remaining parts look very good. Good job again!

unix-ninja:
@jsteube ok, I have more commits coming in soon for some of these things. There are two things I wanted to discuss, though.

> digest[1] = otp_code;

So, I used index 1 because the SHA1 boilerplate ignores index 0 when actually comparing hashes, and if I want to reuse that boilerplate without writing new code I need to start the values I care about at 1. I don't think it's worth reinventing the wheel for this since the algorithm is based on HMAC-SHA1. I'd like to keep using as much of that existing infrastructure as possible.

> plain_ptr = (u8 *) strdup ((const char *) temp_ptr);

There is no memory leak for this one. The buffer is locally scoped and destroyed at the end of the if block. You can't even use the reference outside of that block because it will cause a compiler error. I don't think there's any real value trying to manually manage the memory here.

unix-ninja:
@philsmd omg- thanks! that's definitely a time-saver!

I will get these taken care of in a little while and push a new commit. This help is much appreciated.
🙂

philsmd:
I actually think that @jsteube is correct @ memory leak.

if we have this:
```
  u32 plain_buf[64] = { 0 };

  u8 *plain_ptr = (u8 *) plain_buf;
```

and afterwards:
```
  plain_ptr = (u8 *) strdup ((const char *) temp_ptr);
```

this last "plain_ptr = something" is just a pointer assignment (plain_ptr now "points" to a new address). It's not a memcpy ().
According to this https://en.cppreference.com/w/c/experimental/dynamic/strdup , we always need to use free () when we use strdup ().
plain_ptr = (u8 *) strdup ((const char *) temp_ptr);
will just assign a new address to plain_ptr, but it won't replace the bytes allocated by plain_buf[64] = { 0 }, but instead strdup () will return a very new address (on heap) and this new memory block must be freed to avoid memory leaks.

unix-ninja:
@philsmd I see. Apparently I misunderstood what strdup() was doing under the hood. In that case, I will amend as jens recommended. Thanks for the clarification!

jsteube:
> snprintf (out_buf, out_len - 1, "%06d:%llu", digest_buf[1], tmp_salt_buf);

this needs to be written like this:

> snprintf (out_buf, out_len - 1, "%06d:%" PRIu64, digest_buf[1], tmp_salt_buf);

you also need to add this to interface.h in order to enable PRIu64

#include <inttypes.h>

>    base32_encode (int_to_base32, (const u8 *) plain_ptr, plain_len, (u8 *) temp_buf);
>    plain_len = strlen ((const char *) temp_buf);

I think you can rewrite this to just:

plain_len = base32_encode (int_to_base32, (const u8 *) plain_ptr, plain_len, (u8 *) temp_buf);

>                hashconfig->dgst_pos0      = 1;
>                hashconfig->dgst_pos1      = 2;
>                hashconfig->dgst_pos2      = 3;
>                hashconfig->dgst_pos3      = 4;

Please rewrite it to use 0, 1, 2 and 3 instead. I see your point with staying stick to sha1-hmac
but in my view its not correct. The last one who touches the digest is not sha1,
it's the algorithm you use. Therefore the compiler can create not auto optimized code by assuming
the second element is the first element.

Don't forget to change the code in parser code, too

> DGST_SIZE_4_5

I think you can set it to DGST_SIZE_4_4 afterwards and save some memory

>

unix-ninja:
@jsteube changed as requested. the smaller digest size made a small performance bump to my benches. It's not a whole like, just 50k-100k or so, that's still cool. thanks!

PR #1719, opened 2018-10-17
stdin: add read timeout checks (abort if no input for a long time)

philsmd:
The problem we try to fix here is whenever the user forgets to specify any attack mode or dictionary file, the stdin mode is automatically used and it could happen that the user doesn't understand that hashcat expects the input from stdin (a pipe).
With these few code changes we detect if nothing can be read from stdin and if we exceed a certain wait limit (currently it is set to 120 seconds, which is 2 minutes), we abort the cracking job.
Before we abort the cracking job the user is warned (several times on a regular time frame) that hashcat was unable to read from the pipe/stdin.

Examples of when this could happen (wrong commands):
```
hashcat -m 0 hash.txt
```
(here hashcat waits for input on stdin because no dictionary file was provided)

second example:
```
while :; do sleep 180; echo; done | hashcat -m 0 hash.txt
```
(here the pipe is way too slow and hashcat can't use the full GPU power. We can say that the "password candidat generator is way too slow")

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -36,6 +36,7 @@
 - Increased the maximum size of edata2 in Kerberos 5 TGS-REP etype 23
 - Allow hashfile for -m 16800 to be used with -m 16801
 - Make the masks parser more restrictive by rejecting a single '?' at the end of the mask (use ?? instead)
+- Add a periodic check for read timeouts in stdin/pipe mode and abort if no input was provided
 
 ##
 ## Bugs
--- a/include/monitor.h
+++ b/include/monitor.h
@@ -6,6 +6,9 @@
 #ifndef _MONITOR_H
 #define _MONITOR_H
 
+#define STDIN_TIMEOUT_MIN  20 // warn  after no input from stdin for x seconds
+#define STDIN_TIMEOUT_MAX 120 // abort after no input from stdin for x seconds
+
 int get_runtime_left (const hashcat_ctx_t *hashcat_ctx);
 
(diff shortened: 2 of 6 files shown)


jsteube:
Great! Thanks!

magnumripper:
I can picture several use cases where this is definitely not wanted - we'll need an option to disable it!

roycewilliams:
Also a way to change the timeout period. Since the default used to be to never time out, maybe a --stdin-timeout=[seconds] option, and when the user uses "0" (zero) for [seconds], it is disabled?

s3inlc:
But when the default is to never timeout, then the idea of alerting user forgetting the attack mode or a dictionary would not work anymore.

magnumripper:
He said it **used** to be the default, we're not suggesting it should be now.

s3inlc:
Ah sorry, my fault, I misunderstood this part.

jsteube:
https://github.com/hashcat/hashcat/commit/d4123333c0a6f7836df8ce40b56b9dc308da2925 should help

gavriilos:
(Thanks pointing me here @magnumripper)

I second @magnumripper and @roycewilliams opinion that this should have both an option to disable and an option to increase the timeout. As I described on #1744 I am faced with a real use case that was broken by this stdin timeout.

d4123333c0a6f7836df8ce40b56b9dc308da2925 does not help on my use case because the candidate generator I am using takes a long time before outputting the first candidate.

PR #1726, opened 2018-10-24
Fixed Streebog hash parser

Naufragous:


s3inlc:
Are you sure that this should be switched? Because there might be some confusion between some implementations (e.g. https://github.com/adegtyarev/streebog) which for some reason mirrored the example hashes compared to the original RFC (https://tools.ietf.org/html/rfc6986#section-10.1.2). And as far as I see, the current way hashcat is doing is the way how it's given by the examples in the RFC.

Naufragous:
Indeed, you are right.

There definitely is some confusion, given that Alexey Degtyarev is both the maintainer of that implementation and the co-author of the RFC. I found the rationale behind his approach in the commit message of adegtyarev/streebog@dd005b5cd681ea01665c73298ba908ddf8a62851:
> * Feature: command line options "-e" introduced (switch endianness). The standard does not specifies explicitly how to output the resulting hash. This implementation outputs hash in hexadecimal format in order how bytes are represented in memory - least significant first. With this option set, program outputs digest in hexadecimal format in reversed order - most significant first.

I wonder how accurate this statement is, given that all the example digests are presented in big-endian order. I guess we should stick with big-endian, since it is more intuitive from the user point of view.

What *needs* to be fixed is the `ascii_digest` branch for mode 11800 (Streebog-512). Currently, if you crack the example hash you get the output
```
a4db5b5dee898f8c110e[...]:hashcat
```
instead of
```
5d5bdba48c8f89ee6c0a[...]:hashcat
```
I'll amend the commit so that we won't need to open another PR.

PR #1734, opened 2018-10-27
error correction

honeyakshat999:


jsteube:
Pretty sure "So on" is correct and "Soon" is not

Issue #1707, opened 2018-09-28
Support for Kerberos ASREP type 23


skelsec:
I've made some modifications on the m13100 ( Kerberos 5 TGS-REP etype 23) code to make it work against ASREP tickets. As I don't really have the knowledge to create a PoC for additional changes like create a new module or adding a new parser for the hash etc on the hashcat code, I'd like to ask for a little help. (hence no PR) 

Chanes made:
1. changed the keyusage parameter to get the correct RC4 key
2. changed the verification for the expected ASN1 data, because it's a different type of ticket. This probably needs to be improved, as it's currently only verifying 2 bytes on the start.

IMPORTANT: test data format is not representing the correct data type, but since I dunno how to add additional parsers/module to hashcat I had to improvise :)

Test data: [hashes.txt](https://github.com/hashcat/hashcat/files/2429313/hashes.txt)
Passwords: [passwords.txt](https://github.com/hashcat/hashcat/files/2429314/passwords.txt)
Modified module: [m13100_a0-optimized.cl.txt](https://github.com/hashcat/hashcat/files/2429318/m13100_a0-optimized.cl.txt) -sorry for txt, ovbiously rename it pls-

To make it run, you'd need to:
1. overwrite the original m13100_a0-optimized.cl module with the modified module
2. run a dictionary attack against the test data such as:
```hashcat64.exe -O -m 13100 -a 0 hashes.txt passwords.txt --self-test-disable```

Attack described [here](http://www.harmj0y.net/blog/activedirectory/roasting-as-reps/)

jsteube:
Good info for start, thanks. Looks doable :)

Issue #1622, opened 2018-07-19
Feature Request - Kerberos AS-REP hast type


mubix:
Stealing @kholia 's post from https://github.com/magnumripper/JohnTheRipper/pull/2709:

Background,
- https://github.com/HarmJ0y/ASREPRoast
- http://www.harmj0y.net/blog/activedirectory/roasting-as-reps/
- http://www.zytrax.com/tech/survival/kerberos.html (basics)

Sample data:
- http://www.exumbraops.com/layerone2016/party (sample.krb.pcap, packet number 1114, AS-REP with password P@$$w0rd123).
- 

Example Hash (ASDqwe123):
```
$krb5asrep$23$ima-domainadmin@SITTINGDUCK.INFO:3624689b65105348df911bade0933c5f$442e29367a90072dfbf9dc34586943135a26c394c6e1324204475767fb80c1670305cd19aa2d8debe0eaf515f12fd3a977822964b85c7ea1dff77b0bf926b933ce8e2e56148ba6b43a5987c5f8b2b0661c360c242a42fd2ca61e61dc3921b9fdc9941518af0a81dc9fc8388b128bb7d11c9adf4faf79a5e49e439e68ef8cbd4b30a1e6835579978a2ce27029448e1bd1d4fc78a0e96c8ae572ef8059b170a784fdc52bb9c993411a06db4dbcf688e0f2b6186c4516327ade4a48747635e1a186b16b649d39c5be5e062945bb8b0823694c5cac0dc749e0393b3ff781f0cae066ca132e9bde484ff56238674b92f2288862ee0e94
```

Using JOHN to crack it:
```
Loaded 1 password hash (krb5asrep, Kerberos 5 AS-REP etype 17/18/23 [MD4 HMAC-MD5 RC4 / PBKDF2 HMAC-SHA1 AES 128/128 AVX 4x])
Will run 4 OpenMP threads
Press 'q' or Ctrl-C to abort, almost any other key for status
ASDqwe123        ($krb5asrep$23$ima-domainadmin@SITTINGDUCK.INFO)
1g 0:00:00:00 DONE (2018-07-19 12:20) 100.0g/s 400.0p/s 400.0c/s 400.0C/s ASDqwe123
```

skelsec:
You may merge this into mine :) [issue here](https://github.com/hashcat/hashcat/issues/1707)
@mubix The issue has a solution as well but currently overwrites the -m13100 flag. Bit of a manual work but it's tested.

[...]

PR #1747, opened 2018-10-30
Added hash-modes 18200 (Kerberos 5 AS-REP etype 23)

mohemiv:
#1707 #1622 
Just added AS-REP hashes support


jsteube:
Thanks!

yellow-starburst:
I used the most recent hashcat - 11/29 and used -m 18200 and received an error - "No hashes loaded"

mohemiv:
Could you provide us a sample hash in the format that you are using?

yellow-starburst:
I found out the problem. I needed to add 23$ to the end of the first word. For example the hash needs to be - $krb5asrep$23$ not $krb5asrep$.

PR #1780, opened 2018-11-12
Clean up test suite scripts

Naufragous:
The test scripts have grown to be quite big (over 15000 lines) and
are hard to navigate. There are multiple if branches with over
40 conditional checks chained together. This commit solves some of
those issues.

 - Unite big repetetive if conditions into clean array lookups
 - Move 'install help' commands to a separate shell script
 - Adjust array lookup in test.sh to behave more intuitive
 - Add comments at key points to simplify navigation
 - Code formatting

### DIFF
--- a/tools/install_modules.sh
+++ b/tools/install_modules.sh
@@ -1,0 +1,65 @@
+#!/bin/bash
+
+##
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+## Test suite installation helper script
+
+# Sum of all exit codes
+ERRORS=0
+
+cpan install Authen::Passphrase::LANManager \
+             Authen::Passphrase::MySQL323   \
+             Authen::Passphrase::NTHash     \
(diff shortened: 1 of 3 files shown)


jsteube:
- Can you please replace the smart match operator with a normal function which takes the array and the keyword instead, analogue to is_in_array() in the shellscript?
- Is the pygost working with pip3 as well? In that case, everything is OK. Otherwise please replace the "pip" with "pip2" in the depency install shellscript
- OCD error:
``` 
VC_MODES="    13711 13712 13713 13721 13722 13723 13731 13732 13733 13751 13752\
  13753 13771 13772 13773"
```

Naufragous:
> replace the smart match operator with a normal function

Done.
> Is the pygost working with pip3 as well?

It is, but since the python code is executed with `python2 -c`, pip2 should be used indeed. Done.

> OCD error

Done :)

PR #1783, opened 2018-11-16
Replace kernel parameter lists with macros

Naufragous:
Most of the kernel functions use nearly identical parameter lists. Essentially, there are four parameters that vary. This means that the function declarations can be hidden behind a few simple macros with up to two parameters.
- Unify kernel function parameter names
- Add macros for kernel function declarations
- Substitute long parameter lists in ~2900 kernel function declarations with macros.


jsteube:
Great work again, thanks!

The parameter for hooks is currently set to void for all the macros. I assume you will push another some PR for the one kernel which is using the hooks.

Naufragous:
>I assume you will push another some PR for the one kernel which is using the hooks.

Yes, 7zip and scrypt kernels are next on the TODO list.

PR #1799, opened 2018-12-03
Add swedish keyboard layout mapping

markusklock:


jsteube:
This is great, thanks! Exactly what I was looking for. I'll add the missing space and replace the spaces with the tab character.

PR #1801, opened 2018-12-04
Set github master to non-production mode

philsmd:
We should unset the production flag in the Makefile. We are currently back to dev mode ;)

Thx


jsteube:
Indeed :)

PR #1804, opened 2018-12-05
Add support for Open Document Format 1.2

### LINKED ISSUE #101: support for recovering ODF (document) passwords
This feature request was originally reported on the (now obsolete and offline) **trac** ticketing system of hashcat.net.

---

| Ticket details: |  |
| --- | --- |
| Original reporter (OP): | a.ryabkov |
| Title: | add support for ODF |
| Ticket number: | 657 |
| Date reported: | 2015-09-09T11:48:20+02:00 |
| **Description**: | Maybe [you could] add ODF support in oclhashcat[?]<br><br>Some info about ODF - PBKDF2(HMAC-SHA-1, 1024,SHA256(password)) and then AES more:<br>\- [wikipedia - ODF specification](https://en.wikipedia.org/wiki/OpenDocument_technical_specification)<br>\- [jtr - odf2john.py](https://github.com/piyushcse29/john-the-ripper/blob/master/run/odf2john.py) |

Naufragous:
Contains a kernel for the latest ODF 1.2 encryption implemented in LibreOffice. The algorithm uses a SHA-256 checksum, a PBKDF2-HMAC-SHA1 key derivation with 100000 iterations and key stretching and AES-CBC encryption.

Valid hashes can be extracted with the [libreoffice2john.py](https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/run/libreoffice2john.py) script, available from the John the Ripper Jumbo repository.

Use hashcat with the `--username` option to process those hashes.

 - Add hash-mode 18400 (ODF 1.2 (SHA-256, PBKDF2-HMAC-SHA1, AES))
 - Tests: add hash-mode 18400 (ODF 1.2 (SHA-256, PBKDF2-HMAC-SHA1, AES))

Closes #101


jsteube:
Great work again, thanks!

Some comments:

- Please add a string substring (ODF) to the human readable descriptions for example in usage.c, interface.c, readme.txt and changes.txt. It's more likely people search for ODF instead of "Open Document Format"
- Please add missing entries in docs/readme 
- Please add missing entries in tab_completion/install.sh
- test.sh: Please add entry for 18400 to the SLOW_ALGOS array (it's possible there's some other modes missing as well)
- test.pl: The mode 18400 should not be listed in COMMON_UNSALTED_MODES - at least since this a salted algorithm it creates confusion with the word UNSALTED?! please check previous PR as well
- interface.c: This is ok since out_buf is a lot larger, but it still looks suspicious:

snprintf (&out_buf[159 + 2048], out_len - 1, "%s", odf12->suffix);

I'd have expected something like:

snprintf (&out_buf[159 + 2048], out_len - (1 + 159 + 2048), "%s", odf12->suffix);

but even that looks a bit strange. Maybe theres even a different solution with a temporary buffer

- interface.c:

missing hashconfig->hash_type      =  

- interface.h:

ROUNDS_LIBREOFFICE should be ROUNDS_ODF12 ?

- kernel:

please replace:

[16 * i + 0];

with:

const i16 = i * 16;

[i16 +  0] (notice 2x space because aligning)

- kernel:

please replace:

 pt1[0] = swap32_S (pt1[0] ^ iv[0]);

 with:

 pt1[0] ^= iv[0];

 Do all the swapping in a final block before hashing

Naufragous:
Thank you for the comments, I really did forget about a few things.
>add ODF to the human readable descriptions
>add missing entry in docs/readme.txt
>add missing entry in extra/tab_completion/hashcat.sh
>add 18400 to SLOW_ALGOS
>mode 18400 should not be listed in COMMON_UNSALTED_MODES
>interface.c: snprintf (&out_buf[...
>kernel: replace [16 * i + 0] with const i16
>kernel: group swaps together

All done.
>missing hashconfig->hash_type

Done, although not necessary for this hash mode.
>ROUNDS_LIBREOFFICE should be ROUNDS_ODF12?

Probably not... The ODF 1.2 standard does not require a specific iteration count beyond the 1,024 default. The parser expects anything between 1,000 and 999,999. The current LibreOffice version uses 100,000, that is why I choose it as the benchmark constant.

Updated the PR & rebased on current master (09ec2a9).

tools/test.pl
```diff
@@ -10597,6 +10598,43 @@ sub gen_hash
 
     $tmp_hash = sprintf ('$fvde$%d$%d$%s$%d$%s', $Z_PK, length ($salt_bin), unpack ("H*", $salt_bin), $iterations, unpack ("H*", $blob_bin));
   }
+  elsif ($mode == 18400)
+  {
+    # Complex hash string with IV, salt, checksum, ciphertext and length
+    # Do not use salt_buf, do everything here in python
+    my $python_code = <<"END_CODE";
```
philsmd: do we really need python code here?
I didn't find anything that can't be done with perl directly in the code block below. Maybe I'm missing something.
Please explain why you use python here

philsmd: in general (a general question/discussion): it would also make sense to use a single additional programming language if perl really has absolutely no way/modules to do it. I think using PHP and python is acceptable, but if we can avoid it and stick to 1 additional language it would be even better. (I think the reason why we added the PHP code was just that perl had no modules and we needed that mode during a cmiyc contest etc... therefore it was a quick "hack")

Naufragous: You are right, we don't need python code here. That's the just code I used while writing the kernel, and since I am not a perl monk I wrote it in python. But I have to rewrite this anyway, because right now there is no way to properly use the verify function.

Btw the last time I commited python code to the test suite was because of Streebog hashing. Didn't find anything on CPAN for that either.

philsmd: yeah, I figured that out. so my idea would be to convert the code of -m 11900 = PBKDF2-HMAC-MD5 to python too such that we only have the python dependency. Do you think that makes sense? hashlib.pbkdf2_hmac in python2 probably ?
Furthermore, the base64 conversion of the words/salts (https://github.com/hashcat/hashcat/blob/master/tools/test.pl#L8226-L8227) should be done also for the python code. We do this to avoid any code injection and also to avoid any strange/special characters as input resulting in strange behaviour (this should be pretty easy to do in python too, just encode in perl and decode the strings in python).
we should also update tools/install_modules.sh if we need any additional/special python modules (btw: should we rename the file "install_modules.sh" into something that explains that this is needed for tests ? (install_test_modules.sh ? test_install_modules.sh ?) 

BTW: I do not think everything I mentioned here should be done within this commit/PR, the PHP->python could/should be a separate PR/issue (it's just something I noticed here and wanted to mention and ask if python is really the way to go here)

src/interface.c
```diff
@@ -299,6 +299,7 @@ static const char *ST_HASH_18000 = "2fbf5c9080f0a704de2e915ba8fdae6ab00bbc026b2c
...
...0db7aec89972a32f5daa2a73c5895d1fced01b3ab8e576bd2630eff01cad97781f4966d4b528e1b15f011f28ae907a352073c96b203adc7742d2b79b2e2f440b17e7856ae119e08d15d8bdf951f6d4a3f9b516da2d9a8f9dd93488f8e0119f3da19138ab787f0d7098a652cccd914aa0ff81d375bd6a5a165acc936f591639059287975cfc3ca4342e5f9501b3249a76d14e56d6d56b319e036bc0449ac7b5afa24ffbea11babed8183edf8d4fdca1c3f0d23bfd4a02797627d556634f1a9304e03737604bd86f6b5a26aa687d6df73383e0f7dfe62a131e8dbb8c3f4f13d24857dd29d76984eac6c45df7428fc79323ffa1f4e7962d705df74320141ed1f16d1ad483b872168df60315ffadbfa1b7f4afaed8a0017421bf5e05348cb5c707a5e852d6fee6077ec1c33bc707bcd97b7701ee05a03d6fa78b0d31c8c97ea16e0edf434961bd5cc7cbb7eb2553730f0405c9bd21cee09b3f7c1bc57779fdfc15f3935985737a1b522004c4436b631a39a66e8577a03f5020e6aa41952c0662c8c57f66caa483b47af38b8cb5d457245fd3241749e17433e6f929233e8862d7c584111b1991b2d6e94278e7e6e1908cee5a83d94c78b75a84a695d25aeb9fdde72174fe6dd75e8d406671f44892a385a4a1e249f61ebc993e985607423a0a5742e668d52c1ebf5cecae7c2b7908f4627b92ec49354a9ccff8cb5763ad074a00e65a485a41bf4c25ce7e6fae49358a58547b1c0ca79713e297310c0a367c3de196f1dd685ca4be643bdf1e4f6b034211d020557e37a3b6614d061010b4a3416b6b279728c245d3322:::gecos (optional)::filename.odt";
```
philsmd, line 302: What I noticed here is that you allow the file names to be included within the hash. As you can see here https://hashcat.net/wiki/example_hashes , we never include the file names within the hashes.
I think to be consistent, we also need to stick with -m 18400 to the standard format that hashcat doesn't use file names within the hashes (except if needed by the hashing algorithm). The problem is that if we add this "suffix" here, we would need to add it to many, many other formats too (like ms office, docx, pdf etc).
In my opinion the hash shouldn't be mixed with a file name (we could in theory use the --username feature to add some extra data/info at the beginning of the single lines of a hash file).

src/interface.c
```diff
@@ -10987,6 +10990,175 @@ int oldoffice34cm2_parse_hash (u8 *input_buf, u32 input_len, hash_t *hash_buf, M
...
+  token.attr[7]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_HEX;
+
+  token.len_min[8] = 2;
+  token.len_max[8] = 2;
+  token.sep[8]     = '*';
+  token.attr[8]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_DIGIT;
+
+  token.len_min[9] = 32;
```
philsmd, line 11062: hmm, shouldn't these len_min == len_max all using TOKEN_ATTR_FIXED_LENGTH instead ?

src/interface.c
Naufragous: That's what I initially did. Afterwards I added the suffix buffer to ensure a better UX; you can take the output of libreoffice2john.py and throw it at hashcat without editing. But if this is not required, then I actually would be glad to get rid of the suffix! Simplifies the parsing.

@jsteube, what will it be?

jsteube: OK, remove it.

src/interface.c
Naufragous: I'd be happy to use `FIXED_LENGTH` (like I did for the 2048 chars of ciphertext at the end), but the tokenizer will automatically omit the separating `*` only if I use `VERIFY_LENGTH`. Or I am missing something?

jsteube: The TOKEN_ATTR_FIXED_LENGTH is ideal for hashes when there is no separator character (like descrypt) while at the same time there are still different columns. The macro name maybe is not ideal to reflect that, but TOKEN_ATTR_VERIFY_LENGTH is correct use here.

Naufragous:
Update:
- use perl modules instead of python code in the test suite
- complete test suite with proper verify and passthrough branches
- remove support for hash suffix generated by libreoffice2john.py


Rebased on current master (0db6913).

tools/test.pl
Naufragous: Done! :D

docs/changes.txt
```diff
@@ -1,11 +1,23 @@
...
 ## Bugs
 ##
 
 - Fixed output of IKE PSK (mode 5300 and 5400) hashes to have separators at right position
 
+##
+## Technical
+##
+
+- Tests: Added hash-mode 18400 (Open Document Format (ODF) 1.2 (SHA-256, AES))
```
philsmd, line 19: minor consideration: we normally do not add any lines/entries for new tests in docs/changes.txt (except if the tests were missing in older releases and we finally managed to complete/add the tests). I think it is enough to just have the other line under Algorithms, because usually if we add a new algo it comes accompanied with tests (as in this case). Again, this is not really a problem. It's just a note and a hint to make it more consistent for other/future PRs ;)

jsteube:
Many thanks!

PR #1805, opened 2018-12-05
Electrum Salt-Type 2

### LINKED ISSUE #1510: Electrum salt type 2 support
Feature request for adding support for Electrum salt type 2.
Note: Currently labeled as algorithm "Wallet (Salt-Type 1-3)" but support is limited to type 1 (actually type 2 will be processed but never work).

See also
#273
#1495

mcovalt:
This adds support for Electrum Salt-Type 2 hashes. It follows the [method implemented in JohnTheRipper](https://github.com/magnumripper/JohnTheRipper/issues/2504#issuecomment-309152292). If the hash has successfully been cracked, the first four characters of the plaintext are `xprv` and the following 12 characters are valid [Base58](https://en.wikipedia.org/wiki/Base58) characters.

Closes #1510

### DIFF
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -236,6 +236,26 @@ DECLSPEC int is_valid_hex_32 (const u32 v)
   return 1;
 }
 
+DECLSPEC int is_valid_base58_8 (const u8 v)
+{
+  if (v > 'z') return 0;
+  if (v < '1') return 0;
+  if ((v > '9') && (v < 'A')) return 0;
+  if ((v > 'Z') && (v < 'a')) return 0;
+
+  return 1;
+}
+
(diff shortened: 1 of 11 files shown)


jsteube:
Good stuff, thanks! 

Do I understand it correctly that only version 1 was actually working and with your PR version 2 will be supported as well? In that case I think we need to modify interface.c a bit, too:

In electrum_wallet13_parse_hash:

  if ((salt_type == 1) || (salt_type == 2) || (salt_type == 3))

It should be just:

  if ((salt_type == 1) || (salt_type == 2)) ?

Can you update the PR?

mcovalt:
> Do I understand it correctly that only version 1 was actually working and with your PR version 2 will be supported as well?

Correct.

PR #1829, opened 2018-12-18
Allow cracking non-unique salts for Office 2013

### LINKED ISSUE #1826: Can only find one hash of two - Office 2013
This issue is a placeholder for the problem explained here :
https://hashcat.net/forum/thread-8006-post-43114.html

Hashcat is not able to find multiple Office-hashes in one file; when running one hash at a time, it can do the job without any problem

Naufragous:
With hash-mode 9600 (MS Office 2013) there can be multiple hashes with
the same salt but with different encryption verifiers in `esalt_bufs`.
This commit adds the functionality to execute `_comp` kernels for
different hashes after deriving their common key once.

Fixes #1826


jsteube:
Great, thanks!

PR #1844, opened 2019-01-03
Added test modules for mode 101, 111 and 112

s3inlc:


jsteube:
Salt length constrains in -m 111 cannot be -1

jsteube:
In -m 112, if there's a fixed length  salt it should be fixed in both pure and optimized mode

PR #1848, opened 2019-01-06
Add zlib support for loading hashlist/wordlist

matrix:
Hola,
I coded a new feature, it's now possible load hashlist and/or wordlist in gzip format.
Hashcat detect the file format without using any options from command line.

@jsteube for make it works you need also add as submodule the official zlib github repository:
[madler/zlib](https://github.com/madler/zlib.git)

For build the Mac OSX binary (probably) must be recompile and install manually the zlib library. From the hashcat dir (libz must be present in deps before doing that):

```
mkdir -p deps/libz/b
cd deps/libz/b
cmake ..
make
make install
```

Following the POC from Mac OSX:

```
$ cat example500.gz.sh
./hashcat -m 500 example500.hash.gz example.dict.gz
$ ./example500.gz.sh
hashcat (v5.1.0-43-g69e96b84+) starting...

OpenCL Platform #1: Apple
=========================
* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, skipped.
* Device #2: Iris, 384/1536 MB allocatable, 40MCU

Hashes: 1 digests; 1 unique digests, 1 unique salts
Bitmaps: 16 bits, 65536 entries, 0x0000ffff mask, 262144 bytes, 5/13 rotates
Rules: 1

Applicable optimizers:
* Zero-Byte
* Single-Hash
* Single-Salt

Minimum password length supported by kernel: 0
Maximum password length supported by kernel: 256

ATTENTION! Pure (unoptimized) OpenCL kernels selected.
This enables cracking passwords and salts > length 32 but for the price of drastically reduced performance.
If you want to switch to optimized OpenCL kernels, append -O to your commandline.

Watchdog: Hardware monitoring interface not found on your system.
Watchdog: Temperature abort trigger disabled.

Dictionary cache hit:
* Filename..: example.dict.gz
* Passwords.: 128416
* Bytes.....: 511731
* Keyspace..: 128416

$1$uOM6WNc4$r3ZGeSB11q6UUSILqek3J1:hash234

Session..........: hashcat
Status...........: Cracked
Hash.Type........: md5crypt, MD5 (Unix), Cisco-IOS $1$ (MD5)
Hash.Target......: $1$uOM6WNc4$r3ZGeSB11q6UUSILqek3J1
Time.Started.....: Sun Jan  6 19:31:15 2019 (1 sec)
Time.Estimated...: Sun Jan  6 19:31:16 2019 (0 secs)
Guess.Base.......: File (example.dict.gz)
[...]

PR #1849, opened 2019-01-08
Add zlib support to load hashlist/wordlist

matrix:
Patch from #1848 adapted to hash-mode-plugin branch

### DIFF
--- a/docs/credits.txt
+++ b/docs/credits.txt
@@ -18,13 +18,20 @@ Philipp "philsmd" Schmidt <philsmd@hashcat.net> (@philsmd)
 
 Gabriele "matrix" Gristina <matrix@hashcat.net> (@gm4tr1x)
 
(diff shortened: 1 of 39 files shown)


jsteube:
As discussed, please use LZMA instead.

matrix:
> As discussed, please use LZMA instead.

I think to add also the LZMA support, but don't put in queue this pull request. It could become too big to review.

matrix:
I need to push also the zlib github repository in deps ? (I see lzma SDK it's statically include into that folder)

jsteube:
No that's the reason why I cannot accept the PR because I don't want the zlib SDK deps. For LZMA, currently it's just parts of the LZMA SDK. Please add the remaining files from original LZMA SDK which you need for decompression. If you do not want to update the current PR, please close it and create a new one for LZMA.

PR #1851, opened 2019-01-08
Add SHA2-512 module

matrix:
@jsteube check if I wrong to porting something

### DIFF
--- a/modules/module_01700.c
+++ b/modules/module_01700.c
@@ -1,0 +1,214 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "common.h"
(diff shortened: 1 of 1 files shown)


jsteube:
It's almost good. I think the OPTS_TYPE_PT_GENERATE_LE is wrong, should be OPTS_TYPE_PT_GENERATE_BE. See src/interface_migrate.c for original settings. After change, please update PR. I think you can 1:1 transfer it settings from interface_migrate.c. If 1:1 transfer is not possible, something else has to be changed first. Keep in mind the interface is very fresh :)

Ideally also add the unit test tools/test_modules/m01700.pm then you can automate the check.

PR #1870, opened 2019-01-14
Stdio.h issue, but it's not getting resolved

Raks-coder:
This file wasn't working in mac says include stdio.h


Raks-coder:
Please look into this

PR #1872, opened 2019-01-16
module-20: Added module for hashmode 20.

pellejohnk:
Is it preferred to remove the code for the hashmode from _interface.migrate.c_ ?

And does the hashmode need to be added anywhere else in the code?

### DIFF
--- a/src/modules/module_00020.c
+++ b/src/modules/module_00020.c
@@ -1,0 +1,221 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "common.h"
+#include "types.h"
+#include "modules.h"
+#include "bitops.h"
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 3;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 1;
+static const u32   DGST_SIZE      = DGST_SIZE_4_4;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_RAW_HASH_SALTED;
+static const char *HASH_NAME      = "md5($salt.$pass";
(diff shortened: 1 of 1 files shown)


jsteube:
> Is it preferred to remove the code for the hashmode from interface.migrate.c ?

Absolutely. Also interface_migrate.h

> And does the hashmode need to be added anywhere else in the code?

In this case, no. For other conversion modules, we also need to add unit test module where it does not exist.

For the time after conversion when we add truely new modes we still need to edit the files:

docs/readme.txt
docs/changes.txt

jsteube:
There's a couple of errors in the module. I'll merge this now and fix the errors with the idea to make it easier to spot the changes needed.

jsteube:
Here's the commit which can be used to spot the missing changes: https://github.com/hashcat/hashcat/commit/3b9a26c9f609f20e1720b3b757bfb15c134b124d

- Removed old code from interface_migrate.c
- Fixed typo in HASH_NAME
- Added OPTS_TYPE_STATE_BUFFER_LE (new option) which helps the decoder_apply_options() function to decide if it has work to do. Currently it is active only for OPTS_TYPE_STATE_BUFFER_BE
- Removed salt_t *salt = hash_buf->salt; from old parser function. The variable is set as fixed parameter in the function declaration
- Modified u32 *digest = (u32 *) digest_buf; because digest_buf is now void *. This way we can easily switch between u32 and u64. The same goes for esalt and the hash-mode specific esalt structures
- Fixed compiler warning related to input_tokenizer(). This is something I will maybe change in the future and switch it back to char * but for now let's stick to u8 *
- if (hashconfig->opti_type & OPTI_TYPE_PRECOMPUTE_MERKLE) { ... } : This branch is now handled inside the decoder_apply_optimizer() function and the opposite function for the encoder also exists.
- if (hashconfig->opts_type & OPTS_TYPE_ST_HASH_MD5) { ... } : I will remove this option type as it is relevant only for two or three hash-mode specific parser functions. For them we will add this branch directly into their decoder functions.

Please make sure the module compiles cleanly and and the self-test does not fail before sending the PR.

PR #1876, opened 2019-01-17
Fix Makefile NATIVE build

matrix:


jsteube:
This is a bad Idea. The compiler will compile the code from for example src/shared.c into the module binary. The module size will grow and the they will have (eventual) different code from src/shared.c in case the module was compile for an older version of hashcat (this is a situation in the future). It's better to let modules use functions from host binary instead, which is why I've add -Wl,--export-dynamic in the first place. This concept works fine on linux using gcc and clang, so I hoped it also worked on OSX. It does not work on windows however, therefore I've add the exception on windows.

Other fixes like dnclen please make them separate PR and I'll merge them.

matrix:
I got this error, also with -export-dynamic on OSX

> Undefined symbols for architecture x86_64:
  "_decoder_apply_optimizer", referenced from:
      _module_hash_decode in module_00000-6d3308.o
  "_decoder_apply_options", referenced from:
      _module_hash_decode in module_00000-6d3308.o
  "_encoder_apply_optimizer", referenced from:
      _module_hash_encode in module_00000-6d3308.o
  "_encoder_apply_options", referenced from:
      _module_hash_encode in module_00000-6d3308.o
  "_hex_to_u32", referenced from:
      _module_hash_decode in module_00000-6d3308.o
  "_input_tokenizer", referenced from:
      _module_hash_decode in module_00000-6d3308.o
  "_u32_to_hex", referenced from:
      _module_hash_encode in module_00000-6d3308.o

jsteube:
Maybe you can find a different solution which does not involve compiling the code into the module. That would be great. I cannot check since I do not have OSX computer.

PR #1881, opened 2019-01-20
Add -m 17300 module

mhasbini:


jsteube:
Thanks! There's a lot of invalid removals, probably because you confused sha2-224 with sha3-224? Please double check everything, especially if you actually used keccak_224_parse_hash(). This function usually does not have any optimizers like -= SHA224M_A therefore assume the wrong function was converted.

mhasbini:
Thanks for checking! I confused sha2-224 with sha3-224 and used the wrong parsing function.

I updated the PR.

jsteube:
Perfect, thanks!

PR #1892, opened 2019-01-27
WIP: Add -m 11500 module & unit test

mhasbini:
Still investigating test failures.

### DIFF
--- a/src/modules/module_11500.c
+++ b/src/modules/module_11500.c
@@ -1,0 +1,168 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "common.h"
+#include "types.h"
+#include "modules.h"
+#include "bitops.h"
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_4_4;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_RAW_HASH;
+static const char *HASH_NAME      = "CRC32";
+static const u32   HASH_TYPE      = HASH_TYPE_GENERIC;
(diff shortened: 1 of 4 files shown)


src/modules/module_11500.c
```diff
@@ -0,0 +1,168 @@
...
+  if (parse_rc == false) return (PARSER_SALT_LENGTH);
+
+  return (PARSER_OK);
+}
+
+int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
+{
+  const u32 *digest = (const u32 *) digest_buf;
+
+  const int line_len = snprintf (line_buf, line_size, "%08x:%s", digest[0], (char *) salt->salt_buf);
```
mhasbini: @jsteube `digest[0] ` was originally `byte_swap_32 (digest_buf[0])`, but this produced erroneous hashes. e.g. for input `309273c2:00000000` it gave: `c2739230:`. Thus I removed the swap, although not sure if I'm missing something here.

tools/test_modules/m11500.pm
```diff
@@ -0,0 +1,44 @@
...
+use Digest::CRC qw (crc32);
+
+sub module_constraints { [[0, 255], [8, 8], [0, 31], [8, 8], [-1, -1]] }
+
+sub module_generate_hash
+{
+  my $word = shift;
+  my $salt = shift;
+
+  my $digest = crc32 ($word);
```
mhasbini, line 19: @jsteube I'm not sure if I should concat `$salt` here or not.

jsteube: The perl module does not work as expected, therefore we just set it to fixed zero. You can remove it and use sprintf ("%08x:00000000", $digest);

src/modules/module_11500.c
jsteube: This should do it:

const int line_len = snprintf (line_buf, line_size, "%08x:%08x", digest[0], salt->salt_buf[0]);

mhasbini: Thanks :+1:

PR #1898, opened 2019-02-04
Add -m 00023 module

mhasbini:


jsteube:
Almost OK. Two things:

1. OPTS_TYPE_STATE_BUFFER_LE is missing
2. Writing into existing buffers is unsafe. That's why it's defined as const salt_t *salt.

salt_buf_ptr[salt->salt_len - 8] = 0;

Please create new temporary buffer instead, copy the salt data, then do the modification on there.

src/modules/module_00023.c
```diff
@@ -0,0 +1,205 @@
...
+
+  encoder_apply_optimizer (hashconfig, tmp);
+
+  encoder_apply_options (hashconfig, tmp);
+
+  u8 tmp_buf[128];
+
+  memcpy (tmp_buf, salt->salt_buf, salt->salt_len);
+
+  tmp_buf[salt->salt_len - 8] = 0;
```
mhasbini, line 126: @jsteube Please let me know if this is fine.

mhasbini:
Thanks @jsteube, I pushed a fix to address your notes.

PR #1902, opened 2019-02-09
Add unit test for several modes

mhasbini:


tools/test_modules/m01100.pm
```diff
@@ -11,7 +11,7 @@ use warnings;
 use Digest::MD4 qw (md4 md4_hex);
 use Encode;
 
-sub module_constraints { [[0, 255], [0, 255], [0, 27], [0, 27], [0, 27]] }
+sub module_constraints { [[0, 255], [0, 255], [0, 27], [0, 19], [0, 26]] }
```
mhasbini, line 14: @jsteube I'm trying to make sense of the combination constraint (26), I'm not sure if it's correct.

The digest is: `md4_hex (md4 (encode ("UTF-16LE", $word)) . encode ("UTF-16LE", lc ($salt)));`.

Here's how I calculated the optimized constraints for password & salt:

`md4 (encode ("UTF-16LE", $word))`: limit `$word` to 27 (55 / 2) because of unicode. The result is 16 byte.

For md4 we have 55 bytes limit so `encode ("UTF-16LE", lc ($salt))` must be limited to 55 - 16 = 39. `$salt` in this case will be limited to 19 (39 / 2) (because of unicode).

Though I'm not sure how to calculate the combination constraint.

jsteube: All eventual errors which you get with this mode are because of some error in test.sh. However your math in the constraints is correct, you have the right understanding. Basically it always goes back to the kernel code. If you are unsure, use the single or passthrough to generate some examples and try them manually.

jsteube:
The problem with the -m 1100 is that this mode does not require a comb mode to be set. 

The comb mode is only to be set if the same buffer is kernel for both password and salt.

Fixed here: https://github.com/hashcat/hashcat/commit/b9365b89612320012632766e6ac3e51ca967897d

PR #1918, opened 2019-02-15
Add unit tests for multiple modules:

mhasbini:
15500, 15600, 15700, 16000, 16100, 16200, 16300, 16400, 16500, 16600 & 16700

### DIFF
--- a/tools/test_modules/m15500.pm
+++ b/tools/test_modules/m15500.pm
@@ -1,0 +1,100 @@
+#!/usr/bin/env perl
+
+##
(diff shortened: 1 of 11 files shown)


jsteube:
I'm merging this, just one note: The difference between the || and the // is if the value is "0". For || this is false, for // it is true. I had some problem with this in the past.

jsteube:
And thanks again, this really helps.

mhasbini:
@jsteube TIL, thanks! I'll keep this in mind going further 👍

PR #1920, opened 2019-02-17
Add unit tests for multiple modules:

mhasbini:
10900, 11000, 11100, 11200, 11300 & 11400

### DIFF
--- a/tools/test_modules/m10900.pm
+++ b/tools/test_modules/m10900.pm
@@ -1,0 +1,70 @@
+#!/usr/bin/env perl
+
+##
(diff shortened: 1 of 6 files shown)


jsteube:
Thanks again! Note: https://github.com/hashcat/hashcat/commit/a560b45705f4a2b9ed61d2ef27811cf7dea352fc

If there's no OpenCL/mxxxxx\*optimized\* kernel, then no need for optimized constraints.
Same for OpenCL/mxxxxx\*pure\*, then no need for pure constraints.

mhasbini:
👍 Thanks.

PR #1924, opened 2019-02-20
Fix potfile line splitting for LM

c0d3z3r0:
Lines in a potfile with LM hashes have the format <hash>:<pw>. Splitting for --show is done by finding the last occurrence of the separator : in a line.
This breaks found password with a : in it as they will be marked as [notfound].

Fix this by looking for the FIRST occurrence instead of the LAST one (strchr vs strrchr).

### DIFF
--- a/src/potfile.c
+++ b/src/potfile.c
@@ -535,17 +535,17 @@ int potfile_remove_parse (hashcat_ctx_t *hashcat_ctx)
 
     if (line_len == 0) continue;
 
-    char *last_separator = strrchr (line_buf, hashconfig->separator);
+    char *first_separator = strchr (line_buf, hashconfig->separator);
 
-    if (last_separator == NULL) continue; // ??
+    if (first_separator == NULL) continue; // ??
 
-    char *line_pw_buf = last_separator + 1;
+    char *line_pw_buf = first_separator + 1;
 
     size_t line_pw_len = line_buf + line_len - line_pw_buf;
 
     char *line_hash_buf = line_buf;
 
-    size_t line_hash_len = last_separator - line_buf;
+    size_t line_hash_len = first_separator - line_buf;
 
     line_hash_buf[line_hash_len] = 0;
 


philsmd:
if the plains contain a ":" character, they should use the $HEX[] notation.

I think it's more correct to use the last occurrence, because some hash types have ":" within the hash line because of the salt etc

I think your change breaks some other hash type support, doesn't it ?

Why do you have ":" in the plain after alll ? hashcat should use $HEX[] to output those passwords

c0d3z3r0:
@philsmd I always use hashcat with --outfile-autohex-disable.

> I think it's more correct to use the last occurrence, because some hash types have ":" within the hash line because of the salt etc

Damn. You are right, I just tested this with salted SHA1 (120)...

I will try to find another solution... one idea would be using something like ::: or ;;; or :*: as separator but this would maybe impact performance for big pot files. Another solution could be reverting $HEX for --show. I think the latter one is the correct way. What do you think?

philsmd:
my response to this is that we shouldn't mix up outfiles with potfiles and that potfiles should (and do!) always use the $HEX[] notation, as it is also needed!

I think it's only a problem on your site because you are using outfiles as potfiles, aren't you ?

--outfile-autohex-disable shouldn't change the format of potfiles.

I don't think changing the separator etc is a good idea.
We just should use potfiles (do not use --potfile-disable if you want to use --show/--left and profit from the speed improvement by avoiding duplicate work, i.e. cracking the same hashes and/or trying the same salts again and again) and those potfiles should (continue to) use $HEX[] if needed and the format should stay AS-IS.

I think no changes are needed if you use the potfile correctly, or am I missing something here?
Thx

c0d3z3r0:
> my response to this is that we shouldn't mix up outfiles with potfiles and that potfiles should (and do!) always use the $HEX[] notation, as it is also needed!

ACK. Didn't realized that before...

> I don't think changing the separator etc is a good idea.

ACK

> I think it's only a problem on your site because you are using outfiles as potfiles, aren't you ?

No, I don't. Example: I throw LM hashes on hashcat (using --potfile-path test.pot). Those hashes are splitted into two parts and cracked seperately. Then I use hashcat ... --show. The outfile/output then contains passwords with $HEX[...].

What I'd like to have is an option to get those $HEX stuff decoded in the outfile/output.
For e.g. NTLM hashes I dont' use --show, since the full passwords go to the outfile. Here I use --outfile-autohex-disable.
Unfortunately --outfile-autohex-disable does not work with --show.

philsmd:
there are plenty of ways to convert $HEX[] to non-hex.

e.g.
```
hashcat --show -m 1000 --outfile-format 2 hash.txt | hashcat --stdout
```

I think the problem with -m 3000 (which is very special) is that you could have two parts and hashcat (it seems) currently does not combine 2 $HEX[x],$HEX[y] parts correctly into 1 larger $HEX[x+y].
This could be improved indeed !

jsteube:
Please create a new PR in that case.

PR #1939, opened 2019-03-04
WIP: Add new hash mode: AuthMe - SHA256(SHA256(password) + salt)

### LINKED ISSUE #1486: [New Algorithm] AuthMe - SHA256(SHA256(password) + salt)
**Name of algorithm**: AuthMe default password storage algorithm
**Where it is used**: AuthMe is an authentication plugin for Minecraft cracked servers. It is the most known and used plugin for all kinds of servers. By default, it stores it's passwords in a database in the mentioned format. By implementing the requested algorithm, password recovery/cracking would be available based on the hashes stored in plain text. 
**Source code**: The algorithm is implemented here: [https://github.com/AuthMe/AuthMeReloaded/blob/master/src/main/java/fr/xephi/authme/security/crypts/Sha256.java](url)

**Restrictions**: Salt length is always 16 characters. It is randomly generated but stored together with the password hash. Password length can vary from 1 to 32 characters. 

**Full example of password & hash**: 

> Password:	pantof
Stored entry: $SHA$c7dedf5a36c4a343$05ae3239eee683872ef1cc9096777bf4b1a72a179709efc17d8bf1603b082065

**Other details**: The passwords are stored in the following format: 
`String encryptedPassword = "$SHA$" + salt + "$" + sha256(sha256(password) + salt);`

mhasbini:
Closes #1486


jsteube:
Thanks for adding this mode. It was requested several times before! Some things need to be fixed first.

- This mode differs from the original SHA256 raw hash implementation. Therefore it should get its own hash-mode number. I've reserved hash-mode 19800 for this. Please rewrite.
- I know that there's other applications than AuthMe using this mode. It would be therefore make more sense to have a general kernel and then use the 1980x number for the special applications using this mode.
- The mode is raw and it is also salted at the same time. Therefore the kernel should get the name 19810, the Application AuthMe however should get the name 19811. You can call the 19810 from the 19811 module.
- A very similar structure can be found with the vbull mode (2611).  I recommend to use this mode as template.
- The hash is doing only two iterations of SHA256. IOW, it's too fast to be marked as a slow hash (ATTACK_EXEC_OUTSIDE_KERNEL). It should use ATTACK_EXEC_INSIDE_KERNEL instead. This is required to achieve full GPU utilization. See mode 2611 as an example.
- You can start writing the pure kernels only first. Hashcat accepts either pure kernels only, optimized kernels only, or both. In case you stick to pure kernel only, you can set the module constraints to 0,256 for both password and salt. However AuthMe seems to have a fixed 16 byte salt. In this case you can set the test module to 16,16. I just wanted to explain it generally.
- for parsing the salt, please use the interface generic_salt_decode() and generic_salt_encode() in the module. See 2611 module as an example.
- The DGST_SIZE      = DGST_SIZE_8_8; is invalid for 32 bit based hashes as SHA256. Use DGST_SIZE      = DGST_SIZE_4_8; instead. (The first 8 indicates that this is a 64 bit integer based hash)
- If you configure DGST_POS0-DGST_POS3 correctly (see 1400 as an example) you can allow the JiT to auto optimize the early skip the last 3 steps in the last round of sha256. The marcros are transfered to the kernel automatically. If you stick to them in the kernel you dont need to care about them

jsteube:
So you've started using the same Kernel for "19800". This is fine so far! Does "WIP" mean work in progress?

mhasbini:
@jsteube Thanks for your notes 👍

Yep, I'm trying to get 19800 to pass self test then I'll add a1 & a3 modes. After this I'll add 19811 module.

OpenCL/m19810_a0-pure.cl
```diff
@@ -0,0 +1,284 @@
...
+              | uint_to_hex_lower8_le ((g >> 24) & 255) << 16;
+    ctx.w3[1] = uint_to_hex_lower8_le ((g >>  0) & 255) <<  0
+              | uint_to_hex_lower8_le ((g >>  8) & 255) << 16;
+    ctx.w3[2] = uint_to_hex_lower8_le ((h >> 16) & 255) <<  0
+              | uint_to_hex_lower8_le ((h >> 24) & 255) << 16;
+    ctx.w3[3] = uint_to_hex_lower8_le ((h >>  0) & 255) <<  0
+              | uint_to_hex_lower8_le ((h >>  8) & 255) << 16;
+
+    ctx.len = 64;
+
```
mhasbini, line 272: @jsteube `sha256($pass)` is computed in `ctx0`. What I'm trying to achieve in the above code block is to append the hex values of `ctx0.h` to `s` and compute `sha256` for both of them. Probably there's something wrong in the above block.

jsteube: Please rephrase, I don't understand

olegdenega (approved):

jsteube:
Any updates here?

mhasbini:
@jsteube Sorry didn't have a chance to continue my work on this. Will try to work on it soon.

OpenCL/m19810_a0-pure.cl
purnapattela: Nice job

jsteube:
Note that 19800 has been used now. Please use 199xx instead.

its5Q:
Any updates on your progress, @0xbsec?

jsteube:
This mode was added with PR https://github.com/hashcat/hashcat/pull/2128

Thanks for inspiration!

PR #1941, opened 2019-03-06
Added hash mode 19500

### LINKED ISSUE #70: support for (ruby on rails) restful-authentication
This feature request was originally reported on the (now obsolete and offline) **trac** ticketing system of hashcat.net.

---

| Ticket details: |  |
| --- | --- |
| Original reporter (OP): | joernchen |
| Title: | Support for restful-authentication hashing scheme |
| Ticket number: | 223 |
| Date reported: | 2013-12-12T15:15:22+01:00 |
| **Description**: | Hi,<br><br> restful-authentication [0] is a quite popular Ruby on Rails extension which uses a rather flexible hashing scheme. Basically there is a side-wide and a[n] user-specific salt which both go into a configurable number of SHA1rounds (default 10) together with the password.<br><br>Sample ruby code:<br><br>require 'openssl'<br><br>def secure_digest(*args)<br>Digest::SHA1.hexdigest(args.flatten.join('--'))<br>end<br><br>def password_digest(password, salt, site_salt, rounds)<br>digest = site_salt<br><br>rounds.times do<br>digest = secure_digest(digest, salt, password, site_salt)<br>end<br>digest<br><br>end<br><br>password = ARGV[0]<br>salt = ARGV[1]<br>site_salt = ARGV[2]<br>rounds = ARGV[3]<br>puts password_digest(password, salt, site_salt, rounds.to_i)<br><br><br>Sample output:<br>$ ruby build_hash.rb test test1 test2 10<br>36ca05ee9bc00c59f60096a6d41e336fa27a31d0<br><br>(here test is the password, test1 the salt, test2 the site salt, and 10 the number of rounds)<br><br>I'd very much like to see this algorithm in [ocl]hashcat. Just let me know if this is doable.<br><br>[[0] Restful authentication - github](https://github.com/technoweenie/restful-authentication/blob/master/lib/authentication/by_password.rb#L30) |

---

| Comments: |  |  |
| --- | --- | --- |
| 1. | philsmd | It would be great if you could help me/us to understand some details:<br><br>\- are there some min/max salt/pass length rules<br><br>\- is there an already (de facto) way to recognize / write down all components needed for those hashes, e.g. hash:salt:site_salt:iterations , or any particular signature, also look here: [hashcat example hashes](https://hashcat.net/wiki/doku.php?id=example_hashes). If some other cracker already supports those hashes, we should think about _not_ introducing a different way to do enumerate those hashes (2 different signatures for the same thing is not so good, same holds for the whole line, including position of "site salt" and iterations)<br><br>\- I assume all this info is generally stored in a database where each field (salt,site_salt,hash,iter) has its own column, right? or is site_salt and iter a per-application value -> only salt + hash is in "database" ?<br><br>thx |
| 2. | joernchen | Hi,<br>pass length is 6 to 40 chars, the salt is a SHA-1 hexdigest, so a string of length 40 with the pattern [0-9a-f] (generally no uppercase chars).<br><br>I'm not aware of any cracker for this scheme.<br><br>site_salt and iterations are stored in the applications configuration. hash und salt are stored per user in the database. The default number of iterations is 10.<br><br>Hope that helps & sorry for the delay.<br>joernchen |
| 3. | atom | I'll add this sooner or later |
| 4. | donnet1 | +1<br>I would also like to see this algorithm in hashcat. Any idea when you will implement this atom? |
| 5. | philsmd | We need an example "hash" that contains all informations (but most importantly w/ the correct salt lengths), e.g. of password "hashcat".<br><br>Can you generate an example hash and collect all the info from configuration file and database and post it here in format:<br>hash:salt:site_salt:plain<br><br>As mentioned above the site_salt alone is already 40 chars in length. What about the (user defined) "salt" ? Is the length variable ? Is it randomly generated? |
| 6. | donnet1 | Philsmd,<br>I was searching and stumbled upon this before I created a new trac ticket yesterday, however I just realized that I am looking for [Devise](https://github.com/plataformatec/devise) which is very similar to restful-authentication however much more popular. At the time of writing, it has 10,736 stars on Github and 2,821 forks and wildly more used in software.<br><br>After comparing both, the algorithm they use to hash the password is virtually the same. The only differences are:<br>\- Devise's site_salt is always empty, instead of a SHA1 in restful-authentication.<br>\- Devise's user_salt is generated differently using SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz') which results in 20 character salts, instead of SHA1's in restful-authentication<br>\- Devise passes salt first to the secure_digest function inside the loop, then digest<br><br>Here is the example hash you requested with restful-authentication:<br>require 'digest/sha1'<br>require 'erb'<br><br>def secure_digest(_args)<br>Digest::SHA1.hexdigest(args.flatten.join('--'))<br>end<br><br># stretches is defeault to 10<br>DIGEST_STRECTCHES = 10<br><br># site key (site salt) is generated like so, always a 40 char sha1<br>SITE_KEY = secure_digest(Time.now, (1..10).map{ rand.to_s })<br>puts 'Site Key: ' + SITE_KEY<br><br># user salt is geneated like so, same as site key always a 40 char sha1<br> USER_SALT = secure_digest(Time.now, (1..10).map{ rand.to_s })<br>puts 'User Salt: ' + USER_SALT<br><br>def password_digest(password, salt)<br>digest = SITE_KEY<br>DIGEST_STRECTCHES.times do<br>digest = secure_digest(digest, salt, password, SITE_KEY)<br>end<br><br>digest<br>end<br><br>puts 'Final Hash: ' + password_digest('hashcat', USER_SALT)<br><br><br>hash:salt:site_salt:plain format:<br>9a48e120e736a4f9915da71d4b04634eb68df70f:0ea32eb8e813f797422f7a52305a2ce9c336008f:2097080b915d65d4c78a5d03f6ff095e96a3e615:hashcat<br><br>Here is the same example but with Devise:<br>require 'digest/sha1'<br>require 'erb'<br>require 'securerandom'<br><br>def secure_digest(_args)<br>Digest::SHA1.hexdigest(args.flatten.join('--'))<br>end<br><br># stretches is defeault to 10<br>DIGEST_STRECTCHES = 10<br><br># site key is empty<br>SITE_KEY = ''<br><br># user salt is geneated like so<br>USER_SALT = SecureRandom.urlsafe_base64(15).tr('lIO0', 'sxyz')<br>puts 'User Salt: ' + USER_SALT<br><br>def password_digest(password, salt)<br>digest = SITE_KEY<br><br>DIGEST_STRECTCHES.times do<br>digest = secure_digest(salt, digest, password, SITE_KEY)<br>end<br>digest<br>end<br><br>puts 'Final Hash: ' + password_digest('hashcat', USER_SALT)<br><br><br>hash:salt:plain format:<br>cc4c5cbc2567bb729633c24bbcf9679241d4c234:hxE-93LEABQ9cJ4x-N6k:hashcat<br><br>Should I make a separate trac ticket for Devise, or could you hit two birds with one stone? Let me know if you require anymore information about either.<br><br>Thanks! |
| 7. | philsmd | The main problem w/ restful-authentication is that the "string" we use in sha1 is at least 40 + 2 + 40 + 2 + x + 2 + 40, so at least 126 characters long... This will for sure imply that some optimizations (only a small number of SHA1 transforms etc) are not possible... <br><br>The problem w/ the 'similarities' between restful-authentication and devise is that they seem to be very different indeed.<br><br>While restful-authentication uses the "digest" always at the very beginning, devise has that digest in the middle somewhere and instead at the very beginning is the user salt.<br><br>So it seems that we can't easily use one single GPU kernel for both hash types.<br><br>Thx for the examples |
| 8. | donnet1 | philsmd,<br><br>So with the output of your observations, does that leave us at these algorithms not being implemented into hashcat, or that it just won't be easy to implement? |
| 9. | philsmd | yeah, we need 2 separate trac tickets (this for restful authentication and a new one for devise), then we need to check if:<br>\- restful authentication is still widely used and worth to implement<br>\- is there a minimum/maximum password length also for devise (this needs to be discussed in the new ticket, for this ticket, "restful authentication", it seems to be 6 - 40 chars). |
| 10. | donnet1 | Ok, I've created a new ticket for Devise [~~460 old hashcat trac ticketing system~~] [#77]. I would say that restful authentication is widely used, a couple big open source projects use it.<br><br>As for the minimum/maximum password length for restful authentication, I just checked the code on github repo to confirm and it is 6-40 characters. |

s3inlc:
Closes #70


OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+    const u32x w0 = w0l | w0r;
+
+    w[0] = w0;
+
+    sha1_ctx_t ctx0;
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx0);
+
+    sha1_update_vector (&ctx0, k, site_key_len);
```
jsteube: The use of sha1_update_vector typically requires to use sha1_ctx_vector_t type not sha1_ctx_t type. In this case, it works because you disabled the SIMD by commenting out NEW_SIMD_CODE. I would switch it either to sha1_ctx_vector_t or sha1_update(). Up to you.

OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+  }
+
+  const int salt_len = esalt_bufs[digests_offset].salt_len;
+
+  const int site_key_len = esalt_bufs[digests_offset].site_key_len;
+
+  u32 s[64] = { 0 };
+  u32 k[64] = { 0 };
+
+  const u32 glue[16] = {0x2d2d0000};
```
jsteube: It's possible that this could be misunderstood by the OpenCL runtime and initialize all elements in the array with the value 0x2d2d0000 not just the first one. Please rewrite this as
```
  const u32 glue[16] = { 0x2d2d0000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
```
in all kernels.

OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+  for (int i = 0, idx = 0; i < salt_len; i += 4, idx += 1)
+  {
+    s[idx] = swap32_S (esalt_bufs[salt_pos].salt_buf[idx]);
+  }
+
+  for (int i = 0, idx = 0; i < site_key_len; i += 4, idx += 1)
+  {
+    k[idx] = swap32_S (esalt_bufs[salt_pos].site_key_buf[idx]);
+  }
+
```
jsteube, line 98: You can increase the performance by another 50% in -a 3 mode (probably also -a0 and -a1) if you optimize the strategy of the kernel a bit.

Before entering the inner loop you can precompute the following:

```
  // precompute some stuff

  sha1_ctx_t ctx0;

  sha1_init (&ctx0);

  sha1_update (&ctx0, k, site_key_len);
  sha1_update (&ctx0, glue, 2);
  sha1_update (&ctx0, s, salt_len);
  sha1_update (&ctx0, glue, 2);
```

OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+  for (u32 il_pos = 0; il_pos < il_cnt; il_pos += VECT_SIZE)
+  {
+    const u32x w0r = words_buf_r[il_pos / VECT_SIZE];
+
+    const u32x w0 = w0l | w0r;
+
+    w[0] = w0;
+
+    sha1_ctx_t ctx0;
```
jsteube: To make use of the precomputation you can do use instead:

```
    sha1_ctx_t ctx = ctx0;

    sha1_update (&ctx, w, pw_len);
    sha1_update (&ctx, glue, 2);
    sha1_update (&ctx, k, site_key_len);

    sha1_final (&ctx);
```

OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+      ctx.w1[2] = uint_to_hex_lower8_le ((d >> 16) & 255) <<  0
+                | uint_to_hex_lower8_le ((d >> 24) & 255) << 16;
+      ctx.w1[3] = uint_to_hex_lower8_le ((d >>  0) & 255) <<  0
+                | uint_to_hex_lower8_le ((d >>  8) & 255) << 16;
+      ctx.w2[0] = uint_to_hex_lower8_le ((e >> 16) & 255) <<  0
+                | uint_to_hex_lower8_le ((e >> 24) & 255) << 16;
+      ctx.w2[1] = uint_to_hex_lower8_le ((e >>  0) & 255) <<  0
+                | uint_to_hex_lower8_le ((e >>  8) & 255) << 16;
+
+      ctx.len = 40;
```
jsteube: At this position you can hardcode the glue, it will always be at exactly this point.

```
      ctx.w2[2] = glue[0];

      ctx.len = 40 + 2;
```

OpenCL/m19500_a3-pure.cl
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    const u32 r0 = ctx.h[DGST_R0];
+    const u32 r1 = ctx.h[DGST_R1];
+    const u32 r2 = ctx.h[DGST_R2];
+    const u32 r3 = ctx.h[DGST_R3];
+
+    COMPARE_S_SIMD (r0, r1, r2, r3);
+  }
+}
```
jsteube, line 340: Here's a fully optimized kernel -a3. I'm sure you can backport most of the improvements to -a0 and -a1:

https://pastebin.com/cHFu55ys

jsteube:
Any updates here?

s3inlc:
Sorry for the delay, I updated the kernels based on your comments.

src/modules/module_19500.c
```diff
@@ -0,0 +1,255 @@
...
+  const bool parse_rc2 = generic_salt_decode (hashconfig, token.buf[2], token.len[2], (u8 *) devise_double_salt->site_key_buf, (int *) &devise_double_salt->site_key_len);
+
+  if (parse_rc2 == false) return (PARSER_SALT_LENGTH);
+
+  // make salt sorter happy
+
+  salt->salt_buf[0] = devise_double_salt->salt_buf[0];
+  salt->salt_buf[1] = devise_double_salt->salt_buf[1];
+
+  salt->salt_len = devise_double_salt->salt_len;
```
jsteube: In this case, salt->salt_len should be set to 8, since we only set salt->salt_buf[0] and salt->salt_buf[1]. This is only to make the unique salt sorting in hashcat working correctly to remove duplicate salts.

OpenCL/m19500_a1-pure.cl
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * loop
+   */
+
+  for (u32 il_pos = 0; il_pos < il_cnt; il_pos++)
+  {
+    sha1_ctx_t ctx = ctx0;
+
+    sha1_update_global_swap (&ctx, pws[gid].i, pws[gid].pw_len);
```
jsteube: I think you should be able to move

```
sha1_update_global_swap (&ctx, pws[gid].i, pws[gid].pw_len);
```

to the outer loop (into ctx0). It doesn't change inside the inner loop.

jsteube:
The unit test successfully finished. Please change the code as requested in the two sections. I'll merge afterward.

s3inlc:
Requested changes are applied.

jsteube:
Thanks!

Issue #1951, opened 2019-03-12
Proposal: Changed behaviour of --keep-guessing flag


Chick3nman:
**Problem:**
Currently, the flag `--keep-guessing` allows you to continue an attack after all of the target hashes have been found. With `--keep-guessing` enabled, all new found collisions of the target hashes will be added to the potfile and available after the attack using `--show`. 

With `--keep-guessing` enabled, if all target hashes are already present in the potfile, you will receive the standard information line `INFO: All hashes found in potfile! Use --show to display them.` and the attack will not start. This blocks a user from starting a different attack to generate new collisions for hashes with existing collisions in the potfile. 

To work around this block, a user may enable the `--pofile-disable` flag, which will allow the attack to start, however, with the potfile disabled no new collisions will be added to the potfile and the user is forced to enable `-o outfile` to collect the new collisions.

**Proposed changes/fix:**

The `--keep-guessing` flag should still give the information alert that all or some target hashes are present in the potfile, however, it should then continue with the attack without throwing out the hashes for which collisions exist already. New collisions should still be added to the potfile as they are cracked. A user starting a new attack with `--keep-guessing` enabled shows enough understanding that they are looking for multiple collisions for their target hashes, some of which may be present in the potfile already, and should not be blocked by existing collisions in the potfile.

Chick3nman:
After some review, I have revised my proposed fix to include a warning message in place of the potfile count, skipping the potfile remove step entirely. I've issued a pull request that implements my changes.

jsteube:
Nothing to add here, we can do this.

PR #1952, opened 2019-03-12
Change behaviour of --keep-guessing flag

Chick3nman:
https://github.com/hashcat/hashcat/issues/1951


src/potfile.c
```diff
@@ -373,10 +373,13 @@ int potfile_remove_parse (hashcat_ctx_t *hashcat_ctx)
   const hashes_t       *hashes       = hashcat_ctx->hashes;
   const module_ctx_t   *module_ctx   = hashcat_ctx->module_ctx;
   const potfile_ctx_t  *potfile_ctx  = hashcat_ctx->potfile_ctx;
+  const user_options_t *user_options = hashcat_ctx->user_options;
 
   if (potfile_ctx->enabled == false) return 0;
 
   if (hashconfig->potfile_disable == true) return 0;
+  
+  if (user_options->keep_guessing == true) return 0;
```
jsteube: This would only change the behavior in case the user is using the --keep-guessing command line argument. 

There are however modules which automatically enable the keep-guessing flag automatically, such as 9720 and 18100 and others. Those would not be covered with this test. Please replace this test with the following and we should be good:

```
if (hashconfig->opts_type & OPTS_TYPE_PT_NEVERCRACK) return 0;
```

The reason why this will also cover the use of --keep-guessing on the command line is best explained here:

https://github.com/hashcat/hashcat/blob/master/src/interface.c#L347-L350

Issue #1645, opened 2018-08-02
[Feature Request] DiskCryptor supprt (harddrive encryption and ISO)


KotekCzek:
Make support DiskCryptor please :)
The program is using:
HMAC-SHA-512
Mode XTS

This program has a very good feature - you can encrypt ISO (Aes-Twofish-Serpent) and burn it into CD, and then u can mount it in program. It is very similar to Truecrypt

Thank YOU

KotekCzek:
**Refresh :) 
DiskCryptor is really good encrypting program.**

 I am asking you to look at the program. It supports popular encryption algorithms.

![d](https://user-images.githubusercontent.com/42041983/54862956-85d4b200-4d42-11e9-9ca0-16964f7262e9.jpg)
![d2](https://user-images.githubusercontent.com/42041983/54862991-faa7ec00-4d42-11e9-9c81-104045bb4d8a.jpg)

**I wish you all good :)**

jsteube:
added

PR #1981, opened 2019-04-07
added -m 19800 = DiskCryptor AES

philsmd:
this implements the first encryption algorithm (AES) for DiskCryptor (see #1645 and https://hashcat.net/forum/thread-8012.html, https://hashcat.net/forum/thread-8243.html).

There are multiple (encryption) algorithms supported by DiskCryptor, this patch only addresses the pure AES encryption.

The algorithm is PBKDF2-HMAC-SHA512 with 1000 iterations and AES-XTS (with a starting sector ID of 1).

### DIFF
--- a/OpenCL/inc_diskcryptor_xts.cl
+++ b/OpenCL/inc_diskcryptor_xts.cl
@@ -1,0 +1,70 @@
+DECLSPEC void xts_mul2 (u32 *in, u32 *out)
+{
+  const u32 c = in[3] >> 31;
+
+  out[3] = (in[3] << 1) | (in[2] >> 31);
+  out[2] = (in[2] << 1) | (in[1] >> 31);
+  out[1] = (in[1] << 1) | (in[0] >> 31);
+  out[0] = (in[0] << 1);
+
+  out[0] ^= c * 0x87;
+}
+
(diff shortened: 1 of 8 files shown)


OpenCL/inc_diskcryptor_xts.cl
```diff
@@ -0,0 +1,70 @@
...
+  aes256_set_decrypt_key (ks, ukey1, s_te0, s_te1, s_te2, s_te3, s_td0, s_td1, s_td2, s_td3);
+  aes256_decrypt (ks, out, out, s_td0, s_td1, s_td2, s_td3, s_td4);
+
+  out[0] ^= T[0];
+  out[1] ^= T[1];
+  out[2] ^= T[2];
+  out[3] ^= T[3];
+}
+
+DECLSPEC int decrypt_and_check (__global const u32 *encrypted_data, const u32 *ukey1, const u32 *ukey2, SHM_TYPE u32 *s_te0, SHM_TYPE u32 *s_te1, SHM_TYPE u32 *s_te2, SHM_TYPE u32 *s_te3, SHM_TYPE u32 *s_te4, SHM_TYPE u32 *s_td0, SHM_TYPE u32 *s_td1, SHM_TYPE u32 *s_td2, SHM_TYPE u32 *s_td3, SHM_TYPE u32 *s_td4)
```
jsteube: This is using the old __global syntax which would fail in emulator mode. Please use GLOBAL_AS instead.

src/modules/module_19800.c
```diff
@@ -0,0 +1,254 @@
...
+
+  // first only add the signature and version number:
+
+  int line_len = snprintf (line_buf, line_size, "%s%i*", SIGNATURE_DISKCRYPTOR, DISKCRYPTOR_VERSION);
+
+  // ... then add the full header (in hexadecimal):
+
+  for (u32 i = 0; i < 512; i++)
+  {
+    line_len += snprintf (line_buf + line_len, line_size, "%08x", byte_swap_32 (diskcryptor_esalt->salt_buf[i]));
```
jsteube: should be line_size - line_len

OpenCL/inc_diskcryptor_xts.cl
philsmd: yeah, good catch. fixed

src/modules/module_19800.c
philsmd: yeah, good catch. fixed

PR #1995, opened 2019-04-13
Fix for library compilation failure due to multiple def of sbob_xx64

f0cker:
Renamed these to sbob256_xx64 and sbob512_xx64.


jsteube:
Thanks!

PR #1996, opened 2019-04-13
tests: fix hash generation of -m 8300 = DNSSEC

philsmd:
The salt needs to be passed AS-IS (in hexadecimal) not converted to binary for the perl module Net::DNS::RR::NSEC3::name2hash.

Thank you very much


jsteube:
Nice find!

PR #2005, opened 2019-04-27
bcrypt changes

jsteube:


PenguinKeeper7:
Apologies for resurrecting this but is there a reason for disabling JiT caching on only bcrypt? In which situations would it not be safe to use the cached kernel?

PR #2008, opened 2019-05-01
New modules for Python passlib pbkdf2 sha variants

vlohacks:
Added new module for pbkdf2-sha512 hashes created by Python passlib. The
module is based on the grub2 module (using the same kernel)


jsteube:
Thanks for the contribution!

I have just three comments:

- The hash-mode should be a separate one, please use 20200
- Please provide a unit test in tools/test_modules should be easy to modify original mode 7100
- Is there any well know application using this and can you give a sample extracts of its database?

philsmd:
another very related question would be what should we do with the remaining algos like pbkdf2_sha1
and pbkdf2_sha256 ? are you planning to implement these too ?

see https://passlib.readthedocs.io/en/stable/lib/passlib.hash.pbkdf2_digest.html

I'm okay if we "only" start with pbkdf2_sha512 , but my guess is that if we support one of these passlib formats, we actually should support all (3 ?) of them in the future. Do you agree?

vlohacks:
I stumbled upon the pbkdf2-sha512 hashes in the open source ERP software "odoo" version 8 (https://www.odoo.com/). My samples are production data from a penetration test, therefore I'm afraid that I can't share these. However it should not be too hard to get the thing up and running to generate some example data.

Regarding the other hashes, I'll add them too, give me a few days.

vlohacks:
Hi, I just added the pbkdf2-sha256 and pbkdf2-sha1, hope the naming is correct, I am still not completely familiar with the naming convention here. If not, let me know and I'll rename them.

philsmd:
wow, thank you very, very much for your effort and hard work :)

What is a little bit strange to me, is that the pbkdf2-sha1 hashes only have $pbkdf2$ as a signature... but I tried to generate some hashes myself with python and see that you are kind of right that pbkdf2-hmac-sha1 is somehow different when it comes to passlib.
```
from passlib.hash import pbkdf2_sha1

hash = pbkdf2_sha1.hash ("hashcat")

print hash
```

What makes this even weirder, is that the documentation **emphasizes** that all variants use the **identical** format :

> All of the pbkdf2 hashes defined by passlib follow the same format, $pbkdf2-{digest}${rounds}${salt}${checksum}.

taken from https://passlib.readthedocs.io/en/stable/lib/passlib.hash.pbkdf2_digest.html#passlib.hash.pbkdf2_sha1

I also tried running those $pbkdf2$ through jtr and they are not automatically detected as pbkdf2-hmac-sha1 (@magnumripper), maybe I'm doing something wrong here @ jtr, dunno.

Anyway, back to the actual topic... it would be great if you (@vlohacks ) could modify also the files docs/changes.txt and docs/readme.txt to "announce"/list these new supported formats.

Maybe in the meantime I see more changes that are needed, but it looks already very good. Good job. Thanks

src/modules/module_20200.c
```diff
@@ -0,0 +1,331 @@
...
+  }
+
+  
+  const int salt_len_decoded = base64_decode (base64_to_int, (const u8 *) tmp_buf2, salt_len, tmp_buf);
+
+  u8 *salt_buf_ptr = (u8 *) pbkdf2_sha512->salt_buf;
+  memcpy (salt_buf_ptr, tmp_buf, salt_len_decoded);
+  memcpy (salt->salt_buf, salt_buf_ptr, salt_len_decoded);
+  
+  salt->salt_len    = salt_len_decoded;
```
philsmd: strange identation (spaces). I would suggest to just use salt->salt_len = salt_len_decoded; with one space before and one space after the equal sign

src/modules/module_20200.c
```diff
@@ -0,0 +1,331 @@
...
+  tmp[8] = 0;
+
+  char salt_enc[128] = { 0 };
+  char hash_enc[128] = { 0 };
+
+  int salt_len_enc = base64_encode (int_to_base64, (const u8 *) pbkdf2_sha512->salt_buf, salt->salt_len, (u8 *) salt_enc);
+  int hash_len_enc = base64_encode (int_to_base64, (const u8 *) tmp, HASH_LEN_RAW, (u8 *) hash_enc);
+  
+  
+  // substitute + with . and remove padding =
```
philsmd: this is a little bit strange. doesn't any base64 table in src/convert.c fit ?
maybe we should add it there if not present. Is this a RFC/standard base64 table?

src/modules/module_20400.c
```diff
@@ -0,0 +1,324 @@
...
+  // substitute + with . and remove padding =
+  for (int i = 0; i < salt_len_enc; i++) 
+  {
+    if (salt_enc[i] == '+')
+    { 
+      salt_enc[i] = '.';
+    }
+    if (salt_enc[i] == '=') 
+    {
+	  salt_enc[i] = '\0';
```
philsmd: the spaces at the beginning (indentation) are also not very consistent. but I guess this code can be removed after we add/fix the base64_encode () call with the correct base64 chars/table

src/modules/module_20200.c
philsmd: update: it's probably this one: https://passlib.readthedocs.io/en/stable/lib/passlib.utils.binary.html#passlib.utils.binary.ab64_encode , but I think it's worth adding it to src/convert.c if we use it already in 3 different modules.
Do you agree?

jsteube:
I agree this looks very good. Can you please add the missing convert table?

vlohacks:
OK, will remove code which replaces + with . and create new tables in convert.c. I already thought about this and wondered why there is no table when there is a parser check for this charset (TOKEN_ATTR_VERIFY_BASE64B). Other modules also use this token verification but use itoa32_to_int and int_to_itoa32 as encoding/decoding tables, which however seems to be yet another format.

The indentation mess is misconfiguration of my editor (tabs vs. spaces), I'll will fix this.

What about the code removing the padding '=' from base64 which mangles the base64 output in a similar (strange) way? Should I also handle this differently? I have 3 suggestions here:
- Option 1: Adding a parameter to base64_encode which makes it possible to turn off the padding. This however will change the function signature and will break lots of other modules.
- Option 2: Implementing a base64_encode_nopadding which leaves padding away
- Option 3: Just leaving the padding removal code in the modules.

philsmd:
wow. again a great update! Thank you soooo much

two more small questions/concerns:
1. @jsteube / @vlohacks  should we move this https://github.com/hashcat/hashcat/pull/2008/files#diff-9af4335b0672b095f1cf804f8eb44de0R273 up a little bit to related algos maybe (not adding it at the end) ? I think we have used some kind of grouping/sorting by related algos etc
2. @vlohacks it should be actually possible to use 0x00 instead of 0x3d over here: https://github.com/hashcat/hashcat/pull/2008/files#diff-a8324b5b2e4774a3a8b59742ed8fd635R579, to make sure that we do not use "=" as padding, or is this a stupid/wrong suggestion/assumption ? It feels strange to first add the "=" and then remove it... in my opinion it would make more sense to use 0x00 instead of 0x3d from the beginning (encoding) . Could you @vlohacks  try or at least mention here why that's not doable ?

**update**: I think the problem is that we have hard-coded the '=' here: https://github.com/hashcat/hashcat/blob/b6cc3c7d55df3f5bce9780c0f814809860361d9d/src/convert.c#L824 , maybe this should also use the base64 table instead of the hard-coded char ? I didn't test it, but if this is the case it's of course not a problem of @vlohacks code, but a problem with already existing code which is not as flexible as it should be . any ideas or concerns changing this ?
**update2**: I think I was looking at the wrong table (this is the correct one: https://github.com/hashcat/hashcat/pull/2008/files#diff-a8324b5b2e4774a3a8b59742ed8fd635R561 , I was just blindly looking for 0x3d, hehe) ... it seems we currently only store 64 chars **without** the padding character. That's of course unfortunate, but I think it's okay to make an exception (by replacing = with 0x00) for this new base64 encoding as @vlohacks already did in his/her code... because alternative padding is probably very rare, I guess (this is the first case we found/have). So I'm also fine if we leave the code as @vlohacks suggested

vlohacks:
- regarding 1: Allright, I suggest right below the "plain" PBKDF2-HMAC-.... Algorithms?
- regarding 2: The mentioned table is the decoding table which means that the 0x3d stands for a 'real' value instead of the '=' char. The base64_encode function does not use chars from the substitution tables, instead it pads a hardcoded '=' char (see convert.c, base64_encode (lines 861-866)). It would however be another option to place the padding char outside the regular range (for base64 at position 64) in the encoding table and use this instead of a hardcoded '='.

philsmd:
yeah, I agree. I noticed this too later on

another question: did you @vlohacks  try with much longer salt size (**fixed typo**) and digest size (output of python's pbkdf2) ? maybe a 100 bytes is already very small: https://github.com/hashcat/hashcat/pull/2008/files#diff-cc6414cfbcec7311b3b93dc3255a3eafR154 for SALT_MAX (256 ?)
I think the math is 256 / 4 * 3 = 192 bytes max ? or is my math wrong ?

vlohacks:
Aww, this was really bad. I changed it to 256 since the target buffer also is a hardcoded u32[64] array which is also 256 bytes large. Just in case if someone else uses my code, he/she won't run into the same trap ;-)

philsmd:
I think this pull request is (almost) ready to be merged. very, very good. thx a lot.

I'm currently wondering if we should really call this base64 table alternate_base64 (alternate_base64_to_int () and int_to_alternate_base64 () for instance), maybe something like passlib64_* or passlib_ab64_* would be more correct or at least easier to understand/find in the future.

I'm also not quite sure if the output format of the hashes is okay, since I think we are truncating the hashes to the minimum needed (ideally input hash == output hash).

... but again these are just my last few observations/doubts, maybe @jsteube can jump in and confirm/say if we need to still change it a little bit or if it's already perfect (it for sure is almost perfect in my eyes, thx again).

docs/readme.txt
```diff
@@ -270,6 +270,9 @@ NVIDIA GPUs require "NVIDIA Driver" (418.56 or later)
 - Ethereum Pre-Sale Wallet, PBKDF2-HMAC-SHA256
 - Ansible Vault
 - Plaintext
+- Python passlib pbkdf2-sha1
```
jsteube: Please move this section to the other generic hash modes section (scrypt, pbkdf2)

src/modules/module_20200.c
```diff
@@ -0,0 +1,300 @@
...
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_8_16;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_FORUM_SOFTWARE;
```
jsteube: I think generic KDF is more appropriate

src/convert.c
```diff
@@ -555,6 +555,45 @@ u8 base64_to_int (const u8 c)
   return tbl[c];
 }
 
+// alternate base64 using ./ instead of +/, used in python passlib hashes
+u8 int_to_alternate_base64 (const u8 c)
```
jsteube: int_to_ab64 looks more streamlined than int_to_alternate_base64 in comparison with the other existing buffers. Or something involving substring 'passlib' ?

jsteube:
Great addition and always fast response. Thanks!

PR #2016, opened 2019-05-08
Fix default option name of '--opencl-vector-width'

leonklingele:


jsteube:
That was done intentionally. I wanted to point out that using just --opencl-vector=4 will work, too. It's a feature of getopt().

PR #2020, opened 2019-05-12
Fix bcrypt with AMD on MacOS

marksteward:
The MacOS AMD compiler <s>only accepts up to 6 u32 words when initialising private variables inside a function</s> appears to be buggy, so move the constants out to constant space.

This is my first PR and I haven't done any performance testing - please let me know if there's something I've missed out.

### DIFF
--- a/OpenCL/m03200-pure.cl
+++ b/OpenCL/m03200-pure.cl
@@ -298,6 +298,14 @@ CONSTANT_AS u32a c_sbox3[256] =
   0x90d4f869, 0xa65cdea0, 0x3f09252d, 0xc208e69f,
   0xb74e6132, 0xce77e25b, 0x578fdfe3, 0x3ac372e6
 };
+CONSTANT_AS u32a c_parray[18] =
+{
+    0x243f6a88, 0x85a308d3, 0x13198a2e, 0x03707344,
+    0xa4093822, 0x299f31d0, 0x082efa98, 0xec4e6c89,
+    0x452821e6, 0x38d01377, 0xbe5466cf, 0x34e90c6c,
+    0xc0ac29b7, 0xc97c50dd, 0x3f84d5b5, 0xb5470917,
+    0x9216d5d9, 0x8979fb1b
+};
 
 #define BF_ROUND(L,R,N)                       \
 {                                             \
@@ -455,14 +463,11 @@ KERNEL_FQ void __attribute__((reqd_work_group_size(FIXED_LOCAL_SIZE, 1, 1))) m03
 
   // initstate
 
-  u32 P[18] =
+  u32 P[18];
+  for (u32 i = 0; i < 18; i++)
   {
-    0x243f6a88, 0x85a308d3, 0x13198a2e, 0x03707344,
-    0xa4093822, 0x299f31d0, 0x082efa98, 0xec4e6c89,
-    0x452821e6, 0x38d01377, 0xbe5466cf, 0x34e90c6c,
-    0xc0ac29b7, 0xc97c50dd, 0x3f84d5b5, 0xb5470917,
-    0x9216d5d9, 0x8979fb1b
-  };
+    P[i] = c_parray[i];
+  }
 
   for (u32 i = 0; i < 256; i++)
   {


marksteward:
(I didn't apply the change to modes 9000 and 18600 as they don't fail their self tests. I can look into why if that's a problem.)

philsmd:
interesting, thank you for troubleshooting this.

Is there some related github issue for this PR ? We have some github issues that have to do with macOS but I'm not totally sure if you are referring to one of these here. Please let us know
Thanks

marksteward:
This was just something that stopped me from using mode 3200.

It actually looks like there's something more complex going on - I confirmed the 6 * u32 limit on this kernel, but when I copy/paste the code to a different kernel I don't see the same limit.

I'll do some more playing tonight.

jsteube:
Can you please also apply this on -m 9000 and -m 18600

marksteward:
OK, done some playing, and it seems the issue isn't actually related the the size of the initialiser, but some sort of memory corruption when combining initialised arrays in a loop:

```
  u32 E[4] = { 0 };
  u32 P[4] =
  { 
    0x243f6a88, 0x85a308d3, 0x13198a2e, 0x03707344,
  };

  if (true && atomic_inc (&hashes_shown[digests_offset]) == 0) {
    for (u32 i = 0; i < sizeof(P) / sizeof(P[0]); i++) {
      printf("0x%08lx ", P[i]);
    }
    printf("\n");
  }

  for (u32 i = 2; i < 4; i++)  // if the loop only runs once, no corruption occurs
  { 
    P[i] ^= E[i];
    // P[0] ^= E[i];  // same effect
    // P[i] |= E[i];  // same effect
  }
```

This results in memory corruption, usually all zeroes, even when copied to another kernel:

```
Hashmode: 30 - md5(utf16le($pass).$salt)

0x00000000 0x00000000 0x00000000 0x00000000 
```

Dumping here as I'm out of time for this evening, but I'll try to come back to it over the weekend.

philsmd:
you could try to use some "volatile" variables etc. but this definitely seems to be a very specific compiler bug and maybe just a very specific version of that driver is affected etc.

We shouldn't merge this pull request unless we are sure that the root of the problem was clearly identified and the problem is really fixed with the patch.

jsteube:
> The MacOS AMD compiler only accepts up to 6 u32 words when initialising private variables inside a function, so move the constants out to constant space.

Is this some official statement? Where can I find it? It might have an affect on other kernels, too.

Note that you can apply the changes on hash-mode -m 9000 and -m 18600, too. I checked this on my local system and it did not have any negative side-effect (as theoretically expected).

marksteward:
> Is this some official statement? Where can I find it? It might have an affect on other kernels, too.

Just corrected it. While the tests I did within the kernel seemed to show this, further testing suggests it's some sort of memory corruption/compiler bug.

I'm really struggling to get intermediate output from the MacOS compiler (and it's apparently deprecated by Apple) so maybe we should just close this PR and note that MacOS 10.14 is problematic?

jsteube:
> I'm really struggling to get intermediate output from the MacOS compiler (and it's apparently deprecated by Apple) so maybe we should just close this PR and note that MacOS 10.14 is problematic?

OK, we can do that. But does it mean that you original fix did not work around the problem successfully? Because in this case, we should apply it (as it has no negative effects on other systems, so it's only beneficial). If it does not work around the problem we should not apply it. That's how the decision should be done.

jsteube:
No response from OP so far. I've created a new commit with the changes from this PR: https://github.com/hashcat/hashcat/commit/f689532e4c02370fa5213cfaa6443e17d4a8af56

marksteward:
Ah yeah, sorry about that. It does work around the problem, but given it doesn't work for the right reasons I wanted to confirm first. Unfortunately I've been pretty busy, and will be for the next week too.

What's different in your commit?

jsteube:
No worries, I just wanted to remove it from my todo list. We had this code already in -m 9000 and -m 18600 so I just moved it from there. I just prefer the code to be synchronous across all kernel.

PR #2024, opened 2019-05-15
update the tab completion script

philsmd:
this commits adds some missing flags to the extra/tab_completion/hashcat.sh script (-D and -T were missing amongs others) and updates the script to use the output of hashcat -I (... and includes a new blacklist filter for aliases).

Thank you very much


jsteube:
Many thanks!

PR #2029, opened 2019-05-17
fixes #2026: rule engine hex interpretation problem

### LINKED ISSUE #2026: hex notation broken in insert and overwrite rule
When two or more rules use hex notation hashcat prints `Skipping invalid or unsupported rule in file...`
```
$ cat test.rule 
i0A i1B i2C
i0\x41 i1\x42 i2C
i0\x41 i1\x42 i2\x43
```
```
$ echo test | ./hashcat -r test.rule --stdout
Skipping invalid or unsupported rule in file test.rule on line 2: i0\x41 i1\x42 i2C
Skipping invalid or unsupported rule in file test.rule on line 3: i0\x41 i1\x42 i2\x43
ABCtest
```
The position of these two hex notation rules doesn't matter. For example `0A i1\x42 i2\x43 i3D` throws the same error.
However `i0A i1\x42 i2C i3\x44` seems to work fine.
It appears to only affect the insert and overwrite rule, however I haven't tested it with every rule operation.

Tested with a5301f0 and hashcat 5.1.0

philsmd:
Patch for the rule engine to support the \x00 - \xff notation in some rule functions

### DIFF
--- a/src/rp_cpu.c
+++ b/src/rp_cpu.c
@@ -629,14 +629,14 @@ int _old_apply_rule (const char *rule, int rule_len, char in[RP_PASSWORD_SIZE], 
 
       case RULE_OP_MANGLE_INSERT:
         NEXT_RULEPOS (rule_pos);
-        NEXT_RPTOI (rule, rule_pos, upos);
+        NEXT_RPTOI (rule_new, rule_pos, upos);
         NEXT_RULEPOS (rule_pos);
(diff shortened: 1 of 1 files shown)


philsmd:
after discussing this with @jsteube, we agreed that we should get rid of rule vs rule_new problems in the main loop (they might not be in sync with rule_pos/lengths etc). will probably provide a better fix soon

PR #2048, opened 2019-05-31
Rek

TheREK3R:
Final tests


philsmd:
hey. could you please explain a little bit what you did there with this code and why it is important that we add this to hashcat ?
I see there are a lot of cppunit tests. that's interesting, but why did you decide to implement them ?

we already have some black-box testing tools like test.sh/test.pl, I know it's not 100% the same, but I also do not really get what's going on here with this commit. Could you please give a little bit of information ? thanks

jsteube:
I believe this PR was created unintentionally

TheREK3R:
It was unintentional, I'm sorry, it's for a software engineering project. 

-Ryan

PR #2075, opened 2019-06-22
Add zlib support for loading hashlist/wordlist (v2)

matrix:
Hi,
same as 6 month ago, this patch add the support to load gzip hashes/wordlist.
Following an example using the gzipped version of example0.hash and rockyou ;)

(^._.^)ﾉ

```
$ ./hashcat -t 32 -a 7 example0.hash.gz ?a?a?a?a ~/Downloads/rockyou.txt.gz
hashcat (v5.1.0-1185-g0103c85b) starting...

OpenCL API (OpenCL 1.2 (Jun 15 2018 20:52:09)) - Platform #1 [Apple]
* Device #2: Iris, 384/1536 MB allocatable, 40MCU

Dictionary cache hit:
* Filename..: [...]/Downloads/rockyou.txt.gz
* Passwords.: 14344387
* Bytes.....: 53291009
* Keyspace..: 14344387

Hashes: 6494 digests; 6494 unique digests, 1 unique salts

Applicable optimizers:
* Zero-Byte
* Early-Skip
* Not-Salted
* Not-Iterated
* Single-Salt
* Raw-Hash

Using pure kernels enables cracking longer passwords but for the price of drastically reduced performance.
If you want to switch to optimized backend kernels, append -O to your commandline.

Watchdog: Hardware monitoring interface not found on your system.
Watchdog: Temperature abort trigger disabled.

Host memory required for this attack: 409 MB

Dictionary cache hit:
* Filename..: [...]/Downloads/rockyou.txt.gz
* Passwords.: 14344387
* Bytes.....: 53291009
* Keyspace..: 15041179942912

23d241ee5008678208fd1a750ad23d4d:williloveyou
5221ffa74f1c46b55c8e796a8ad9fba3:sexy123456789
68789d81d4ab0e2858e9d87ac4e0d10e:smll123456789
b2637d7c37876235be5896d5c96d8f32:saha123456789
8205566bf37a6918d682d0ed332ed1f8:giff123456789
3be589bf10c1c6b3a87a59209e5a8d90:sawa123456789
1551d724333954977f67f319da74c62e:djoniloveyou
5b736a8e6d77c4190a8920d6637d54a7:yura123456789
039618d9bd96b3b52c1dab77053d8d12:pide1234567
38508b4ecb421af484c77e84e95f3eee:anyailoveyou
af3babbfd0e8e5abe36ff4dcc2f67eb3:db87password
cf58bcbcb86a9bdb796612fec8d8cff9:2205123456789
eeebbbb0d214666749385f33746b5ea0:gamepassword
a345d7b5f44216b452e266be5e381652:evil123456
3d72c67ae26ba903ab7a7bae4905363b:piel123456
e151194d1b7a64b21daefbb0b5d9c90e:gava123456789
7cddcb3e4cc4e0b392697200b0a3e5e9:lopa123456789
6bedf956ebbde11c2ed1205dc662b1a9:megapassword
ccda5674a0ad1937a46a6cae69c674bb:chan12345678
33355d482a773b8e67fe4d25dc9bb212:missbabygirl
[...]
```


matrix:
Probably the following patch fix cygwin build error (it's a known issue in zlib):

[https://github.com/madler/zlib/pull/248](url)

jsteube:
Thanks for your PR. This is a deep change to the hashcat core and we need to be very careful. Therefore I have some comments:

1. I can see some use for this feature for wordlists. But wordlists only. Having a compressed hashlist makes their handling in other command line tools unpractical. We should avoid supporting this. The same would be the case for wordlists, but wordlists stay more or less the same over long period of time. Therefore the first change would be to remove all special handling for any files which are not a wordlist.

2. Once (1) has been applied, we will see that hc_fread() will stay the standard and not being the exception any longer. Therefore we should not rename hc_fread() to hc_fread_direct(). Instead it would be better to have a hc_fread_compressed() for reading compressed since this function will be the real exception. 

3. There is 'is_gzip' variable in struct fp_tmp (please change from 'short' to 'bool' type). Why not having inside hc_fread() a branch, checking 'is_gzip' for true and one for false and then choose matching code path. This will require to change all FILE* based functions but that's ok for me as it's a simple general rule like 'Do not use FILE*, use fp_tmp_t'. You should now realize that after doing this, it will make (2) obsolete. Therefore (2) is actually not needed. I just listed it to make it possible to follow my thought.

4. The integration should be as seamless as possible to not confuse other developer when they see the use of fp_tmp_t instead of FILE for this the first time. For instance, the name 'fp_tmp_t' radically changes the name convention type. Therefore please rename it to 'HCFILE' instead, even if we typically use xxx_t to mark it as typedef. In this case we do not. In (3) a was referring to 'fp_tmp_t' but actually I meant 'HCFILE' instead. Again this was done only to follow my thoughts.

5. On several places you changed this type of coding from:

```
      FILE *fp = fopen (dictfile, "rb");

      if (fp == NULL)
```

To this type:

```
      fp_tmp_t fp_t;

      if (hc_fopen (&fp_t, dictfile, "rb") == false)
```

It might not look that cool, but I prefer it a bit more conservative. It allows to read the source simply line by line.

```
      fp_tmp_t fp_t;

      const bool rc_whatever = hc_fopen (&fp_t, dictfile, "rb");

      if (rc_whatever == false)
```

6. The zlib patch you linked need to get merged into zlib master. We can not change the original sources from external libraries just to make it work in ours. This is serious as it could kill your entire PR to not be merged. BUT: Maybe you find a way to workaround the problem without the need for this zlib patch to be merged. For instance, by adding/removing/deleting other macros from outside (outside means the hashcat source files which include the zlib). If you can not find any solution 'from outside', you can still disable the compression feature for cygwin in general.

7. If not already checked, please make sure your PR runs with mingw, too.

Please do the changes as needed. I'll then review it again.

matrix:
Morning Jens,
thanks for the initial review.
I agree on almost everything and I answer you point by point.

(1) and (2) ok, I will do

(3) initially I thought the same thing but it's not practical, I think the best way is the one I have already undertaken. Check the source code of the new "hc_fopen" function to better understand. The check must be done immediately after the "open" call for several reasons.

(4) ok, I will do

(5) If I understand well do you prefer have a "const bool rc*" for each if statement ? Bro it's not only not cool. I think what you say it's acceptable if you do it rarely.
An example of "bad use" of this convention it's in the "backend.c" code.

(I've already rewritten it but not yet pushed. I think in the end, even if I removed almost all the "rc" you'll like it because it has also become much more readable than before.
After I do all the mods to this patch I can send to you the new backend.c version, so you can full compare)

Example 1:

```
int CL_rc;

CL_rc = hc_clSetKernelArg (hashcat_ctx, kernel, 0, sizeof (cl_mem),   (void *) &buf);                         if (CL_rc == -1) return -1;
CL_rc = hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof (cl_uint),  device_param->kernel_params_memset[1]); if (CL_rc == -1) return -1;
CL_rc = hc_clSetKernelArg (hashcat_ctx, kernel, 2, sizeof (cl_ulong), device_param->kernel_params_memset[2]); if (CL_rc == -1) return -1;
 
```

I think it's better the following version:

```
if (hc_clSetKernelArg (hashcat_ctx, kernel, 0, sizeof (cl_mem),   (void *) &buf) == -1) return -1;
if (hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof (cl_uint),  device_param->kernel_params_memset[1]) == -1) return -1;
if (hc_clSetKernelArg (hashcat_ctx, kernel, 2, sizeof (cl_ulong), device_param->kernel_params_memset[2]) == -1) return -1;
```

Example 2:

```
int CL_rc;

CL_rc = hc_clSetKernelArg (hashcat_ctx, kernel, 0, sizeof (cl_mem), (void *) &buf);

if (CL_rc == -1) return -1;

CL_rc = hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof (cl_ulong), device_param->kernel_params_atinit[1]);

if (CL_rc == -1) return -1;

CL_rc = hc_clEnqueueNDRangeKernel (hashcat_ctx, device_param->opencl_command_queue, kernel, 1, NULL, global_work_size, local_work_size, 0, NULL, NULL);

if (CL_rc == -1) return -1;

CL_rc = hc_clFlush (hashcat_ctx, device_param->opencl_command_queue);

if (CL_rc == -1) return -1;

CL_rc = hc_clFinish (hashcat_ctx, device_param->opencl_command_queue);

if (CL_rc == -1) return -1;
```

Again I think it's better the following version:

```
if (hc_clSetKernelArg (hashcat_ctx, kernel, 0, sizeof (cl_mem),   (void *) &buf) == -1) return -1;
if (hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof (cl_ulong), device_param->kernel_params_atinit[1]) == -1) return -1;

if (hc_clEnqueueNDRangeKernel (hashcat_ctx, device_param->opencl_command_queue, kernel, 1, NULL, global_work_size, local_work_size, 0, NULL, NULL) == -1) return -1;

if (hc_clFlush  (hashcat_ctx, device_param->opencl_command_queue) == -1) return -1;
if (hc_clFinish (hashcat_ctx, device_param->opencl_command_queue) == -1) return -1;
```

(6) I known but the author of zlib do not push from 2017 ... I can try to involve it or I can fork zlib repository and apply this patch to understand if the patch I found in the pull request actually solves the problem on cygwin. I think there is no workaround due to the nature of the problem but as always, you are the code optimizer on windows, anything could happen :D

(7) mingw works, check the appveyor build logs of this PR ;)

Thanks again bro, I hope I was clear.
Let me know what you think of my comments before I start editing everything.

jsteube:
> (3) initially I thought the same thing but it's not practical, I think the best way is the one I have already undertaken. Check the source code of the new "hc_fopen" function to better understand. The check must be done immediately after the "open" call for several reasons.

Maybe you've understood different. The hc_fopen() is fine. But you need to add special codepath for both results of 'is_gzip' to all other functions that make use of the HCFILE struct to make it transparent. Then you can full replace all calls to fopen(), fclose(), fread() etc with own wrapper function, no matter if the file is compressed or not.

> (5) If I understand well do you prefer have a "const bool rc*" for each if statement ? Bro it's not only not cool. I think what you say it's acceptable if you do it rarely.

OK, but if you do, do it everywhere in the code (not just backend.c). Having code style in different convention across all code creates chaos. Please do this change in a separate PR. For this PR, stick to my convention.

> (6) I known but the author of zlib do not push from 2017...

Be more tricky handling macros, maybe you find a solution. Like hacking the source, but without hacking the source. If that's not an option then just deactivate the compression feature in cygwin. However we should stick to original source.

matrix:
> Maybe you've understood different. The hc_fopen() is fine. But you need to add special codepath for both results of 'is_gzip' to all other functions that make use of the HCFILE struct to make it transparent. Then you can full replace all calls to fopen(), fclose(), fread() etc with own wrapper function, no matter if the file is compressed or not.

It's already like this :) But I need to double check if I can switch all file ops using the new hc_* functions included by this patch in shared.*

```
bool   hc_fopen (HCFILE *fp, const char *path, char *mode);
size_t hc_fread_compress (void *ptr, size_t size, size_t nmemb, HCFILE *fp);
size_t hc_fread (void *ptr, size_t size, size_t nmemb, FILE *stream);
size_t hc_fwrite_compress (void *ptr, size_t size, size_t nmemb, HCFILE *fp);
void   hc_fwrite (const void *ptr, size_t size, size_t nmemb, FILE *stream);
int    hc_fseek (HCFILE *fp, off_t offset, int whence);
void   hc_rewind (HCFILE *fp);
off_t  hc_ftell (HCFILE *fp);
int    hc_fgetc (HCFILE *fp);
int    hc_fputc (int c, HCFILE *fp);
char  *hc_fgets (char *buf, int len, HCFILE *fp);
int    hc_fileno (HCFILE *fp);
int    hc_feof (HCFILE *fp);
//void   hc_fflush (HCFILE *fp);
void   hc_fclose (HCFILE *fp);
```

If yes I can remove hc_fread_compress/hc_fwrite_compress and using hc_fread/hc_fwrite by using a "void *fp" instead "FILE *fp" / "HCFILE *fp" and etc ...

> OK, but if you do, do it everywhere in the code (not just backend.c). Having code style in different convention across all code creates chaos. Please do this change in a separate PR. For this PR, stick to my convention.

Ok but after this patch is applied otherwise I will have to spend more time to do it again.
I started from backend.c because it's the ugliest :D

> Be more tricky handling macros, maybe you find a solution. Like hacking the source, but without hacking the source. If that's not an option then just deactivate the compression feature in cygwin. However we should stick to original source.

Ok, if you know how to solve the problem with zlib without patch the source and also without disable this new feature please do it. I don't have the environment ready to try build hashcat using cygwin/windows.

Thanks

example0.gz.sh
```diff
@@ -0,0 +1 @@
+./hashcat -t 32 -a 7 example0.hash ?a?a?a?a example.dict.gz
```
jsteube: Please rename back to example0.sh

include/shared.h
```diff
@@ -61,8 +62,24 @@ bool hc_string_is_digit (const char *s);
...
+void   hc_rewind   (HCFILE *fp);
+off_t  hc_ftell    (HCFILE *fp);
+int    hc_fgetc    (HCFILE *fp);
+int    hc_fputc    (int c, HCFILE *fp);
+char  *hc_fgets    (char *buf, int len, HCFILE *fp);
+int    hc_fileno   (HCFILE *fp);
+int    hc_feof     (HCFILE *fp);
+void   hc_fflush   (HCFILE *fp);
+void   hc_fclose   (HCFILE *fp);
+size_t hc_fwrite_compress (void *ptr, size_t size, size_t nmemb, HCFILE *fp);
```
jsteube: Why is there still _compress functions? Please remove. If this is related to potfile problem, then please explain the error in potfile and how to reproduce. Then update PR so that error occurs. I'll then try to fix.

include/types.h
```diff
@@ -988,6 +989,21 @@ typedef struct link_speed
...
+
+typedef struct hc_fp
+{
+  union
+  {
+    FILE *fp;
+    gzFile gfp;
+  } f;
+
+  short is_gzip;
```
jsteube: Please change to bool

include/types.h
```diff
@@ -988,6 +989,21 @@ typedef struct link_speed
 
 } link_speed_t;
 
+// handling gzip files
+
+typedef struct hc_fp
+{
+  union
```
jsteube: Please remove union, it's just easier to read.

include/types.h
```diff
@@ -1543,7 +1559,8 @@ typedef struct debugfile_ctx
 {
   bool enabled;
 
-  FILE *fp;
+//  FILE *fp;
```
jsteube: obsolete comment

src/backend.c
```diff
@@ -137,26 +137,31 @@ static int ocl_check_dri (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
 
   // Now we need to check if this an AMD vendor, because this is when the problems start
 
-  FILE *fd_drm = fopen (drm_card0_vendor_path, "rb");
+  HCFILE fp_drm;
+//  FILE *fd_drm = fopen (drm_card0_vendor_path, "rb");
```
jsteube: Please use the format which you've commented out

src/brain.c
```diff
@@ -1651,28 +1668,36 @@ bool brain_server_write_hash_dump (brain_server_db_hash_t *brain_server_db_hash,
...
   {
     brain_logging (stderr, 0, "%s: %s\n", file, strerror (errno));
 
     return false;
   }
   else
   {
-    const size_t nwrite = fwrite (brain_server_db_hash->long_buf, sizeof (brain_server_hash_long_t), brain_server_db_hash->long_cnt, fd);
+    // force gzip to false
+    fp.is_gzip = 0;
```
jsteube: After switching is_gzip to bool, do not forget to replace 0 with false and 1 with true.

jsteube:
Please change the PR according to my comments. I'll review again afterward.

include/shared.h
matrix: Because the porting it's not end. If you want check the potfile bug you need switch to new hc_* functions (now are all commented, check potfile source)

include/types.h
matrix: mmm, sure? I think union here it's the right choise. If you don't like fp->f.fp we can add a new macro to access the member's of union.

matrix:
> Please change the PR according to my comments. I'll review again afterward.

Yep, I will do asap. I'm curious if it works on windows :) have you tried?

src/backend.c
matrix: I need to switch back to fopen in "ocl_check_dri" ? There are similar functions for nvidia too, do I have to change those too?

example0.gz.sh
matrix: In the current version there're both example0.sh and example0.gz.sh
The only difference is that one uses the wordlist in plaintext and the other uses gz.
Do you want only the example0.sh with plaintext wordlist (original) or update example0.sh with new example.dict.gz ?

include/types.h
matrix: If I change it to bool I can't check for "fp->is_gzip == -1". Please double check the hc_fopen function to understand what I mean.

include/types.h
matrix: The "debugfile_ctx" struct is correct if you apply the patch. Please double check also by apply the patch, not only by reading the diff

```
typedef struct debugfile_ctx
{
  bool enabled;

//  FILE *fp;
  HCFILE *fp;
  char *filename;
  u32   mode;

} debugfile_ctx_t;
```

include/shared.h
matrix: You need also update the "struct potfile_ctx" to test it

matrix:
I think potfile can be use standard fopen/fread/fwrite/fclose/etc functions.
Handling gzip here is useless.

example0.gz.sh
jsteube: Only the example0.sh with plaintext wordlist since the goal is to demonstrate a very easy use of hashcat

include/types.h
jsteube: Yes, please remove it. We don't care about a few extra bytes at this point.

include/types.h
jsteube: What I mean is, please remove the comment itself. It makes it easier to read.

```
//  FILE *fp;
```

Same applies for all comments which you are using in all other files.

include/types.h
jsteube: I was reviewing hc_fopen() but I do not see where you need to set -1. There's no mentioning of -1 in the function body at all.

jsteube:
I've added some more comments. Please update the PR accordingly then I'll review again.

include/types.h
matrix: ok

include/types.h
matrix: ok

include/types.h
matrix: I remove it with the last patch, now it's bool and all is_gzip are true/false only

example0.gz.sh
matrix: ok, as soon as it works everything will remove both this file and the example wordlist in gzip

matrix:
@jsteube can you try to build with cygwin to see if it works ? Thanks ;)

matrix:
> Be more tricky handling macros, maybe you find a solution. Like hacking the source, but without hacking the source. If that's not an option then just deactivate the compression feature in cygwin. However we should stick to original source.

Done without deactivate the compression feature (tested on win7 x64 vm)
!(•̀ᴗ•́)و ̑̑

src/backend.c
```diff
@@ -440,28 +440,37 @@ static bool opencl_test_instruction (hashcat_ctx_t *hashcat_ctx, cl_context cont
...
     }
 
+    is_gzip = fp.is_gzip;
+
     #define EXTRASZ 100
 
-    char *buf = (char *) hcmalloc (st.st_size + 1 + EXTRASZ);
+    size_t klen = st.st_size;
+
+    if (is_gzip) klen *= 10; // must be >= of uncompress len
```
jsteube: Please remove for writing files compressed completely. Having kernels in readable format helps debugging a lot, especially on nvidia.

You can also get rid of this:

```
  if (is_gzip) klen *= 10; // must be >= of uncompress len
````

Which is super ugly.

File: src/backend.c [hunk 9]
```diff
@@ -503,29 +512,30 @@ static bool write_kernel_binary (hashcat_ctx_t *hashcat_ctx, char *kernel_file,
 {
   if (binary_size > 0)
   {
-    FILE *fp = fopen (kernel_file, "wb");
+    HCFILE fp;
 
-    if (fp == NULL)
+    // change HCFILE_FORMAT_GZIP to HCFILE_FORMAT_PLAIN to write kernel binary uncompressed
+    if (hc_fopen (&fp, kernel_file, "wb", HCFILE_FORMAT_GZIP) == false)
```
jsteube: Good thing is, after removing writing files compressed, you can also remove third parameter for hc_fopen() which kind of defeats a normal wrapper functionality.

src/debugfile.c
```diff
@@ -105,31 +105,35 @@ int debugfile_init (hashcat_ctx_t *hashcat_ctx)
...
     }
 
-    debugfile_ctx->fp = fp;
+    debugfile_ctx->fp = &fp;
   }
   else
   {
-    debugfile_ctx->fp = stdout;
+    fp.is_gzip = false;
+    fp.pfp = stdout;
```
jsteube: What is .pfp variable?

src/locking.c
```diff
@@ -6,35 +6,40 @@
...
 #include "types.h"
 #include "locking.h"
+#include "shared.h"
 
 #if defined (F_SETLKW)
 
-int lock_file (FILE *fp)
+int hc_lockfile (HCFILE *fp)
 {
+  if (!fp) return -1;
```
jsteube: please use (fp == NULL)

src/locking.c
```diff
@@ -6,35 +6,40 @@
...
     if (errno != EINTR) return -1;
   }
 
   return 0;
 }
 
-int unlock_file (FILE *fp)
+int hc_unlockfile (HCFILE *fp)
 {
+  if (!fp) return -1;
```
jsteube: Same here and probably other places, too. The idea is to make it easier to read by not need to interpret NOT while reading.

src/loopback.c
```diff
@@ -105,16 +105,16 @@ int loopback_write_open (hashcat_ctx_t *hashcat_ctx)
 
   hc_asprintf (&loopback_ctx->filename, "%s/%s.%d_%u", induct_ctx->root_directory, LOOPBACK_FILE, (int) now, random_num);
 
-  FILE *fp = fopen (loopback_ctx->filename, "ab");
+  HCFILE fp;
 
-  if (fp == NULL)
+  if (hc_fopen (&fp, loopback_ctx->filename, "ab", HCFILE_FORMAT_PLAIN) == false)
```
jsteube: Just a reminder to remove third parameter here as well.

src/main.c
```diff
@@ -60,19 +57,9 @@ static void main_log (hashcat_ctx_t *hashcat_ctx, FILE *fp, const int loglevel)
...
 
-  if (msg_newline == true)
-  {
-    event_ctx->prev_len = 0;
-  }
-  else
-  {
-    event_ctx->prev_len = msg_len;
-  }
+  event_ctx->prev_len = (msg_newline == true) ? 0 : msg_len;
```
jsteube: Please go back to old format, it's easier to read in block.

src/shared.c
```diff
@@ -602,14 +603,311 @@ void hc_string_trim_trailing (char *s)
   s[new_len] = 0;
 }
 
-size_t hc_fread (void *ptr, size_t size, size_t nmemb, FILE *stream)
+#if defined (__CYGWIN__)
+// workaround for zlib with cygwin build
+int _wopen(const char *path, int oflag, ...)
```
jsteube, line 608: Can you explain why you need this? Please also fix indents etc.

src/shared.c
```diff
@@ -602,14 +603,311 @@ void hc_string_trim_trailing (char *s)
...
+  {
+    unsigned char check[3] = { 0 };
+
+    int fd_tmp = open (path, O_RDONLY);
+
+    lseek (fd_tmp, 0, SEEK_SET);
+
+    size_t s = read (fd_tmp, check, sizeof(check));
+
+    if (s == 3 && (check[0] == 0x1f && check[1] == 0x8b && check[2] == 0x08)) file_format = HCFILE_FORMAT_GZIP;
```
jsteube: Branching on two different source variables is super hard to read. Please rewrite it to make it more beautiful. Maybe with two branches and/or multiline formating.

src/shared.c
```diff
@@ -602,14 +603,311 @@ void hc_string_trim_trailing (char *s)
...
+}
+
+size_t hc_fread (void *ptr, size_t size, size_t nmemb, HCFILE *fp)
 {
-  return fread (ptr, size, nmemb, stream);
+  size_t n = -1;
+
+  if (fp == NULL) return n;
+
+  if (fp->is_gzip)
```
jsteube, line 708: Please fix missing { and } blocks. It's easier to read.

src/shared.c
```diff
@@ -602,14 +603,311 @@ void hc_string_trim_trailing (char *s)
...
-size_t hc_fwrite (const void *ptr, size_t size, size_t nmemb, FILE *stream)
+size_t hc_fwrite (void *ptr, size_t size, size_t nmemb, HCFILE *fp)
 {
-  return fwrite (ptr, size, nmemb, stream);
+  size_t n = -1;
+
+  if (fp == NULL) return n;
+
+  if (fp->is_gzip)
+    n = gzfwrite (ptr, size, nmemb, fp->gfp);
```
jsteube, line 728: Same here

src/shared.c
```diff
@@ -602,14 +603,311 @@ void hc_string_trim_trailing (char *s)
...
+}
+
+int hc_fseek (HCFILE *fp, off_t offset, int whence)
+{
+  int r = -1;
+
+  if (fp == NULL) return r;
+
+  if (fp->is_gzip)
+    r = gzseek (fp->gfp, (z_off_t) offset, whence);
```
jsteube, line 748: And here... And so on for all other occurrences of this type.

jsteube:
Excellent work so far. We're almost there. Please fix the comments. Is potfile support still broken?

src/backend.c
matrix: I know :) ok

src/backend.c
matrix: yep

src/debugfile.c
matrix: because of "fp.fp" is ugly and unreadable ...
.pfp is the FILE * pointer (plain)
.gfp is the gzFile pointer (zlib)

src/locking.c
matrix: ok

src/locking.c
matrix: y

src/loopback.c
matrix: k

src/main.c
matrix: I love ternary operator :)
it's used also in other sections but if you prefer I switch back to old format

src/shared.c
matrix: it's the hack to make zlib build against cygwin !

src/shared.c
matrix: ok

src/shared.c
matrix: You mean for "if (fp->is_gzip)" ?

matrix:
> Excellent work so far. We're almost there. Please fix the comments. Is potfile support still broken?

Thanks ! All is working but in the potfile source I still using fopen/fread/fwrite/fclose/etc functions instead HCFILE

src/shared.c
jsteube: Sorry what I mean is make this:

```
  if (fp->is_gzip)
    gzrewind (fp->gfp);
  else
    rewind (fp->pfp);
```

like this:

```  
  if (fp->is_gzip)
  {
    gzrewind (fp->gfp);
  }
  else
  {
    rewind (fp->pfp);
  }

```

src/backend.c
```diff
@@ -440,28 +440,30 @@ static bool opencl_test_instruction (hashcat_ctx_t *hashcat_ctx, cl_context cont
 
 static bool read_kernel_binary (hashcat_ctx_t *hashcat_ctx, const char *kernel_file, size_t *kernel_lengths, char **kernel_sources, const bool force_recompile)
 {
-  FILE *fp = fopen (kernel_file, "rb");
+  HCFILE fp;
 
-  if (fp != NULL)
+  if (hc_fopen (&fp, kernel_file, "rb") != false)
```
jsteube: Please review all changes and try to -not- use negative comparison statements, in general try to avoid ! operator.

Replace with

```
if (hc_fopen (&fp, kernel_file, "rb") == true)
```

src/shared.c
```diff
@@ -602,14 +603,370 @@ void hc_string_trim_trailing (char *s)
...
+  else
+  {
+    fp->fd = open (path, oflag, fmode);
+  }
+
+  if (fp->fd == -1) return false;
+
+  if (fp->is_gzip)
+  {
+    if (!(fp->gfp = gzdopen (fp->fd, mode))) return false;
```
jsteube: Same here, please remove ! operator and rewrite

jsteube:
Almost ready, just two or three more comments. But please make sure to check all changed code for occurrences of ! operator and rewrite it. 

Please also give me example commandline (and files?) how to reproduce the problem with potfile.

matrix:
> Almost ready, just two or three more comments. But please make sure to check all changed code for occurrences of ! operator and rewrite it.
> 
> Please also give me example commandline (and files?) how to reproduce the problem with potfile.

Fixed :)

matrix:
@jsteube I miss something to fix or it's ok ?

jsteube:
Great addition, many thanks!

PR #2077, opened 2019-06-28
allow up to 64 bytes of salt in module 01460

arrtchiu:
Test case hashes.txt:
```
1f9dba8e10e8155fd684ca4ee8b416004a0729fae9472316ac2d9bfff743d1bb:aa73150989391fe8ffdc19b930f6059d029a9442f4c7e0da2609afc1ccb0ef05f85d37de3774826de38eb2a4c058fc3b6bc872a4afc518d5e581a0a858c48a92
```

Invocation:
```
hashcat -m 1460 -a 3 --hex-salt --custom-charset1=hasct hashes.txt '?1?1?1?1?1?1?1'
```

Result before patch:
```
Hashfile 'hashes.txt' on line 1 [...]: Salt-length exception
```

Result after patch:
Successfully broken
```
1f9dba8e10e8155fd684ca4ee8b416004a0729fae9472316ac2d9bfff743d1bb:aa73150989391fe8ffdc19b930f6059d029a9442f4c7e0da2609afc1ccb0ef05f85d37de3774826de38eb2a4c058fc3b6bc872a4afc518d5e581a0a858c48a92:hashcat
```

### Further details

When generating a HMAC-SHA256, the algorithm's first few operations are like this (pseudo-code):

```
if (key.size_in_bytes > 64) {
  key = sha256(key)
}

if (key.size_in_bytes < 64) {
  key = pad_with_nulls_up_to_size(key, 64)
}

do_hmac_operation(key, payload)
```

This means for keys > 64 bytes, they get SHA256'd back down to 32 bytes, then padded with zeroes.

A simple fix for users who want to use > 64 bytes could be some documentation which instructs the user to simply hash their key (or Hashcat could do it by precomputing for them, but I don't see an easy way of fitting that into the existing code)

### DIFF
--- a/src/modules/module_01460.c
+++ b/src/modules/module_01460.c
@@ -16,6 +16,7 @@ static const u32   DGST_POS1      = 7;
 static const u32   DGST_POS2      = 2;
 static const u32   DGST_POS3      = 6;
 static const u32   DGST_SIZE      = DGST_SIZE_4_8;
+static const u32   DGST_BLK_SIZE  = 64;
 static const u32   HASH_CATEGORY  = HASH_CATEGORY_RAW_HASH_AUTHENTICATED;
 static const char *HASH_NAME      = "HMAC-SHA256 (key = $salt)";
 static const u64   KERN_TYPE      = 1460;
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks! I think the same check can be done for hash-mode 60, 160 and 1760. For 1760 it will be 128 instead of 64. Can you please also provide a PR for them?

PR #2086, opened 2019-07-08
Fix #2083: hashcat not using multiple GPUs on MacOS

### LINKED ISSUE #2083: hashcat not using multiple GPUs on MacOS
After commit https://github.com/hashcat/hashcat/commit/e3500ff4aa608839d4667c5dca252b2edcb39111 hashcat won't make use of the Intel and AMD GPU on a MBP 2016 running MacOS Mojave 10.14.5.

At the last working commit, running gave:
hashcat (v5.1.0-968-ge3500ff4) starting in benchmark mode...

[snip CUDA warnings ...]

```
OpenCL Platform #1: Apple
=========================
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped.
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, 512/2048 MB allocatable, 12MCU
```

The first commit that introduces the bad behaviour is https://github.com/hashcat/hashcat/commit/6fd936b43a1aa18207727dfda49b1c878087defb), where running gives:

```
hashcat (v5.1.0-969-g6fd936b4) starting in benchmark mode...

OpenCL API (OpenCL 1.2 (Apr  7 2019 18:38:19)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, skipped
```

As you can see only device 2 is used, and the second GPU device 3 is skipped. Device 3 can be forced with -d3, but then device 2 will be skipped. Specifying -d2,3 will just use device 2.

matrix:
as the title, check the issue #2083 for details

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -6646,10 +6646,10 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
   // using force here enables both devices, which is the worst possible outcome
   // many users force by default, so this is not a good idea
 
-  //if (user_options->force == false)
-  //{
-  backend_ctx_find_alias_devices (hashcat_ctx);
-  //}
+  if (user_options->force == false)
+  {
+    backend_ctx_find_alias_devices (hashcat_ctx);
+  }
 
   if (backend_ctx->backend_devices_active == 0)
   {
--- a/src/backend.c
+++ b/src/backend.c
@@ -82,7 +82,7 @@ static int backend_ctx_find_alias_devices (hashcat_ctx_t *hashcat_ctx)
       {
         if (device_param_dst->skipped == false)
         {
-          device_param_dst->skipped = true;
+          device_param_src->skipped = true;
 
           backend_ctx->opencl_devices_active--;
 
--- a/src/backend.c
+++ b/src/backend.c
@@ -43,6 +43,10 @@ static bool is_same_device (const hc_device_param_t *src, const hc_device_param_
   if (src->pcie_device   != dst->pcie_device)   return false;
   if (src->pcie_function != dst->pcie_function) return false;
 
+  if (src->device_processors         != dst->device_processors)         return false;
+  if (src->device_maxclock_frequency != dst->device_maxclock_frequency) return false;
+  if (src->device_maxworkgroup_size  != dst->device_maxworkgroup_size)  return false;
+
   return true;
 }
 
@@ -6226,6 +6230,10 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
           }
         }
 
+        device_param->pcie_bus      = 0;
+        device_param->pcie_device   = 0;
+        device_param->pcie_function = 0;
+
         if (device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
         {
           if ((device_param->opencl_platform_vendor_id == VENDOR_ID_AMD) && (device_param->opencl_device_vendor_id == VENDOR_ID_AMD))
@@ -6646,10 +6654,7 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
   // using force here enables both devices, which is the worst possible outcome
   // many users force by default, so this is not a good idea
 
-  if (user_options->force == false)
-  {
-    backend_ctx_find_alias_devices (hashcat_ctx);
-  }
+  backend_ctx_find_alias_devices (hashcat_ctx);
 
   if (backend_ctx->backend_devices_active == 0)
   {


philsmd:
the question I ask myself is if this is really a valid fix.

It's good that you identified the root of the problem (thanks, very very good), but an actually good fix would be to try to understand why backend_ctx_find_alias_devices () thinks that 2 different devices are seen as identical/aliases while they are not (at least I think that is the problem here, but please correct me if that's wrong !).

We also should consider the comment here: https://github.com/hashcat/hashcat/pull/2086/files#diff-7af547b52769371de74af0bfca268a62R6645 , which really tells us to not use this check ```user_options->force == false```
because it leads to some very dangerous problems. Therefore I think that we need to digg deeper here and try to understand and fix the problem with aliases instead.

matrix:
> the question I ask myself is if this is really a valid fix.
> 
> It's good that you identified the root of the problem (thanks, very very good), but an actually good fix would be to try to understand why backend_ctx_find_alias_devices () thinks that 2 different devices are seen as identical/aliases while they are not (at least I think that is the problem here, but please correct me if that's wrong !).
> 
> We also should consider the comment here: https://github.com/hashcat/hashcat/pull/2086/files#diff-7af547b52769371de74af0bfca268a62R6645 , which really tells us to not use this check `user_options->force == false`
> because it leads to some very dangerous problems. Therefore I think that we need to digg deeper here and try to understand and fix the problem with aliases instead.

Hi @philsmd , probably the backend_ctx_find_alias_devices() do something wrong, as you write, but it's not only this functionality to have some issue.
Before think how to rewrite the code we can use this workaround initially and then go deeper with some "hidden" bugs in the code :)

I'm waiting the merge of #2075 before start rewriting something. If you have another solution for that problem please do it :)

Thanks

matrix:
@philsmd this patch make -D1,2 works without set --force but honestly ... I don't even know what this function is for

```
diff --git a/src/backend.c b/src/backend.c
index b9143b68..872f7cf9 100644
--- a/src/backend.c
+++ b/src/backend.c
@@ -66,7 +66,7 @@ static int backend_ctx_find_alias_devices (hashcat_ctx_t *hashcat_ctx)

       if (device_param_dst->skipped_warning == true) continue;

-      if (is_same_device (device_param_src, device_param_dst) == false) continue;
+      if (is_same_device (device_param_src, device_param_dst) == true) continue;

       device_param_src->device_id_alias_buf[device_param_src->device_id_alias_cnt] = device_param_dst->device_id;
       device_param_src->device_id_alias_cnt++;
```

philsmd:
yeah the alias check is trying to find out if the two devices are identical (like using the OpenCL backend together with the CUDA backend and that's only one device, not 2 different devices). This means that is_same_device () doesn't work for your device list, but I'm not sure why.... we would need to troubleshoot why the values for those devices can be the same even if you have different devices (are we even sure that you really have 3 physical different devices ? how can be distinguish them differently by fixing the is_same_device () function ?)

matrix:
> yeah the alias check is trying to find out if the two devices are identical (like using the OpenCL backend together with the CUDA backend and that's only one device, not 2 different devices). This means that is_same_device () doesn't work for your device list, but I'm not sure why.... we would need to troubleshoot why the values for those devices can be the same even if you have different devices (are we even sure that you really have 3 physical different devices ? how can be distinguish them differently by fixing the is_same_device () function ?)

Why not by the name?

```
diff --git a/src/backend.c b/src/backend.c
index b9143b68..107730b4 100644
--- a/src/backend.c
+++ b/src/backend.c
@@ -42,6 +42,7 @@ static bool is_same_device (const hc_device_param_t *src, const hc_device_param_
   if (src->pcie_bus      != dst->pcie_bus)      return false;
   if (src->pcie_device   != dst->pcie_device)   return false;
   if (src->pcie_function != dst->pcie_function) return false;
+  if (src->device_name   != dst->device_name)   return false;

   return true;
 }
```

matrix:
These info are not saved because of the CPU type not handled well in backend.c

By debugging is_same_device() ..

```
SRC : 0, 0, 0, Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz
DST : 0, 0, 0, Iris

```

Here is where these vars are set, but only for GPU devices 
https://github.com/hashcat/hashcat/blob/master/src/backend.c#L6239

philsmd:
I am not sure, maybe the only thing missing is to add a check for VENDOR_ID_APPLE for the Iris GPU similar to the 2 checks we already have for AMD and Nvidia:
- https://github.com/hashcat/hashcat/blob/a714ce645fbd58cf26f59dac35ae5a9190bc966c/src/backend.c#L6239
- https://github.com/hashcat/hashcat/blob/a714ce645fbd58cf26f59dac35ae5a9190bc966c/src/backend.c#L6252

If I understood this correctly, the problem is not the CPU, but the missing assignment of pcie_bus, pcie_device and pcie_function in case of VENDOR_ID_APPLE. So we should fix this instead, I think !?

matrix:
> I am not sure, maybe the only thing missing is to add a check for VENDOR_ID_APPLE for the Iris GPU similar to the 2 checks we already have for AMD and Nvidia:
> 
> * https://github.com/hashcat/hashcat/blob/a714ce645fbd58cf26f59dac35ae5a9190bc966c/src/backend.c#L6239
> * https://github.com/hashcat/hashcat/blob/a714ce645fbd58cf26f59dac35ae5a9190bc966c/src/backend.c#L6252
> 
> If I understood this correctly, the problem is not the CPU, but the missing assignment of pcie_bus, pcie_device and pcie_function in case of VENDOR_ID_APPLE. So we should fix this instead, I think !?

Already tried and I got the following error 

`clGetDeviceInfo(): CL_INVALID_OPERATION`

when I try to retrive these values. Have you a Mac to confirm this?
Thanks!

src/backend.c
```diff
@@ -78,7 +82,7 @@ static int backend_ctx_find_alias_devices (hashcat_ctx_t *hashcat_ctx)
       {
         if (device_param_dst->skipped == false)
         {
-          device_param_dst->skipped = true;
+          device_param_src->skipped = true;
```
jsteube, line 85: Please explain switch to device_param_src?

jsteube:
I'm not 100% certain since I do not have access to OSX to confirm. I think @philsmd is right that for non-nvidia devices the pci-express information is set to zero and the is_same_device () should not be called.

```
  if (src->device_processors         != dst->device_processors)         return false;
  if (src->device_maxclock_frequency != dst->device_maxclock_frequency) return false;
  if (src->device_maxworkgroup_size  != dst->device_maxworkgroup_size)  return false;
```

This will work only since the frequency is very likely to be not same for CPU and AMD GPU. It's however not the right way to check it. We should aim to have the most minimal checks. The reason, for instance, is to avoid problems with frequency scaling by GPU bios. These days they change clock speed multiple times per second to save power consumption and it's possible to have a bad outcome if we use it as a reference.

@matrix can you please do the change as @philsmd suggested and update the PR?

src/backend.c
matrix: because with APPLE the right device is skipped instead of get selected

jsteube: OK, but will not be needed if fixed based on vendor_id is used.

matrix:
I can't because also with APPLE the pci-express info are not set. The only way I found is what I pushed in this patch. Without I don't know how to make it work with APPLE.

jsteube:
Can you please try and post here what happens if you apply only this patch and nothing else:

```
diff --git a/src/backend.c b/src/backend.c
index 2f3c2523..28f02a1e 100644
--- a/src/backend.c
+++ b/src/backend.c
@@ -39,6 +39,8 @@ static double TARGET_MSEC_PROFILE[4] = { 2, 12, 96, 480 };
 
 static bool is_same_device (const hc_device_param_t *src, const hc_device_param_t *dst)
 {
+  if (src->opencl_device_vendor_id != dst->opencl_device_vendor_id) return false;
+
   if (src->pcie_bus      != dst->pcie_bus)      return false;
   if (src->pcie_device   != dst->pcie_device)   return false;
   if (src->pcie_function != dst->pcie_function) return false;
```

This one will work because on Windows and Linux for NV based systems. The opencl_device_vendor_id value is set fixed to VENDOR_ID_NV for backward compatibility with some OpenCL kernel code. I think it will also solve OSX problem. If not, maybe I did not fully understand the problem.

matrix:
with APPLE don't work (same vendor_id)

```
OpenCL Info:
============

OpenCL Platform ID #1
  Vendor..: Apple
  Name....: Apple
  Version.: OpenCL 1.2 (Apr 18 2019 20:03:31)

  Backend Device ID #1 (Alias: #2)
    Type...........: CPU
    Vendor.ID......: 4
    Vendor.........: Intel
    Name...........: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz
    Version........: OpenCL 1.2
    Processor(s)...: 4
    Clock..........: 3000
    Memory.........: 2048/8192 MB allocatable
    OpenCL.Version.: OpenCL C 1.2
    Driver.Version.: 1.1

  Backend Device ID #2 (Alias: #1)
    Type...........: GPU
    Vendor.ID......: 4
    Vendor.........: Intel
    Name...........: Iris
    Version........: OpenCL 1.2
    Processor(s)...: 40
    Clock..........: 1200
    Memory.........: 384/1536 MB allocatable
    OpenCL.Version.: OpenCL C 1.2
    Driver.Version.: 1.2(Jun  2 2019 21:52:17)
```

jsteube:
Wait, the device vendor ID should not be the same. From the original issue, we have Apple and AMD:

```
OpenCL API (OpenCL 1.2 (Apr  7 2019 18:38:19)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, skipped
```

In your case it's only Apple, but it's CPU and GPU. In this case we don't want the CPU to be enabled automatically by default. You should be able to enable both if you use -D1,2 which is what we'd expect the user to do. Is this not working, too?



Is there anything else I need to

jsteube:
What I mean is that for the original issue in #2083 this should fix the issue because of value for vendor_id will be Intel and Apple. The device number 3 will no longer be skipped.

For your case it's something completely different and we need to make sure to not confuse this. You have GPU and CPU. In this situation, the CPU should be disabled by default no matter which vendors are involved. That's the way we want it to work. 

Only if the user sets -D1,2 then both devices should be enabled at the same time. Can you confirm this works?

matrix:
> Only if the user sets -D1,2 then both devices should be enabled at the same time. Can you confirm this works?

The bug it's here, if I set -D1,2 only the first is enabled

singe:
> Can you please try and post here what happens if you apply only this patch and nothing else:

I applied just that patch to the latest commit to master (6ecc662451be3b4867a35716281d5aec71833e8b). While it fixes the multiple GPU problem, now nothing cracks:

```
▶ ./hashcat -b -m1000
hashcat (v5.1.0-1285-g6ecc6624+) starting in benchmark mode...

OpenCL API (OpenCL 1.2 (Jun 23 2019 21:50:55)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, 512/2048 MB allocatable, 12MCU

Benchmark relevant options:
===========================
* --optimized-kernel-enable

Hashmode: 1000 - NTLM

Speed.#2.........:        0 H/s (0.00ms) @ Accel:0 Loops:0 Thr:0 Vec:256
Speed.#3.........:        0 H/s (0.00ms) @ Accel:0 Loops:0 Thr:0 Vec:0
Speed.#*.........:        0 H/s

Started: Wed Jul 31 12:30:32 2019
Stopped: Wed Jul 31 12:30:34 2019
```

```
▶ ./hashcat -b -m500 
hashcat (v5.1.0-1285-g6ecc6624+) starting in benchmark mode...

OpenCL API (OpenCL 1.2 (Jun 23 2019 21:50:55)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, 512/2048 MB allocatable, 12MCU

Benchmark relevant options:
===========================
* --optimized-kernel-enable

Hashmode: 500 - md5crypt, MD5 (Unix), Cisco-IOS $1$ (MD5) (Iterations: 1000)

Speed.#2.........:        0 H/s (0.00ms) @ Accel:0 Loops:0 Thr:0 Vec:256
Speed.#3.........:        0 H/s (0.00ms) @ Accel:0 Loops:0 Thr:0 Vec:0
Speed.#*.........:        0 H/s

Started: Wed Jul 31 12:30:46 2019
Stopped: Wed Jul 31 12:30:52 2019
```

matrix:
@singe please restart from cloning the official repository (do not apply other PR not jet approved). It should work.

singe:
I’m confused. If master isn’t the official repo, what is?

matrix:
> I’m confused. If master isn’t the official repo, what is?

I thought you applied some extra patches :) As soon as I can I do some tests and let you know

matrix:
> Can you please try and post here what happens if you apply only this patch and nothing else:
> 
> ```
> diff --git a/src/backend.c b/src/backend.c
> index 2f3c2523..28f02a1e 100644
> --- a/src/backend.c
> +++ b/src/backend.c
> @@ -39,6 +39,8 @@ static double TARGET_MSEC_PROFILE[4] = { 2, 12, 96, 480 };
>  
>  static bool is_same_device (const hc_device_param_t *src, const hc_device_param_t *dst)
>  {
> +  if (src->opencl_device_vendor_id != dst->opencl_device_vendor_id) return false;
> +
>    if (src->pcie_bus      != dst->pcie_bus)      return false;
>    if (src->pcie_device   != dst->pcie_device)   return false;
>    if (src->pcie_function != dst->pcie_function) return false;
> ```
> 
> This one will work because on Windows and Linux for NV based systems. The opencl_device_vendor_id value is set fixed to VENDOR_ID_NV for backward compatibility with some OpenCL kernel code. I think it will also solve OSX problem. If not, maybe I did not fully understand the problem.

@singe without this hashcat works?

matrix:
https://hashcat.net/forum/thread-8525-post-45441.html#pid45441

jsteube:
Commit https://github.com/hashcat/hashcat/commit/57a149276c5a20d5af1b2364628b3a305772ab3d will replace this PR

singe:
Looks like this issue is back with the latest 6.0.0 release and commit. I did a quick check, and applying the same patch @matrix did earlier to allow --force fixes it, so there's no technical reason why it shouldn't work, just looks like an error in the selection logic I haven't had time to dig into yet.

The diff:

```
diff --git a/src/backend.c b/src/backend.c
index 80fdbb38..e0c9246d 100644
--- a/src/backend.c
+++ b/src/backend.c
@@ -6606,7 +6606,7 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
     // using force here enables both devices, which is the worst possible outcome
     // many users force by default, so this is not a good idea
 
-    //if (user_options->force == false)
+    if (user_options->force == false)
     //{
     backend_ctx_find_alias_devices (hashcat_ctx);
     //{
```

For completeness sake, here's the output from hashcat 5.1.0, 6.0.0 and 6.0.0 with my patch and --force:

5.1.0 (homebrew)
```
/usr/local/Cellar/hashcat/5.1.0/bin/hashcat -m1000 --benchmark
hashcat (v5.1.0) starting in benchmark mode...

OpenCL Platform #1: Apple
=========================
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped.
* Device #2: Intel(R) HD Graphics 530, 384/1536 MB allocatable, 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, 512/2048 MB allocatable, 12MCU

Benchmark relevant options:
===========================
* --optimized-kernel-enable

Hashmode: 1000 - NTLM

Speed.#2.........:   739.5 MH/s (32.47ms) @ Accel:128 Loops:32 Thr:256 Vec:1
Speed.#3.........:  4081.5 MH/s (95.39ms) @ Accel:512 Loops:256 Thr:256 Vec:1
Speed.#*.........:  4821.0 MH/s

Started: Wed Jun 17 13:01:44 2020
Stopped: Wed Jun 17 13:01:55 2020
```

6.0.0
```
./hashcat -m1000 --benchmark
hashcat (v6.0.0) starting in benchmark mode...

OpenCL API (OpenCL 1.2 (May  3 2020 20:15:19)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 1472/1536 MB (384 MB allocatable), 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, skipped

Benchmark relevant options:
===========================
* --optimized-kernel-enable

Hashmode: 1000 - NTLM

Speed.#2.........:   669.5 MH/s (73.83ms) @ Accel:1024 Loops:256 Thr:8 Vec:1

Started: Wed Jun 17 13:00:49 2020
Stopped: Wed Jun 17 13:00:54 2020
```

6.0.0 patched with --force
```
./hashcat -m1000 --benchmark --force
hashcat (v6.0.0) starting in benchmark mode...

OpenCL API (OpenCL 1.2 (May  3 2020 20:15:19)) - Platform #1 [Apple]
* Device #1: Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz, skipped
* Device #2: Intel(R) HD Graphics 530, 1472/1536 MB (384 MB allocatable), 24MCU
* Device #3: AMD Radeon Pro 455 Compute Engine, 1984/2048 MB (512 MB allocatable), 12MCU

Benchmark relevant options:
===========================
* --force
* --optimized-kernel-enable

Hashmode: 1000 - NTLM

Speed.#2.........:   674.1 MH/s (73.28ms) @ Accel:1024 Loops:256 Thr:8 Vec:1
Speed.#3.........:  4171.1 MH/s (95.53ms) @ Accel:1024 Loops:512 Thr:64 Vec:1
Speed.#*.........:  4845.2 MH/s

Started: Wed Jun 17 13:01:06 2020
Stopped: Wed Jun 17 13:01:11 2020
```

PR #2098, opened 2019-07-11
restyling hwmon

matrix:
- moved functions from hwmon.c to .c/.h
- code style


jsteube:
I did not go into detail yet, but this breaks use on my Linux Box. 

Before:

```
Hardware.Mon.#1..: Temp: 81c Fan: 45% Util: 96% Core:1126MHz Mem:3004MHz Bus:16
```

After:

```
Hardware.Mon.#1..: N/A
```

matrix:
I will do a recheck as soon as possible. Thanks

matrix:
Fixed, now works

jsteube:
Yeah, it works on Linux now. It also works with MinGW but on cygwin, I'm getting a segfault. Without this patch, it works fine.

matrix:
probably I miss something from the original patch ...
as soon as I understand the problem I will update the PR.
Thanks

matrix:
I can't reproduce it because the monitor hardware is not activated inside the VM ... @philsmd have you a windows machine native with hashcat to check what is wrong with cygwin here ?
Thanks!

jsteube:
Can you fix the conflict so that we can continue to work on this?

jsteube:
@matrix Please update status on this PR

matrix:
check here #2806

PR #2101, opened 2019-07-12
Fix #2084: bug with -S -O arguments set together

### LINKED ISSUE #2084: Brain Server Failed to Crack NTLM
HashCat Output log

```shell
Session..........: hashcat (Brain Session/Attack:0xdcddafec/0x49bb773c)
Status...........: Exhausted
Hash.Name........: NTLM
Hash.Target......: 086d31b727ecc2d671ed3809eb0e1bc5
Time.Started.....: Mon Jul  8 20:01:39 2019 (1 sec)
Time.Estimated...: Mon Jul  8 20:01:40 2019 (0 secs)
Guess.Mask.......: ?1?1?1?1?1?1 [6]
Guess.Charset....: -1 ?l?u, -2 Undefined, -3 Undefined, -4 Undefined
Guess.Queue......: 2/2 (100.00%)
Speed.#1.........:        0 H/s (0.00ms) @ Accel:256 Loops:1 Thr:256 Vec:1
Recovered........: 0/1 (0.00%) Digests, 0/1 (0.00%) Salts
Progress.........: 19770609664/19770609664 (100.00%)
Rejected.........: 19770609664/19770609664 (100.00%)
Brain.Link.All...: RX: 0 B, TX: 16 B
Brain.Link.#1....: N/A
Restore.Point....: 0/19770609664 (0.00%)
Restore.Sub.#1...: Salt:0 Amplifier:0-0 Iteration:0-1
Candidates.#1....: [Generating]
Hardware.Mon.#1..: Temp: 33c Fan: 28% Core:1284MHz Mem:1000MHz Bus:0
Started: Mon Jul  8 20:01:36 2019
Stopped: Mon Jul  8 20:01:41 2019
```

> or

```shell
Session..........: hashcat (Brain Session/Attack:0xdcddafec/0x49bb773c)
Status...........: Exhausted
Hash.Name........: NTLM
Hash.Target......: 086d31b727ecc2d671ed3809eb0e1bc5
Time.Started.....: Mon Jul  8 20:01:39 2019 (1 sec)
Time.Estimated...: Mon Jul  8 20:01:40 2019 (0 secs)
Guess.Mask.......: ?1?1?1?1?1?1 [6]
Guess.Charset....: -1 ?l?u, -2 Undefined, -3 Undefined, -4 Undefined
Guess.Queue......: 2/2 (100.00%)
Speed.#1.........:        0 H/s (0.00ms) @ Accel:256 Loops:1 Thr:256 Vec:1
Recovered........: 0/1 (0.00%) Digests, 0/1 (0.00%) Salts
Progress.........: 19770609664/19770609664 (100.00%)
Rejected.........: 19770609664/19770609664 (100.00%)
Brain.Link.All...: RX: 0 B, TX: 16 B
Brain.Link.#1....: N/A
Restore.Point....: 0/19770609664 (0.00%)
Restore.Sub.#1...: Salt:0 Amplifier:0-0 Iteration:0-1
Candidates.#1....: $HEX[730020002000] -> $HEX[20007e007e00]
Hardware.Mon.#1..: Temp: 33c Fan: 28% Core:1284MHz Mem:1000MHz Bus:0
Started: Mon Jul  8 20:01:36 2019
Stopped: Mon Jul  8 20:01:41 2019
```

> start brain server code:

```shell
hashcat --brain-server --brain-host=<ip> --brain-port=<port> --brain-password=<pwd>
```

> run code:

```shell
hashcat -O --brain-client --brain-client-features=2 --brain-host=<ip> --brain-port=<port> --brain-password=<pwd> -a 3 -m 1000 '596b7e1a3c020eccac807f4009273eb8' --increment --increment-min 2 --increment-max 3 -1 ?l?u ?1?1?1 --status --status-timer 3
```

> show crack results code:

```shell
hashcat -O --brain-client --brain-client-features=2 --brain-host=<ip> --brain-port=<port> --brain-password=<pwd> -a 3 -m 1000 '596b7e1a3c020eccac807f4009273eb8' --increment --increment-min 2 --increment-max 3 -1 ?l?u ?1?1?1 --status --status-timer 3 --show
```

matrix:
Fix issue #2084

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -353,7 +353,7 @@ int hashconfig_init (hashcat_ctx_t *hashcat_ctx)
         {
           if (user_options->quiet == false) event_log_warning (hashcat_ctx, "%s: Optimized OpenCL kernel requested but not needed - falling back to pure OpenCL kernel", source_file);
         }
-        else if (user_options->slow_candidates == false)
+        else
         {
           hashconfig->opti_type |= OPTI_TYPE_OPTIMIZED_KERNEL;
         }
--- a/src/interface.c
+++ b/src/interface.c
@@ -353,7 +353,7 @@ int hashconfig_init (hashcat_ctx_t *hashcat_ctx)
         {
           if (user_options->quiet == false) event_log_warning (hashcat_ctx, "%s: Optimized OpenCL kernel requested but not needed - falling back to pure OpenCL kernel", source_file);
         }
-        else
+        else if (user_options->slow_candidates == false)
         {
           hashconfig->opti_type |= OPTI_TYPE_OPTIMIZED_KERNEL;
         }
--- a/src/mpsp.c
+++ b/src/mpsp.c
@@ -1297,13 +1297,16 @@ int mask_ctx_update_loop (hashcat_ctx_t *hashcat_ctx)
         return -1;
       }
 
-      if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+      if (user_options->slow_candidates == false)
       {
-        if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
-      }
-      else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)
-      {
-        if (mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+        {
+          if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
+        }
+        else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)
+        {
+          if (mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        }
       }
 
       u32 css_cnt_orig = mask_ctx->css_cnt;


philsmd:
what's the idea behind this ?

I don't think it makes much sense if the user is using -S -O but the -O option is ignored silently. I am missing something here ?

matrix:
> what's the idea behind this ?
> 
> I don't think it makes much sense if the user is using -S -O but the -O option is ignored silently. I am missing something here ?

ola :) I tested the cmdline in the related issue against almost all the revision from the current up to  the revision that add the '-S' feature, and It has never worked :|

philsmd:
I understand that, but we need a fix that actually solves the problem.

It's very bad if we just ignore the user choice of using optimized kernels (< 32 length passwords in general). That's a crucial choice that shouldn't be ignored, because it affects the performance a lot !

matrix:
> I understand that, but we need a fix that actually solves the problem.
> 
> It's very bad if we just ignore the user choice of using optimized kernels (< 32 length passwords in general). That's a crucial choice that shouldn't be ignored, because it affects the performance a lot !

try analyzing it yourself, there're many "if" to check inside "backed.c" ... if you want we can do it in two

philsmd:
maybe I just do not understand your fix.

but for me it looks like this:
1. the user really wants to use -O -S (both !)
2. your fix (silently) converts the user choice from this  -O -S by removing -O => to just -S

=> the problem is that the optimized kernels are not used (now, with your patch), but the user requested them and s/he doesn't know what is going on and why -O is just ignored (and no optimized kernels are used).
This is not a good fix to ignore the command line options, in my opinion... but maybe I am misunderstanding your fix that's why I am asking why you fix it like that and where the actual problem is

maybe we should also analyzed if other hash types are affected the same way or only utf16-based kernels... maybe that's the real reason and the root of the problem is something completely different (like utf16 encoding etc... btw I didn't troubleshoot/debug it yet, so I have no clue)

matrix:
ok, so I can close this or you want go deeper to try to fix it? let me know

roycewilliams:
It feels worth fixing - I'm not clear on *how*, but it feels like leaving it as-is is also sub-optimal.

matrix:
the problem is that in my opinion it never worked, so it's not only a "fix" ..

philsmd:
I'm totally with you guys. we need to fix it that's why I've investigated it a little bit already here: https://github.com/hashcat/hashcat/issues/2084#issuecomment-509759131

... but it's also important that the fix actually fixes the problem with optimized kernels and not just silently doesn't use the optimized kernels (implying a huge speed drop).

I've just tested with a -m 100 hash and -O -S works in -a 3 mode.

so to summarize the problem is only triggered if -a 3 AND -O AND -S are selected and only if the kernel uses utf16 (OPTS_TYPE_PT_UTF16LE ?).... so it's very specific (and actually no wonder that we didn't catch this problem yet with all those prerequisite needed ;) )

In theory we would need to look into the src/slow_candidates.c file for cases like this, but it doesn't really do much in case of ATTACK_MODE_BF (maybe that's the problem ?).

You can also see from the hashcat status
```Candidates.#1....: $HEX[390031003100] -> $HEX[390031003100]```
that something is wrong in the host because it should just say 911 (in my opinion), because the kernel already does the utf16le conversion (and therefore it's done twice I think ! and that might be the main problem that we need to fix !?)

**Comment by matrix** [comment]
Very good :) Which kernel produce this output ? 

`Candidates.#1....: $HEX[390031003100] -> $HEX[390031003100]
`

**Comment by philsmd** [comment]
I think the problem is that we have OPTS_TYPE_PT_UTF16LE for -m 1000 and the call to the "mask processor" in https://github.com/hashcat/hashcat/blob/d1f473d6bab953fe575d77814661e066df043c2f/src/slow_candidates.c#L328
already converts the password candidates here https://github.com/hashcat/hashcat/blob/d1f473d6bab953fe575d77814661e066df043c2f/src/mpsp.c#L1300-L1303
and the actual kernel code does also the utf16 conversion... therefore (as I already guessed above) we are basically doing the utf16 conversion twice (once in the slow candidate code and once in the kernel code, e.g here https://github.com/hashcat/hashcat/blob/d1f473d6bab953fe575d77814661e066df043c2f/OpenCL/m01000_a0-optimized.cl#L195-L196

**Comment by matrix** [comment]
Great! :) You found it 👍

**Comment by matrix** [comment]
it seems that this is enough, is it to you too?

```
diff --git a/src/mpsp.c b/src/mpsp.c
index 0fce4d5f..56faef68 100644
--- a/src/mpsp.c
+++ b/src/mpsp.c
@@ -1297,13 +1297,16 @@ int mask_ctx_update_loop (hashcat_ctx_t *hashcat_ctx)
         return -1;
       }

-      if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+      if (user_options->slow_candidates == false)
       {
-        if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
-      }
-      else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)
-      {
-        if (mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+        {
+          if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
+        }
+        else
+        {
+          if ((hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE) && mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        }
       }

       u32 css_cnt_orig = mask_ctx->css_cnt;
```

**Comment by philsmd** [comment]
seems at least that this fix is targeting the actual problem better than the patch in PR.

I'm not sure if there are any side-effects of this fix, e.g. we would need to check when exactly we need the mp_css_utf16le_expand () on the host etc and if it is never needed if slow candidates are enabled.

btw: why do you not use "else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)" and get rid of that weird long && at the same time ? It looks better to me that way (even though I know the outcome is identical)

**Comment by matrix** [comment]
> seems at least that this fix is targeting the actual problem better than the patch in PR.
> 
> I'm not sure if there are any side-effects of this fix, e.g. we would need to check when exactly we need the mp_css_utf16le_expand () on the host etc and if it is never needed if slow candidates are enabled.
> 
> btw: why do you not use "else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)" and get rid of that weird long && at the same time ? It looks better to me that way (even though I know the outcome is identical)

Hi, if you have a better patch share it, so let's choose @jsteube

**Comment by jsteube** [comment]
Great find and analysis. 

The fix would be to not call mp_css_utf16le_expand()  in case this is a slow_candidate session. Also note that the problem should exist in mp_css_utf16be_expand() which makes fixing a bit more easy. too. Therefore I'd suggest the following fix:

```
diff --git a/src/mpsp.c b/src/mpsp.c
index 0fce4d5f..2ebf72d7 100644
--- a/src/mpsp.c
+++ b/src/mpsp.c
@@ -1297,13 +1297,16 @@ int mask_ctx_update_loop (hashcat_ctx_t *hashcat_ctx)
         return -1;
       }
 
-      if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+      if (user_options->slow_candidates == false)
       {
-        if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
-      }
-      else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)
-      {
-        if (mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16LE)
+        {
+          if (mp_css_utf16le_expand (hashcat_ctx) == -1) return -1;
+        }
+        else if (hashconfig->opts_type & OPTS_TYPE_PT_UTF16BE)
+        {
+          if (mp_css_utf16be_expand (hashcat_ctx) == -1) return -1;
+        }
       }
 
       u32 css_cnt_orig = mask_ctx->css_cnt;
```

If you're fine with this, please update the PR and I'll merge

matrix:
done

Issue #2103, opened 2019-07-13
[Alg Request] SHA256(MD5('password'))


vido89:
This algorithm is used on Huawei routers to encrypt user passwords can we have it on hashcat ?

philsmd:
How can you **encrypt** passwords with a hashing algorithm ? Can you explain the details ?

vido89:
I gues I used bad terminology.
So I go here https://md5hashing.net/hash and paste my hash "402931e04c03e24d360477a9f90b9eb15777e154360f06228be15c37679016ef" and I got I guess un-hashed string and paste it here https://hashtoolkit.com/reverse-hash?hash=a016b28a2669f876cc24e9b3dac1bd73 and I got clear text password

vido89:
Its the same as 4700 | sha1(md5($pass)) just insted `sha1` I need `sha256`

matrix:
@vido89 I'm pushing the new algo. Have you time to test it?

vido89:
sure I will be glade to test it

matrix:
Try using the following repository, the only that not working it's a3 with -O (i'm debugging), the others works

https://github.com/matrix/hashcat/tree/sha256_md5_huawei

Let me know :)

matrix:
fixed -a3 -O :)

vido89:
Great it works
./hashcat -m 4710 -a 3 -i --increment-min=6 --increment-max=10 402931e04c03e24d360477a9f90b9eb15777e154360f06228be15c37679016ef ?l?l?l?l?l?l?l?l?l?l

Edit: Tnx a lot

jsteube:
Added support with PR: https://github.com/hashcat/hashcat/pull/2110

Note: Mode was changed from 4710 to 20800

collerok:
for those come there to findout whats the password hidden for user telecomadmin and hash is 

402931e04c03e24d360477a9f90b9eb15777e154360f06228be15c37679016ef

password is:    **admintelecom**

PR #2110, opened 2019-07-13
Added hash-mode: sha256(md5(pass))

matrix:
added the algo proposed here #2103


philsmd:
didn't look at the code yet. but docs/changes.txt and docs/readme.txt changes are missing for sure

jsteube:
Please update hash-mode number to next one free, 4710 can't be used since 4700 is used. 4710 would be salted version of 4700.

matrix:
> Please update hash-mode number to next one free, 4710 can't be used since 4700 is used. 4710 would be salted version of 4700.

ok, could you give me a list of free hash-mode to be used ? thanks!

matrix:
done

PR #2100, opened 2019-07-12
fix bug in mpsp

matrix:
Hi,
it's a clear (and my) error :) 

Without '-O' works but :

```
$ ./hashcat -t 32 -a 7 example0.hash ?a?a?a?a example.dict --quiet -O
Integer overflow detected in keyspace of mask: ?a?a?a?a
```

PR #2118, opened 2019-07-18
Remove copy-pasted 'return -1;' to fix -a 6 and 7

### LINKED ISSUE #2117: attack-mode 6 and 7 broken in master
One gets the error message "Integer overflow detected in keyspace of mask" for all non-huge keyspaces due to a copy-paste typo in [src/mpsp.c line 1235](https://github.com/hashcat/hashcat/commit/013b914e2240220dd854bcf30e691a2f3dafd027#diff-22610ad22085d5eaa33f122303492fe0R1235) from commit 013b914e2240220dd854bcf30e691a2f3dafd027
The error message is not shown if the error condition actually occurs.

jackschmidt:
Fix #2117 by removing the 'return -1;' before the original body of the conditional.


matrix:
Hi, it's a duplicate of #2100

jsteube:
Thanks for the PR! Was merged with #2100

PR #2120, opened 2019-07-18
nsec3 m 8300: salt can be empty

### LINKED ISSUE #2119: Mode 8300, salt can be empty
In mode 8300, nsec3, it is permissible for the salt to be empty. This is currently done by the top-level-domain ".fun" for example. The fix is to change the min length of that token to 0.

jackschmidt:
Fixes #2119 -- the salt for an nsec3 record can be empty according to the standard, and this does occur. Change the minimum token length to 0 to allow this.

### DIFF
--- a/src/modules/module_08300.c
+++ b/src/modules/module_08300.c
@@ -61,7 +61,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
   token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH;
 
   token.sep[2]     = ':';
-  token.len_min[2] = 1;
+  token.len_min[2] = 0;
   token.len_max[2] = 32;
   token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH;
 


jsteube:
Thanks for the PR. Did you try with a real example hash with salt length zero if it works? Can you also update tools/test_modules/m08300.pm?

jackschmidt:
Tested: Yes, it works with the top level domains .fun .ink .ooo and .space

Update tests: I'll look into that

jackschmidt:
For the tests, I think no changes are needed as the tests are not very thorough. I could add more tests as a separate PR.

Test: Salt is hard coded as a random 8 digit numeric string
Hashcat: Salt is a [0 now, 2 before] to 32 hexit hex string
RFC: Salt is 0 to 255 byte string
Reality: Most salts are 4 or 8 bytes

Test: Iterations is hard coded to be 1
Hashcat: Iterations can be 1 to 999999
RFC: Iterations can be 0 to 65535, but values larger than 150 may be rejected as invalid so are not used
Reality: This is sha1, iteration counts this small are misguided. Most domains are set to 1.

jsteube:
It's fine. Just wanted to make sure it works with zero (but you already confirmed).

PR #2125, opened 2019-07-20
Added hash-mode: sha256(sha256_bin(pass))

matrix:
inspired by 
https://hashcat.net/forum/thread-8298.html


jsteube:
two comments:

- Please update hash-mode number to next one free, xx70 can't be used they are reserved for hmac variants. 
- That this mode will be very simar to sha256(sha256(p).s) which I saw you also send in a PR. For those modes there's special handling, see mode 2710 (and 2711 respectively). The 2710 and 2711 modes use the general 2710 kernel. The 2710 is the general mode ha256(sha256(p).s) and the last number is a specific application using this mode.  This way we can add constraints like salt length, password length, etc from the application itself into the module.

jsteube:
One more comment:

- Please do not use _hex. In case we omit that, we always refer to _hex (as in all other modules).

jsteube:
I just saw that this module is not sha256 (hex) twice. Therefore this need to be a standalone mode. The following action is needed to merge:

- Rename the hashmode to the next one free
- Rename the mode from sha256_hex(sha256_bin(pass)) to sha256(sha256_bin(pass))
- Fix conflicts with docs/*

Issue #293, opened 2016-04-10
Request new algorithm md5(sha1($password).md5($password).sha1($password))


Maxus777:
I really need the support of this type of hash: md5(sha1($password).md5($password).sha1($password))

SunShop Shopping Cart & Ecommerce Software

here is some examples

> 379ffc4ddd1498db21fcdd8a7b0a8e3a:agape
> 04196603ca6c89e6fef84490d7524ebb:Lemons18
> 9c772a3165aa0e52510bec091b5eccbf:M1n1b1k3
> 4fc0b4139d2a1da24e0a101495f99acd:magnolia101
> 066c5957a141c2155341c637675c4329:80Monkey10

Fist0urs:
Where does this format is used? 

Anw, it is easy to implement this kind of format and there are plenty of new algorithms examples. You can pick one and implement it :)

Maxus777:
> Where does this format is used? 

SunShop Shopping Cart & Ecommerce Software

> pick one and implement it

Friend, I am not a programmer. I have no idea how to do it :)

Maxus777:
$ 50 a person, who will do what I need

ghost:
Hey ! Did yo found a solution ? thx

Maxus777:
No :(

jsteube:
The requested algorithm was added to hashcat. You can either wait for release version or use the binary beta from https://hashcat.net/beta or compile from GitHub master source yourself. Thanks for the request!

PR #2127, opened 2019-07-25
Added hash-mode: md5(sha1(pass).md5(pass).sha1(pass))

matrix:
Details in #293


jsteube:
Please use the next one free hash-mode number.

matrix:
done

PR #2128, opened 2019-07-25
Added hash-mode: AuthMe - sha256(sha256(pass).salt)

matrix:
Details in #1486 and #1939


jsteube:
The mode 19400 is reserved for something else. Please use 19800 instead.

matrix:
19800 is used

```
static const char *HASH_NAME      = "Kerberos 5, etype 17, Pre-Auth";
static const u64   KERN_TYPE      = 19800;
```

jsteube:
Ok, then the next one free would be 20700. Not sure if you have used this one with the other PR. If yes, please use the next one which is free after 20600.

matrix:
done

jsteube:
I can't merge anymore, since I've merged #2129 which now creates a conflict in docs/changes.txt and docs/credits.txt. Can you please remove all entries in these files in all open PR and create another PR with only entries to those files? It will be easier to handle.

matrix:
done

jsteube:
This is almost ready to merge, but there are still some problems.

The module is not a raw mode, because "AuthMe" is some named service. The kernel, however, can act as a raw one. The same situation as with Vbull, just it is SHA256 in this case. The way to deal with this is the following:

- First to rename the kernel to 20710, which flags it as salted. If there would be sha256(sha256(p)) then this would be 20700.
- Rename the existing module to 20711. The last number one flags it as application-specific. Make it use the kernel 20710 inside the module. Also, update the hash category to EAS or whatever AuthMe is in real
- Create a new module 20710. This one is the real generic one. Remove constraints with salt length and make it use kernel 20710 as well.

matrix:
I should have done everything you described, please re-check it ;)
Thanks

docs/readme.txt
```diff
@@ -92,6 +92,7 @@ NVIDIA GPUs require "NVIDIA Driver" (418.56 or later) and "CUDA Toolkit" (10.1 o
 - sha512($salt.$pass)
 - sha512(utf16le($pass).$salt)
 - sha512($salt.utf16le($pass))
+- AuthMe - sha256(sha256($pass).$salt)
```
jsteube: We want two entries:

- sha256(sha256($pass).$salt)
- AutheMe sha256

src/modules/module_20710.c
```diff
@@ -0,0 +1,279 @@
...
+#include "memory.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 3;
+static const u32   DGST_POS1      = 7;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 6;
+static const u32   DGST_SIZE      = DGST_SIZE_4_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_EAS;
+static const char *HASH_NAME      = "AuthMe - sha256(sha256($pass).$salt)";
```
jsteube: This one should be 

```
static const char *HASH_NAME      = "sha256(sha256($pass).$salt)";
```

src/modules/module_20710.c
```diff
@@ -0,0 +1,279 @@
...
+#include "shared.h"
+#include "memory.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 3;
+static const u32   DGST_POS1      = 7;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 6;
+static const u32   DGST_SIZE      = DGST_SIZE_4_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_EAS;
```
jsteube: This one needs to be"

```
static const u32   HASH_CATEGORY  = HASH_CATEGORY_RAW_HASH_SALTED;
```

src/modules/module_20710.c
```diff
@@ -0,0 +1,279 @@
...
+const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
+
+u32 module_pw_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  const u32 pw_max = 32;
+
+  return pw_max;
+}
+
+static const char *SIGNATURE_AUTHME = "$SHA$";
```
jsteube: Raw modes do not have a signature

src/modules/module_20710.c
```diff
@@ -0,0 +1,279 @@
...
+
+int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
+{
+  const u32 *digest = (const u32 *) digest_buf;
+
+  int line_len = 0;
+
+  // signature
+
+  line_buf[line_len++] = '$';
```
jsteube: Same here

src/modules/module_20711.c
```diff
@@ -0,0 +1,293 @@
...
+#include "memory.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 3;
+static const u32   DGST_POS1      = 7;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 6;
+static const u32   DGST_SIZE      = DGST_SIZE_4_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_EAS;
+static const char *HASH_NAME      = "AuthMe - sha256(sha256($pass).$salt)";
```
jsteube: Rename:

```
static const char *HASH_NAME      = "AuthMe sha256";
```

tools/test_modules/m20700.pm
```diff
@@ -0,0 +1,48 @@
...
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::SHA  qw (sha256_hex);
+
+sub module_constraints { [[1, 31], [16, 16], [-1, -1], [-1, -1], [-1, -1]] }
```
jsteube: I think constraints for generic mode should be more open. Also note that for this mode you have pure and optimized kernels, therefore you should make use of all paramters not just for pure kernels.

jsteube:
I've commented on some stuff. I also think the unit test .pm for 20711 is missing.

tools/test_modules/m20700.pm
matrix: like this?

`sub module_constraints { [[1, 31], [16, 16], [1, 31], [16, 16], [-1, -1]] }`

matrix: another thing, I need create two test modules right?
I need to rename this m20700.pm to m20711.pm and create the m20710 ?

tools/test_modules/m20711.pm
```diff
@@ -0,0 +1,48 @@
...
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::SHA  qw (sha256_hex);
+
+sub module_constraints { [[1, 31], [16, 16], [1, 31], [16, 16], [-1, -1]] }
```
jsteube: Almost there. Please change this restrictions to:

```
sub module_constraints { [[0, 256], [16, 16], [0, 55], [16, 16], [-1, -1]] }
```

src/modules/module_20711.c
```diff
@@ -0,0 +1,293 @@
...
+}
+
+u32 module_salt_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  const u32 salt_max = 16;
+
+  return salt_max;
+}
+
+u32 module_pw_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
```
jsteube: This configuration to limit pw_max is not needed. The kernel supports the normal operation mode (32 for -a 0, 256 for -a 3, etc). Please remove completely.

src/modules/module_20710.c
```diff
@@ -0,0 +1,258 @@
...
+u64         module_opts_type      (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return OPTS_TYPE;       }
+u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return SALT_TYPE;       }
+const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
+const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
+
+u32 module_pw_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
```
jsteube: This configuration to limit pw_max is not needed. The kernel supports the normal operation mode (32 for -a 0, 256 for -a 3, etc). Please remove completely.

tools/test_modules/m20710.pm
```diff
@@ -0,0 +1,46 @@
...
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::SHA  qw (sha256_hex);
+
+sub module_constraints { [[1, 31], [16, 16], [1, 31], [16, 16], [-1, -1]] }
```
jsteube: Almost there. Please change this restrictions to:

```
sub module_constraints { [[0, 256], [0, 256], [0, 55], [16, 16], [-1, -1]] }
```

I was wondering why the salt has be exactly length 16. In a generic mode, this should be more variable. Can you change the kernel? The pure kernel supports it already, it's just the optimized kernel.

matrix:
done, please re-check.
Thanks

Issue #1572, opened 2018-05-13
BitShares v0.x wallet password hash type: sha512_hex(sha512_binary(password))


Cortexelus:
### Application 
[BitShares](https://en.wikipedia.org/wiki/BitShares) is a blockchain-based decentralized asset exchange. It started in 2014 and today has a market cap of $641 million. BitShares uses the cryptocurrency token BTS. Like other cryptocurrencies, BitShares users use a secure digital wallet to 'store', send, and receive BTS. However, if a user forgets their BitShares wallet password, they lose access to their BitShares funds. It is common to see these users ask in forums for help recovering their lost wallet password. It is easy to extract the `password_checksum` from a locked-out wallet. A user could give their `password_checksum` to a password recovery service, which may be able to crack it and recover the password. 

Note: cracking the `password_checksum` is not sufficient to steal that user's cryptocurrency. One also needs the `encrypted_private_key` associated with the wallet. A user can keep their `encrypted_private_key` secret, while asking a password recovery service to crack the `password_checksum`, without risking theft. 

### Hash function
```
password_checksum = sha512_hex(sha512_binary(password))
```

Note: this is for the [older version](https://github.com/bitshares/bitshares1-core/releases) of the BitShares client, up to and including v0.9.3c. The BitShares-2.0 client wallet encrypts passwords differently.

### Steps to extract the password checksum 
1. Download [BitShares client v0.9.3c](https://github.com/bitshares/bitshares1-core/releases)
2. Create a wallet and enter a `password`
3. Export wallet to JSON file
4. Read `password_checksum` from JSON file

### Example
```
password123
=>
b335d77abf42f6da1b6d0864129fa176c7c499b70c389b273e719473e5f029f54b9b8914e96a6585db2f16c987e6b988e78a0e497bbe8bc88d62df0eb81b7010
```

###
Thanks

jsteube:
Something's strange here!

While your example from JSON works, bit it seems to be using a client from Sep 22, 2015, but this is the latest version 0 (or 1?) on GitHub (did it move somewhere else?).

Now if you take a look at JtR which support cracking BitShares.Setup.2.0.180115.exe it access a sqlite database. The KDF is the same, but no more password_checksum field. It now needs an additional AES descrypt (plus it became salted) this way to find out if the password is correct.

The question is now what's the right algorithm and if there's actually 3 different algorithms (including the one from the backup files).

Cortexelus:
Yes this is for the older version. 

[...]

PR #2131, opened 2019-07-26
Added hash-mode: BitShares v0.x - sha512(sha512(pass))

matrix:
Details here #1572


jsteube:

jsteube:
In -a 3 mode with pure kernel I'm getting this error:

```
main_kernel(65): error: argument of type "sha512_ctx_t *" is incompatible with parameter of type "sha512_ctx_vector_t *"
```

The optimized kernel works fine.

matrix:
I pushed the fix but I don't see the new patch here ...

matrix:
https://github.com/matrix/hashcat/commit/d045c0f62a837bd741cf47d4ee6d6f0f72cfab72

matrix:
done

Issue #1446, opened 2017-11-14
Support for sha1(md5($pass.$salt)) new algorithm


anmakar111:
Add support new algo sha1(md5($pass.$salt))

philsmd:
you could just use -m 4700 = sha1(md5($pass)) and append the salt with rules etc. This of course only works sufficiently if the hash list is small (ideally only one hash is used).

If you really want that this new hashing algorithm gets implemented you need to provide much more information.
Important information that is required would be: where is this hash type used in? are there any limitations of salt length/password length etc

jsteube:
The requested algorithm was added to hashcat. You can either wait for release version or use the binary beta from https://hashcat.net/beta or compile from GitHub master source yourself. Thanks for the request!

PR #2132, opened 2019-07-27
Added hash-mode: sha1(md5(pass.salt))

matrix:
Details here: #1446


jsteube:

matrix:
done

Issue #490, opened 2016-09-08
Add new algorithm: md5(sha1($salt).md5($pass))


blandyuk:
Seen this algorithm come up a few times now with large lists so would be very handy to add to hashcat. Cheers.

blandyuk:
Example hash:salt:pass for testing:

33d00ac42f7f141469ed7d2147d5574a:02f76533cfc1fdcdf20:sommer123

hops:
This would be fairly easy to add.
Do you know what software is using this scheme?
Is there a min/max length of the salt?

blandyuk:
Originally seen on HashKiller forum here:

https://forum.hashkiller.co.uk/topic-view.aspx?t=13829

Not sure on any software name but the salt seems to be always 19 hex chars as I have a private list of 600k+ which always has a 19 char salt. I can beta test if you can compile to .exe thanks.

TreyJenkins:
I would also appreciate this algorithm

jsteube:
The requested algorithm was added to hashcat. You can either wait for release version or use the binary beta from https://hashcat.net/beta or compile from GitHub master source yourself. Thanks for the request!

PR #2133, opened 2019-07-27
Added hash-mode: md5(sha1(salt).md5(pass))

matrix:
Details here #490


jsteube:

matrix:
done

Issue #1980, opened 2019-04-07
Support for md5($salt.sha1($salt.$pass)


peterstavrou:
Can you please add support for md5($salt.sha1($salt.$pass)) in the binaries version of the next update?
{EHF} seems to have already created a kernal for this.
https://forum.hashkiller.co.uk/topic-view.aspx?t=23819&m=195585#195585

philsmd:
what do you mean by "created a kernel" ? where is this kernel ? Added to which software ? to hashcat ?
Where is the code?

peterstavrou:
For hashcat 5.1
Sorry I should have included [direct link](https://drive.google.com/file/d/1Y34ULJC52ndD-VqzQUFPTJCucPzJCIiW/view?usp=sharing).

philsmd:
interesting, but why didn't the author send a PR. it's probably not the best idea to just copy the code from a google drive link and add it as if it was our code... we probably either need to implement it from scratch or ask the author to send a pull request with the permission to integrate it.

Where is this code used?
Answers for these questions would be interesting https://hashcat.net/faq#i_want_to_request_some_new_algorithms_or_features_how_can_i_accomplish_this . e.g. if there are some limits and where it is used

jsteube:
The requested algorithm was added to hashcat. You can either wait for release version or use the binary beta from https://hashcat.net/beta or compile from GitHub master source yourself. Thanks for the request!

PR #2134, opened 2019-07-27
Added hash-mode: md5(salt.sha1(salt.pass))

matrix:
Details here #1980


jsteube:

matrix:
done

jsteube:
The -a 3 kernel fails with this error:

```
main_kernel(119): error: argument of type "md5_ctx_vector_t *" is incompatible with parameter of type "sha1_ctx_vector_t *"

main_kernel(119): error: argument of type "md5_ctx_t *" is incompatible with parameter of type "sha1_ctx_t *"
```

Issue #2136, opened 2019-07-31
Mode 4710 fails on CPU


jsteube:
Hey @matrix please check this:

Works on GPU:

```
root@ht:~/hashcat# ./hashcat -b -m 4710 -D 2
hashcat (v5.1.0-1299-g41dc5035) starting in benchmark mode...

Benchmarking uses hand-optimized kernel code by default.
You can use it in your cracking session by setting the -O option.
Note: Using optimized kernel code limits the maximum supported password length.
To disable the optimized kernel code in benchmark mode, use the -w option.

/root/git/hashcat/OpenCL/m04710_a3-optimized.cl: Optimized kernel requested but not needed - falling back to pure kernel
CUDA API (CUDA 10.1)
====================
* Device #1: GeForce GTX 980, 4041 MB, 16MCU

OpenCL API (OpenCL 2.1 LINUX) - Platform #1 [Intel(R) Corporation]
==================================================================
* Device #2: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz, skipped

OpenCL API (OpenCL 1.2 CUDA 10.1.120) - Platform #2 [NVIDIA Corporation]
========================================================================
* Device #3: GeForce GTX 980, skipped

Benchmark relevant options:
===========================
* --opencl-device-types=2
* --optimized-kernel-enable

Hashmode: 4710 - sha1(md5($pass).$salt)

Speed.#1.........:  1767.0 MH/s (75.34ms) @ Accel:16 Loops:512 Thr:1024 Vec:1

Started: Wed Jul 31 14:00:37 2019
Stopped: Wed Jul 31 14:00:46 2019
```

Fails on CPU:

```
root@ht:~/hashcat# ./hashcat -b -m 4710 -D 1
hashcat (v5.1.0-1299-g41dc5035) starting in benchmark mode...

Benchmarking uses hand-optimized kernel code by default.
You can use it in your cracking session by setting the -O option.
Note: Using optimized kernel code limits the maximum supported password length.
To disable the optimized kernel code in benchmark mode, use the -w option.

/root/git/hashcat/OpenCL/m04710_a3-optimized.cl: Optimized kernel requested but not needed - falling back to pure kernel
CUDA API (CUDA 10.1)
====================
* Device #1: GeForce GTX 980, skipped

OpenCL API (OpenCL 2.1 LINUX) - Platform #1 [Intel(R) Corporation]
==================================================================
[...]

Issue #431, opened 2016-07-11
support for sha1(md5($pass).$salt) hashes


Carlos7777:
is it possible to add support for this algorithm:

sha1(md5($pass).$salt)

4780273b56d495904a225131211ea20c9dccc567:bEpY4bEr:13lucky
00f7f21a49313d75b36f451e11606e148106637d:rVzYynKa:047925
003e0b6c2d4f9b6e1d1adc650192679402c725a5:RJdPbqfz:901603
0054c29fff4efa9c4085fa71dfd29aec59cb8f24:u9Z.PVVW:890926
005bad6fe73622226e86780a40bd1926ffedbcfb:rMwA4KNU:250273
0092930759fba02263e1ad81ec2dda46c1bc99ca:nEjvYaGb:311205
47655fd371071d9a5c1c3c2030c6caa90fe96e14:n.=V4rY=:nirvana123

Ins1ghtLabs:
+1

forhuan:
+1

jsteube:
Support added with PR https://github.com/hashcat/hashcat/pull/2129

PR #2137, opened 2019-07-31
fix bug in 4710 a3 kernel

matrix:
Fix issue #2136


src/modules/module_04710.c
```diff
@@ -46,6 +46,20 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig,
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+u32 module_salt_min (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  const u32 salt_min = 8;
```
jsteube, line 51: Why have such a constraint in a generic hash module?

matrix: Because the test hashes that I found here #431 the salt is always 8

PR #2142, opened 2019-08-01
Avoid -Waddress-of-packed-member warnings

phymod0:
Getting these warnings in GCC 9.1.0

As far as I checked, my changes didn't modify the compiled binaries


jsteube:
Many Thanks!

PR #2145, opened 2019-08-04
sh

neheb:


jsteube:
Please see here, something broke after merging PR:

```
root@ht:~/hashcat# tools/test.sh -t multi -a all -m all -V 1
ls: cannot access '/root/hashcat/tools/test_modules/*.pm': No such file or directory
grep: /root/hashcat/tools/../src/modules/module_*.c: No such file or directory
grep: /root/hashcat/tools/../src/modules/module_*.c: No such file or directory
[ test_1564992399 ] > Init test for hash type 6211.
Could not load test module: m06211.pm
```

Can you fix?

neheb:
Bah some quoting bug. I will look later today.

Issue #1736, opened 2018-10-28
Feature Request - New-Algo - SolarWinds


mubix:
As described here: https://www.atredis.com/blog/2018/10/24/fun-with-the-solarwinds-orion-platform

Specific quote from the post:
> Orion password hashing is a variant of a salted SHA512 hash. The hash is computed by first generating a salt that consists of the lowercase username. If the salt is less than 8 bytes long, it is appended with bytes from the string "1244352345234" until it is 8 bytes. For example, the salt for username "ADMIN" would become "admin124", while the salt for "Bo" would become "bo124435". Once the salt has been calculated, a RFC2898 PBKFD2 is generated using the default iteration count of 1000 and the SHA1 hash algorithm. Finally, a SHA512 hash of the PBKDF2 output is taken and encoded using Base64. It doesn't appear that any existing tools support cracking passwords in this format, but Hashcat comes close with PBKDF2-HMAC-SHA1(sha1:1000) support, and is only missing the final call to SHA512(). This hashing function has been implemented in the Ruby script [hash-password.rb](https://github.com/atredispartners/solarwinds-orion-cryptography/blob/master/hash-password.rb).

- `admin` username with blank password is:
```
/+PA4Zck3arkLA7iwWIugnAEoq4ocRsYjF7lzgQWvJc+pepPz2a5z/L1Pz3c366Y/CasJIa7enKFDPJCWNiKRg==
```

roycewilliams:
JtR reference: https://github.com/magnumripper/JohnTheRipper/pull/3449

jsteube:
The requested algorithm was added to hashcat. You can either wait for release version or use the binary beta from https://hashcat.net/beta or compile from GitHub master source yourself. Thanks for the request!

PR #2146, opened 2019-08-04
Added hash-mode 21500 - SHA512(PBKDF2-HMAC-SHA1)

matrix:
Hi,
I wrote this in order to support the "SolarWinds Orion" algo, details here #1736 

Before making this new hash-mode completely generic we must decide whether to adopt the format chosen by John The Ripper or not.

https://github.com/magnumripper/JohnTheRipper/blob/ae24a410baac45bb36884d793c429adeb7197336/src/solarwinds_common_plug.c#L14

Let me know :)

### DIFF
--- a/OpenCL/m21500-pure.cl
+++ b/OpenCL/m21500-pure.cl
@@ -30,7 +30,7 @@ typedef struct pbkdf2_sha1_tmp
 
 typedef struct pbkdf2_sha1
 {
-  u32 salt_buf[4];
+  u32 salt_buf[5];
 
 } pbkdf2_sha1_t;
 
--- a/OpenCL/m21500-pure.cl
+++ b/OpenCL/m21500-pure.cl
@@ -30,7 +30,7 @@ typedef struct pbkdf2_sha1_tmp
 
 typedef struct pbkdf2_sha1
 {
-  u32 salt_buf[64];
+  u32 salt_buf[4];
 
 } pbkdf2_sha1_t;
 
--- a/OpenCL/m21500-pure.cl
+++ b/OpenCL/m21500-pure.cl
@@ -30,7 +30,7 @@ typedef struct pbkdf2_sha1_tmp
 
 typedef struct pbkdf2_sha1
 {
-  u32 salt_buf[5];
+  u32 salt_buf[64];
 
 } pbkdf2_sha1_t;
 
@@ -243,7 +243,7 @@ KERNEL_FQ void m21500_comp (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1
 
   sha512_init (&ctx);
 
-  sha512_update (&ctx, (const u32 *)tmps[gid].out, 1024);
+  sha512_update_global (&ctx, tmps[gid].out, 1024);
 
   sha512_final (&ctx);
 
--- a/OpenCL/m21500-pure.cl
+++ b/OpenCL/m21500-pure.cl
@@ -1,0 +1,260 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha1.cl"
+#include "inc_hash_sha512.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+typedef struct pbkdf2_sha1_tmp
+{
+  u32  ipad[5];
+  u32  opad[5];
+
+  u32  dgst[260];
+  u32  out[260];
+
+} pbkdf2_sha1_tmp_t;
+
+typedef struct pbkdf2_sha1
+{
+  u32 salt_buf[64];
+
+} pbkdf2_sha1_t;
+
+DECLSPEC void hmac_sha1_run_V (u32x *w0, u32x *w1, u32x *w2, u32x *w3, u32x *ipad, u32x *opad, u32x *digest)
+{
+  digest[0] = ipad[0];
+  digest[1] = ipad[1];
+  digest[2] = ipad[2];
+  digest[3] = ipad[3];
+  digest[4] = ipad[4];
+
+  sha1_transform_vector (w0, w1, w2, w3, digest);
+
+  w0[0] = digest[0];
+  w0[1] = digest[1];
+  w0[2] = digest[2];
+  w0[3] = digest[3];
+  w1[0] = digest[4];
+  w1[1] = 0x80000000;
+  w1[2] = 0;
+  w1[3] = 0;
+  w2[0] = 0;
+  w2[1] = 0;
+  w2[2] = 0;
+  w2[3] = 0;
+  w3[0] = 0;
+  w3[1] = 0;
+  w3[2] = 0;
+  w3[3] = (64 + 20) * 8;
+
+  digest[0] = opad[0];
+  digest[1] = opad[1];
+  digest[2] = opad[2];
+  digest[3] = opad[3];
+  digest[4] = opad[4];
+
+  sha1_transform_vector (w0, w1, w2, w3, digest);
+}
+
+KERNEL_FQ void m21500_init (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1_t))
+{
+  /**
+   * base
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= gid_max) return;
+
+  sha1_hmac_ctx_t sha1_hmac_ctx;
+
+  sha1_hmac_init_global_swap (&sha1_hmac_ctx, pws[gid].i, pws[gid].pw_len);
+
+  tmps[gid].ipad[0] = sha1_hmac_ctx.ipad.h[0];
+  tmps[gid].ipad[1] = sha1_hmac_ctx.ipad.h[1];
+  tmps[gid].ipad[2] = sha1_hmac_ctx.ipad.h[2];
+  tmps[gid].ipad[3] = sha1_hmac_ctx.ipad.h[3];
+  tmps[gid].ipad[4] = sha1_hmac_ctx.ipad.h[4];
+
+  tmps[gid].opad[0] = sha1_hmac_ctx.opad.h[0];
+  tmps[gid].opad[1] = sha1_hmac_ctx.opad.h[1];
+  tmps[gid].opad[2] = sha1_hmac_ctx.opad.h[2];
+  tmps[gid].opad[3] = sha1_hmac_ctx.opad.h[3];
+  tmps[gid].opad[4] = sha1_hmac_ctx.opad.h[4];
+
+  sha1_hmac_update_global_swap (&sha1_hmac_ctx, esalt_bufs[digests_offset].salt_buf, salt_bufs[salt_pos].salt_len);
+
+  for (u32 i = 0, j = 1; i < 256; i += 5, j += 1)
+  {
+    sha1_hmac_ctx_t sha1_hmac_ctx2 = sha1_hmac_ctx;
+
+    u32 w0[4];
+    u32 w1[4];
+    u32 w2[4];
+    u32 w3[4];
+
+    w0[0] = j;
+    w0[1] = 0;
+    w0[2] = 0;
+    w0[3] = 0;
+    w1[0] = 0;
+    w1[1] = 0;
+    w1[2] = 0;
+    w1[3] = 0;
+    w2[0] = 0;
+    w2[1] = 0;
+    w2[2] = 0;
+    w2[3] = 0;
+    w3[0] = 0;
+    w3[1] = 0;
+    w3[2] = 0;
+    w3[3] = 0;
+
+    sha1_hmac_update_64 (&sha1_hmac_ctx2, w0, w1, w2, w3, 4);
+
+    sha1_hmac_final (&sha1_hmac_ctx2);
+
+    tmps[gid].dgst[i + 0] = sha1_hmac_ctx2.opad.h[0];
+    tmps[gid].dgst[i + 1] = sha1_hmac_ctx2.opad.h[1];
+    tmps[gid].dgst[i + 2] = sha1_hmac_ctx2.opad.h[2];
+    tmps[gid].dgst[i + 3] = sha1_hmac_ctx2.opad.h[3];
+    tmps[gid].dgst[i + 4] = sha1_hmac_ctx2.opad.h[4];
+
+    tmps[gid].out[i + 0] = tmps[gid].dgst[i + 0];
+    tmps[gid].out[i + 1] = tmps[gid].dgst[i + 1];
+    tmps[gid].out[i + 2] = tmps[gid].dgst[i + 2];
+    tmps[gid].out[i + 3] = tmps[gid].dgst[i + 3];
+    tmps[gid].out[i + 4] = tmps[gid].dgst[i + 4];
+  }
+}
+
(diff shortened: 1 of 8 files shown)


roycewilliams:
Is it really specific to SolarWinds Orion, or do other platforms use the same algo?

matrix:
> Is it really specific to SolarWinds Orion, or do other platforms use the same algo?

never seen anywhere else

tools/test_modules/m21500.pm
```diff
@@ -0,0 +1,82 @@
...
+
+use strict;
+use warnings;
+
+use MIME::Base64 qw (encode_base64 decode_base64);
+use Crypt::PBKDF2;
+use Digest::SHA qw (sha512);
+use Encode;
+
+sub module_constraints { [[0, 256], [8, 16], [-1, -1], [-1, -1], [-1, -1]] }
```
jsteube, line 16: From what I can see with JtR mode is that the length of the hash is always the same. But here you use dynamic salt length 8 - 16. How does that match together?

src/modules/module_21500.c
```diff
@@ -0,0 +1,274 @@
...
+static const u64   KERN_TYPE      = 21500;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_USES_BITS_64
+                                  | OPTI_TYPE_SLOW_HASH_SIMD_LOOP;
+static const u64   OPTS_TYPE      = OPTS_TYPE_PT_GENERATE_LE
+                                  | OPTS_TYPE_ST_BASE64
+                                  | OPTS_TYPE_HASH_COPY;
+static const u32   SALT_TYPE      = SALT_TYPE_EMBEDDED;
+static const char *ST_PASS        = "hashcat";
+static const char *ST_HASH        = "sha1:1000:aGFzaGNhdDE=:Dnz04T9ptYINnj9mckXTuk80w6f1vkZLwv02HiufklgnDw53BU1bgn4gsHia/Q6kxYSqK4PDbxRMwa6HB5KA/w==";
```
jsteube: Can you please stick to the original format from JtR.

OpenCL/m21500-pure.cl
```diff
@@ -0,0 +1,260 @@
...
+  u32  opad[5];
+
+  u32  dgst[260];
+  u32  out[260];
+
+} pbkdf2_sha1_tmp_t;
+
+typedef struct pbkdf2_sha1
+{
+  u32 salt_buf[64];
```
jsteube, line 33: This is probably oversized if salt is always length 8, salt_buf[2] should be ok.

OpenCL/m21500-pure.cl
```diff
@@ -0,0 +1,260 @@
...
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+typedef struct pbkdf2_sha1_tmp
+{
+  u32  ipad[5];
+  u32  opad[5];
+
+  u32  dgst[260];
```
jsteube, line 26: Any reason why not to stick to 256?

jsteube:
I've added some comments. This mode should not be generic, there's too much specific stuff going on, especially with the 1024 byte output size which is kind of huge/unique.

src/modules/module_21500.c
```diff
@@ -0,0 +1,274 @@
...
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32   DGST_POS0      = 14;
+static const u32   DGST_POS1      = 15;
+static const u32   DGST_POS2      = 6;
+static const u32   DGST_POS3      = 7;
+static const u32   DGST_SIZE      = DGST_SIZE_8_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_GENERIC_KDF;
+static const char *HASH_NAME      = "SHA512(PBKDF2-HMAC-SHA1)";
```
jsteube: Please use 'SolarWinds Orion'

tools/test_modules/m21500.pm
matrix: Because in the details is written:

> The hash is computed by first generating a salt that consists of the lowercase username. If the salt is less than 8 bytes long, it is appended with bytes from the string "1244352345234" until it is 8 bytes. For example, the salt for username "ADMIN" would become "admin124", while the salt for "Bo" would become "bo124435".

I don't see any reference to fixed len of salt. The "username" length is dynamic from my point of view

OpenCL/m21500-pure.cl
matrix: yes, without the digest buffer is not complete

tools/test_modules/m21500.pm
jsteube: ok

OpenCL/m21500-pure.cl
jsteube: Sorry, what do you mean with "without digest buffer is not complete"?

matrix: I mean than if you set 256 instead of 260 it don't work

tools/test_modules/m21500.pm
matrix: jtr handle also the padding of the salt/username but I wanted to hear your opinion first :)

OpenCL/m21500-pure.cl
matrix: what am I doing here? I leave it as it is or should I change it?

src/modules/module_21500.c
matrix: HASH_CATEGORY_GENERIC_KDF or EAS ?

jsteube:
I think the unit test still outputs the hash in the old format, can you please update?

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32   DGST_POS0      = 14;
+static const u32   DGST_POS1      = 15;
+static const u32   DGST_POS2      = 6;
+static const u32   DGST_POS3      = 7;
+static const u32   DGST_SIZE      = DGST_SIZE_8_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_GENERIC_KDF;
```
jsteube: I'm not really sure what SolarWinds actually is, but I think HASH_CATEGORY_EAS fits better than HASH_CATEGORY_GENERIC_KDF. Please update.

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+  u32  opad[5];
+
+  u32  dgst[260];
+  u32  out[260];
+
+} pbkdf2_sha1_tmp_t;
+
+typedef struct pbkdf2_sha1
+{
+  u32 salt_buf[64];
```
jsteube, line 61: I think you changed this in the kernel but not in the module. In Multihash situations, this can create evil out of bound reads

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+  token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_BASE64A;
+
+  const int rc_tokenizer = input_tokenizer ((const u8 *) line_buf, line_len, &token);
+
+  if (rc_tokenizer != PARSER_OK) return (rc_tokenizer);
+
+  // iter
+
+  const u32 iter = 1000;
```
jsteube: Typically, for hash-mode specific fixed iteration count, we create a variable for this. It's not really required, just convention. See here:

```
root@ht:~/hashcat# grep ^static.*ROUNDS_ src/modules/* 
src/modules/module_00500.c:static const u32   ROUNDS_MD5CRYPT    = 1000;
src/modules/module_00501.c:static const u32   ROUNDS_MD5CRYPT    = 1000;
src/modules/module_01600.c:static const u32   ROUNDS_MD5CRYPT    = 1000;
src/modules/module_01800.c:static const u32   ROUNDS_SHA512CRYPT    = 5000;
src/modules/module_02500.c:static const u32 ROUNDS_WPA_PBKDF2 = 4096;
src/modules/module_02501.c:static const u32 ROUNDS_WPA_PMK = 1;
src/modules/module_05800.c:static const int ROUNDS_ANDROIDPIN = 1024;
```

... and so on.

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+
+  salt->salt_iter = iter - 1;
+
+  // salt
+
+  char *salt_pos = (char *) token.buf[1];
+  int salt_len = token.len[1];
+
+  if (salt_len > 16) return (PARSER_SALT_LENGTH);
```
jsteube: This should be covered already with this line:

```
token.len_max[1] = 16;
```

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+
+  char custom_salt[17];
+
+  memset (custom_salt, 0, sizeof (custom_salt));
+
+  for (int i = 0; i < salt_len; i++, salt_pos++)
+  {
+    if (*salt_pos >= 'A' && *salt_pos <= 'Z') {
+      custom_salt[i] = *salt_pos + 32;
```
jsteube: What is this loop about? I couldn't find it in the original description of the algorithm. Why not using existing base64 decoder?

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+  {
+    strncat (custom_salt, SALT_PADDING, 8 - salt_len);
+    salt_len = 8;
+  }
+
+  salt_pos = custom_salt;
+
+  memcpy (pbkdf2_sha1->salt_buf, salt_pos, salt_len);
+
+  salt->salt_len = salt_len;
```
jsteube, line 164: FYI, if you are setting 16 (or even 20) bytes in the salt->salt_buf buffer, then you need to set salt->salt_len to 16 (or 20), too. The reason is that hashcat will only compare data of length salt->salt_len in order to detect duplicate salts. If you are OK with this you can leave it as it is.

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+
+  salt->salt_buf[0] = pbkdf2_sha1->salt_buf[0];
+  salt->salt_buf[1] = pbkdf2_sha1->salt_buf[1];
+  salt->salt_buf[2] = pbkdf2_sha1->salt_buf[2];
+  salt->salt_buf[3] = pbkdf2_sha1->salt_buf[3];
+  salt->salt_buf[4] = salt->salt_iter;
```
jsteube: Is the iteration in binary part of the salt? If not, why add it?

src/modules/module_21500.c
```diff
@@ -0,0 +1,307 @@
...
+
+  return (PARSER_OK);
+}
+
+int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
+{
+  const u64 *digest = (u64 *) digest_buf;
+
+  const pbkdf2_sha1_t *pbkdf2_sha1 = (const pbkdf2_sha1_t *) esalt_buf;
+
```
jsteube: Typically if you have any strange code to add in the decoder (like you did here):

```
custom_salt[i] = *salt_pos + 32;
```

we expect to do the opposite in the encoder. Why is not here?

tools/test_modules/m21500.pm
```diff
@@ -0,0 +1,82 @@
...
+    output_len => 1024
+  );
+
+  my $key = $kdf->PBKDF2 ($salt, $word);
+
+  my $key_b64 = encode_base64 (sha512($key), "");
+
+  my $salt_b64 = encode_base64 ($salt, "");
+
+  my $hash = sprintf ("sha1:%i:%s:%s", $iter, $salt_b64, $key_b64);
```
jsteube: Please update format to new format

src/modules/module_21500.c
matrix: this convert the salt (username) to lowercase, one byte at once.

> [...] The hash is computed by first generating a salt that consists of the lowercase username [...]

src/modules/module_21500.c
matrix: as described above it's about the lowercase conversion of salt (username). It must not be converted back in the encoder

src/Makefile
```diff
@@ -4,7 +4,7 @@
 ##
 
 SHARED                  := 0
-DEBUG                   := 0
+DEBUG                   := 1
```
jsteube: Some debug leftover

OpenCL/m21500-pure.cl
```diff
@@ -0,0 +1,260 @@
...
+
+  if (gid >= gid_max) return;
+
+  const u64 lid = get_local_id (0);
+
+  sha512_ctx_t ctx;
+
+  sha512_init (&ctx);
+
+  sha512_update (&ctx, (const u32 *)tmps[gid].out, 1024);
```
jsteube: I think we can switch this to sha512_update_global() in order to save the cast. 

Note: Whenever we need to cast something in kernels to access pure data it's a sign of doing something wrong.

src/modules/module_21500.c
```diff
@@ -0,0 +1,298 @@
...
+static const u32   DGST_SIZE      = DGST_SIZE_8_8;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_EAS;
+static const char *HASH_NAME      = "SolarWinds Orion";
+static const u64   KERN_TYPE      = 21500;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_USES_BITS_64
+                                  | OPTI_TYPE_SLOW_HASH_SIMD_LOOP;
+static const u64   OPTS_TYPE      = OPTS_TYPE_PT_GENERATE_LE
+                                  | OPTS_TYPE_ST_BASE64
+                                  | OPTS_TYPE_HASH_COPY;
```
jsteube, line 27: The OPTS_TYPE_HASH_COPY is not needed since you do provide a proper _encode function.

src/modules/module_21500.c
```diff
@@ -0,0 +1,298 @@
...
+
+  const u8 *salt_pos = token.buf[1];
+
+  int salt_len = token.len[1];
+
+  char custom_salt[17];
+
+  memset (custom_salt, 0, sizeof (custom_salt));
+
+  for (int i = 0; i < salt_len; i++, salt_pos++)
```
jsteube: Please replace with

void lowercase (u8 *buf, const size_t len);

from convert.h

src/modules/module_21500.c
```diff
@@ -0,0 +1,298 @@
...
+    }
+  }
+
+  if (salt_len < 8)
+  {
+    strncat (custom_salt, SALT_PADDING, 8 - salt_len);
+    salt_len = 8;
+  }
+
+  memcpy (pbkdf2_sha1->salt_buf, custom_salt, salt_len);
```
jsteube: I think this is not valid. You're copying the lowercase version of the username and if the hash gets cracked, the encoder will show the username but in lowercase only. When a cracked hash is displayed we want to have it in 1:1 the same output as the input was. 

The ideal way to deal with such a situation is to store the salt twice. One time in the original format, one time in a separate field in the esalt. The data which is actually used by the kernel should be stored into salt->salt_buf and then used from there in the kernel. By doing this it also becomes more clear what to use as the right salt->salt_len.

The original buffer can be set to pbkdf2_sha1->salt_buf along with the right pbkdf2_sha1->salt_len (currently missing).

jsteube:
Currently, the kernel fails to crack in multi hash mode. As soon as the salt buffer is fixed as described it should crack.

matrix:
> Currently, the kernel fails to crack in multi hash mode. As soon as the salt buffer is fixed as described it should crack.

It's strange because I tried it more than once before committing ...

PR #2155, opened 2019-08-06
test.sh: Fix bad quoting

neheb:
${*_MODES} should not be passed quoted. Each element must be processed individually.


neheb:
Closed in favor of https://github.com/hashcat/hashcat/pull/2156

PR #2159, opened 2019-08-10
tools.sh

neheb:


jsteube:
The changes from https://github.com/hashcat/hashcat/pull/2159/commits/0965e25943bedb13866d240430526d0baf8cdd0c are fine, however I cannot merge a single commit from a PR. The other commit https://github.com/hashcat/hashcat/pull/2159/commits/0dd4ffbd6b6022490fff08386d6425157b425419 can you please explain what's the reason to do this? I totally prefer for() over while() since all important info is in a single line, makes it much easier to read. Plus it matches hashcat coding style.

neheb:
One way is to go to the commit, append .patch to the URL, download and git am.

It's a shellcheck warning which is only really applicable with less capable shells. numeric for loops are a bash extension (actually they first appeared elsewhere).

jsteube:
OK, I didn't know. In this case, it actually makes sense to replace the for() with the while().

Issue #2072, opened 2019-06-21
web2py pbkdf2(1000,20,sha512) format conversion details 


blacktraffic:
I suspect this doesn't make sense to implement as a separate module - it's the same as 12100 bar encoding. But you might have different opinions. 

pbkdf2(1000,20,sha512)$a2a2ca127df6bc19$77bb5a3d129e2ce710daaefeefef8356c4c827ff

Conversion script to hashcat type 12100 below if that's easier to read.

```
#!/bin/python

#str(LazyCrypt(crypt, 'mysecret'))
#
#input:
#pbkdf2(1000,20,sha512)$a2a2ca127df6bc19$77bb5a3d129e2ce710daaefeefef8356c4c827ff
#
#output, and crack with -m 12100
#sha512:1000:YTJhMmNhMTI3ZGY2YmMxOQ==:d7taPRKeLOcQ2q7+7++DVsTIJ/8=:mysecret

import fileinput
import base64

for line in fileinput.input():
    line=line.rstrip()
    [a,b,c]=line.split('$',3)

    print('sha512:1000:'+base64.b64encode(b)+':'+base64.b64encode(c.decode("hex")))
```

Chick3nman:
It may be possible to add this conversion step to the parser internally and allow hashcat to do the conversion for the user without telling them. That said, i'm not sure what @jsteube wants to do about such things.

blacktraffic:
no worries either way, i just made a note of it as i couldn't find it on the web elsewhere. it doesn't seem to be particularly common.

jsteube:
That's what the ****x modes are for. Like -m 12001 is just a parser to kernel 12000 (PBKDF2-HMAC-SHA1), this could be the same for SHA512 based PBKDF2. So this totally makes sense to add. Feel free to add it as -m 12101. Only the parser is needed (src/modules/) and optional unit test (tools/test_modules/).

blacktraffic:
ok, i'll have a pop this weekend - i've been meaning to write a custom mode anyway, and this looks like an easy way to start.

jamieriden:
This works for the example I found - we came across this on an engagement, and don't have that much info about it unfortunately. 

I'll do a proper pull request soon-ish, but it's attached for now . 


[module_12101.c.txt](https://github.com/hashcat/hashcat/files/3489175/module_12101.c.txt)

blacktraffic:
There's a PR in there now - it works for the public example I found. sadly, not a lot of info about it.

[...]

PR #2165, opened 2019-08-15
second go at web2py pbkdf2-hmac-sha512 variant

blacktraffic:
I've tried to address the comments from the first PR... hopefully correctly :)

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -51,6 +51,7 @@
 - Added hash-mode: sha256(md5($pass))
 - Added hash-mode: sha256(sha256_bin(pass))
 - Added hash-mode: sha256(sha256($pass).$salt)
+- Added hash-mode: pbkdf2(1000,20,sha512)
 
 ##
 ## Bugs
--- a/docs/credits.txt
+++ b/docs/credits.txt
@@ -84,5 +84,8 @@ Rick "Minga" Redman and KoreLogic <mingakore@gmail.com> (@CrackMeIfYouCan)
 Brandon Chalk <brandon@casaba.com> (@brandoncasaba)
 * Kerberos Pre-Auth 17/18 kernel module, ported from @Fist0urs TGS kernel modules
 
+Jamie <jamie@blacktraffic.co.uk> (@pedantic_hacker)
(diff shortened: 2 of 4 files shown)


philsmd:
Looks almost perfect, but I have some further comments/recommendations:
1. docs/changes.txt and docs/readme.txt needs to be updated too, please do so
2. we can and should do a hash signature check, like in other hash formats, search for SIGNATURE in the modules. in this case it would be enough to match "pbkdf2" or even "pbkdf2(" at the beginning
3. this is some strange formatting, it's not consistent with other formattings, so let get rid of this:
  - https://github.com/hashcat/hashcat/pull/2165/files#diff-1718c964f3889b6da78225ee22553a8dR199
  - https://github.com/hashcat/hashcat/pull/2165/files#diff-1718c964f3889b6da78225ee22553a8dR213
4. I also see no check for "sha512" (strcmp () or something like this, this missing check could be kinda dangerous)
5. the "20" for digest length/size is hard-coded (https://github.com/hashcat/hashcat/pull/2165/files#diff-1718c964f3889b6da78225ee22553a8dR195). should we make this more dynamic/flexible ? should we also check for "20" in the input hash line ? I would say that we either make everything dynamic or everything hard-coded, but we should try to be consistent here, or?

blacktraffic:
ok, let me do an update. I agree entirely, but I'm slightly limited in that
I have only 1 or two examples of the hash - but I'll assume the '20' can
vary.

thanks!
 Jamie

philsmd:
sounds great.

btw, I just realized that the test modules are missing too. i.e. we would need a perl module named tools/test_modules/m12101.pm which is able to generate hashes and verify cracks.
This would be perfect to have because it would make sure that everything is working correctly because test.pl is able to interact with test.sh which is a test suite to test all hash modes with different inputs, attack modes, vector widths etc

of course it could be added later on, but it shouldn't be too difficult since we already have similar code for pbkdf2-based algos in the test_modules folder (just needs to be adapted to the "pbkdf2(...)" signature of the hash etc.
Thanks again. great work


btw: it's not mentioned here anywhere, but this is related to this (still open at the time of this writing) issue: #2072

PR #2181, opened 2019-09-05
Fixed bug in mode 9500 

Chick3nman:
Fixed issue causing hashes with the same salt to fail to crack.


jsteube:
Great fix, thanks!

PR #2187, opened 2019-09-11
third go at web2py pbkdf2-hmac-sha512

blacktraffic:
formatting fixes as suggested by phil, checks for "pbkdf", "20,sha512" strings (I can generalise if I have more examples, but don't have any different here for now), updated changes, credits, readme files.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -51,6 +51,7 @@
 - Added hash-mode: sha256(md5($pass))
 - Added hash-mode: sha256(sha256_bin(pass))
 - Added hash-mode: sha256(sha256($pass).$salt)
+- Added hash-mode: pbkdf2(1000,20,sha512)
 
 ##
 ## Bugs
--- a/docs/credits.txt
+++ b/docs/credits.txt
@@ -84,5 +84,8 @@ Rick "Minga" Redman and KoreLogic <mingakore@gmail.com> (@CrackMeIfYouCan)
 Brandon Chalk <brandon@casaba.com> (@brandoncasaba)
 * Kerberos Pre-Auth 17/18 kernel module, ported from @Fist0urs TGS kernel modules
 
+Jamie <jamie@blacktraffic.co.uk> (@pedantic_hacker)
+* module_12101.c - web2py pbkdf2 variant
+
 !!! All the package maintainer of hashcat !!!
 
--- a/docs/readme.txt
+++ b/docs/readme.txt
@@ -133,6 +133,7 @@ NVIDIA GPUs require "NVIDIA Driver" (418.56 or later) and "CUDA Toolkit" (10.1 o
 - PBKDF2-HMAC-SHA1
 - PBKDF2-HMAC-SHA256
 - PBKDF2-HMAC-SHA512
+- web2py PBKDF2-HMAC-SHA512
 - scrypt
 - phpass
 - Ansible Vault
--- a/src/modules/module_12101.c
+++ b/src/modules/module_12101.c
@@ -42,6 +42,9 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig, 
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+static const char *SIGNATURE_WEB2PY = "pbkdf2";
+static const char *SIGNATURE_WEB2PY_PARAMS = "20,sha512)";
+
 typedef struct pbkdf2_sha512
 {
   u32 salt_buf[64];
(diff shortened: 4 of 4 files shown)


jsteube:
Hey I've just noticed the unit tests are still missing. Please add so that I can merge this PR.

blacktraffic:
Sorry, I'm being really dense this morning - can't see what to actually implement for a unit test. Can you point me at a unit test that I should follow as an example please?

philsmd:
hey @blacktraffic ,

I kind of already explained what is missing over here: https://github.com/hashcat/hashcat/pull/2165#issuecomment-523337096

Test modules are in tools/test_modules/ and the name of the module is basically m[hash_type].pm

you can run tests with: ./tools.test.sh -m 12101 -a all -t all

or generate hashes with perl tools/test.pl passthrough 12101 or verify hashes with perl tools/test.pl verify 12101 only_hash.txt cracks_to_verify.txt  successfully_verified_cracks.txt; cat successfully_verified_cracks.txt

There is some documentation about the test framework here (but it's mostly about the salt/pass lenght limits): https://github.com/hashcat/hashcat/blob/master/tools/test_modules/README.md

We often call this "unit tests", but at the end it's just some random hashes that get generated and hopefully successfully cracked ( so the term might be not 100% correct, but who cares... it's just our test framework, mostly only run by devs)

Thank you !

btw: just a friendly reminder that you actually do not need to close PRs all the time (not sure if you know this)... you can just push changes to the same git branch and therefore update the code... This is better for everyone here, because I already totally forgot about what we discussed within the closed pull requests (closed PRs do not really get that much of an attention of course and there os actually no direct link to the old PR here either, so it's difficult to follow what we said should be changed etc)

blacktraffic:
Thank you very much! 

It's halfway there now - just wondering if there were any common mistakes that cause the multi to fail but single to work OK? e.g. maybe i'm not correctly initialising everything in the module. Anyway, I'll have another bash at it as soon as I can, and hopefully get it working completely. 

$ ./tools/test.sh -m 12101 -a all -t all
[ test_1573570847 ] > Init test for hash type 12101.
[ test_1573570847 ] [ Type 12101, Attack 0, Mode single, Device-Type Gpu, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout
[ test_1573570847 ] [ Type 12101, Attack 0, Mode multi,  Device-Type Gpu, Vector-Width 1 ] > Error : 1/1 not found, 0/1 not matched, 0/1 timeout
[ test_1573570847 ] [ Type 12101, Attack 0, Mode single, Device-Type Gpu, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout
[ test_1573570847 ] [ Type 12101, Attack 0, Mode multi,  Device-Type Gpu, Vector-Width 4 ] > Error : 1/1 not found, 0/1 not matched, 0/1 timeout

jsteube:
This is a very good example of why the unit tests are so important. This is actually a bug in the implementation. Typically if there are problems only in multi-hash but not in single-hash, then it's often a problem related to salt. The problem is that both struct datatypes between your new mode and the kernel from 7100 are not in sync. See here:

https://github.com/hashcat/hashcat/pull/2187/files#diff-1718c964f3889b6da78225ee22553a8dR48-R56
and
https://github.com/hashcat/hashcat/blob/master/OpenCL/m07100-pure.cl#L30-L34

This means the original idea that the 7100 kernel can be reused is no longer valid. There are two possible solutions. It's up to you how you want to fix it. The first is to copy the 7100 kernel as-is into a new kernel 12101 so that you can change the datatype in the kernel. You will also need to change the kernel names (m07100_init -> m12101_init etc). The other solution would be to change the 7100 kernel, but then you also need to update all modules which are utilizing the 7100 kernel. I'd prefer the first solution, but then the hash-mode numbering idea no longer applies because the hash-mode becomes a "standalone" mode. This mean you will need to filenames and all occurances of 12101 to the next free hash-mode. The next free would be 21600.

blacktraffic:
thank you very much! - i do appreciate the guidance and will give it a rework as suggested.

philsmd:
some minor notes: I see some whitespace " " at the end of some lines and some " ;" instead of just ";" at the end of some lines. there should be no empty line with only spaces in the code " " (without any code within the line).

here we normally add spaces: token.buf[0]+7 => token.buf[0] + 7; 
we also use spaces between the function name and the parenthesis e.g.
```
strncmp((const... => strncmp ((const
```

Another note: we have salt_t which already stores a lot of info... please look at OpenCL/inc_types.h for salt_t and try to use all the fields you need like salt_len (instead of pbkdf2_sha512->salt_len)... therefore avoid to use salt_len in both pbkdf2_sha512_t and in salt_t (it's not needed to duplicate the value in both or add this salt_len field at all in pbkdf2_sha512_t).

I do not understand why hash_len is stored and used as it was a dynamic/flexible variable, but at the end we force it to be 20 in the signature check SIGNATURE_WEB2PY_PARAMS (20,sha512).
Could this be done more flexible and most of all done in a way that is more logical (why store it if 20 is hard-coded?).

The rest seems really good for now... I'm looking forward to the test module and the new kernel files (btw: they should be pretty much straight forward: just copy the m07100-pure.cl and name m21600-pure.cl, inside the file replace all 7100 with 21600 and change the pbkdf2_sha512_t struct, in the src/modules/ file you need to also use 21600 everywhere now and adapt the struct (remove salt_len?))

this is great
Thank you so much.

blacktraffic:
Thanks very much!

I wanted to generalise the signature, but so far I only have one or two
examples, both with "20,sha512", hence I've currently left it requiring
this - and I will need to test with further examples before trying to
generalise it properly I think.

thanks,
 Jamie

jsteube:
Replaced by https://github.com/hashcat/hashcat/pull/2239

PR #2189, opened 2019-09-12
bools

neheb:


neheb:
Got rid of the second commit. It's mostly a style issue.

Issue #2219, opened 2019-11-09
Bitcoin wallet.dat (-m 11300) limits salt size to 8 bytes


solardiz:
This is mostly JFYI, feel free to act on it or ignore it:

https://github.com/hashcat/hashcat/blob/11c8a4040b0b439b520952a461cbfa8f1d12898e/src/modules/module_11300.c#L147 requires that the salt size be exactly 16 hexadecimal characters. All test vectors that we have in JtR match this requirement. However, I've just learned that the legacy wallet for Nexus https://github.com/Nexusoft/Legacy used salt size of 18 bytes, or 36 hexadecimal characters. A minor revision of JtR's `bitcoin2john.py` script (dropping the unneeded parsing of transaction data from it, for which I opened an issue with JtR) appears to be able to parse such wallet, producing a "hash" that JtR accepts as-is, but hashcat rejects for wrong salt length (hashcat wants exactly 16, but the "hash" has 36 in there). Maybe hashcat's check should be relaxed, after confirming that the OpenCL kernel accepts a variable salt length correctly. I've tried this, and hashcat accepts the "hash" and appears to run fine. Unfortunately, I do not currently have a test vector with a known password. I suppose one can be produced with the legacy wallet.

```
$ fgrep -r _SALT_ .
./src/wallet/crypter.cpp:        if (nRounds < 1 || chSalt.size() != WALLET_CRYPTO_SALT_SIZE)
./src/wallet/crypter.h:    const unsigned int WALLET_CRYPTO_SALT_SIZE = 18;
./src/wallet/wallet.cpp:        kMasterKey.vchSalt.resize(WALLET_CRYPTO_SALT_SIZE);
./src/wallet/wallet.cpp:        RAND_bytes(&kMasterKey.vchSalt[0], WALLET_CRYPTO_SALT_SIZE);
```

solardiz:
Update: it's not that simple regarding that Nexus wallet - they seem to have blindly changed the key and salt size constants in the code relative to Bitcoin's (32 to 72, 8 to 18), but the code wouldn't handle the new values correctly. So we should _not_ blindly accept 36 as valid and try to crack that - it doesn't work.

I'll close this issue for now since the actual issue, if any, would be different. Maybe we'll get away with a `bitcoin2john.py` update, or maybe an update to JtR and hashcat will also be needed - we'll see.

Sorry for the noise.

PR #2222, opened 2019-11-12
Add Nexus legacy wallet support to -m 11300

solardiz:
Related to #2219, this adds Nexus legacy wallet support to mode 11300, which amounts to two things:

1. Accept salt size of 36 hex characters, but treat it the same as 16 (only use first 16 hex characters) except it indicating this wallet type.

2. Expect different padding in the last AES block - 8 bytes of 0x08 instead of 16 bytes of 0x10 - because this wallet's full mkey size is 72 (which gives 8 modulo AES block size of 16) instead of the usual 32.

Test vector I used:

```
$bitcoin$64$6b0fbcd048e791edbab30408e14ee24cc51493b810afb61a1e59bc633993a093$36$74fc96a47606814567f02c7df532f6079cbd$169021$2$00$2$00
```

The password is "openwall". The test wallet was generated with `Nexus-Qt-Linux-0_2_0_5.tar.gz` and processed with current/work-in-progress revision of JtR's `bitcoin2john.py`. Older versions of that script would also process the wallet correctly, including for use with hashcat with this PR's changes, but they would include extraneous data. The above has placeholders in the last two fields, which are unneeded for password cracking.

While at it, I also changed a few things for all wallet types supported in this mode:

1. Instead of decrypting the full mkey (or whatever trailing portion of it is available in the input "hash"), decrypt only one final block using the previous block as IV. This is very slightly faster since it saves at least one AES block decryption (we keep at least two blocks in the "hash" since the penultimate block is needed as IV for the final block), although the difference is negligible given that it's done after having done (usually) thousands of iterations of SHA-512. As a side-effect of this change, the kernel now requires mkey size to be at least 32 bytes, whereas previously it could also handle 16, but I'm unaware of any wallet with mkey size less than 32. This side-effect is easy to avoid (just set IV like the code did before when the mkey size is 16 bytes) if there's any reason to.

2. Correct mkey length validation. Previously, the code checked for `% 16` giving zero, but that check worked on the number of hex characters, not bytes, whereas AES block size is 16 _bytes_. This is now changed to `% 32`, and now we also check that there are at least 64 hex characters (two AES blocks, which is required per the point above).

3. Allow 3-digit numbers for encoded mkey size, since the code previously allowed for up to 256 for the next field's length, and that's a 3-digit number. We could reduce that limit instead, which would also achieve consistency, but I saw no reason to. Nexus wallet's mkey size is 72 bytes, which is 144 hex characters. Our current `bitcoin2john.py` script would only encode the last 64 hex characters, but revisions from ~2+ years ago could encode the full 144.

### DIFF
--- a/OpenCL/m11300-pure.cl
+++ b/OpenCL/m11300-pure.cl
@@ -102,7 +102,7 @@ KERNEL_FQ void m11300_init (KERN_ATTR_TMPS_ESALT (bitcoin_wallet_tmp_t, bitcoin_
 
   sha512_update_global_swap (&ctx, pws[gid].i, pws[gid].pw_len);
 
-  sha512_update_global_swap (&ctx, salt_bufs[salt_pos].salt_buf, salt_bufs[salt_pos].salt_len);
+  sha512_update_global_swap (&ctx, salt_bufs[salt_pos].salt_buf, 8);
 
   sha512_final (&ctx);
 
@@ -293,13 +293,6 @@ KERNEL_FQ void m11300_comp (KERN_ATTR_TMPS_ESALT (bitcoin_wallet_tmp_t, bitcoin_
   key[6] = h32_from_64_S (dgst[3]);
   key[7] = l32_from_64_S (dgst[3]);
 
-  u32 iv[4];
-
-  iv[0] = h32_from_64_S (dgst[4]);
-  iv[1] = l32_from_64_S (dgst[4]);
-  iv[2] = h32_from_64_S (dgst[5]);
-  iv[3] = l32_from_64_S (dgst[5]);
-
   #define KEYLEN 60
 
   u32 ks[KEYLEN];
@@ -308,10 +301,18 @@ KERNEL_FQ void m11300_comp (KERN_ATTR_TMPS_ESALT (bitcoin_wallet_tmp_t, bitcoin_
 
   u32 out[4];
 
-  for (u32 i = 0; i < esalt_bufs[digests_offset].cry_master_len; i += 16)
   {
-    u32 data[4];
+    u32 i = esalt_bufs[digests_offset].cry_master_len - 32;
 
+    u32 iv[4];
+    iv[0] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 0]);
+    iv[1] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 1]);
+    iv[2] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 2]);
+    iv[3] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 3]);
+
+    i += 16;
+
+    u32 data[4];
     data[0] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 0]);
     data[1] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 1]);
     data[2] = hc_swap32_S (esalt_bufs[digests_offset].cry_master_buf[(i / 4) + 2]);
@@ -323,17 +324,19 @@ KERNEL_FQ void m11300_comp (KERN_ATTR_TMPS_ESALT (bitcoin_wallet_tmp_t, bitcoin_
     out[1] ^= iv[1];
     out[2] ^= iv[2];
     out[3] ^= iv[3];
-
-    iv[0] = data[0];
-    iv[1] = data[1];
-    iv[2] = data[2];
-    iv[3] = data[3];
   }
 
-  if ((out[0] == 0x10101010)
-   && (out[1] == 0x10101010)
-   && (out[2] == 0x10101010)
-   && (out[3] == 0x10101010))
+  u32 pad;
+  if (salt_bufs[salt_pos].salt_len != 18) /* most wallets */
+  {
+    pad = 0x10101010;
+    if (out[0] != pad || out[1] != pad)
+      return;
+  } else { /* Nexus legacy wallet */
+    pad = 0x08080808;
+  }
+
+  if (out[2] == pad && out[3] == pad)
   {
     if (atomic_inc (&hashes_shown[digests_offset]) == 0)
     {
--- a/src/modules/module_11300.c
+++ b/src/modules/module_11300.c
@@ -127,12 +127,12 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[1]     = '$';
   token.len_min[1] = 2;
-  token.len_max[1] = 2;
+  token.len_max[1] = 3;
   token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_DIGIT;
 
   token.sep[2]     = '$';
-  token.len_min[2] = 16;
+  token.len_min[2] = 64;
   token.len_max[2] = 256;
   token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_HEX;
@@ -145,7 +145,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[4]     = '$';
   token.len_min[4] = 16;
-  token.len_max[4] = 16;
+  token.len_max[4] = 36;
   token.attr[4]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_HEX;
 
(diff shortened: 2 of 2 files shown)


solardiz:
I tested these changes on an older hashcat tree, from mid-June, and then re-applied them to the current tree for this PR. I didn't re-test on the current tree (sorry!), but I see no reason why they wouldn't work just as well. Nothing important related to this format changed since June, as far as I can see (only a warning on macOS was added).

solardiz:
> I'm unaware of any wallet with mkey size less than 32.

Moreover, for a wallet to have an encrypted master key size of less than 32, the original master key size would need to be less than 16, which is unreasonable. Exactly 16 would result in 32 seen by us. This is because the padding adds at least 16 bytes. So e.g. for a typical master key size of 32 bytes, the encrypted master key size seen by our tools is 48 bytes (96 hex, but recent `bitcoin2john.py` leaves just the last 64, which is all we need).

> Nexus wallet's mkey size is 72 bytes, which is 144 hex characters

Actually, it's 72 before encryption, but with the padding it's 80, and thus is 160 hex characters - but recent `bitcoin2john.py` leaves just the last 64, which is all we need.

jsteube:
Thanks for the PR! 

I'm not too much into the details of this hash-mode but from your description your changes make sense. 

There's just this one thing that confuses me a bit, maybe you can explain it?

```
-  sha512_update_global_swap (&ctx, salt_bufs[salt_pos].salt_buf, salt_bufs[salt_pos].salt_len);
+  sha512_update_global_swap (&ctx, salt_bufs[salt_pos].salt_buf, 8);
```

Is this maybe an optimization?

solardiz:
No, this isn't an optimization, but is a required change:

OpenSSL's `EVP_BytesToKey` that these wallets use assumes the salt is exactly 8 bytes, and doesn't accept a salt size parameter. Nexus legacy wallet changed (relative to Bitcoin's) their salt size constant from 8 to 18, but this call only made use of the first 8 bytes anyway. What we do here is pass the wallet's specified salt size into the kernel so that we can use this to detect Nexus legacy wallet to do the right padding check (since that wallet's master key size mod 16 is also different from Bitcoin's), but we need to change the above to hard-coded 8 in order for this salt size not to matter otherwise, like it doesn't matter in the wallet.

As an alternative, we could set the salt size to 8 in the host code and introduce a flag or whatever for the kernel to detect Nexus legacy wallet. This might be cleaner, but would be more code.

Curiously, a bigger difference between the wallets is in how the initial IV is derived through Skein and Keccak, but luckily this doesn't matter for us since it only affects the first AES block, and we only need the last. (Quite possibly our old code, in both JtR and hashcat, was also buggy in not using the correct initial IV - but this also didn't matter for the same reason.)

solardiz:
Two questions I had during development of this patch:

1. If I continued to pass `cry_salt_buf_len` into `generic_salt_decode()`, this appeared to cause some kind of memory overwrite(?) resulting in no match on the padding check when the salt size is 18 bytes. I didn't track this down, but simply changed the line to only decode what's needed:

```c
  const bool parse_rc = generic_salt_decode (hashconfig, cry_salt_buf_pos, 16 /* instead of cry_salt_buf_len */, (u8 *) salt->salt_buf, (int *) &salt->salt_len);
```

2. Before my changes, `salt_bufs[salt_pos].salt_len` was only used in `m11300_init()`. Now it's used in `m11300_comp()`. This works OK in my testing, but is it allowed per hashcat conventions - that is, is the value guaranteed to stay the same across these calls? If not, the value would need to be cached in a field (to be) added to `struct bitcoin_wallet_tmp` (in fact, I had a code revision like that, but then simplified this out as I saw no convincing reason for this complication).

solardiz:
BTW, it's unfortunate that hashcat uses only one test vector per hash mode. This results in incomplete code coverage by the self-test here, so any potential breakage of Nexus legacy wallet support by future code changes or by a miscompile would go undetected.

philsmd:
hey @solardiz,
this is some great contribution. thank you very much.

The self-test hashes are not meant to test for bugs in the code (kernel or host), but are only used to detect driver issues on the user end. We have test.sh to test for bugs in the code
```
./tools/test.sh -m 11300 -t all -a all
```

this makes more sense because devs run this tests periodically and before releases etc to find regressions or new problems.


I have another question for you, what do you mean by:
`a warning on macOS was added`

are you referring to a compiler warning on macOS because of the changes? I don't really understand this statement (probably it's something completely different, hehe). If it is a compiler warning can you post it here since I have currently can't test with macOS.

Here we normally use a little bit different coding style
```
if (out[0] != pad || out[1] != pad)
      return;
```
=>
```
if (out[0] != pad || out[1] != pad) return;
```
or
```
if (out[0] != pad || out[1] != pad)
{
  return;
}
```

and
```
} else {
```
=>
```
}
else
{
```
I also think we should remove the extra block {} and indentation where AES256_decrypt () is done (this indentation, extra block is not needed, right?).

but this is not a big deal, we can always make these changes later on

yeah, "salt_len" is guaranteed to be correct in the comp () kernel

jsteube:
> OpenSSL's EVP_BytesToKey that these wallets use assumes the salt is exactly 8 bytes, and doesn't accept a salt size parameter. Nexus legacy wallet changed (relative to Bitcoin's) their salt size constant from 8 to 18, but this call only made use of the first 8 bytes anyway.

OK, I understood this systematic.

> As an alternative, we could set the salt size to 8 in the host code and introduce a flag or whatever for the kernel to detect Nexus legacy wallet. This might be cleaner, but would be more code.

Indeed it has a bit a character of a hack, therefore your proposal seem legit. The right struct would be `struct bitcoin_wallet`. This is the buffer that holds all the static data for use in the kernel which does never change. Note that this struct exists twice in the module and in the kernel and needs to be synchronized manually on both sides.

> If I continued to pass cry_salt_buf_len into generic_salt_decode(), this appeared to cause some kind of memory overwrite(?) resulting in no match on the padding check when the salt size is 18 bytes.

No worries there is no overwrite of memory. To understand it one needs to see this first:

https://github.com/hashcat/hashcat/blob/master/OpenCL/inc_hash_sha512.cl#L37

This goes back to a general concept in hashcat to not use some sort of memcpy() emulation but to use large switch() statements to help the Jit to keep data on registers. All crypto helper functions from the GPU libraries follow this concept.

If you choose to use cry_salt_buf_len, then the buffer in salt->salt_buf will contain more than the 8 byte data. As a result, the final AES key will be completely different. Probably it's just confusion with a problem in padding check.

> ... is the value guaranteed to stay the same across these calls

Yes, this value is guaranteed to stay the same (except if one changes it from the kernel itself)

> BTW, it's unfortunate that hashcat uses only one test vector per hash mode. 

Hashcat architecture is a bit different to JtR. The goal with the runtime self-test in hashcat plugin is not to deeply check the module and kernel but to check for driver bugs. I'm sure you have experienced yourself how often the drivers produce invalid results. However, if they do, it's always invalid. This check is only done once on startup to inform the user he's probably using a not correctly working OpenCL driver.

For real testing, hashcat has a dedicated testing suite. It consists mainly of three components test.sh, test.pl and a hash-mode specific perl module. One write the algorithm-specific details into a module for instance tools/test_modules/m11600.pm. This module is loaded at runtime from test.pl program which creates random passwords or reads user passwords, salts, etc and then calls the hash mode specific functions and print out the resulting hash. Finally, the test.sh is a wrapper script to test.pl which then compares the calculated test.pl hashes and compares them with the expected results. It also permutes multiple different attack options. To get everything started the user just calls something like this: 

```
tools/test.sh -t multi -a all -m 11600 
```

This will create a ton of different checks with different salt and password length etc so it will take a while.

solardiz:
@philsmd Thank you for your review and comments!

> what do you mean by: `a warning on macOS was added`

I mean commit 32cb191837e841a33aa42f47a4dee072d0da0dd9. Nothing to worry about here.

> I also think we should remove the extra block {} and indentation where AES256_decrypt () is done (this indentation, extra block is not needed, right?).

I had thought of this, but decided to keep the extra {} block to reduce the number of lines changed and to keep the scope of `i`, `iv`, and `data` limited to that block. If hashcat's coding style is to avoid such scope limiting, this can indeed be changed.

solardiz:
> your proposal seem legit. The right struct would be `struct bitcoin_wallet`.

Do you suggest we implement this before merging this PR or after? I don't know whether I'll have more time to work on this, and certainly not right away.

Instead of a flag, maybe we can make this `u32 pad`, with the expected padding value in there, and have the code infer how many `out` elements to check for that value from the value itself.

Regarding the tests, in this case - and I guess in some others - hashcat's builtin self-test might also miss some driver bugs, because the individual test vectors result in limited code coverage. With this PR, the (few) lines of code specific to Nexus wouldn't be tested, so a miscompile of them would go undetected. Hopefully, there are few enough of those lines that a miscompile won't happen.

Looking at `tools/test_modules/m11300.pm` I don't see how we'd enhance it (within the existing API) to produce more than one test hash per input word. Maybe it's invoked multiple times and we can alter between the two hash sub-types for different input words?

jsteube:
> Do you suggest we implement this before merging this PR or after? I don't know whether I'll have more time to work on this, and certainly not right away.

OK, I'll merge the PR and do the changes as discussed.

> Regarding the tests, in this case - and I guess in some others - hashcat's builtin self-test might also miss some driver bugs, because the individual test vectors result in limited code coverage. With this PR, the (few) lines of code specific to Nexus wouldn't be tested, so a miscompile of them would go undetected. Hopefully, there are few enough of those lines that a miscompile won't happen.

There's indeed a change for that but typically if it happens it fails in the first instructions.

> Looking at tools/test_modules/m11300.pm I don't see how we'd enhance it (within the existing API) to produce more than one test hash per input word. Maybe it's invoked multiple times and we can alter between the two hash sub-types for different input words?

Yes, test.pl calls the function multiple times, each time with different password value/length, salt length etc. We just call it salt, but this also means IV or any other configuration items like version number or something similar. So we can have different branches in the perl module to produce hashes to cover all possible code paths.

jsteube:
As discussed: https://github.com/hashcat/hashcat/commit/08a74596c1f289805f0e6c45b5575e4dd77c008b

PR #2227, opened 2019-11-16
Fixes #1806: added -m 21700 = Electrum 4 and -m 21800 = Electrum 5

### LINKED ISSUE #1806: Electrum 3.1.3
Hi guys,

I forgot my wallet password, running Electrum 3.1.3.

So Hashcat still doesn't support this wallet format?

Anyway to work around it?

Thanks,
Libre

philsmd:
With this commit we implement the missing Electrum salt type 4 and salt type 5 "hashes" as mentioned in #1806 :
-m 21700 = Electrum Wallet (Salt-Type 4)
-m 21800 = Electrum Wallet (Salt-Type 5)

Most of the technical details are explained in #1806 by @philsmd (me).

I've decided to split the 2 hash types into 2 separate one because we can leverage full GPU power for Electrum 4 by doing the final sha512 and hmac-sha256 () on the kernel too.

For Electrum 5 it's a little bit different because we need to AES decrypt the data and decompress it (zlib) to find the fixed validation string at the beginning.

This was a **longer** journey than I first thought (my opinions I also expressed over here: #1806) , because as you can see over here https://github.com/philsmd/hashcat/commits/electrum_4_and_5  I was messing around with all types of bignum libraries (and reasoning about library licensing problems) until I got totally rid of the bignum libraries.

The remaining library is secp256k1 from Pieter Wuille (bitcoin core developer) which should in theory be very optimized (it is! but of course GPU support for ECC would also be great) and zlib only for Electrum 5 (truncated, encrypted and compressed data).

Thank you very much


jsteube:
Excellent contribution, many thanks!

jsteube:
No it's GPU and CPU, speed was improved with newer commit, retry.

PR #2242, opened 2019-12-05
electrum 4/5: improve speed by removing hook

philsmd:
This patch adds our own OpenCL kernel code for elliptic curve multiplication by a scalar/tweak: OpenCL/inc_ecc_secp256k1.cl

The affected hash types for now are:
- -m 21700 = Electrum Wallet (Salt-Type 4) and
- -m 21800 = Electrum Wallet (Salt-Type 5)

With this code we can perform everything on GPU (OpenCL/CUDA) without the need of external libs and/or hook code. This should (and does) improve speed by a lot (but of course be aware that secp256k1 by Pieter Wuille is already among the most optimized codes for ECC operations on the secp256k1 curve).

The main multiplication code point_mul () was highly optimized by using some precomputed buffers (coordinates) by the host code... and it is also NOT meant to be used in security critical environment that are at risk of timing/side-channel attacks (non-constant time). Shout out go also to several online resources and in particular to micro-ecc (https://github.com/kmackay/micro-ecc, BSD) which I sometimes used as a reference to optimize/implement some BigNum and ECC operations (I've listed more interesting projects at the top of OpenCL/inc_ecc_secp256k1.cl).

Thank you

this is also in relation to #1806 where Electrum 4/5 were requested as feature requests


jsteube:
Excellent!

ghost:
Error running $4$

image

Progress stop on **Initialized device kernels and memory...**

philsmd:
I think it's **not** very clever and helpful to just post some questions/problems/issues under commits (https://github.com/philsmd/hashcat/commit/d078aa3f08c83ed693dd5054bb1a99a09e953976, message deleted) or pull requests (like here).

I also **DO NOT** like that you delete messages after you resolve the problem or just randomly after some time. We have a hashcat forum user with 100% warning level called Kulahin that does exactly the same, they just ask something and as soon as somebody spend couple of minutes/hours to think and troubleshoot the problem (which often turns out to be just PEBCAK) and answer the question... the forum post is gone. I'm not sure if this is the same person... but they are also very interested in Electrum and have the same (in my opinion) very "dangerous" and rude/disrespectful attitude to waste the time of developers/mods/users which spend time to explain problems and after just a glance .... seeing that the whole thread was deleted just after the user gets/accepts/understand the answer (public posts should be there to help also **other** users). Again, I'm not sure if it is you... but there are some evidences that you've previously posted the same "Electrum only works on CPU" at the same time on github as Kulahin posts in the forum etc...

Back to the original problem...
it would make much more sense to get attention and some help by some users/mods/developers if you posted a detailed description of the problem with your hashcat version number (or beta version), your hardware/setup, CPU/GPU model/make, driver version, command line on the forum. also test --benchmark, test with -D 1 etc   ... instead of posting under some **closed**/merged PR.

Don't get me wrong, I'm very grateful that users report problems... but you also need to do it in an acceptable way and let other users also learn and profit from the outcomes/answers etc (and don't waste the time of other people)

Thank you

Issue #2123, opened 2019-07-19
Can't crack PDF 1.7 Ext8 hash (-m 10700)


vizzdoom:
I am testing pdf hash cracking on hashcat 5.1.0 and hashcat-BETA v5.1.0-1223-ga7fd1e40.

I created a PDF document using Acrobat Reader 10.1.16. I have enabled document encryption with the following encryption compatibility options:
- Acrobat X and never
- Encode the whole document with the metadata
- Set encryption password: `mamatata`
- Encoding: AES-256

The output document has PDF 1.7 Extension 8 format. 

The pdf2john output is the following:
```
$pdf$5*6*256*-1028*1*16*a9f7206e48ba1d45b9ead8a55d5d6e17*127*d157c5663d0cba76cf6b59166cfbaa56cafd379b8ad0d6f908fcc04f4415a90795015f2e49dfb527aa6c677334285b9500000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000*127*45c06f4bc847d1298f06213b6ed65c6c6e7ab82e52c08e38f99fe727fd7002a4392e666dcc3da1b828bcc673870213f400000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000*32*073a209c0e8d438f5dffe1ff32777b6131fdf61f26f2da72e9c4882f67f2a8e5*32*83c200fe60a5803c7b7f291cce8a557d0cbfbfa9d2043229a265027141c0ea1c
```
Cracking command (tested modes: 10700, 10600, 10500, 10400
```
PS C:\hashcat> .\hashcat64.exe -m 10700 C:\Desktop\hashes-pdf.txt D:\Wordlists\testwordlist.txt --potfile-disable
hashcat (v5.1.0) starting...

* Device #1: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
* Device #2: Intel's OpenCL runtime (GPU only) is currently broken.
             We are waiting for updated OpenCL drivers from Intel.
             You can use --force to override, but do not report related errors.
nvmlDeviceGetFanSpeed(): Not Supported

OpenCL Platform #1: NVIDIA Corporation
======================================
* Device #1: GeForce RTX 2070, 2048/8192 MB allocatable, 36MCU

[...]

PR #2262, opened 2020-01-04
PDF module: -m 10700 missing assignment of tmp_size

philsmd:
The missing assignment of tmp_size is the culprit for crashes.

It still seems to NOT crack without -O (i.e. with the pure kernel, instead of the optimized kernel) with some drivers... we would need to investigate this separately... but this is definitely a serious bug that leads to wrong memory access etc. Luckily I've found this pretty quickly when investigating https://github.com/hashcat/hashcat/issues/2123 ... but there is still work to do to make pure kernels not-fail on some drivers (#2123 not yet completely fixed AFAIK).

Thanks


jsteube:
Eagle Eye

PR #2268, opened 2020-01-09
Added optional modules separation for easying package mantainers work

TanyaEleventhGoddess:


jsteube:
How is it making live easier for maintainer? Can you please explain what's behind this?

jsteube:
I don't see how this is making live more easier for maintainer. Copy modules to /usr/local/share/hashcat/ seems to be a good solution. For instance, perl is doing the same for all of it's perl modules. Please clarify the changes and explain your thoughts in detail then we can reopen the PR.

TanyaEleventhGoddess:
Sun, 12 Jan 2020 09:22:44 +0000 (UTC) Jens Steube <notifications@github.com>:
> I don't see how this is making live more easier for maintainer. Copy modules to /usr/local/share/hashcat/ seems to be a good solution. For instance, perl is doing the same for all of it's perl modules. Please clarify the changes and explain your thoughts in detail then we can reopen the PR.

Usually libs are separated in /usr/lib, and in /usr/share resides architecture-independent data

PR #2282, opened 2020-01-15
Fixes #1534: Added Telegram and sha256($s.$p.$s)

### LINKED ISSUE #1534: Support for Telegram Passcode (android)
The popular chat-app Telegram has a feature in the settings-menu to protect the app with a passcode. The source code for this protection can be found on the [Telegram-github (line 411)](https://github.com/DrKLO/Telegram/blob/4ebcbf61cd1ce04685c82f93b962d1094c5ffac2/TMessagesProj/src/main/java/org/telegram/messenger/UserConfig.java).

One finds the passcodeHash and the passcodeSalt in the user file *userconfig.xml*.

When reading the code, it seemed to me that it is very easy to find the passcode. I wrote the following script to retrieve the passcode. [See my Github-page for the details.](https://github.com/Banaanhangwagen/Telegram-Passcode-Bruteforce)

So, the used algorithm is `sha256($salt:key:$salt) = $pass
`Correct?

Can this be added to Hashcat?

philsmd:
This PR adds the generic mode -m 22300 for sha256 ($salt . $pass . $salt) and the specific hash type -m 22301 = Telegram (SHA256) as requested in #1534 .

The format for -m 22300 is of course just hash:salt, but the format for Telegram is similar to the one generated by telegram2john.py (see https://raw.githubusercontent.com/magnumripper/JohnTheRipper/bleeding-jumbo/run/telegram2john.py), but with the exception that I've replaced "dynamic_1528" with "$telegram$" and the version number 1 (in case we need a newer format in the future)... futheremore it of course has no file names in it and no "HEX" string.

An example is like this:
```
$telegram$1*518c001aeb3b4ae96c6173be4cebe60a85f67b1e087b045935849e2f815b5e41*25184098058621950709328221838128
```
(password for this hash is hashcat)

where the first 1 is the (our internal) "version number" of the format, 518c00... is the sha256 digest and 25184... is the salt (in hex).

Thank you very much

### DIFF
--- a/OpenCL/m22300_a0-optimized.cl
+++ b/OpenCL/m22300_a0-optimized.cl
@@ -1,0 +1,574 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_rp_optimized.h"
+#include "inc_rp_optimized.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha256.cl"
+#endif
+
+#define SHA256_STEP_REV(a,b,c,d,e,f,g,h)        \
+{                                               \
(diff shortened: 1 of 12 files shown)


philsmd:
note: we had to change the format to $telegram$0 (see https://github.com/hashcat/hashcat/pull/2322)

Issue #1664, opened 2018-08-17
feature: next generation of outfile format


roycewilliams:
The outfile format has much potential, but we need a new way to express it.

The current method consists of IDs like the following:

```
  1 | hash[:salt]
  2 | plain
  3 | hash[:salt]:plain
  4 | hex_plain
  5 | hash[:salt]:hex_plain
  6 | plain:hex_plain
  7 | hash[:salt]:plain:hex_plain
  8 | crackpos
  9 | hash[:salt]:crack_pos
 10 | plain:crack_pos
 11 | hash[:salt]:plain:crack_pos
 12 | hex_plain:crack_pos
 13 | hash[:salt]:hex_plain:crack_pos
 14 | plain:hex_plain:crack_pos
 15 | hash[:salt]:plain:hex_plain:crack_pos
```

This method has worked well up to this point, but growing beyond it will be increasing difficult. Adding just a single additional field would be very expensive, as it would require many new additional outfile format IDs.

Instead, I propose a new way to express outfile format fields - one that is extensible, using "variables" for the components.

My suggested approach:

1. Change "--outfile-format" to "--outfile-format-classic", using the original syntax (for backwards compatibility, but making it clear that this is the new way forward, just like '--markov-classic' was the old default, but must now be explicitly requested).

2. Create a new "--outfile-format" that uses atomic variables, as follows:

| field | field variable |
| ----- | --------------|
| hash[:salt] | h |
| plain | p |
| hex_plain | hp |
| crackpos |  cp |

(These are just for demonstration purposes - there may be better variables. That being said, I suggest shorter variables for easier typing).

This method will allow new "variables" to be more easily added later, without having to add a large number of additional outfile formats.

All current outfile formats would be expressed in the new syntax as follows:

| classic outfile type | outfile format | new syntax |
| -------------------- | -------------- | ---------- |
|  1 | hash[:salt] | h |
|  2 | plain | p |
|  3 | hash[:salt]:plain | h:p |
|  4 | hex_plain | hp |
|  5 | hash[:salt]:hex_plain | h:hp |
|  6 | plain:hex_plain | p:hp |
|  7 | hash[:salt]:plain:hex_plain | h:p:hp |
|  8 | crackpos | cp |
|  9 | hash[:salt]:crack_pos | h:cp |
| 10 | plain:crack_pos | p:cp |
| 11 | hash[:salt]:plain:crack_pos | h:p:cp |
| 12 | hex_plain:crack_pos | hp:cp |
[...]

PR #2292, opened 2020-01-23
Fixes #2255: new --outfile-format with support for timestamps

### LINKED ISSUE #2255: [Feature Request] Timestamps for cracked hashes
It would be nice to know exactly when or how long from start a password was cracked for statistics purposes and oddly enough report outs people are very interested in how much better one password is than another. I know that a simple timestamp doesn't correctly convey if one password is better than another, but it's something I have been asked for multiple times.

philsmd:
This commit changes the way we use the --outfile-format command line parameters by changing it from a single number to a comma-separated list/string of (sub-)formats:
``` 
  1 | hash[:salt]
  2 | plain
  3 | hex_plain
  4 | crack_pos
  5 | timestamp absolute
  6 | timestamp relative
```
e.g. by specifying --outfile-format 1,2,4 the hash+salt, plain and crack position will be shown.

The default remains to always show hash[:salt]:plain (i.e. --outfile-format 1,2).

This commit also adds two new outfile formats, i.e. 2 different timestamps:
- if the --outfile-format list contains the number 5, the **unix epoch** time will be shown
- if the --outfile-format list contains the number 6, the seconds past **since the start** of hashcat will be shown (for instance combined with --outfile-format containing 1, the relative timestamp will be next to the hash)

For now, the order of the fields (or sub-formats) in the output is pre-determined (fixed, can't be changed by the user). This should be acceptable/okay in most of the cases and this way we avoid seeing strange outputs e.g. if the users post their outputs in the hashcat format (without showing the --outfile-format string), we kind of know which field is which and comes first etc.
The current order is:

- first field in output (if enabled): timestamp absolute
- second field in output (if enabled): timestamp relative
- third field in output (if enabled): hash[:salt]
- fourth field in output (if enabled): plain
- fifth field in output (if enabled): hex_plain
- sixth field in output (if enabled): crack_pos

This means, that if (and only if) the absolute timestamp was specified it will be for sure at the start in the output.... a similar logic holds for crack_pos: if (and only if) the sub-format 4 was enabled it will be always at the end of the outfile-format etc etc etc

As mentioned in #2255 this issue also is somehow related to #1664 , we still need to decide if we close this other github issue or need to make further changes for #1664 .

Thanks


philsmd:
Well, actually the assignment of OUTFILE_FORMAT is **NOT** the way it should be done.
compare:
https://github.com/hashcat/hashcat/pull/2292/files#diff-3aaeee3ff1fae36ab5344d073d02228fR222
vs
https://github.com/hashcat/hashcat/pull/2292/files#diff-fdb264e240292589f40e0eea0f91ef89R22

we have the problem of char* vs u32 and we shouldn't use NULL as default and shouldn't set OUTFILE_FORMAT in the parser function. This is very ugly.

This needs to be done much better and cleaner.

jsteube (approved):
Looks great!

PR #2293, opened 2020-01-23
Implement sha256($salt.sha256_bin($pass)) kernel

seger672:
This patch implements a pure `sha256($salt.sha256_bin($pass))` kernel, as used in Bitcoin (P2WSH addresses). It is envisioned that `$salt` will be set to `0x0020` and `$pass` to the candidate redeem script, with converted Bech32 address used as hashes. However, it may also have other uses.

### DIFF
--- a/src/modules/module_21420.c
+++ b/src/modules/module_21420.c
@@ -1,0 +1,248 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "common.h"
+#include "types.h"
+#include "modules.h"
+#include "bitops.h"
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 3;
+static const u32   DGST_POS1      = 7;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 6;
+static const u32   DGST_SIZE      = DGST_SIZE_4_8;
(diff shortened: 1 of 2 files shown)


src/modules/module_21420.c
```diff
@@ -0,0 +1,248 @@
...
+                                  | OPTI_TYPE_EARLY_SKIP
+                                  | OPTI_TYPE_NOT_ITERATED
+                                  | OPTI_TYPE_PREPENDED_SALT;
+static const u64   OPTS_TYPE      = OPTS_TYPE_PT_GENERATE_BE
+                                  | OPTS_TYPE_PT_ADD80
+                                  | OPTS_TYPE_PT_ADDBITS15
+                                  | OPTS_TYPE_PT_HEX
+                                  | OPTS_TYPE_ST_HEX;
+static const u32   SALT_TYPE      = SALT_TYPE_EMBEDDED;
+static const char *ST_PASS        = "6f72676f6e6c6a75732076692067c3a5";
```
philsmd: in your other PR (#2294) you changed the password to plain text instead of hexadecimal. Can we do the same also here ?
Why is the password hexadecimal at all (is this again some bitcoin specific stuff) ?

philsmd:
The combinator (-a 1) and mask attack (-a 3) kernels are missing also here (similar to #2294).

Another quesion that we should ask ourself is whether we should add optimized kernels for passwords < 55. Just have a look at the OpenCL/ folder, it contains optimized kernels for most of the fast hash types, especially meaningful if the passwords are very often < 55 bytes long. Did you already have a look at this or did you not implement the optimized kernels for some specific reason (maybe because of the typical length of the password for the bitcoin case etc ) ?

Thx

Ammarsultan997 (approved):

PR #2294, opened 2020-01-23
Implement ripemd160(sha256_bin($salt.sha256_bin($pass))) kernel

seger672:
This patch implements a pure `ripemd160(sha256_bin($salt.sha256_bin($pass)))` kernel, as used in Bitcoin (P2SH-P2WSH addresses). It is envisioned that `$salt` will be set to `0x0020` and `$pass` to the candidate redeem script, with converted Base58 addresses used as hashes. Feeding big-endian data to RIPEMD-160 is intentional; this is how Bitcoin does it.


seger672:
Please do not merge this.

seger672:
The matter has been resolved; please see latest commit.

jsteube:
Many thanks for the contributions it's very welcome. Do you think you can add the other optional changes to make it complete? like docs/changes, docs/readme, and most important the unit test module in tools/test_modules/?

seger672:
The necessary changes have been made, please review.

seger672:
Paging @jsteube

jsteube:
OK this looks very good actually, but it's still not ready for merge. 

- The main problem is that there's only an _a0 kernel but for full testing we also need _a1 and _a3 kernel. But this leads to another question of how hashcat can help to find the missing piece, which is connected to the second problem
- If I understood this correctly to goal is it to guess "and $pass to the candidate redeem script" which looks like a very large random string. Of course it's not feasible to guess a 64 byte random string, as we can see here https://www.soroushjp.com/2014/12/20/bitcoin-multisig-the-hard-way-understanding-raw-multisignature-bitcoin-transactions/ but maybe this is why you only implemented the _a0 kernel? Could you please clarify?

seger672:
You can use it if there's multiple candidate keys, like in a dictionary attack.

a1 and a3 are now added. Tests pass.

OpenCL/m21920_a3-pure.cl
```diff
@@ -0,0 +1,274 @@
...
+    _w3[0] = 0;
+    _w3[1] = 0;
+    _w3[2] = 0;
+    _w3[3] = 0;
+
+    sha256_update_vector_64 (&ctx_mid, _w0, _w1, _w2, _w3, 32);
+
+    sha256_final_vector (&ctx_mid);
+
+    _v0[0] = hc_swap32(ctx_mid.h[0]);
```
philsmd, line 106: please add spaces: `hc_swap32 (ctx_mid.h[0]);`

this is also true for the following lines and other hc_swap32 () etc function calls.

philsmd:
Could you please share how a "normal password" and hash would look like for the bitcoin case ?

How long are those redeem scripts ("password" ?) ? is the length of these redeem scripts even feasible to crack ? are these random bytes or what is behind those scripts?

I know that this is not really important for adding this new hash type... but since this hash type with the byte swaps and this combination of ripemd+sha256 is quite rarely used outside of bitcoin, it's propably important/interesting to know how an user would normally be able to crack these hashes (a typical scenario of what s/he tries to brute-force or what pattern s/he typical have, a substring/part of the password maybe ?). Thx

bukastillo:
I want to test it but how can give hash ?
`RIPEMD160:???`

Can you share an example for any BTC address ?

PR #2306, opened 2020-02-06
Fixes #1538: Added -m 22500 = MultiBit Classic .key (MD5)

### LINKED ISSUE #1538: support for Multibit Classic
Would be great to have support for "Multibit Classic" key hashes

Test-hash
sleepy.key:$multibit$1*935f8d9b235bff03*e25936c998d095467b727db1028d9d6e4a6660bb8c48b4b6fb43ab8c348c45f2
Password:WinSuxx

https://github.com/magnumripper/JohnTheRipper/blob/7dd60616a4f08bbe26cf529997a827792fe7da79/run/multibit2john.py

Keyfiles
https://github.com/magnumripper/JohnTheRipper/files/1004226/MultiBitWallets.zip

This is not a standard hash but have support as well
https://github.com/gurnec/btcrecover/blob/master/extract-scripts/extract-multibit-privkey.py

philsmd:
This implements the MD5 variants (for MultiBit Bitcoin Wallet, bitcoinj and KnCGroup Bitcoin Wallet) of the MultiBit Classic .key (MD5) algorithm, see request #1538 .

MultiBit has some other variants too (see $multibit$2 and $multibit$3 from multibit2john.py), these other algorithms are currently NOT supported (yet).

The algorithm basically uses 3 calls of MD5 to generate the AES key and IV to decrypt some data.

Even if it is actually quite unlickely, the data length and missing base58check checksum from the JTR hash could make this format produce some false positives (again, very unlikely). I would still suggest running it with --keep-guessing and test the passwords manually with the MultiBit Wallet software.

Thank you very much

### DIFF
--- a/OpenCL/m22500_a0-optimized.cl
+++ b/OpenCL/m22500_a0-optimized.cl
@@ -1,0 +1,1220 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+//#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_rp_optimized.h"
+#include "inc_rp_optimized.cl"
+#include "inc_simd.cl"
+#include "inc_hash_md5.cl"
+#include "inc_cipher_aes.cl"
+#endif
+
+DECLSPEC int is_valid_bitcoinj_8 (const u8 v)
+{
+  // .abcdefghijklmnopqrstuvwxyz
+
+  if (v > (u8) 'z') return 0;
+  if (v < (u8) '.') return 0;
+
+  if ((v > (u8) '.') && (v < (u8) 'a')) return 0;
+
+  return 1;
+}
+
+KERNEL_FQ void m22500_m04 (KERN_ATTR_RULES ())
+{
+  /**
+   * modifier
+   */
+
+  const u64 gid = get_global_id (0);
+  const u64 lid = get_local_id (0);
+  const u64 lsz = get_local_size (0);
+
+  /**
+   * aes shared
+   */
+
+  #ifdef REAL_SHM
+
+  LOCAL_VK u32 s_td0[256];
+  LOCAL_VK u32 s_td1[256];
+  LOCAL_VK u32 s_td2[256];
+  LOCAL_VK u32 s_td3[256];
+  LOCAL_VK u32 s_td4[256];
+
+  LOCAL_VK u32 s_te0[256];
+  LOCAL_VK u32 s_te1[256];
+  LOCAL_VK u32 s_te2[256];
+  LOCAL_VK u32 s_te3[256];
+  LOCAL_VK u32 s_te4[256];
+
+  for (u32 i = lid; i < 256; i += lsz)
+  {
+    s_td0[i] = td0[i];
+    s_td1[i] = td1[i];
+    s_td2[i] = td2[i];
+    s_td3[i] = td3[i];
+    s_td4[i] = td4[i];
+
+    s_te0[i] = te0[i];
+    s_te1[i] = te1[i];
+    s_te2[i] = te2[i];
+    s_te3[i] = te3[i];
+    s_te4[i] = te4[i];
+  }
+
+  SYNC_THREADS ();
+
+  #else
+
+  CONSTANT_AS u32a *s_td0 = td0;
+  CONSTANT_AS u32a *s_td1 = td1;
+  CONSTANT_AS u32a *s_td2 = td2;
+  CONSTANT_AS u32a *s_td3 = td3;
+  CONSTANT_AS u32a *s_td4 = td4;
+
+  CONSTANT_AS u32a *s_te0 = te0;
+  CONSTANT_AS u32a *s_te1 = te1;
+  CONSTANT_AS u32a *s_te2 = te2;
+  CONSTANT_AS u32a *s_te3 = te3;
+  CONSTANT_AS u32a *s_te4 = te4;
+
+  #endif
+
+  if (gid >= gid_max) return;
+
+  u32 pw_buf0[4];
+  u32 pw_buf1[4];
+
+  pw_buf0[0] = pws[gid].i[0];
+  pw_buf0[1] = pws[gid].i[1];
+  pw_buf0[2] = pws[gid].i[2];
+  pw_buf0[3] = pws[gid].i[3];
+  pw_buf1[0] = pws[gid].i[4];
+  pw_buf1[1] = pws[gid].i[5];
+  pw_buf1[2] = pws[gid].i[6];
+  pw_buf1[3] = pws[gid].i[7];
+
+  const u32 pw_len = pws[gid].pw_len & 63;
+
+  /**
+   * salt
+   */
+
+  u32 salt_buf[2];
+
+  salt_buf[0] = salt_bufs[salt_pos].salt_buf[0];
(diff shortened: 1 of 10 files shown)


kost123:
Any benchmark on this? Just wanna know how it compares to JTR, havn't got a rig to check atm.

jsteube:
There's no GPU support for multibit in JtR, only CPU. Running it on my i7-4700k (against the same hash which hashcat is using) in mask mode I am getting with latest JtR GitHub master version with OMP support:

```
0g 0:00:00:49 0.00% (ETA: 2020-08-29 10:16) 0g/s 4047Kp/s 4047Kc/s 4047KC/s iFEN1aa..AFEN1aa
```

Running the same attack on the same CPU hashcat is getting this:

```
Speed.#9.........:  5917.9 kH/s
```

However, hashcat has GPU support. On my GTX1080:

```
Speed.#1.........:   501.0 MH/s
```

kost123:
Jesus christ man, this is going to help so much, I have been getting 5mil a second too and now I will be getting about 1.5bil/s, such a massive jump. I have been trying for years to get my wallet and now I feel like I have a shot, if I do get it, I will send a nice donation to you atom and phil.

philsmd:
hey, I just accidentally discovered (actually just by looking at the changes of hashcat forks: https://github.com/hashcat/hashcat/network) that you did some changes to your fork by adding/introducing "y" as one of the first letters of the (decrypted/raw) private key (WIF format): https://github.com/kost123/hashcat/commit/30b7977bc3333047b232679e3c0dc596b5a74509 etc

Would you mind letting us know if this is something we need to add too, what software/cryptocurrency does use the "y" as the first char (I know 5,L,K are for BTC and Q is for Dogecoin/MultiDoge), but why do you add y ? should we add this to our code too, @kost123 ?

BTW: I also think your changes are kind of incomplete/incorrect: you have 2 branches (master/patch-1), but none of them has all the changes (pure a0, pure a1, pure a3, optimized a0, optimized a1, optimized a3)... Furthermore some changes seem to be incomplete, for instance here: https://github.com/kost123/hashcat/commit/3c9963caaa8596d1f43c5195fc7e1b6a51dab098

https://github.com/kost123/hashcat/blob/3c9963caaa8596d1f43c5195fc7e1b6a51dab098/OpenCL/m22500_a3-pure.cl#L229-L234

there is no "y" (0x79) added.

Would be great if we get some feedback here and maybe a reason if and why we should add this (or even better a PR ?). Thx

kost123:
Hey Phil, your code is fine as is(working great), I just needed to change it to "y" myself for a personal thing, it won't benefit anyone else remotely. Thanks for pointing out the incompleteness though, I didn't catch that.

philsmd:
thanks for the feedback/explanation :)

kost123:
@philsmd you don't happen to know if the multibit2john.py key hashes are safe to send do you(for others to attempt to crack)?

philsmd:
It's **NOT** safe to send these types of hashes to anyone you don't trust (or also sending it using an untrusted/unencrypted services is dangerous).

I already mentioned my thought and the facts a while back in here https://hashcat.net/forum/thread-8928-post-47437.html#pid47437
`I think even with the first 2 blocks (2*16=32 bytes) of the private key, you can steal the balance quite easily.... so I don't think the output of multibit2john is really meant to give to untrusted people`

The decrypted data is basically "just" the WIF format of the private key (starting with the first byte of the output). The encoding that is used is base58check (https://github.com/bitcoinj/bitcoinj/blob/2ec193f8479425c3a66bebf5f2d3493e39e88f7c/core/src/main/java/org/bitcoinj/core/Base58.java#L176-L177 , same for bitcoinj/MultiBit client software: https://github.com/Multibit-Legacy/multibit/blob/21e2e9d653d291a7dc36d21b6fc14b2a0da48985/src/main/java/org/multibit/file/PrivateKeysHandler.java#L339).

From the decrypted output I've seen during investigating these formats is that we always have something like this:
```
base58checkPrivateKey1 2020-03-06T15:21:45Z
base58checkPrivateKey2 2020-03-06T15:22:10Z
base58checkPrivateKey3 2020-03-06T15:23:59Z
```
etc
i.e. in each and every line there is always the base58 checksum (52 bytes), then a space, and after that a timestamp (date/time/timezone, "Zulu time" (UTC)) and they always end with "\x0a" (new line).

In theory, one could think to use these bytes and check for the ending, instead of the start of the decrypted output (or to exploit the PKCS#5 padding used for the AES CBC encryption/decryption), but the problem is exactly that the mode of operation is CBC: https://en.wikipedia.org/wiki/File:CBC_encryption.svg
and therefore you would need to have the other/previous bytes too to have the correct "last" IV (each block depends on the previous one unfortunately).

Therefore, I don't think there is much you can do here... only the first IV/salt is known and the later blocks (bytes with index > 16) need the "result" of the previous decrypt.

TL;DR: it's not safe at all to let others (untrusted 3rd parties) crack these types of hashes for you, because the first 2 blocks reveal 32 bytes of the private key... In theory you could say, it's "just" 32 of 52... which is just 60% of the private key, but actually the last 4 bytes for instance are just checksum etc... I wouldn't risk it.

You might be able to change the code that you reveal only the first 16 bytes (1st block), but this has the disadvantage to produce a lot of false positives (wrong results), but 16/52 would be probably less dangerous (it would be more difficult to brute-force the remaining part). I still wouldn't recommend it.... it's probably better to come up with a better strategy and try to crack them yourself (and try to remember as much as possible from the password).

kost123:
That sucks for me, but thank you so much for the write up, seems I am going to have to go about it myself.

kost123:
I believe I have found  a bug and that is that its not finding passwords for the "# keep your private keys safe" mode, it's also happening on JTR aswell, here is an example wallet
[final.txt](https://github.com/hashcat/hashcat/files/4341475/final.txt)

Btcrecover cracks it fine, but JTR and HC don't.

I'm also not sure if it's cracking the .org ones also but need more testing.

WELTEINZ:
Hi, is a $multibit$2-support planned in hashcat?

philsmd:
As far as I know, it uses scrypt which is a very slow algorithm and hardly benefits from GPU acceleration (well, it depends on the cost factor of course).

There is also no feature request for the scrypt multibit format, therefore nobody from the hashcat developer team (as far as I know) really looked into this. If you think it would make sense to support this in hashcat and there is some benefit in supporting this scrypt variant in hashcat (did you try to run the scrypt hash type -m 8900 with similar cost factors that are used for $multibit$2 ?), then please go ahead and open a github issue.
Thx

WELTEINZ:
Hi,

thank you for your fast answer!
I tried -m 8900 before, sadly does not work.
What do you mean by 'similar cost factors'.
I open the issue above like recommended.

Regards

philsmd:
As you can see here: https://hashcat.net/wiki/doku.php?id=example_hashes
for scrypt we have some N, r, p values at the beginning of the hash:
```
SCRYPT:1024:1:1....
```

John The Ripper (JTR) for instance tells us that $multibit$2... is using this parameters:
https://github.com/magnumripper/JohnTheRipper/blob/c18515e7136a8df8f8df128a1f73dd6755a3bb6b/src/multibit_fmt_plug.c#L374
N = 16384
r = 8
p = 1

Therefore, it would make sense to test a -m 8900 hash with these cost factors, parameters to get a rough idea about the speed and how well hashcat works with those hashes (of course for multibit there is more to it than just scrypt... but scrypt is the most costly part of the algorithm)

WELTEINZ:
I can't get to the point where I can test a speed because the hashes won't load with the prefixes I tried. If you want, I can send you the  hash which I read with JtR via PN. I may be doing something wrong..

PR #2322, opened 2020-03-06
solve $telegram$1 format conflict with JTR

philsmd:
We have recently added the mobile client Telegram APP algorithm to hashcat with https://github.com/hashcat/hashcat/pull/2282 .
Unfortunately nobody mentioned that there are 2 different algorithms in use (see the discussion in here: https://hashcat.net/forum/thread-8888-post-47765.html#pid47765).

The problem is that JTR already uses $telegram$1... for the Desktop algorithm, while it uses a dynamic format for the other mode (for which I suggest using $telegram$0 in hashcat).

The hash format basically remains the same as mentioned in here: https://github.com/hashcat/hashcat/pull/2282 , just using $telegram$0 as the signature+version number.

I've think we can't and shouldn't use $telegram$2 because it might probably give a further conflict in the future if any new format will be implemented in hc/jtr .

Thanks


jsteube:
OK. Btw, what happened to the $0 format? I can see from the JtR extraction tool there's no more $1 exported.

PR #2332, opened 2020-03-16
fixes #2067: 40-bit oldoffice false positive problem

### LINKED ISSUE #2067: open Office 2003 doc error: using 9820 pwd can't open doc file
use 9810 get office 2003 rc4 key
use 9820 get the password.
but can't use the password open the doc.
test file as attach
correct password is :^99998888~!

thanks.

[3.zip](https://github.com/hashcat/hashcat/files/3304303/3.zip)

philsmd:
As discussed in #2067 , we had several reports (forum/github) about "collisions"/false positives with the oldoffice formats using 40-bit keys:
-m 9800
-m 9820
(note: 9810 doesn't use the RC4 decryption step internally, it's the collider)

Thanks to @magnumripper we now also have an updated office2john.py (see https://github.com/magnumripper/JohnTheRipper/commit/400e959d7c3e00594b61f75088f9407aa826dc23 and https://github.com/magnumripper/JohnTheRipper/pull/4202) and there I would suggest that we also "allow" this new format (optional variant), which reduces false positives by checking the second "block" (512 bytes are the full blocks, we check the first 2 RC4 blocks = 2 * 16 = 32 bytes).

The modules and tests have been updated too, to check also allow/check these variants.

The hash format is basically the same, but just with a "*" separator and the 32 bytes in hexadecimal added to the old hash format. This variant is of course only a suggested format (but recommended to avoid false positives, therefore we suggest using latest office2john.py) and the old hashes will continue to work.

Thanks

### DIFF
--- a/OpenCL/m09800_a0-optimized.cl
+++ b/OpenCL/m09800_a0-optimized.cl
@@ -17,11 +17,15 @@
 #include "inc_hash_sha1.cl"
 #endif
 
+#define MIN_NULL_BYTES 10
+
 typedef struct oldoffice34
 {
   u32 version;
   u32 encryptedVerifier[4];
   u32 encryptedVerifierHash[5];
+  u32 secondBlockData[8];
+  u32 secondBlockLen;
   u32 rc4key[2];
 
 } oldoffice34_t;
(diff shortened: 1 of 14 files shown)


jsteube:
Many thanks @philsmd @magnumripper

PR #2350, opened 2020-03-31
fixes #2341: electrum 4/5 mod_512 () infinite loop fix

### LINKED ISSUE #2341: Hashcat freezes in end
Hashcat version **v5.1.0-1755-gddb641b8**
Progress freezes on **98.55** percentage. 
`hashcat64.exe --machine-readable --quiet --runtime=30 --restore-disable --potfile-disable --session=hashtopolis -p "\t" ..\\..\\hashlists\\19 -a 3 ?l?l?l?l?l?l --hash-type=21700 -o ..\\..\\hashlists\\19.out`

This situation on 21700 and 21800 algos

philsmd:
Fixes a problem reported in #2341 and https://hashcat.net/forum/thread-9081.html which lead to a infinite loop only when testing with some specific password candidates (which made mod_512 () loop forever).

Thanks for the reports guys


ahvahsky2008:
Where download release files?

ahvahsky2008:
https://hashcat.net/beta/ when we see it in beta

PR #2353, opened 2020-03-31
Patch all inconsistent separators in modules

s3inlc:
After discovering a module where the separator was not set to hashconfig->separator consistently, I went through all modules and fixed all of them where this was also the case.


jsteube:
Thank you

PR #2372, opened 2020-04-23
Allow module_hash_binary_parse to report a fatal error

mpalmer:
If module_hash_binary_parse is completely unable to successfully parse out
any hashes, up until now the output has been

```
Hashfile 'foo': Success
```

which is less than helpful.

This patch allows (but does not require) m_h_binary_parse to report a useful
error response, by returning a negative value.  Modules which continue to
return '0 hashes' will get the same less-than-useful behaviour they always
hace.

I've also modified the LUKS module to report a useful error, as a proof of
concept.

Further expansions on this could include:

* Applying similar behaviour to module_hash_binary_count, so it too can
  report errors when trying to count hashes.  This would require more
  co-ordinated change, because m_h_binary_count already uses -1 to indicate
  a system error.

* Allow and encourage modules to print their own errors and warnings during
  parsing.  This would allow for situations where a single hash in a
  multi-hash file could be reported as malformed, without having to fail the
  whole parse.  However, implementing this would, I expect, require modules
  to have access to `hashcat_ctx`, which...  yeah.  Not so straightforward.

### DIFF
--- a/src/hashes.c
+++ b/src/hashes.c
@@ -1296,9 +1296,13 @@ int hashes_init_stage1 (hashcat_ctx_t *hashcat_ctx)
         {
           hashes_cnt = hashes_parsed;
         }
-        else
+        else if (hashes_parsed == 0)
         {
           event_log_warning (hashcat_ctx, "Hashfile '%s': %s", hashes->hashfile, strerror (errno));
+        }
+        else
+        {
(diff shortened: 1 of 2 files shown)


jsteube:
This is great, many thanks for contribution

PR #2373, opened 2020-04-23
Add some more generic error messages

mpalmer:
These are almost all taken from error messages that are specific to one particular module, and turned into generic errors that can be used for any module.  It seemed like a better idea to provide generic messages than to encourage infinite proliferation of error codes (which would probably end up blowing the 255 value threshold sooner or later).  It doesn't seem necessary to provide module-specific error messages for things like "Invalid <X> filesize", since users should already know what sort of file they're asking to be parsed.

These error messages are required for better parsing failure reporting for my PKCS#1 module.

I haven't cleaned up existing uses of module-specific error messages, because that would change existing behaviour, but I can certainly do that (in this or another PR) if it should be deemed proper.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -534,6 +534,10 @@ typedef enum parser_rc
   PARSER_TOKEN_LENGTH         = -35,
   PARSER_INSUFFICIENT_ENTROPY = -36,
   PARSER_PKZIP_CT_UNMATCHED   = -37,
+  PARSER_KEY_SIZE             = -38,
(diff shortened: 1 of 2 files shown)


jsteube:
You are right, these are mostly relics of a time when the code was not modularized and no general concept existed. But now that hashcat has this plugin interface it makes sense to remove the module specific error codes such as PARSER_LUKS_FILE_SIZE. Good thing is it should not affect too many plugins.

Issue #2379, opened 2020-04-28
Overwriting TMP environment variable breaks AMD ROCm in Linux


ahesford:
In [`src/folder.c`](https://github.com/hashcat/hashcat/blob/c5d2d5396fb7324ab60465c46a27dc8452ff2761/src/folder.c#L465), a comment notes that a good workaround for problematic handling of `-I` parameters is to set the `TMP` environment variable before loading the runtime. This is not the case for the AMD ROCm runtime in Linux, at least with the latest version (3.3.0), unless hashcat is run as the superuser.

The [ROCm Code Object Manager](https://github.com/RadeonOpenCompute/ROCm-CompilerSupport) and [ROCm OpenCL Runtime](https://github.com/RadeonOpenCompute/ROCm-OpenCL-Runtime) use the value of the `TMP` environment variable to determine where compiled objects should be placed. [Lines 465 and 467 of `src/folder.c`](https://github.com/hashcat/hashcat/blob/c5d2d5396fb7324ab60465c46a27dc8452ff2761/src/folder.c#L465) set this value to `${PREFIX}/share/hashcat/OpenCL`, which is not typically writeable by ordinary users. Thus, OpenCL kernel compilation fails when the build products can not be stored.

While I'm not sure about other OpenCL runtimes (it certainly doesn't seem to hurt ocl-icd with the amdgpu-pro OpenCL ICD), the proper behavior for ROCm is either to leave `TMP` alone or explicitly set it to somewhere like `/tmp` that can be written by any user running hashcat.

jsteube:
That's not the case for me. You are right that TMP is set by hashcat as a workaround for buggy OpenCL runtimes, but this has no effect on ROCM. 

I've just tested this locally with a normal (not superuser) user.

```
$ strace -f -v -o x hashcat -b -m 0
...
$ grep local/share x
```

Output:

```
[...]

PR #2387, opened 2020-05-07
Fix POSIX temporary directory semantics

ahesford:
This change sets the value of `TMPDIR` on POSIX systems to restore correct temporary directory semantics when overwriting `TMP` as a workaround for poor OpenCL implementations. It is an attempt to address #2379. 

When setting `TMP` to a non-writable directory, ROCm 3.3.0 breaks on a Void Linux installation because the OpenCL compiler tries to write temporary files to `$TMP` unless the POSIX-ly correct `TMPDIR` is set (`TMPDIR` supersedes `TMP` in this instance). The patch makes the following behavior changes *only* for POSIX systems:
1. If `TMPDIR` is already set, do nothing; otherwise,
2. If `TMP` is set before it is overridden in hashcat, copy the original value of `TMP` to the `TMPDIR` environment value; otherwise,
3. Enforce an explicit default `TMPDIR=/tmp` to restore expected POSIX temporary directory behavior.

I have no idea if `TMPDIR` supersedes `TMP` on these broken OpenCL implementations because I have none to test. If so, this probably negates the workaround. However, if the broken implementations tend to be on Windows, maybe this isn't an issue anyway.

If any OpenCL implementations break as a result of this change, I recommend adding something like a `--broken-opencl` command argument that would need to be specified before employing the `TMP` overwrite hack. That way, temporary directory semantics won't be broken by default.

### DIFF
--- a/src/folder.c
+++ b/src/folder.c
@@ -462,9 +462,24 @@ int folder_config_init (hashcat_ctx_t *hashcat_ctx, MAYBE_UNUSED const char *ins
   {
     char *tmp;
 
-    hc_asprintf (&tmp, "TMP=%s", cpath_real);
+    #if defined (_POSIX)
+    // Setting TMP may break POSIX temporary directory semantics
+    // Fix it by setting the POSIX-ly correct TMPDIR if it is not already set
+    if (getenv ("TMPDIR") == NULL) {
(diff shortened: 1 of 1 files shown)


ahesford:
I believe this is obsolete with the decision to remove the `TMP` workaround.

PR #2395, opened 2020-05-11
Adding escaping for --status-json

### LINKED ISSUE #2369: --status-json doesnt escape " in target and produces invalid json
example target md5($pass.$salt) a9f52a88089bc71e185a2e027a35417c:E#6dsdn[z/K""&|OvS

s3inlc:
closes #2369

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -99,6 +99,7 @@
 - Fixed some memory leaks in case mask-files are used in optimized mode
 - Fixed the 7-Zip parser to allow the entire supported range of encrypted and decrypted data lengths
 - Fixed the validation of the --brain-client-features command line argument (only values 1, 2 or 3 are allowed)
+- Fixed --status-json to correctly escape certain characters in hashes
 
 ##
 ## Improvements
--- a/src/terminal.c
+++ b/src/terminal.c
@@ -1020,15 +1020,41 @@ void status_display_status_json (hashcat_ctx_t *hashcat_ctx)
     end = time_now + sec_etc;
   }
 
+  char *tmp_target = (char *) hcmalloc (strlen (hashcat_status->hash_target) * 2);
+
+  unsigned long i, j;
+
+  for (i = 0, j = 0; j < strlen (hashcat_status->hash_target); i++, j++)
+  {
+    char rep = hashcat_status->hash_target[j];
+
+    switch (hashcat_status->hash_target[j])
+    {
+      case '\b': rep =  'b'; tmp_target[i] = '\\'; i++; break;
+      case '\t': rep =  't'; tmp_target[i] = '\\'; i++; break;
+      case '\n': rep =  'n'; tmp_target[i] = '\\'; i++; break;
+      case '\f': rep =  'f'; tmp_target[i] = '\\'; i++; break;
+      case '\r': rep =  'r'; tmp_target[i] = '\\'; i++; break;
+      case '\\': rep = '\\'; tmp_target[i] = '\\'; i++; break;
+      case  '"': rep =  '"'; tmp_target[i] = '\\'; i++; break;
+    }
+
+    tmp_target[i] = rep;
+  }
+
+  tmp_target[i] = 0;
+
   printf ("{ \"session\": \"%s\",", hashcat_status->session);
   printf (" \"status\": %d,", hashcat_status->status_number);
-  printf (" \"target\": \"%s\",", hashcat_status->hash_target);
+  printf (" \"target\": \"%s\",", tmp_target);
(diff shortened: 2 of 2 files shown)


philsmd:
Hey @s3inlc , this is a fantastic, very elegant solution. I like it a lot !

I must say, that I didn't test it yet... maybe also @Chick3nman could have a look at it (since he implemented basically all the JSON output support), that would be great if we double-check it... but again, I don't suspect any problems, it just looks great .


If I really would have to find and choose something in the patch that could/should be changed,  I probably would come up with some minor suggestions/review points:
1. out of curiosity, what does "rep" stand for ? We normally just use "c" for "character" or something like this to represent a single "extracted char" (byte etc).
2. should we rename tmp_target in something more meaningful like "hash_json_encoded" (or even better name ?)
3. why don't you just use "switch (rep)" or if changed "switch (c)" ?
4. this is of course very minor problem: but normally we use the "i" variable for the main loop variable and j,k,l etc for the secondary variables... e.g. we would use "char c = hashcat_status->hash_target[i];" (i.e. switch every j with i)
5. should we also add a short comment above this block of code to mention *WHY* we (need to) do this (because of JSON encoding rules) and maybe if available link to some JSON spec or table of chars etc (I didn't look for it, but maybe something like this is available and can be linked to by including an URL). 
6. "free(tmp_target);" should be "free (tmp_target);" (add a space)

Again, this are just some of my few observation... but these are very, very minor nit-picking "problems". It would be completely fine, if we just would merge it like this... (we just sometimes tend to get very strict on formatting/naming convention/coding style... to make the code looks nicer and easier to read... I know it's sometimes annoying, but again please do not misunderstand me, the code is actually really nice... I like how compact that block is !!! ).

Chick3nman:
I believe "rep" in this case stand for replace or replacement? Just guessing on that though, will need confirmation from @s3inlc. 

As for the fix, I may suggest moving this to a slightly different area in the code. I believe the values in `hashcat_status->hash_target` at the time of replacement will have already passed through the line compression found here: https://github.com/hashcat/hashcat/blob/master/src/status.c#L342

I'm not sure if returning the compressed line in this case even makes sense, though it does mean that excessively long hashes such as 7zip hashes will at least be manageable in the output. To me, it makes sense to do the replacement in a conditional above the line linked above, catching both the default and non-default Module_Encode cases, so an uncompressed but still manageable hash can be returned instead. If someone is using --status-json, i would think they would want the full target value if only a single hash is being run, except of course in the case of excessive length hashes. Perhaps a less restrictive compression could be added that allows for full length hashes out to say, 200 characters, but not to the extreme sizes of 7zip or other long hash types.

s3inlc:
Thanks for the feedback, let me go through the unclear points :)

1. As @Chick3nman suspected, 'rep' was thought as short for 'replacement'. But of course, if you prefer to use a more common naming, I can rename it to 'c'.
2. Agreed, the name is not telling very much, something like your suggestion sounds much better.
3. I guess that's a small leftover as the first approach was slightly differently structured. Of course this should be replaced with rep/c.
4. Makes sense, I can switch these two names then.
5. I can add a comment explaining the reason for this special handling. I took the list of characters to escape from here: https://www.freeformatter.com/json-escape.html and I added all of them for the sake of completion, even if probably something like backspace will never occur in a hash.
6. Damn, I tried to make all spaces correctly, this one I overlooked :)

Coming to the point of @Chick3nman. I cannot imagine any reason why someone would need to have the target in the status output in a longer (or complete) form. In the normal --machine-readable flag, the target is not even included, and obviously, any program reading from that status output should read the hash (if needed) from the hash file which was given to hashcat.

From my understanding of the proposal and what I read on the issue (correct me if I'm wrong), moving this encoding to an earlier stage would cause that a lot more code is affected by this change and also for all further modules which are not using the default module encode need to be adapted for that (someone needs to make sure that for new modules this is handled properly).

philsmd:
good points. I also wouldn't bother too much about the "complete hash" if it makes the code much more complicated.

I think this is actually a quite nice solution for a quick "JSON-parsing-problem" fix... if users really need a feature to have the full hash, they should probably later on request it on github (but I kind of agree with @s3inlc that the full one is not needed, it would make the JSON output longer and it would also  be kind of inconsistent compared to the "normal" status prompt)...

To avoid making things too complicated, I would say this version is fixing the main problem quite nicely. I think both of you agree and we can merge it as a hotfix.

Thx

s3inlc:
I now updated the code based on @philsmd's comments, let me know if there is anything else missing. Otherwise it should be good to be merged.

Chick3nman:
I am fine with this fix being merged, the point that MACHINE READABLE doesn't even give target information is convincing enough for me not to want the uncompressed password in the output. If you only have 1 candidate, in theory you don't need that information in the status anyway.

Good to go.

jsteube:
Looks very good, thanks for contribution.

PR #2407, opened 2020-05-19
Blowfish Advanced CS module

mpalmer:
https://hashcat.net/forum/thread-9205.html

(Depends on #2408)

### DIFF
--- a/OpenCL/inc_bfacs_common.cl
+++ b/OpenCL/inc_bfacs_common.cl
@@ -1,0 +1,627 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "inc_types.h"
+#include "inc_hash_sha1.cl"
+
+#define BFACS_ENCRYPTED_MAGIC 0x75191114
+#define BFACS_SALT_LEN 11
+
+typedef struct bfacs
+{
+  u32 block_size;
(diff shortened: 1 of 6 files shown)


mpalmer:
I think that the appveyor failure is probably spurious:

```
error: could not open file /var/cache/pacman/pkg/msys2-runtime-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
error: could not open file /var/cache/pacman/pkg/msys2-runtime-devel-3.1.4-1-x86_64.pkg.tar.zst: Child process exited with status 127
```

jsteube:
Two numbers in here disagree. The module declares KERN_TYPE 24300 while the kernels are named m24308, so hashcat looks for "m24300_a0-pure.cl", which is not there. 24300 has since been taken as well, by sha1($salt.sha1($pass.$salt)), so the number the module names is no longer free either.

Five years have also moved underneath it. There is no test module, which every mode carries now, and the kernel interface, the module hooks and the comparison path have all changed shape since 2020. What is here would be rewritten against the current tree rather than rebased onto it.

So I am closing this one. The work is not the problem, and Blowfish Advanced CS is still unsupported, so a fresh PR against master would be welcome if you are still interested. Thank you for putting it up in the first place.

PR #2423, opened 2020-05-27
added WSL build documentation

thatux:
This explains how you can use WSL to build hashcat for Windows.

### DIFF
--- a/BUILD_WSL.md
+++ b/BUILD_WSL.md
@@ -1,0 +1,31 @@
+# Compiling hashcat for Windows with Windows Subsystem for Linux.
+
+Tested on a Windows 10 (x64 build 18363), should also work to build hashcat for Windows on Linux.
+
+### Installation ###
+
+Enable WSL.
+
+Run bash (win+r bash) and make sure to install additional dependencies necessary for hashcat compilation
+```
+sudo apt install gcc-mingw-w64-x86-64 make git
+git clone https://github.com/hashcat/hashcat
+git clone https://github.com/win-iconv/win-iconv 
+cp hashcat/tools/win-iconv-64.diff win-iconv/
+cd win-iconv/
+sudo make install
+```
+
+### Building ###
+
+You've already cloned the latest master revision of hashcat repository above, so switch to the folder and type "make win" to start compiling hashcat
+```
+cd ../hashcat
+make win
+```
+
+The process may take a while, please be patient. Once it's finished, run a Windows command prompt (win+r cmd) and start hashcat by typing "hashcat.exe"
+
+```
+hashcat.exe
+```
--- a/BUILD_WSL.md
+++ b/BUILD_WSL.md
@@ -1,31 +1,36 @@
 # Compiling hashcat for Windows with Windows Subsystem for Linux.
 
-Tested on a Windows 10 (x64 build 18363), should also work to build hashcat for Windows on Linux.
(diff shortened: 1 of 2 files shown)


philsmd:
Thanks for the contribution. I have some minor points that I would like to understand and probably change (I will review the source code lines below):

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
+# Compiling hashcat for Windows with Windows Subsystem for Linux.
+
+Tested on a Windows 10 (x64 build 18363), should also work to build hashcat for Windows on Linux.
```
philsmd: should we really mention a build version here ? I think it will look outdated very soon, within some months/years. I think mentioning windows 10 64-bit is enough

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Installation ###
+
+Enable WSL.
+
+Run bash (win+r bash) and make sure to install additional dependencies necessary for hashcat compilation
```
philsmd: press the win + r key on your keyboard simultaneously and in the "Run" popup window type bash

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
...
+```
+sudo apt install gcc-mingw-w64-x86-64 make git
+git clone https://github.com/hashcat/hashcat
+git clone https://github.com/win-iconv/win-iconv 
+cp hashcat/tools/win-iconv-64.diff win-iconv/
```
philsmd: why copy this diff file without applying it with the "patch" command ? this step makes no sense to just copy a diff file... There is something missing here like "patch < win-iconv-64.diff" etc... but this needs to be tested. The way it's explained here to just copy a file makes no sense

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
...
+cd win-iconv/
+sudo make install
+```
```
philsmd, line 18: I would prefer to use "cd ../" here to complete these "prerequisite" steps

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Building ###
+
+You've already cloned the latest master revision of hashcat repository above, so switch to the folder and type "make win" to start compiling hashcat
+```
+cd ../hashcat
```
philsmd: if we agree to use cd ../ above, we only would need to use "cd hashcat/" in here. Does this make sense to you ?

BUILD_WSL.md
```diff
@@ -0,0 +1,31 @@
...
+make win
+```
+
+The process may take a while, please be patient. Once it's finished, run a Windows command prompt (win+r cmd) and start hashcat by typing "hashcat.exe"
```
philsmd: press the win + r keys together and type cmd in the popup window

philsmd (changes requested):

BUILD_WSL.md
thatux: sure will remove, I'd guess that this will stay true for multiple versions

BUILD_WSL.md
thatux: good spot, copy pasted wrong probably

thatux:
Thank for your comments, I think I've resolved your comments?

jsteube:
Thanks for the contribution @thatux

PR #2425, opened 2020-05-31
fixes #2410: added -m 2300x = SecureZIP

### LINKED ISSUE #2410: Feature Request - Add support for zip3 hashes ?
zip2john outputs a new (?) $zip3$ hash : "The new format for PKWARE's Strong Encryption Specification".

https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/src/zip2john.c#L96

Can you please add support for this hash?
Thanks.

philsmd:
This PR adds AES-192 on GPU and implements issue #2410 = SecureZIP.

There are 3 variants: 128-bit AES, 192-bit AES, 256-bit AES. For performance reasons, I've splitted all of these variants into separate hash types (because if you need less bits, you also need to perform less SHA1 transforms and different decryption algo, see algo description in #2410 ).

Implementing this took again a lot of time (18 different/separate highly-optimized OpenCL kernels for 3 variants * pure/optimized * a0/a1/a3 = 3 * 2 * 3 = 18 kernel files).

As always, tests have also been added to this commit to check if everything is working correctly.

Hope this work is helpful to at least a couple of hashcat users, feedback is much appreciated. Thank you very much

### DIFF
--- a/OpenCL/inc_cipher_aes.cl
+++ b/OpenCL/inc_cipher_aes.cl
@@ -1040,6 +1040,389 @@ DECLSPEC void aes128_decrypt (const u32 *ks, const u32 *in, u32 *out, SHM_TYPE u
   out[3] = hc_swap32_S (out[3]);
 }
 
+// 192 bit key
+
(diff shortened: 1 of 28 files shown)


jsteube:
All tests passed: NV, AMD, Intel, both optimized and pure. Many thanks!

PR #2434, opened 2020-06-05
Added option --verify-hashlist-only

davidbolvansky:
This option allows to use hashcat as a "hash format validator" tool. 
Hashcat can verify hash format without need for opencl drivers, but currently it cannot do it on devices without opencl drivers and fails with an error.

This patch introduces new option to run only "hashlist parser" - this means that we don't need opencl at all.

PS: I am not very familiar with codebase and I just spend some time looking at the codebase to produce minimal code to implement this feature..


davidbolvansky:
Ok, it seems --show hashlist -m hashtype works too. This patch is not needed.

PR #2458, opened 2020-06-17
Allow GNU/Linux maintainers to select options without patching

xvitaly:
Allow GNU/Linux maintainers to select options without patching of Makefile.


jsteube:
Thanks!

PR #2459, opened 2020-06-17
Fixed linkage against packaged version of minizip (part of zlib)

xvitaly:
Fixed linkage against packaged version of minizip (part of zlib).


jsteube:
I think the change is correct, but additional changes need to made. I will merge this and commit the other changes needed.

PR #2466, opened 2020-06-22
Re-implement alias check

singe:
This is a repeat of
https://github.com/hashcat/hashcat/commit/57a149276c5a20d5af1b2364628b3a305772ab3d
which was undone in this commit
https://github.com/singe/hashcat/commit/34f71aaea3550d146b4a3f2a7b55e140e9bd3da5#diff-7af547b52769371de74af0bfca268a62R6599

The original issue was #2083 and discussion of a fix was #2086.

The problem is with Apple devices with an Intel and AMD GPU and using both
simultaneously and the alias check.

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -6625,8 +6625,8 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
 
   // find duplicate devices
 
-  //if ((cuda_devices_cnt > 0) && (opencl_devices_cnt > 0))
-  //{
+  if ((cuda_devices_cnt > 0) && (opencl_devices_cnt > 0))
+  {
     // using force here enables both devices, which is the worst possible outcome
     // many users force by default, so this is not a good idea
 
@@ -6634,7 +6634,7 @@ int backend_ctx_devices_init (hashcat_ctx_t *hashcat_ctx, const int comptime)
     //{
     backend_ctx_find_alias_devices (hashcat_ctx);
     //{
-  //}
+  }
 
   if (backend_ctx->backend_devices_active == 0)
   {


philsmd:
These are the cases where we need to be very careful.

If there was some commit that changed something, it most of times was because of a **very good reason**. In this case I'm pretty sure that we discovered that it's not a good test that we need to have both a CUDA **and** a OpenCL device, because we could also have just 2 OpenCL Runtimes for CPUs (like Intel and POCL driver) that are aliases (same device, but different OpenCL platform and/or driver).

It's interesting that also recently we got a hashcat forum post about (I think) the same problem: https://hashcat.net/forum/thread-9321-post-49267.html#pid49267 (but different user reporting it). I suggested that we should search again about some key differences between the devices and find out if we can reliable distinguish them (but of course we would also need to make sure this is not a macOS Apple fix that breaks other systems, like windows drivers, linux drivers, ROCm, POCL, Intel CPU drivers etc etc etc... unfortunately there are many different configurations we need to think about, because hashcat supports a variety of systems/OpenCL platforms/device types etc).

I think the diff mentioned there could be something we could investigate a little more, but something  similar was actually already mentioned here: https://github.com/hashcat/hashcat/pull/2086#issuecomment-516020380 . I'm not sure if any user tested this diff, because the following posts seem to only focus on the `src->opencl_device_vendor_id != dst->opencl_device_vendor_id` change. Maybe nobody tested a `device_processors` check. I don't know.

Again, if possible I would like that we come up with a fix that tries to fix the **root** of the problem, not just work around it.... if we find nothing, it's probably better to say on macOS Apple devices we consider all the devices different (that would be quite radical, but it should work most of the time, I guess, because you can't really install POCL or ROCm etc on Apple, as far as I know... not sure about other alternative OpenCL Runtimes... but my guess is that on Apple everything most be allowed/signed by Apple, everything that deals with those low-level hardware stuff, so it's kind of unlikely that alternative drivers exist on macOS... I could be wrong: I'm not an macOS expert)

jsteube:
Please retry with hashcat 6.0.0 + 32 beta from https://hashcat.net/beta/ and reopen the issue if it still exists.

PR #2471, opened 2020-06-25
Makefile: update clean/distclean targets

philsmd:
With this commit I suggest some minor changes to the `src/Makefile` which make the clean vs distclean target in the Makefile more meaningful.

In my opinion it makes sense to keep the .restore file and .potfile while debugging / testing, but if an user chooses to `make distclean` it can/should be deleted.

Repetitions between `distclean` and `clean` do not make sense, because `clean` is a dependency of `distclean` (and therefore it's redundant and therefore I removed instances of repetition between clean and distclean).

This change also added the test files `test_${unix_timestamp}` to the `distclean` target, because they can get annoying sometimes and I think it makes sense to delete these folders with `distclean`.

What do you think ? It seems good to me.

Thanks

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -374,7 +374,10 @@ distclean: clean
 	$(RM) -f *.restore
 	$(RM) -f *.potfile
 	$(RM) -f *.out
-	$(RM) -rf test_[0-9]*/
+	$(RM) -f hashcat.dictstat2
+	$(RM) -f brain.*
+	$(RM) -rf test_[0-9]*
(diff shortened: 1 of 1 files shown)


jsteube:
Looks good to me. In addition to distclean we can add these files, what do you think?

```
- hashcat.dictstat2
- tools/luks_tests
- *.admp or brain.*
```

philsmd:
I totally agree, added these 3 files/folders to the distclean Makefile target with the latest commit above

PR #2523, opened 2020-08-16
-m 21200 bugfix: Array declared too small.

magnumripper:
We're using indeces 0..4 so it has to be `u32 pc[5]`.


jsteube:
Nice find, thanks!

PR #2531, opened 2020-08-26
fixes #2518: call clear_prompt () more often to avoid misaligned prompt

### LINKED ISSUE #2518: status line sometimes showing to the right of a found
It's not consistent, but recently, hashcat is now occasionally putting the status info to the right of one of the founds, like this (hashes and founds redacted):
```
Hardware.Mon.#3..: Temp: 64c Fan:100% Util:100% Core:1809MHz Mem:4513MHz Bus:16
Hardware.Mon.#4..: Temp: 62c Fan:100% Util: 90% Core:1847MHz Mem:4513MHz Bus:4
Hardware.Mon.#5..: Temp: 54c Fan:100% Util:  0% Core:1847MHz Mem:4513MHz Bus:1
Hardware.Mon.#6..: Temp: 56c Fan:100% Util:  0% Core:1860MHz Mem:4513MHz Bus:1

Cracking performance lower than expected?

* Update your backend API runtime / driver the right way:
  https://hashcat.net/faq/wrongdriver

* Create more work items to make use of your parallelization power:
  https://hashcat.net/faq/morework

HHHHHHHHHHHHH:ffffffff                           [s]tatus [p]ause [b]ypass [c]heckpoint [q]uit =>
HHHHHHHHHHHHH:ffffffff
HHHHHHHHHHHHH:ffffffff
HHHHHHHHHHHHH:ffffffff
HHHHHHHHHHHHH:ffffffff
```

philsmd:
As mentioned in #2518 and very intensively tested / debugged by @roycewilliams (thank you very much), we have/had a minor cosmetic problem when some warning messages interfered and popped up almost simultanously with the status prompt.

This problem also occurred more often when using `--increment` with increasing mask length and/or by using the `.hcmask` feature (several mask running quickly one after the other). When doing so the prompt wasn't really cleared/gone after one length/mask was finished and therefore the interference happened (rarely, but still annoying).

This fix of course is not meant to be 100% guaranteed to be complete, because in theory we could have many more warnings/infos popping up... but it's definitely addressing very many (hopefully almost all) of the cases where simultanous updates to the user interface (i.e. output in the shell) could happen.

Thanks again to royce for testing

:+1:

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -18,6 +18,7 @@
 - Fixed internal access on module option attribute OPTS_TYPE_SUGGEST_KG with the result that it was unused
 - Fixed race condition resulting in out of memory error on startup if multiple hashcat instances are started at the same time
 - Fixed unexpected non-unique salts in multi-hash cracking in Bitcoin/Litecoin wallet.dat module which lead to false negatives
+- Fixed rare case of misalignment of the status prompt when other user warnings are shown within the hashcat output
 
 ##
 ## Improvements
--- a/src/backend.c
+++ b/src/backend.c
@@ -23,6 +23,7 @@
 #include "event.h"
 #include "dynloader.h"
 #include "backend.h"
+#include "terminal.h"
 
 #if defined (__linux__)
 static const char *dri_card0_path = "/dev/dri/card0";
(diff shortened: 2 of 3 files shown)


roycewilliams:
Looks like it addressed my issue - thanks!

PR #2535, opened 2020-08-26
Dev guide --force edit and printf addition

thatux:
add --force to the CLI example in the dev-guide and an example that prevented me to see my printfs when developing a kernel.

### DIFF
--- a/docs/hashcat-plugin-development-guide.md
+++ b/docs/hashcat-plugin-development-guide.md
@@ -537,20 +537,24 @@ Additionally there is a couple of command line parameters that you want to use:
 * -d 1: In case you have multi compute devices in your system, limit it to a single compute device. This is to reduce startup and JiT compile time.
 
 Typically a developer command line for hashcat looks the following:
+
 ```
 $ rm -rf kernels $HOME/.nv; ./hashcat -m XXXXX hash.txt word.txt --potfile-disable --self-test-disable -n 1 -u 1 -T 1 --quiet --backend-vector-width 1 -d 1 --force
 ```
 
-If you need to printf from without a _loop kernel, keep in mind that you need to add a branching manually for a specific loop position.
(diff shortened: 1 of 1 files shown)


jsteube:
The first correction with the --force is good to merge, but the second commit seems pretty clear to me. Could you please rewrite that section and only write a short note about that people keep in mind that this is all zero based that should be sufficient.

docs/hashcat-plugin-development-guide.md
```diff
@@ -537,17 +537,21 @@ Additionally there is a couple of command line parameters that you want to use:
...
+$ rm -rf kernels $HOME/.nv; ./hashcat -m XXXXX hash.txt word.txt --potfile-disable --self-test-disable -n 1 -u 1 -T 1 --quiet --backend-vector-width 1 -d 1 --force
 ```
 
 If you need to printf from without a _loop kernel, keep in mind that you need to add a branching manually for a specific loop position.
-
 ```
 if ((loop_pos + i) == 0) printf ("%08x\n", a);
 ```
 
+When using  -n 1 -u 1 -T 1 you shouldn't add
```
philsmd: I agree with @jsteube here: it's better to avoid "shouldn't" and we should avoid telling the user what NOT to do (the problem is that users sometimes find code and just copy-paste it, yeah it's bad, but it's unfortunately the truth/case most of the time). Instead, it would make more sense to mention how to do it correclty in the guide and emphasize that the `gid`/`lid` are zero-based.

Please also stick to the coding style we use:

```
if ((gid == 0) && (lid == 0))
{
  printf ("%08x\n", a);
}
```

I would suggest to just show how to do printf and afterwards mention something like this: "Please note that the variables gid and lid are zero-based and therefore we start counting with 0", or something like this.

Thank you very much

thatux:
Agreed, I've added an example for a kernel without _loop, this removes the need for explaining gid an lid are zero indexed. I've used the same codestyle as already present in the file.

PR #2542, opened 2020-09-08
Fixes #1334: Added -m 23700/23800 = RAR3-p cracking

### LINKED ISSUE #1334: [Feature Request] $RAR3$*1 Support
Just tried this type of hash today only to find out it doesn't work :( Is there any plan on implementing support for this?

philsmd:
This new pull request adds both uncompressed (-m 23700) and compressed (-m 23800) password-protected RAR3 cracking support (where the headers are **not encrypted**, RAR3-hp was already supported by -m 12500), as requested in #1334 

```
  23700 | RAR3-p (Uncompressed)                            | Archives
  23800 | RAR3-p (Compressed)                              | Archives
```

The -m 23700 mode computes everything on GPU and therefore no decompression is needed, while -m 23800 uses the UnRAR dependency and a hook function to decrypt & decompress the data before computing the CRC32 checksum.

This (especially -m 23800) is of course a much more advanced algo to implement and involved a lot of try-and-error and a lot of testing. A huge thanks goes to @jsteube for helping to get everything working and also making several changes in the architecture (like module_hook_extra_param_init () and module_hook_extra_param_term (), see https://github.com/hashcat/hashcat/commit/98aef2a) to facilitate the implementation of -m 23800 with the new hook function.

The -m 23800 hash type of course depends also on your CPU speed (UnRAR, buffer on-the-fly decompression) and therefore it's recommended to have all: enough RAM, a fast CPU and fast GPU.

Thank you very much


jsteube:
Great work, many thanks!

mblackshore:
Does this mean that after you create the new binaries we will be able to use $RAR3$*1? :O 
Do I need to download unrar from rarlab or are you able place that into package?

philsmd:
As always, it's enough to use the latest beta version from https://hashcat.net/beta . This already has support for the new hash types and includes all the code needed (that's why we had to double check the copyright situation etc etc, don't you remember ?).

It's important to remove the file names from the output of `rar2john` such that the examples look similar to the ones mentioned here https://hashcat.net/wiki/example_hashes or the example hash from:
```
hashcat -m 23800 --example-hashes
```

mblackshore:
Aha, you are also allowed to redistribute the code from rarlab. I was unaware of that. I thought you just created a dependency.
 
So if I had a compressed rar filed which I extracted the hash from I would need to open up the outputed .lst file and remove the filename and create a new file and then run it?

So if I wanted to run a bruteforce operation with for RAR3-p (Compressed) with lowercase, uppercase, numbers and symbols should it look like this?

hashcat.exe -m 23800 -a3 newhash.lst --increment ?a


Do I need to do something inorder save a session like John?
From what I can find I close the active session as needed and then I type hashcat --restore inorder to continue? where is this work file saved and how does hashcat know which work file to choose? Does the Pseudo CPU\GPU solution save this temp file differently?


`
  ? | Charset
 ===+=========
  l | abcdefghijklmnopqrstuvwxyz
  u | ABCDEFGHIJKLMNOPQRSTUVWXYZ
  d | 0123456789
  h | 0123456789abcdef
  H | 0123456789ABCDEF
  s |  !"#$%&'()*+,-./:;<=>?@[\]^_{|}~
  a | ?l?u?d?s
  b | 0x00 - 0xff
`

mblackshore:
I tried to run and I got this:
What should I do?

`

===========================================
hashcat.exe -m 23800 -a3 NewHash.lst --increment ?a


hashcat (v6.1.1-83-g90fb4aad) starting...

Successfully initialized NVIDIA CUDA library.

Failed to initialize NVIDIA RTC library.

* Device #1: CUDA SDK Toolkit installation NOT detected or incorrectly installed.
             CUDA SDK Toolkit installation required for proper device support and utilization
             Falling back to OpenCL Runtime

* Device #1: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
Successfully initialized NVIDIA CUDA library.


OpenCL API (OpenCL 1.2 CUDA 11.0.140) - Platform #1 [NVIDIA Corporation]
========================================================================
* Device #1: GeForce GTX 1070, 7936/8192 MB (2048 MB allocatable), 15MCU


Maximum password length supported by kernel: 127

Counted lines in NewHash.lst...
Oversized line detected! Truncated 536570967 bytes

Oversized line detected! Truncated 536570967 bytes
Hashfile 'NewHash.lst' on line 1 ($RAR3$...[TEXTREPLACEMENT]): Separator unmatched
No hashes loaded.
`

philsmd:
I think this is not the best place to discuss problems like this. Most importantly, because this is a closed/merged pull request, but also because these seem just some misunderstandings and user errors.

It would make much more sense to search for or mention your problems in the hashcat forum: https://hashcat.net/forum/

I will mention these 2 things anyway, just to make this reply not completely useless: the first problem seems that you just didn't have applied the `timeout patch` and you also didn't really install the `CUDA Toolkit` from https://developer.nvidia.com/cuda-downloads.
The second problem is that your file is veeeeeeery long and hashcat only supports a maximum hash line length of arounds 320 KiB. This is kind of a hard limit, because the (hash) line lengths and buffers are restricted to these sizes. I think this is acceptable in the most cases, but of course there could always be exceptions (the good think is, that compressed data is actually most of the time quite small and the rar files allow to select single files and therefore you normally could just use the smallest file if there are multiple files and others are "too long"). I don't think we can change this hard limit very easily and this is not a `rar3-p` only "problem" but other hash types like 7-zip/winzip etc have the same limit. Thx

sagetoo:
Hi @philsmd ,

Thanks for the update: -m 23700/23800 mode

I have a "RAR3-p (Compressed)" file, get a very long hash code from rar2john.exe (latest 1.9.0-jumbo-1 64-bit Windows), 
as below:
`$RAR3$*1*e7e214e350de0031*af29d399*2960*4224*1*9588e43961593c...((VERY_VERY_LONG))....9a5562fcd98*33`

After execute the hashcat (latest V6.1.1):
`hashcat -m 23800 -a 3 filehash.txt --force -o output.txt`
Results:
`Hashfile "filehash.txt" on line 1 ($RAR3$...c9909d8c6e4accb7af9a5562fcd98*33): Salt-value exception`
`No hashes loaded.`

It's OK when using the hash example:
`23800 | RAR3-p (Compressed) * | $RAR3$*1*ad56eb40219c9da2*834064ce*32*13*1*eb47b1abe17a1a75bce6c92ab1cef3f4126035ea95deaf08b3f32a0c7b8078e1*33`

Don't know what happened?
It's caused by rar2john tranformation problem or somthing?
Thanks for hlep!

philsmd:
@sagetoo  there is a maximum data length that hashcat is able to support right now that currently is about 320KB , see:

https://github.com/hashcat/hashcat/blob/fe91f6276db5aff017573c38d0d586cd3b8a0525/src/modules/module_23800.c#L479

(the 655056 is the hexadecimal length, so you need to halve it: 655056 / 2 = 320KiB)

The problem with increasing such a limit is that 1. we would need to make some further architectural changes to hashcat (this is about the maximum line-length a hash can have in general in hashcat), 2. it would probably come with some speed loss for other users because larger buffers etc are used (but this needs to be confirmed, because decompression is host/CPU only and with a system with fast RAM and CPU the increase of buffers could be also very tiny)

sagetoo:
Hi @philsmd ,
Thanks for you reply!
Wish to have a maximum length upgrade in the future.

sagetoo:
Hi @philsmd ,
Thanks for your reply!
Wish to have a maximum length upgrade in the future.

celikj:
will we have a max length upgrade? i have 700KiB of hash :cry:

DeniR:
I have 2Mb of hash in 23700. Rar file was created at 2006. Need to upgrade max length limit.

PR #2543, opened 2020-09-08
Appveyor: test if it started to work without our workaround

philsmd:
This is just a test to see if we still need the signing key workaround added by https://github.com/hashcat/hashcat/pull/2521 . Let's see if Appveyor succeeds or fails. fingers crossed


philsmd:
still doesnt work :(

Why do they not update to latest MSYS ?

Will try it again in a few weeks/months

Issue #2360, opened 2020-04-13
Bestcrypt Volume Encryption - New Format


trounce1:
Hi,
I was working over on the JTR github and we have come up with the derivation for BestCrypt volume encryption and i was wondering if anyone was interested in implementing it in hashcat.

https://github.com/magnumripper/JohnTheRipper/issues/4242

The V3 format fills up an buffer of approx 0x10000 bytes
Then a round of sha256
Then a round of AES-cbc
Then a final round of sha256
Then a comparisson.

working python code for version 3 below

```python
import hashlib
from Crypto.Cipher import AES
from sys import argv

buffsize = 0x10000
buff = b''
password = b'password'

bc_file = open(argv[1], 'rb').read(0x512)

salt = bc_file[488:496]
enc_data = bc_file[38:134]

while len(buff) < buffsize:
    buff += salt
    buff += password
buff = buff[:0x10000]
dk = hashlib.sha256(buff).digest()

cipher = AES.new(dk, AES.MODE_CBC, b'\0' * 16)

plain = cipher.decrypt(enc_data)

hash_of_keys = hashlib.sha256(plain[:0x40]).digest()
hashval = plain[0x40:]

if hash_of_keys == hashval:
	print('correct')
```

the version 4 format works as follows

```python
import hashlib
from Crypto.Cipher import AES
from sys import argv

password = b'password'

bc_file = open(argv[1], 'rb').read(0x512)

salt = bc_file[0x1e8:0x1f0] + bc_file[0x166:0x176]
enc_data = bc_file[38:134]

p = 1
r = 16
N = pow(2, 15)
m = 128 * r * (N + p + 2)
dk = hashlib.scrypt(password, salt=salt, p=p, r=r, n=N, maxmem=m)

cipher = AES.new(dk, AES.MODE_CBC, b'\0' * 16)

plain = cipher.decrypt(enc_data)

hash_of_keys = hashlib.sha256(plain[:0x40]).digest()
hashval = plain[0x40:]

if hash_of_keys == hashval:
	print('correct')
```

Cheers

UPDATE: updated code to have a hard 0x10000 buffer

philsmd:
The sha256 variant (v 3) seems to be doable and worth to implement... but I'm not sure if this is still a very commonly used format/encryption. Maybe you can give more information about that (when was it first introduced and is it still very popular... I honestly never heard about these formats altogether).

[...]

PR #2559, opened 2020-09-26
Added -m 23900 = BestCrypt v3 Volume Encryption

philsmd:
This PR implements the V3 hash type requested in #2360 : 
```
-m 23900 = BestCrypt v3 Volume Encryption
```

For now, you can extract the hashes with https://github.com/trounce1/JohnTheRipper/blob/patch-1/run/bestcryptve2john.py

There is probably still some little room for optimizations (@jsteube already promised me that he will have a look at it soon), but for now it's implemented with the approach we discussed in #2360.

cc: @trounce1, @NilsOlze

### DIFF
--- a/OpenCL/m23900-pure.cl
+++ b/OpenCL/m23900-pure.cl
@@ -1,0 +1,454 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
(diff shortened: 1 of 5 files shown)


jsteube:
Super

PR #2568, opened 2020-10-03
Speed improvement in bcrypt

Sc00bz:
Avoids bank conflicts in bcrypt. I ran a test of 1,000,000 runs each and average bank conflicts for 11, 12, 16, 24, 32 threads are 2.00, 2.09, 2.39, 2.99, 3.53. This means we can save 1 to 2.5 clocks per look up. Also on write it's an N-way bank conflict thus this saves N-1 clocks per write. This assumes the number of shared/local memory banks is 32 for both Nvidia and AMD which has been the case since 2010.

It is possible to do every number of threads but 23, 27, 29, 31 in exactly 4KiB\*threads with no bank conflicts. This only does 12, 16, 24, and 32 in 4KiB\*threads and 11 in 48KiB-4 bytes. By default 11 threads is off because the host is expecting it to use 44KiB when in most cases it can use 48KiB-1 byte. In this case the host should define `BCRYPT_AVOID_BANK_CONFLICTS` and if `DYNAMIC_LOCAL` then allocate 48KiB-4 bytes for it.

### DIFF
--- a/OpenCL/m03200-pure.cl
+++ b/OpenCL/m03200-pure.cl
@@ -13,6 +13,21 @@
 #define COMPARE_S "inc_comp_single.cl"
 #define COMPARE_M "inc_comp_multi.cl"
 
+// FIXED_LOCAL_SIZE must be 11, 12, 16, 24, 32, or 40
+// If FIXED_LOCAL_SIZE is 11 then this will use 48*1024-4 bytes with BCRYPT_AVOID_BANK_CONFLICTS
+// By default this is off because the host is expecting it to use 44KiB when in most cases it can use 48KiB-1 byte
+// It is possible to do every number of threads but 23, 27, 29, 31 in exactly 4KiB*threads with no bank conflicts
+// Number of shared/local memory banks has been 32 for both Nvidia and AMD since 2010
+#ifdef FIXED_LOCAL_SIZE
+#if FIXED_LOCAL_SIZE == 12 || FIXED_LOCAL_SIZE == 16 || FIXED_LOCAL_SIZE == 24 || FIXED_LOCAL_SIZE == 32 || FIXED_LOCAL_SIZE == 40
+#define BCRYPT_AVOID_BANK_CONFLICTS
+#elif FIXED_LOCAL_SIZE != 11
+#undef BCRYPT_AVOID_BANK_CONFLICTS
+#endif
+#else
+#undef BCRYPT_AVOID_BANK_CONFLICTS
+#endif
+
 typedef struct bcrypt_tmp
 {
   u32 E[18];
@@ -309,6 +324,24 @@ CONSTANT_VK u32a c_pbox[18] =
   0x9216d5d9, 0x8979fb1b
 };
 
+#ifdef BCRYPT_AVOID_BANK_CONFLICTS
+#define BF_ROUND(L,R,N)                       \
+{                                             \
+  u32 tmp;                                    \
+                                              \
+  const u32 r0 = unpack_v8d_from_v32_S ((L)); \
+  const u32 r1 = unpack_v8c_from_v32_S ((L)); \
+  const u32 r2 = unpack_v8b_from_v32_S ((L)); \
+  const u32 r3 = unpack_v8a_from_v32_S ((L)); \
+                                              \
+  tmp  = S0[S0_mul * r0];                     \
+  tmp += S1[S1_mul * r1];                     \
+  tmp ^= S2[S2_mul * r2];                     \
+  tmp += S3[S3_mul * r3];                     \
+                                              \
+  (R) ^= tmp ^ P[(N)];                        \
+}
+#else
 #define BF_ROUND(L,R,N)                       \
 {                                             \
   u32 tmp;                                    \
@@ -325,6 +358,7 @@ CONSTANT_VK u32a c_pbox[18] =
                                               \
   (R) ^= tmp ^ P[(N)];                        \
 }
+#endif
 
 #define BF_ENCRYPT(L,R) \
 {                       \
@@ -357,7 +391,7 @@ CONSTANT_VK u32a c_pbox[18] =
 }
 
 #ifdef DYNAMIC_LOCAL
-extern __shared__ u32 lm[];
+extern __shared__ u32 S_all[];
 #endif
 
 DECLSPEC void expand_key (u32 *E, u32 *W, const int len)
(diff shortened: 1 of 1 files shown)


jsteube:
Hey there, I tried to contact you several times on IRC but without response. First time shortly after you send the PR and we had a short discussion, that was still on freenode, 2020-10-02 11:08:52, but without any response from you. Then I tried again on libera, 2021-05-31 16:10:34 also without response.

The problem with this PR is that it is doing out-of-boundary reads or writes. It doesn't work!

```
cuStreamSynchronize(): an illegal memory access was encountered
```

I have started a different memory access pattern which is not so efficient but it works without errors. The commit is https://github.com/hashcat/hashcat/commit/ce8c121b5000ccc94357a4e052d6330b50fb44d8

I believe there's still a way to make it more efficient, but it has work. The updated m03200-pure.cl kernel also makes it a bit more easy to integrate better access pattern because is has the GET_KEY and SET_KEY functionality and all CUDA devices are guaranteed to use DYNAMIC_LOCAL so it's generally easier to modify it.

While not really important I also write a small tool for better visualization and drop it here so it's not lost:

```
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include <stdlib.h>

// gcc -Wall -W -o bank_conflicts bank_conflicts.c

#define LOCAL_SZ      (48 * 1024)                          // available local memory: 48k
#define KEYS_VALUES   256                                  // 8 bit
#define KEYS_GROUPS   4                                    // we have 4 lookup tables
#define KEY_SZ        4                                    // of type u32
#define SBOX_SIZE     (KEYS_GROUPS * KEYS_VALUES * KEY_SZ) // 4096 byte
#define BANKS         32                                   // fixed on NV and AMD
#define BANK_CHUNK_SZ 4                                    // fixed on NV and AMD
#define CHUNKS        (LOCAL_SZ / BANK_CHUNK_SZ)           // 12288
#define THREADS       (LOCAL_SZ / SBOX_SIZE)               // 12
#define LOOKUP_GAP    THREADS                              // can be 12 if THREADS is 11 for 48k-1 case for OpenCL
//#define THREADS       11
//#define LOOKUP_GAP    12

typedef struct
{
  int t; // thread
  int c; // column
  int k; // key

} entry_t;

typedef uint8_t  u8;
typedef uint16_t u16;

int make_idx (const int lid, const int col, const int key)
{
  // linear
  // 32k              :  7 conflicts
  // 48k (12 threads) : 11 conflicts
  // 48k (11 threads) : 10 conflicts
  // 64k              : 15 conflicts
  // key extra work   :  0 instructions
  //return (lid * KEYS_VALUES * KEYS_GROUPS) + (col * KEYS_VALUES) + key;

  // 32k              :    0 conflicts
  // 48k (12 threads) :    2 conflicts
  // 48k (11 threads) : 1.75 conflicts
  // 64k              :    0 conflicts
  // key extra work   :    2 instructions
  //return (key * KEYS_GROUPS * LOOKUP_GAP) + (col * LOOKUP_GAP) + lid;

  // 32k              :    0 conflicts
  // 48k (12 threads) :    2 conflicts
  // 48k (11 threads) : 1.75 conflicts
  // 64k              :    0 conflicts
  // key extra work   :    1 instructions
  return (col * LOOKUP_GAP * KEYS_VALUES) + (key * LOOKUP_GAP) + lid;
}

int main ()
{
  entry_t m[CHUNKS];

  for (int i = 0; i < CHUNKS; i++)
  {
    m[i].t = -1;
    m[i].c = -1;
    m[i].k = -1;
  }

  // apply the pattern

  for (int t = 0; t < THREADS; t++)
  {
    for (int c = 0; c < KEYS_GROUPS; c++)
    {
      for (int k = 0; k < KEYS_VALUES; k++)
      {
        const int idx = make_idx (t, c, k);

        m[idx].t = t;
        m[idx].c = c;
        m[idx].k = k;
      }
    }
  }

  // visualize mapping
  // 1. pos = 1 byte,  thread
  // 2. pos = 1 byte,  sbox group
  // 3. pos = 2 bytes, sbox key

  int banks_to_threads[BANKS][THREADS];

  for (int row = 0; row < CHUNKS; row += BANKS)
  {
    if ((row % (CHUNKS / BANK_CHUNK_SZ)) == 0)
    {
      printf ("byte offset: %d\n", row * BANK_CHUNK_SZ);
    }

    for (int col = 0; col < BANKS; col++)
    {
      int idx = row + col;

      printf ("%01x%01x%02x ", m[idx].t & 0xf, m[idx].c & 0xf, m[idx].k & 0xff);

      banks_to_threads[col][m[idx].t] = 1;
    }

    printf ("\n");
  }

  // verify correctness of the pattern
  // each entry has to be unique

  for (int m1 = 0; m1 < CHUNKS; m1++)
  {
    if (m[m1].t == -1) continue; // ignore uninitialized

    for (int m2 = m1 + 1; m2 < CHUNKS; m2++)
    {
      if (m[m2].t == -1) continue; // ignore uninitialized

      if (m[m1].t != m[m2].t) continue;
      if (m[m1].c != m[m2].c) continue;
      if (m[m1].k != m[m2].k) continue;

      printf ("access pattern incorrect\n");

      return -1;
    }
  }

  // visualize bank conflicts
  // the less per line the better

  printf ("thread access at bank :\n");

  for (int col = 0; col < BANKS; col++)
  {
    printf ("%2d: ", col);

    for (int t = 0; t < THREADS; t++)
    {
      if (banks_to_threads[col][t] == 1)
      {
        printf ("%2d ", t);
      }
    }

    printf ("\n");
  }

  return 0;
}
```

jsteube:
Just for completeness. I tested this PR speed and compared it with commit https://github.com/hashcat/hashcat/commit/ce8c121b5000ccc94357a4e052d6330b50fb44d8

That's possible because this PR seems to work for configuration that result in 12 or 16 threads (Note that configuration 24, 32 and 40 threads I couldn't test because I have no device to handle that many threads). So probably there's a bug only in the macro branches which handle 11 threads. 

However, many GPU's shared memory resources are too small to reach the 16 threads. For instance, my GTX980 only has 48k and so it will never reach 16 threads. In 11 thread OpenCL it fails to the out-of-boundary read or write error described in the previous comment.

OpenCL:

```
v6.2.1: 11244 H/s
PR: - H/s
commit: 11247 H/s (shares code with v6.2.1)
```

CUDA:

```
v6.2.1: 11599 H/s
PR: 11244 H/s
commit: 11735 H/s (new code)
```

The RTX2080Ti (and probably many other new NV devices) offers 64k, but only in CUDA mode. In OpenCL the NV OpenCL driver artificially limits (I think for backward compatibility) the shared memory to 48k. To overcome this limit we added the dynamic mode at some point. So in OpenCL mode we end up using 11 threads and the same error as with the GTX980 occurs. 

OpenCL:

```
v6.2.1: 52207 H/s
PR: - H/s
commit:  52210 H/s (shares code with v6.2.1)
```

CUDA:

```
v6.2.1: 52207 H/s
PR: 56040 H/s
commit: 55861 H/s (new code)
```

And then there's modern AMD GPU. I'm testing with a Vega64. They also have 64k local memory, but they do not suffer from the 48k limit in OpenCL mode (at least not by using ROCm driver). So we can fully use 64k in OpenCL mode, resulting in 16 threads and that allows us to run a benchmark:

OpenCL:

```
v6.2.1: 26004 H/s
PR: 25559 H/s
commit: 26006 H/s (shares code with v6.2.1)
```

So in general the PR shines in the case where we have 16 threads. Still I think the commits code base is a better one because it's more clear what's happening, it works on all compute devices in all backend driver configurations and it passes the unit tests (tools/test.sh). Additionally, for many devices it is faster than the PR code changes.

Still I'd like to see if you can adapt the PR pattern to the new code base which came with the commit. Maybe there's a way to do it.

jsteube:
Updated the visualization tool in previous comment. Is now flexible to cover 32k, 48k, 48k-1 and 64k

PR #2571, opened 2020-10-07
Declare phony targets in Makefile

ventaquil:
According to [GNU `make` Phony Targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) here you have a little change in `Makefile`.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -395,8 +395,10 @@ endif
 ## Targets: Native Compilation
 ##
 
+.PHONY: default
 default: $(HASHCAT_FRONTEND) modules
 
+.PHONY: clean
 clean:
 	$(RM) -f $(HASHCAT_FRONTEND)
 	$(RM) -f $(HASHCAT_LIBRARY)
@@ -417,6 +419,7 @@ clean:
 	$(RM) -rf *.outfiles
 	$(RM) -rf kernels
 
+.PHONY: distclean
 distclean: clean
 	$(RM) -f *.restore
 	$(RM) -f *.potfile
@@ -443,6 +446,7 @@ Only Linux, FreeBSD and Darwin can use the 'install' target
   endif
 endif
 
+.PHONY: install
 ifeq ($(SHARED),1)
 install: install_docs install_shared install_library install_library_dev install_kernels install_modules install_hashcat
 else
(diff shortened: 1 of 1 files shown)


philsmd:
Hey @ventaquil , was this closed on accident or did you intent closing this pull request.

I think that it could make sense, even though it's not very, very important or is it ? it for sure doesn't really make things fail, but I see some advantages, like cleaner "Makefile target list" etc

ventaquil:
Hey @philsmd - I broke this PR because I tried to make another one (#2585)  and force pushed to `master` branch 😢

I think it's not very important - it's more like "nice-to-have" feature.

ventaquil:
I thought about rewrite some install targets like

```makefile
install_hashcat: $(HASHCAT_FRONTEND)
	$(INSTALL) -m 755 -d                                                    $(DESTDIR)$(INSTALL_FOLDER)
	$(INSTALL) -m 755 $(HASHCAT_FRONTEND)                                   $(DESTDIR)$(INSTALL_FOLDER)/
```

into

```makefile
install_hashcat: $(DESTDIR)$(INSTALL_FOLDER)/$(HASHCAT_FRONTEND)

$(DESTDIR)$(INSTALL_FOLDER):
	$(INSTALL) -m 755 -d $(DESTDIR)$(INSTALL_FOLDER)

$(DESTDIR)$(INSTALL_FOLDER)/$(HASHCAT_FRONTEND): $(DESTDIR)$(INSTALL_FOLDER) $(HASHCAT_FRONTEND)
	$(INSTALL) -m 755 $(HASHCAT_FRONTEND) $(DESTDIR)$(INSTALL_FOLDER)/
```

It can work better when you use `--jobs` flag. An additional advantage is that after some change only related targets are executed.

What do you think about that @philsmd?

philsmd:
not sure if it really makes sense for our use case. It seems to be just much longer and more difficult to understand/read.
We try to keep things easy to read and with as little code/lines as possible, if it still works and has only minor disadvantages/cons.

The problem is what if we need to add dozens of more lines to `install_hashcat`, it would just make the `Makefile` very long with this alternative approach (even though it might be a better approach/design and more correct in terms of Makefile targets etc).

I wouldn't really be against any such change if devs decide to use this approach, but in my opinion it's just a little bit more difficult to understand/read compared to the list of `$(INSTALL)` one after the other.

Maybe other devs have a completely different opinion about this, dunno

PR #2576, opened 2020-10-12
docs: fix simple typo, paramater -> parameter

timgates42:
There is a small typo in deps/LZMA-SDK/C/Lzma2DecMt.h, deps/LZMA-SDK/C/Lzma2Enc.h, deps/LZMA-SDK/C/Lzma86.h, deps/LZMA-SDK/C/LzmaEnc.h, deps/LZMA-SDK/C/LzmaLib.h.

Should read `parameter` rather than `paramater`.

### DIFF
--- a/deps/LZMA-SDK/C/Lzma2DecMt.h
+++ b/deps/LZMA-SDK/C/Lzma2DecMt.h
@@ -31,7 +31,7 @@ void Lzma2DecMtProps_Init(CLzma2DecMtProps *p);
 SRes:
   SZ_OK           - OK
   SZ_ERROR_MEM    - Memory allocation error
-  SZ_ERROR_PARAM  - Incorrect paramater in props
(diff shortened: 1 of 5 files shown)


philsmd:
Hey,
thanks for this suggestion.

The problem here is that we do normally **not** change any code that comes from a dependency (`deps`) because these are libraries that the original author should update. Only in very exceptional cases where we really have to make changes for hashcat to make it work, we make minimal changes that we try to document, otherwise a new `upstream` update would undo it etc.

So my suggestion here is that we close this issue and ask the original author to fix the typos and whenever a new `LZMA SDK` upstream libary version is available we will update our dependency.

I think you should ask for a typo fix and update here: https://sourceforge.net/p/sevenzip/patches/

Thank you very much

I think we should close this here, please close @timgates42 , @jsteube . Thx

PR #2582, opened 2020-10-15
Added new feature: guessing hash type

matrix:
Hi,

I added a feature to identify possible hash types starting from a single hash and / or hash list.
Example with single hash:

```
$ echo -n hashcat | openssl dgst -sha1
b89eaac7e61417341b710b727768294d0e6a277b
$ ./hashcat -M 'b89eaac7e61417341b710b727768294d0e6a277b' --quiet
Starting guessing mode...

(100) SHA1 : b89eaac7e61417341b710b727768294d0e6a277b

(300) MySQL4.1/MySQL5 : b89eaac7e61417341b710b727768294d0e6a277b

(4500) sha1(sha1($pass)) : b89eaac7e61417341b710b727768294d0e6a277b

(4700) sha1(md5($pass)) : b89eaac7e61417341b710b727768294d0e6a277b

(6000) RIPEMD-160 : b89eaac7e61417341b710b727768294d0e6a277b

(18500) sha1(md5(md5($pass))) : b89eaac7e61417341b710b727768294d0e6a277b

Started: Thu Oct 15 13:17:26 2020
Stopped: Thu Oct 15 13:17:26 2020
```
Example with hashlist

```
$ echo -n hashcat | openssl dgst -sha1 > test.hashes
$ echo -n hashcat | openssl dgst -sha512 >> test.hashes
$ ./hashcat -M test.hashes --quiet
Starting guessing mode...

(100) SHA1 : b89eaac7e61417341b710b727768294d0e6a277b

(300) MySQL4.1/MySQL5 : b89eaac7e61417341b710b727768294d0e6a277b

(1700) SHA2-512 : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

(4500) sha1(sha1($pass)) : b89eaac7e61417341b710b727768294d0e6a277b

(4700) sha1(md5($pass)) : b89eaac7e61417341b710b727768294d0e6a277b

(6000) RIPEMD-160 : b89eaac7e61417341b710b727768294d0e6a277b

(6100) Whirlpool : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

(11800) GOST R 34.11-2012 (Streebog) 512-bit, big-endian : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

(17600) SHA3-512 : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

(18000) Keccak-512 : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

(18500) sha1(md5(md5($pass))) : b89eaac7e61417341b710b727768294d0e6a277b

(21000) BitShares v0.x - sha512(sha512_bin(pass)) : 82a9dda829eb7f8ffe9fbe49e45d47d2dad9664fbb7adf72492e3c81ebd3e29134d9bc12212bf83c6840f10e8246b9db54a4859b7ccd0123d86e5872c1e5082f

Started: Thu Oct 15 13:18:24 2020
Stopped: Thu Oct 15 13:18:25 2020
```

Works also with "--machine-readable" option

```
$ echo -n hashcat | openssl dgst -sha256
127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935
$ ./hashcat -M '127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935' --quiet --machine-readable
Starting guessing mode...

SHA2-256:1400:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

GOST R 34.11-94:6900:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

GOST R 34.11-2012 (Streebog) 256-bit, big-endian:11700:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

SHA3-256:17400:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

Keccak-256:17800:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

sha256(md5($pass)):20800:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

sha256(sha256_bin($pass)):21400:127e6fbfe24a750e72930c220a8e138275656b8e5d8f48a98c3c92df2caba935

Started: Thu Oct 15 13:20:37 2020
Stopped: Thu Oct 15 13:20:38 2020
```

Exit status is set to return "0" if something is found, else is "1".

You like it ? ;)

Bye,
matrix

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -593,6 +593,7 @@ typedef enum user_options_defaults
   DEBUG_MODE               = 0,
   EXAMPLE_HASHES           = false,
   FORCE                    = false,
+  GUESS_HASH_MODE          = false,
   HWMON_DISABLE            = false,
   HWMON_TEMP_ABORT         = 90,
   HASH_MODE                = 0,
@@ -1933,6 +1934,7 @@ typedef struct user_options
   bool         limit_chgd;
 
   bool         advice_disable;
+  bool         guess_hash_mode;
   bool         benchmark;
   bool         benchmark_all;
   #ifdef WITH_BRAIN
@@ -2021,7 +2023,6 @@ typedef struct user_options
   u32          debug_mode;
   u32          hwmon_temp_abort;
   int          hash_mode;
-  int          guess_hash_mode;
   u32          hccapx_message_pair;
   u32          hook_threads;
   u32          increment_max;
--- a/include/types.h
+++ b/include/types.h
@@ -698,6 +698,7 @@ typedef enum user_options_map
   IDX_ENCODING_TO               = 0xff13,
   IDX_EXAMPLE_HASHES            = 0xff14,
   IDX_FORCE                     = 0xff15,
+  IDX_GUESS_HASH_MODE           = 'M',
   IDX_HWMON_DISABLE             = 0xff16,
   IDX_HWMON_TEMP_ABORT          = 0xff17,
   IDX_HASH_MODE                 = 'm',
@@ -2020,6 +2021,7 @@ typedef struct user_options
   u32          debug_mode;
   u32          hwmon_temp_abort;
   int          hash_mode;
+  int          guess_hash_mode;
   u32          hccapx_message_pair;
   u32          hook_threads;
   u32          increment_max;
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -471,10 +471,17 @@ static int outer_loop (hashcat_ctx_t *hashcat_ctx)
   {
     if (hashes->hashes_cnt == 0)
     {
-      event_log_error (hashcat_ctx, "No hashes loaded.");
+      if (user_options->guess_hash_mode == false) event_log_error (hashcat_ctx, "No hashes loaded.");
 
       return -1;
     }
+  }
+
+  if (user_options->guess_hash_mode == true)
+  {
+    EVENT (EVENT_OUTERLOOP_MAINSCREEN);
+
+    return hashes->hashes_cnt;
   }
 
   /**
@@ -1194,6 +1201,50 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
 
     user_options->quiet = false;
   }
+  else if (user_options->guess_hash_mode == true)
+  {
+    u64 rc = 0;
+
+    rc_final = 1;
+
+    event_log_info (hashcat_ctx, "Starting guessing mode...\n");
+
+    user_options->quiet = true;
+
+    if (user_options->hash_mode_chgd == true)
+    {
+      if (outer_loop (hashcat_ctx) != -1) rc++;
+    }
+    else
+    {
+      char *modulefile = (char *) hcmalloc (HCBUFSIZ_TINY);
+
+      for (int i = 0; i < MODULE_HASH_MODES_MAXIMUM; i++)
+      {
+        // skip STDOUT, Plaintext and Password Safe v2 modes
+        if (i == 2000 || i == 99999 || i == 9000) continue;
+
+        user_options->hash_mode = i;
+
+        module_filename (folder_config, i, modulefile, HCBUFSIZ_TINY);
+
+        if (hc_path_exist (modulefile) == false) continue;
+
+        if (outer_loop (hashcat_ctx) != -1) rc++;
+
+        if (status_ctx->run_main_level1 == false) break;
+      }
+
+      hcfree (modulefile);
+    }
+
+    user_options->quiet = false;
+
+    if (rc > 0)
+    {
+      rc_final = 2;
+    }
+  }
   else
   {
     if (user_options->speed_only == true) user_options->quiet = true;
@@ -1245,6 +1296,13 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
     {
       rc_final = 0;
     }
+  }
+
+  // special case for --guess-hash-type
+
+  if (user_options->guess_hash_mode == true && rc_final == 2)
+  {
+    rc_final = 0;
   }
 
   // done
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -1207,7 +1207,14 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
 
     rc_final = 1;
 
-    event_log_info (hashcat_ctx, "Starting guessing mode...\n");
+    event_log_info (hashcat_ctx, "Starting to guess the hash-type ...");
+
+    if (user_options->quiet == false)
+    {
+      event_log_info (hashcat_ctx, "For each hash-type guessed, the following fields will be printed: type, description and hash.");
+    }
+
+    event_log_info (hashcat_ctx, NULL);
 
     user_options->quiet = true;
 
@@ -1237,6 +1244,8 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
 
       hcfree (modulefile);
     }
+
+    event_log_info (hashcat_ctx, NULL);
 
     user_options->quiet = false;
 
--- a/src/hashes.c
+++ b/src/hashes.c
@@ -639,7 +639,7 @@ int hashes_init_filename (hashcat_ctx_t *hashcat_ctx)
     {
       if (hc_path_read (user_options_extra->hc_hash) == false)
       {
-        event_log_error (hashcat_ctx, "%s: %s", user_options_extra->hc_hash, strerror (errno));
+        if (user_options->guess_hash_mode == false) event_log_error (hashcat_ctx, "%s: %s", user_options_extra->hc_hash, strerror (errno));
 
         return -1;
       }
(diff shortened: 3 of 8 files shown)


src/hashcat.c
```diff
@@ -1194,6 +1201,50 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
...
+      if (outer_loop (hashcat_ctx) != -1) rc++;
+    }
+    else
+    {
+      char *modulefile = (char *) hcmalloc (HCBUFSIZ_TINY);
+
+      for (int i = 0; i < MODULE_HASH_MODES_MAXIMUM; i++)
+      {
+        // skip STDOUT, Plaintext and Password Safe v2 modes
+        if (i == 2000 || i == 99999 || i == 9000) continue;
```
ventaquil, line 1232: There is no better way to do this? I think this is not maintainable in long term.

src/hashes.c
```diff
@@ -1320,17 +1416,23 @@ int hashes_init_stage1 (hashcat_ctx_t *hashcat_ctx)
...
-          event_log_warning (hashcat_ctx, "Hashfile '%s': %s", hashes->hashfile, strerror (errno));
-        }
         else
         {
-          event_log_warning (hashcat_ctx, "Hashfile '%s': %s", hashes->hashfile, strparser (hashes_parsed));
+          if (user_options->guess_hash_mode == false)
+          {
+            if (hashes_parsed == 0)
+            {
+	      event_log_warning (hashcat_ctx, "No hashes loaded.");
```
ventaquil: Indentation is wrong.

```suggestion
              event_log_warning (hashcat_ctx, "No hashes loaded.");
```

include/types.h
```diff
@@ -2020,6 +2021,7 @@ typedef struct user_options
   u32          debug_mode;
   u32          hwmon_temp_abort;
   int          hash_mode;
+  int          guess_hash_mode;
```
ventaquil: Why not `bool` like `benchmark`?

```suggestion
  bool         guess_hash_mode;
```

src/user_options.c
```diff
@@ -186,6 +187,7 @@ int user_options_init (hashcat_ctx_t *hashcat_ctx)
   user_options->encoding_to               = ENCODING_TO;
   user_options->example_hashes            = EXAMPLE_HASHES;
   user_options->force                     = FORCE;
+  user_options->guess_hash_mode           = false;
```
ventaquil: I see that other default options are declared in `user_options_defaults_t`.

ventaquil:
Maybe the best way of storing selected mode is to create simple `mode` field and enum with available modes.

```c
typedef enum mode {
  BENCHMARK,
  GUESS,
  // ...
} mode_t;
```

philsmd:
I think it's a nice feature... but I must also say that I'm a little bit scared what happens if the devs team adds this... most importantly because I very often had to tell users to avoid using something like `HashID` or similar tools, because users try to misinterpret/misunderstand the output.

The most important thing would be to tell/warn the users that this list is not a complete list and it's not a 100% guaranteed list.

There are also 2 different categories of hashes:
* hashes that have a clear signature like `$telegram$`
* hashes that do only have some base64 or hex chars and could be many, many hash types

The problems are that users could use this feature and hashcat tells them that the hash is of type x , but even after wasting hours and hours to crack the hash with hash type x they can't crack it (even though the hash might just correspond to a simple password like "123456").

So we have at least these problems here:
* the hash type is not supported by hashcat yet: maybe because it uses a cascade like `md5 ($salt . sha512 (sha1 (md5 ($pass . $salt))))` (you know how many possibilities exists and how strange some software/CMS uses these combinations, especially when they upgrade the hash type etc)
* truncated hashes
* way too many results
etc etc

So maybe a warning would be enough, dunno... but it needs to be very clear and the most difficult to solve problem is: the user must read/see it (and users unfortunately tend to ignore or just not see/read the warnings).

matrix:
Hi @philsmd,

your remarks are right, but this feature only "**guess**" and not "**identify**" (unless special hash syntax).
I didn't want to implement autodetect (to omit the hash type from the command line), because as you have well said someone could get confused.

Hashcat is not for dummies, a minimum of background you need to have it before use. Surely we can improve the feedback to the user (more verbose message) to minimize misunderstandings.
The key for all these types of issue is ... documentation :)

My advice is ... it might be useful, it can also be used to split a hash list into multiple lists based on what type was guessed, quickly understand if a hash is supported by hashcat or not (something that everyone does manually for now), force the hash-mode to understand if your manual guessing is good or not, using hashcat instead HashID, etc ...

If the documentation is clear and you know what you are doing it is useful, otherwise not :)
Users have to read, study, etc ... it's not our problem if they don't.

What do you think about it? We also await @jsteube opinion :)

Thanks

src/hashcat.c
matrix: There are a thousand ways to make everything better.
First I always make a prototype, then I improve it ;)

philsmd:
good. Any suggestion on how to make this more clear to the user who uses the -M feature ?

maybe a note at the start, an additional note at the end and when finding a match ("could be 1400")

only with --quiet the output is less verbose (without these additional warnings and explanations on how to interpret this output) ?

Could you make these changes too here and show below how the output would look like with and without --quiet ?

src/hashcat.c
ventaquil: Okay, I thought that it's a final version ;)

Anyway, I am still learning Hashcat source code but maybe a new field `module_flags` in `module_ctx_t` struct would be a nice solution.
If flag `MODULE_GUESS_MODE_SKIP` (or something similar) is set then skip.

Anyway, nice feature!

davidbolvansky:
Cool feature!

matrix:
@philsmd , see if you like this type of output:

* standard output

```
$ cat example[4-5]00.hash > example.test.hash
$ ./hashcat -M example.test.hash -D1
hashcat (v2.01-7057-g68edafad7+) starting...

* Device #2: Unstable OpenCL driver detected!

This OpenCL driver has been marked as likely to fail kernel compilation or to produce false negatives.
You can use --force to override this, but do not report related errors.

* Device #1: Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz, 23754/23818 MB (5954 MB allocatable), 4MCU

OpenCL API (OpenCL 3.0 ) - Platform #2 [Intel(R) Corporation]
=============================================================
* Device #2: Intel(R) Graphics Gen9 [0x5917], skipped

Starting to guess the hash-type ...
For each hash-type guessed, the following fields will be printed: type, description and hash.

   400 |                                             phpass | $H$9y5boZ2wsUlgl2tI6b5PrRoADzYfXD1
   500 |          md5crypt, MD5 (Unix), Cisco-IOS $1$ (MD5) | $1$uOM6WNc4$r3ZGeSB11q6UUSILqek3J1

Started: Fri Nov 27 21:02:26 2020
Stopped: Fri Nov 27 21:02:26 2020
```

* using machine-readable

```
$ ./hashcat -M example.test.hash -D1 --machine-readable
hashcat (v2.01-7057-g68edafad7+) starting...

* Device #2: Unstable OpenCL driver detected!

This OpenCL driver has been marked as likely to fail kernel compilation or to produce false negatives.
You can use --force to override this, but do not report related errors.

Starting to guess the hash-type ...
For each hash-type guessed, the following fields will be printed: type, description and hash.

400:phpass:$H$9y5boZ2wsUlgl2tI6b5PrRoADzYfXD1
500:md5crypt, MD5 (Unix), Cisco-IOS $1$ (MD5):$1$uOM6WNc4$r3ZGeSB11q6UUSILqek3J1

Started: Fri Nov 27 21:04:50 2020
Stopped: Fri Nov 27 21:04:50 2020
```

* using quiet

```
$ ./hashcat -M example.test.hash -D1 --quiet
* Device #2: Unstable OpenCL driver detected!

Starting to guess the hash-type ...

   400 |                                             phpass | $H$9y5boZ2wsUlgl2tI6b5PrRoADzYfXD1
   500 |          md5crypt, MD5 (Unix), Cisco-IOS $1$ (MD5) | $1$uOM6WNc4$r3ZGeSB11q6UUSILqek3J1

Started: Fri Nov 27 21:05:11 2020
Stopped: Fri Nov 27 21:05:12 2020
```

Let me know ;)

jsteube:
Many thanks for this contribution, but I will not merge it for the follow reasons:

- The user needs to know about what type of hash they are targeting. If this is unknown, there are other tools much better suited for this task (for instance: mdxfind or hashID). Hashcat is for once you know the algorithm, not for when you want to figure it out.
- It gives the user poor advice. Yes we can warn them, but they will ignore or intentionally not read. They will just waste a lot of energy.
- Too many unresolved conflicts with most recent version

As an alternative I suggest you can have this packed into an external tool.

PR #2585, opened 2020-10-20
Fix types.h include guard

ventaquil:
Causes fire when double-include `types.h` in several files


jsteube:
Many thanks!

ventaquil:
Glad that I can contribute - even such small change!

PR #2586, opened 2020-10-20
Improve warning message wording

llamasoft:
The current warning message makes it sound like the optimized kernel exists but isn't being used.  This rewording makes it clear that the optimized kernel simply doesn't exist.

### DIFF
--- a/src/interface.c
+++ b/src/interface.c
@@ -342,7 +342,7 @@ int hashconfig_init (hashcat_ctx_t *hashcat_ctx)
           if (user_options->quiet == false)
           {
             event_log_warning (hashcat_ctx, "Kernel %s:", source_file);
-            event_log_warning (hashcat_ctx, "Optimized kernel requested but not needed - falling back to pure kernel");
+            event_log_warning (hashcat_ctx, "Optimized kernel requested but not available - falling back to pure kernel");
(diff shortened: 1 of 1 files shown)


roycewilliams:
As I understand it, the wording was deliberate - there's no optimized kernel because it would be redundant (no better than the non-optimized kernel). Changing the wording this way could make people ask "Why isn't it available? Is there something wrong with my install? Do I need to put in a PR to request one?" etc.

Don't get me wrong, I think you're looking for good clarification here, and there may be a better way than "not needed". I'm just not sure that "not available" is the way to go.

llamasoft:
That's fair.  I was just going by the comment I found here: https://github.com/hashcat/hashcat/blob/ee3b5da25b7a3bbe996eb3942375dd8254a2b9c0/src/interface.c#L323-L324

It made it sound like an optimized kernel may be unavailable because A) it hasn't been written yet or B) it provides no benefit.  If it really is the latter case where it provides no benefit then I'm not even sure the user should be told about it.  That is to say, if the regular kernel _is_ the optimized kernel, don't tell the user otherwise.

roycewilliams:
Ah, I'd never noticed that! So there's actually more than one reason. How about "not needed or not available" ?

roycewilliams (approved):

PR #2590, opened 2020-10-23
Fix some include headers macros

philsmd:
This fix is similar to https://github.com/hashcat/hashcat/pull/2585 , I double checked some header files and came to the conclusion that there are some minor, but nasty hidden problems with these macros. 

I think that this fix together with #2585 should fix all these problems.

Credits of course go to @ventaquil for finding these problems.

Thx

### DIFF
--- a/include/ext_lzma.h
+++ b/include/ext_lzma.h
@@ -4,6 +4,7 @@
  */
 
 #ifndef _EXT_LZMA_H
+#define _EXT_LZMA_H
 
 #include <LzmaDec.h>
 #include <Lzma2Dec.h>
--- a/include/folder.h
+++ b/include/folder.h
@@ -3,6 +3,9 @@
  * License.....: MIT
  */
 
+#ifndef _FOLDER_H
+#define _FOLDER_H
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <errno.h>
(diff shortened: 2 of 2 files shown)


ventaquil:
Thanks. Can I ask you why `types.h` is not included in header files? It requires to include `types.h` before all other headers - what is the motivation behind this approach?

philsmd:
This is very off-topic so I wouldn't really want to start a huge discussion here... but I think it's just a matter of design decisions and how we include `*.h` files.

In our case, we try to not use a lot of includes within includes (there are some exceptions and of course some system headers are included also in `*.h` files. but less cross-includes for our own header files).

So our general/"normal" approach is to just include something like `include/common.h` or `include/types.h` in the main `*.c` source code file. So every .c file has to include the header (often of course it's important that the includes are in the correct order). This also makes sure that the devs see immediately which include files are needed/necessary/important/used in the hash type modules or other source files.

So I would say it shouldn't be completely wrong how we do it, it's just one way to look at it and how to include header files in your code. Of course you could say that cross-includes are good to avoid situations where you forget to include some headers etc... but most of the time the compiler will return a (compile time) error, if some important header files are not included correctly etc.

Thanks for the suggestion, though

jsteube:
Many thanks!

PR #2596, opened 2020-10-26
Added -m 24600 = SQLCipher

trounce1:
new mode to support brute force of sqlcipher v4 databases.

some help with the perl module would be appreciated until i get up to speed with perl

### DIFF
--- a/OpenCL/m24600-pure.cl
+++ b/OpenCL/m24600-pure.cl
@@ -1,0 +1,442 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */ 
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha512.cl"
+#include "inc_cipher_aes.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+typedef struct pbkdf2_sha512_tmp
+{
+  u64  ipad[8];
+  u64  opad[8];
+
+  u64  dgst[16];
+  u64  out[16];
+
+} pbkdf2_sha512_tmp_t;
+
+typedef struct pbkdf2_sha512
+{
+  u32 salt_buf[64];
+  u32 ciphertext[64];
+
+} pbkdf2_sha512_t;
+
+DECLSPEC void hmac_sha512_run_V (u32x *w0, u32x *w1, u32x *w2, u32x *w3, u32x *w4, u32x *w5, u32x *w6, u32x *w7, u64x *ipad, u64x *opad, u64x *digest)
+{
+  digest[0] = ipad[0];
+  digest[1] = ipad[1];
+  digest[2] = ipad[2];
+  digest[3] = ipad[3];
+  digest[4] = ipad[4];
+  digest[5] = ipad[5];
+  digest[6] = ipad[6];
+  digest[7] = ipad[7];
+
+  sha512_transform_vector (w0, w1, w2, w3, w4, w5, w6, w7, digest);
+
+  w0[0] = h32_from_64 (digest[0]);
+  w0[1] = l32_from_64 (digest[0]);
+  w0[2] = h32_from_64 (digest[1]);
+  w0[3] = l32_from_64 (digest[1]);
+  w1[0] = h32_from_64 (digest[2]);
+  w1[1] = l32_from_64 (digest[2]);
+  w1[2] = h32_from_64 (digest[3]);
+  w1[3] = l32_from_64 (digest[3]);
+  w2[0] = h32_from_64 (digest[4]);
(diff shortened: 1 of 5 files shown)


OpenCL/m24600-pure.cl
```diff
@@ -0,0 +1,442 @@
...
+
+  u8 counter = 0;
+  for (u8 i = 0; i < 64; i++)
+  {
+    if (res[i] == 0)
+    {
+      counter +=1;
+    }
+  }
+  if (counter >= 2)
```
philsmd, line 438: this seems very strange ? can you explain how good such a check works ? is there no better way to check for a successful crack (verification) ? This seems to be very prone to a lot of collisions (2 bytes in 64 is way too low and they can be in any position too, that happens a lot with very wrong keys, right ?)

src/modules/module_24600.c
```diff
@@ -0,0 +1,267 @@
...
+  tmp_len2 = base64_decode (base64_to_int, ciphertext_pos, ciphertext_len, tmp_buf2);
+
+  memcpy (pbkdf2_sha512->ciphertext, tmp_buf2, tmp_len2);
+
+  return (PARSER_OK);
+}
+
+int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
+{
+  return snprintf (line_buf, line_size, "%s", hash_info->orighash);
```
philsmd, line 190: why do you use `orighash` here ? is this really needed ? Why ?
I don't think that `OPTS_TYPE_HASH_COPY` should be used with this hash type. any good reason why we need this ?

tools/sqlcipher2hashcat.py
```diff
@@ -0,0 +1,7 @@
+from base64 import b64encode
```
philsmd, line 2: there is no shebang here (`#!/usr/bin/env python3`).

Furthermore, this script assumes that the file is readable and the user knowns what files are expected and what the command line arguments should be.

I would at least make a check if the length of `sys.argv` is greater 1 and check also if the file is readable. otherwise (with no arguments, or not readable file) we should show a usage prompt where the use of this tool is described

tools/sqlcipher2hashcat.py
```diff
@@ -0,0 +1,7 @@
...
+import sys
+
+database = open(sys.argv[1], "rb").read(272)
```
philsmd: here is the `open()` function call without any further sanity checks.

Furthermore, all our `*.py`, `*.pl`, `*.sh` ... files are executable (the file flag chmod +x ), we should also make this file executable

philsmd (changes requested):
Thanks a lot for your contribution.

I only found (at a first glance) some minor problems that I commented.
Thx

philsmd:
Well, it would also be great if you post your `POC` (proof of concept) in python or whatever here, such that it's maybe easier to understand for everyone how that algo and most importantly the verification works... especially that 2 in 64 bytes check seems quite strange/wrong to me... that's not a very good proof if a password is correct (collisions)

OpenCL/m24600-pure.cl
```diff
@@ -0,0 +1,442 @@
...
+    res[i + 2] = hc_swap32_S (out[2] ^ iv[2]);
+    res[i + 3] = hc_swap32_S (out[3] ^ iv[3]);
+
+    iv[0] = data[0];
+    iv[1] = data[1];
+    iv[2] = data[2];
+    iv[3] = data[3];
+  }
+
+  u8 counter = 0;
```
philsmd: we normally do not use `u8` in the kernel code except if really, really required ("char" manipulation without any u32 and therefore "faster" alternative).
I think there is no reason for using `u8` here, you even did use `u32` in the loop above, so we should be consistent and just use `u32` for the `counter` variable and the `i` variable in the loop below etc. thx

OpenCL/m24600-pure.cl
trounce1: it should be 8 bytes in 256 which is quite a large number. do you think i should use the entropy checker like the Truecrypt modules?

src/modules/module_24600.c
trounce1: i can modify, i had just copied this .c file from anothor kernel

trounce1: @philsmd whats the problem with the orighash?

also without OPTS_TYPE_HASH_COPY. the kernel has segfault

jsteube:

I had this hash-mode implemented in the last days but I didn't realize you had it sent already as PR. The result is we have two different implementation of the same hash-mode. In order to honor your contribution I will merge this PR, so you can take the credit. However, I will remove some of the code afterwards and replace it with my implementation. My version also support v3 and v4 as well as SHA1, SHA256 and SHA512 (user can customize it). It also comes with an separate extraction tool and a unit test module. 

Thanks again for your contributions and please keep up with it :)

jsteube:
https://github.com/hashcat/hashcat/commit/524cb20703be709981631bdf3e8fbaffdee8455c

PR #2602, opened 2020-11-01
test.pl: add 'potthrough' (like passthrough, but hash:plain)

roycewilliams:
New option 'potthrough' - like passthrough, but in classic potfile format (hash:plain) instead.


jsteube:
Many thanks!

Issue #2636, opened 2020-12-09
support for bcrypt(md5($pass)) / bcryptmd5


roycewilliams:
I've recently learned that vBulletin can be (and now often is) configured to hash passwords client-side with MD5 before sending them to the server to be hashed as bcrypt. The end result / effective storage method is therefore bcrypt(md5($pass)), and must be attacked that way.

matrix:
Hi @roycewilliams :)

I modified bcrypt (3200) by inserting the part related to md5. Is this the change you expected?

Below is the poc.
```
$ cat 3201.hash 
$2a$05$78w9nhQpwg44lZcMHlh9n.wvuYSMprS7J9Gx7Tbjw9mMMjiRuh0vu
$ cat 3201.pass 
hashcat
$ ./hashcat -m 3201 3201.hash 3201.pass --quiet
$2a$05$78w9nhQpwg44lZcMHlh9n.wvuYSMprS7J9Gx7Tbjw9mMMjiRuh0vu:hashcat
```

roycewilliams:
Ah, yes - exactly!

```
$ echo -n 'hashcat' | md5sum | awk '{print $1}' | tee 3201md5.list
8743b52063cd84097a65d1633f5c74f5
$ hashcat --quiet -m 3200 3201.hash 3201md5.list
$2a$05$78w9nhQpwg44lZcMHlh9n.wvuYSMprS7J9Gx7Tbjw9mMMjiRuh0vu:8743b52063cd84097a65d1633f5c74f5

```

matrix:
This first version only works with password length <= 64 instead of 72. I have to work a little to understand how to extend the length up to the maximum length of bcrypt.

matrix:
Everything is fine now. Since md5 is in the way, the len coming to bcrypt is fixed at 32. Consequently the password limit is no longer 72, but it depends on md5 :)

cagdasb:
> 
> 
> This first version only works with password length <= 64 instead of 72. I have to work a little to understand how to extend the length up to the maximum length of bcrypt.

Hello, could you teach me how to download/install/build that version of hashcat on windows?
What I do is download latest version of hashcat then run it as "hashcat.exe -a0 -m0 ......" 
I tried latest version it doesn't have -m 3201 (bcryptmd5)
Could you please teach me that or make it executable on windows. Thanks for your contribution to community

matrix:
[...]

PR #2639, opened 2020-12-11
Added hash-mode 3202 - bcrypt(sha1(pass)) / bcryptsha1

matrix:
taking inspiration from #2636


roycewilliams:
Wow, anticipating the customer's need before they can even ask. :D

PR #2644, opened 2020-12-14
fix m04510_mxx

matrix:
```
$ echo ciao | ./hashcat -m 4510 -a0 '3f786850e387550fdab836ed7e6dc881de23001b:abc'
[...]
clBuildProgram(): CL_BUILD_PROGRAM_FAILURE

<program source>:88:31: error: passing 'const __global u32 *' (aka 'const __global unsigned int *') to parameter of type 'const __constant u32 *' (aka 'const __constant unsigned int *') changes address space of pointer
    tmp.pw_len = apply_rules (rules_buf[il_pos].cmds, tmp.i, tmp.pw_len);
                              ^~~~~~~~~~~~~~~~~~~~~~
[...]/OpenCL/inc_rp.cl:769:50: note: passing argument to parameter 'cmds' here
DECLSPEC int apply_rules (CONSTANT_AS const u32 *cmds, u32 *buf, const int in_len)
                                                 ^

* Device #1: Kernel [...]/OpenCL/m04510_a0-pure.cl build failed.

```


jsteube:
I don't know why GitHub does not show an error on this PR, because it was already fixed. So it should not find the string in the kernel and instead show some error. Let's see what happens if we merge it.

PR #2646, opened 2020-12-16
Fix --keep-guessing mode

Slattz:
New passwords would not be outputted to the potfile or console after one password was found.
This was broken by commit https://github.com/hashcat/hashcat/commit/04d5e5a119ba4c44bedb5bcccd5a42f82463cca3.


jsteube:
Thanks for the fix!

Issue #2401, opened 2020-05-13
Does hashcat support snmp v3 password cracking?


MG-302:
There are authentication password and encrypted password in snmp v3. Does hashcat support the decoding of these passwords?

What should I do if I want to develop the hashcat algorithm library? Is there any instructions for source code development?

philsmd:
Hey, yeah it's not supported yet.

John The Ripper (JTR) seems to support it with https://github.com/magnumripper/JohnTheRipper/blob/bleeding-jumbo/src/snmp_fmt_plug.c

Would be nice if you could do some research (or even develop a short python/perl proof of concept, POC, script) to show how the algorithm works internally and how "hashes" are generated or cracked.

It seems also that there are some variants and the JTR format lists these hashing components: HMAC-MD5-96/HMAC-SHA1-96 32 .

What do you use to extract these hashes ? just wireshark ? it also seems the network2john.lua file has some SNMP hash extraction functions. Is this the way how you normally extract these hashes ?
maybe other network dumping/sniffing tools, like metasploit/nmap etc, are also able to recognize/generate/extract these hashes ? Is there any common format (otherwise I would suggest to use the JTR, without file names, as we do most of the time in cases we do not already have the algorithm implemented in hashcat) ?

and yeah, we have a guide for developers to implement new hash types/algorithms: https://github.com/hashcat/hashcat/blob/master/docs/hashcat-plugin-development-guide.md
This guide is also quite new... would be great if you could have a look at it and maybe help as with both improving/double-checking the guide and help implementing the SNMP hash type. That would be fantastic. (but note: the guide is not intended to be a tutorial. I think there are some - older - blog posts that explain step-by-step how some users added a hash... but we can't really recommend them here... they might very quickly get out of date etc... so we want to stick to the dev guide).
Thank you very much

matrix:
@MG-302, I'm working on your request :)



```
OpenCL API (OpenCL 1.2 (May  7 2020 00:10:14)) - Platform #1 [Apple]
====================================================================
* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

Benchmark relevant options:
===========================
* --opencl-device-types=1
* --optimized-kernel-enable

Hashmode: 25200 - SNMPv3 HMAC-SHA1 (Iterations: 1)

[...]

PR #2661, opened 2020-12-22
Added hash-mode 25100 - SNMPv3 HMAC-MD5-96

matrix:
see issue #2401

### DIFF
--- a/OpenCL/m25100-pure.cl
+++ b/OpenCL/m25100-pure.cl
@@ -1,0 +1,196 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+//#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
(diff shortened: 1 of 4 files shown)


jsteube:
Reverted. please see: https://github.com/hashcat/hashcat/pull/2662#issuecomment-846528035

RLStrug:
Hi, I don't know if this is the place, but there seem to be an issue with this mode (and probably with the other SNMP modes).
Indeed, this piece of code
```C
u32 module_pw_min (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
{
  const u32 pw_min = 8;

  return pw_min;
}
```
sets a minimum password length of 8 bytes.
I don't think such a constraint is specified in RFCs 2571, 3414, or 2574.

Maybe setting the minimum size to 1 would be more appropriate?

As a side question, I have been looking for a way to remove this constraint at runtime, but found nothing. Is there a way, or is it only possible through re-compiling the module?

matrix:
> 3414

Hi,

check this from RFC 3414

Screen Shot 2022-12-03 at 16 19 36

PR #2662, opened 2020-12-22
Added hash-mode 25200 - SNMPv3 HMAC-SHA1-96

matrix:
see issue #2401

### DIFF
--- a/OpenCL/m25200-pure.cl
+++ b/OpenCL/m25200-pure.cl
@@ -1,0 +1,204 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+//#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha1.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+#define SNMPV3_SALT_MAX 375
+#define SNMPV3_ENGINEID_MAX 32
+#define SNMPV3_MSG_AUTH_PARAMS_MAX 12
+
+typedef struct hmac_sha1_tmp
+{
+  u32  dgst[5];
+  u32  out[3];
+
+} hmac_sha1_tmp_t;
+
+typedef struct snmpv3
+{
+  u32 salt_buf[SNMPV3_SALT_MAX];
+  u32 salt_len;
+
+  u8  engineID_buf[SNMPV3_ENGINEID_MAX];
+  u32 engineID_len;
+
+} snmpv3_t;
+
+KERNEL_FQ void m25200_init (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
+{
+  /**
+   * modifier
+   */
+
+  const u64 lid = get_local_id (0);
+  const u64 gid = get_global_id (0);
+
+  if (gid >= gid_max) return;
+
+  /**
+   * base
+   */
+
+  const global u8 *pw_buf = (global u8 *) pws[gid].i;
+
+  const u32 pw_len = pws[gid].pw_len;
+
+  const global u8 *engineID_buf = esalt_bufs[DIGESTS_OFFSET].engineID_buf;
+
+  u32 engineID_len = esalt_bufs[DIGESTS_OFFSET].engineID_len;
+
+  /**
+   * authkey
+   */
+
+  u32 idx = 0;
+
+  u8 tmp_buf[72] = { 0 };
+
+  sha1_ctx_t ctx;
+
+  sha1_init (&ctx);
+
+  for (int j = 0; j < 16384; j++)
+  {
+    for (int i = 0; i < 64; i++)
+    {
+      tmp_buf[i] = pw_buf[idx++];
+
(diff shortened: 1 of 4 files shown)


jsteube:
I had accidentally merged -m 25100 (and then reverted afterwards). But basically there's the same type of error in all 3 PR related to SNMPv3. I'll list them here and then close the other SNMPv3 PR, waiting for the fixed to all 3 of them.

---

```
error: patch failed: docs/changes.txt:14
error: docs/changes.txt: patch does not apply
```

This section:

```
Added hash-mode: SNMPv3 HMAC-SHA1-96
```

This is simple, just move up the to the current block

---

```
error: patch failed: docs/readme.txt:336
error: docs/readme.txt: patch does not apply
```
This section:

```
- SNMPv3 HMAC-SHA1-96
```

The ordering in docs/readme.txt is the same as if the user runs --help. So in order to find the correct stop just run --help.

---

The kernel do not work with CUDA. The problematic sections are:

```

main_kernel(58): error: identifier "global" is undefined
main_kernel(58): error: const variable "u8" requires an initializer
main_kernel(58): error: expected a ";"
main_kernel(62): error: identifier "global" is undefined
main_kernel(62): error: "u8" has already been declared in the current scope
main_kernel(62): error: const variable "u8" requires an initializer
main_kernel(62): error: expected a ";"
main_kernel(72): error: expected a ";"
main_kernel(82): error: identifier "tmp_buf" is undefined
main_kernel(82): error: identifier "pw_buf" is undefined
main_kernel(87): error: identifier "tmp_buf" is undefined
main_kernel(100): error: expected a ";"
main_kernel(104): error: identifier "buf" is undefined
main_kernel(104): error: identifier "as_uchar4" is undefined
main_kernel(163): error: identifier "engineID_buf" is undefined

15 errors detected in the compilation of "main_kernel".

```

So basically the problem is that in CUDA there's no function as_uchar4(). I can see that you have put SNMPV3_OPT1, so that the code switches to bitwise operations:

```
  sha1_buf[ 0] =  h_sha1[0] & 0xff;
  sha1_buf[ 1] = (h_sha1[0] >> 8) & 0xff;
```

We have a better way for extracting 8 bit from 32 bit variables, that is by using:

```
DECLSPEC u8 v8a_from_v32_S   (const u32 v32);
DECLSPEC u8 v8b_from_v32_S   (const u32 v32);
DECLSPEC u8 v8c_from_v32_S   (const u32 v32);
DECLSPEC u8 v8d_from_v32_S   (const u32 v32);
```

These function will work for both OpenCL and CUDA. Note the _S is for single scalar. For future plugins, please refer to OpenCL/inc_common.h. You can find a rich collections of these functions. In your particular case, the above four should be used.

---

Serious Bug:

```
  u8  md5_buf[72] = { 0 };
  u8 sha1_buf[72] = { 0 };
```

and later again here:

```
  u32x  s_md5[375] = { 0 };
  u32x s_sha1[375] = { 0 };
```

Especially OpenCL runtime with intensive optimizations settings will result in false negatives. It is -essential- that whenever one uses the crypto library, they need to stick to the block boundary of the crypto primitive. See section about Hashcat Crypto Library in https://github.com/hashcat/hashcat/blob/v6.2.1/docs/hashcat-plugin-development-guide.md.

```
The buffer you provide must be padded to a size of a multiple of the block size of whatever crypto primitive you are using. For example 64 byte for MD5, SHA256, etc. and 128 byte for SHA512 and others. If you do not know the block sizes, check the algorithm specs. For instance, if you want to statically append a 5 byte string to a password from inside the kernel, you could use sha1_update (&ctx, buf, 5);, the important thing here is that buf[] must be declared as u32 buf[16]. That is because the block size of SHA1 is 64 byte.
```

--

In your case, this declaration:

```
  u8  md5_buf[72] = { 0 };
  u8 sha1_buf[72] = { 0 };
```

should be written as:

```
  u32  md5_buf[32] = { 0 };
  u32 sha1_buf[32] = { 0 };
```

Not that I have also switch from u8 to u32.

---

The switch from u8 to u32 should be done all the time. I can see the reason is this:

```
  sha1_buf[ 0] = as_uchar4 (h_sha1[0]).x;
  sha1_buf[ 1] = as_uchar4 (h_sha1[0]).y;
  sha1_buf[ 2] = as_uchar4 (h_sha1[0]).z;
  sha1_buf[ 3] = as_uchar4 (h_sha1[0]).w;
```

But that's nothing else than a byte swap, right. So you should replace it with hc_swap32_S() instead.

---

In the loop kernel, there is:

```
  const u64 gid = get_global_id (0);

  if ((gid * VECT_SIZE) >= gid_max) return;
```

This construct is only valid if you use a SIMD loop kernel (vector datatypes). But you have it disabled by commenting out this:

```
//#define NEW_SIMD_CODE
```

In this case, it's better to regular construct to not avoid any developers reading your code.

---

Same here:

```
  u32x  md5_key[16] = { 0 };
  u32x sha1_key[16] = { 0 };
```

If you stick to scalar code, no need to use u32x. It's only confusing.

---

In the loop kernel I cannot find and loops. Instead the iteration can be found in the init kernel. This needs to be changed. Please move the heavy sections like this:

```
  for (int j = 0; j < 16384; j++)
  {
    for (int i = 0; i < 64; i++)
    {
      tmp_buf[i] = pw_buf[idx++];

      if (idx >= pw_len) idx = 0;
    }

    md5_update (&ctx, (u32 *)tmp_buf, 64);
  }
```

To the loop section. You can then set in the module a fixed loop count, for example 1024. Then you can use `for (int j = 0; j < 1024; j++)` instead.

---

After you have them fixed in all 3 modes, please send in a new PR.

PR #2663, opened 2020-12-22
Added hash-mode 25000 - SNMPv3 HMAC-MD5-96/HMAC-SHA1-96

matrix:
see issue #2401


jsteube:
See https://github.com/hashcat/hashcat/pull/2662#issuecomment-846528035

PR #2668, opened 2020-12-26
Correct check for gz header.

RealEnder:
gzip format is described in rfc1952.
From there, first 2 bytes (0x1f8b) are header;
next is Compression method (0x08 for deflate, this is the general used method);
and 4th byte is Flags. Some compression tools don't set this and we can't process the gzips.
zlib plays well in this cases, so we can just drop the check for the 4th byte.


ZerBea:
Check the magic number (0x1f8b) and the compression method  (0x08) is also the suggested method found in the example files:
https://github.com/madler/zlib/blob/master/examples/gzjoin.c#L208

jsteube:
Many thanks!

PR #2669, opened 2020-12-27
Removed code not useful in user_options.c

matrix:
As I left written in the comments, the check done in the now commented code is done just before, by hashconfig_init(), using the correct parameters instead of using the wrong ones.

### DIFF
--- a/src/user_options.c
+++ b/src/user_options.c
@@ -2825,6 +2825,8 @@ int user_options_check_files (hashcat_ctx_t *hashcat_ctx)
     return -1;
   }
 
+  hcfree (modulefile);
+
   const bool quiet_save = user_options->quiet;
 
   user_options->quiet = true;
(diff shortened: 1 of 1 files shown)


jsteube:
I think this is important:

```
generate_source_kernel_filename (false, ATTACK_EXEC_OUTSIDE_KERNEL, ATTACK_KERN_STRAIGHT, 400, 0, folder_config->shared_dir, kernelfile);
```

Because we need to use a guaranteed existing kernel. By using the one provided by the user it is not guaranteed. The other changes look ok. If I understand it's only about fixing memory leaks. Please update this PR.

matrix:
> I think this is important:
> 
> ```
> generate_source_kernel_filename (false, ATTACK_EXEC_OUTSIDE_KERNEL, ATTACK_KERN_STRAIGHT, 400, 0, folder_config->shared_dir, kernelfile);
> ```
> 
> Because we need to use a guaranteed existing kernel. By using the one provided by the user it is not guaranteed. The other changes look ok. If I understand it's only about fixing memory leaks. Please update this PR.

Hi, probably if user provide a wrong kernel hashconfig_init() fail ?
https://github.com/hashcat/hashcat/blob/d53913f4449d1a1e30332e41a1b10535af7e8b26/src/interface.c#L107-L109

jsteube:
The fixed hash-mode 400 is intentionally, because we know it has to exist. The idea is to verify the JiT can reach the file from the current working directory before it actually comes to that point in the code. This can take a while, for example because hashcat had to parse through a very large wordlist and count the words. So hashcat runs a minute and then fails with permission error is something the user wouldn't expect.

I understand the Idea. However, to fix the issue you describe we should add special code to verify the user selected hash-mode is a valid one. Please remove the change from this PR so I can merge the other fixes.

jsteube:
Implemented with https://github.com/hashcat/hashcat/commit/4930105ce3c4acca35d0c8fa54bd4685eedf022d

PR #2677, opened 2021-01-08
Makefile: Check Darwin major version instead of macOS minor version

RAN1:
Since macOS 11.0, the Makefile should test the Darwin version. The clock_gettime check uses Darwin 15 (OS X 10.11).


jsteube:
Many thanks! I don't have Apple computer so I can't test. I just hope it will work.

matrix:
> Many thanks! I don't have Apple computer so I can't test. I just hope it will work.

works ;)

PR #2723, opened 2021-03-07
Added hash-mode 24900 - Dahua Authentication MD5

Chick3nman:
This is equivalent to the Dahua MD5 format in JohnTheRipper(https://github.com/openwall/john/blob/b81ed703ceb7ca62df50c2fa0d4ea366ef713a4a/src/dahua_fmt_plug.c) but drops the superfluous `$dahua$` tag that prepends the hashes. If we want to stay directly compatible, I can add that to the tokenizer but I don't see a need to add the tag otherwise.

This is a collision prone algorithm, many passwords will match a given hash.

### DIFF
--- a/OpenCL/m24900_a0-optimized.cl
+++ b/OpenCL/m24900_a0-optimized.cl
@@ -1,0 +1,401 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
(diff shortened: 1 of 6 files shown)


jsteube:
Thanks for the contribution! I did some small changes, see this commit: https://github.com/hashcat/hashcat/commit/de19c31dee849bfe00d9c93dacdf47866e89762a

Basically I removed the lookup table to make it easier to use SIMD datatypes without macros. The trick was to instead encoding the binary hash into the dahua encoding in the kernel to decoded the dahue values in the module and stored the binary values instead. The removal of the lookup table gives a small speed bonus, but not relevant.

The kernel failed for CPU use but that wasn't detected because there was no unit test. I also added the unit test.

Thanks again for contribution!

Chick3nman:
Awesome, I'll build the unit tests next time, thanks for the fix.

PR #2739, opened 2021-04-01
Added attack mode 25400

thatux:
Which cracks a pdf edit password. It's largely duplicate code of 10500. Based on  https://hashcat.net/forum/thread-6233.html

### DIFF
--- a/OpenCL/m25400-pure.cl
+++ b/OpenCL/m25400-pure.cl
@@ -1,0 +1,456 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_hash_md5.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+CONSTANT_VK u32a padding[8] =
+{
+  0x5e4ebf28,
+  0x418a754e,
+  0x564e0064,
+  0x0801faff,
+  0xb6002e2e,
+  0x803e68d0,
+  0xfea90c2f,
+  0x7a695364
+};
+
+typedef struct pdf
+{
+  int  V;
+  int  R;
+  int  P;
+
+  int  enc_md;
+
+  u32  id_buf[8];
+  u32  u_buf[32];
+  u32  o_buf[32];
+
+  int  id_len;
+  int  o_len;
+  int  u_len;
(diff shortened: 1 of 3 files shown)


jsteube:
Very nice! Could you please add the unit test, too?

thatux:
based on the forum post I had swapped the user and owner part of the hash that is returned by pdf2john.pl
@philsmd where did you find this terminology: user & owner part?

I think it'd make more sense to do that in Hashcat self instead of having to rely on the user to swap that.
So that's a TODO. 

I'm working on understanding the Perl testsuite and attack itself to try make a unit test.

@philsmd any pointers to details of the workings of the pdf edit part of the algorithm i.e. to a reference implementation?

thatux:
Algorithm 3.6 and 3.7 seem to explain the terminology https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf we seem to be attacking the owner password, and m10500 is attacking the user password


https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/PDF32000_2008.pdf

thatux:
hmmpf. Apparently the owner password is dependent on the user password. I'll first implement it with an empty user password. This would probably make sense, as typically you'd probably need the user password first anyway.

thatux:
pdf edit password attack now works if no user password is set. we can directly attack the output from pdf2john.pl
there are two TODOs that could be worked upon in the future:
1. TODO use user password as input for md5 of o_digest if no owner password is set
2. TODO dynamically add user password including padding to the RC4 input for the computation of the pdf o-value

However, I think there is no problem with merging this already.

I'm not sure how to go about suppling a previously cracked password (user password in this case) an attack.
That'd mean we need to add this to the hash probably, therefore we'd no longer be able to use the output of pdf2john.pl directly.
It'd probably be best if supplying the user password would be optional?
Is it possible to have an optional part in the hash @jsteube?

jsteube:
Going to merge this, great work! 

About user-password: We can treat this as some kind of 'salt'. I suggest we add another column to the hash line so that the user can add the user-password to it. As a consequence this will break compatibility with the current hash line (as well as the pdf2john output). Therefore we should modify the module to accept both formats as valid (and inform JtR developers). In case the user password column is not existing, use the empty username as implemented now. If it is set, compute the MD5 in the module and replace the o_digest as described.

An example of how the tokenizer can be used to allow two different hash line formats can be found here: src/modules/module_22000.c. The idea is just to parse the old format first and if the tokenizer reports PARSER_OK, then modify the input line artificially to match the new input line format. If it reported not PARSER_OK, simply ignore it because it might be already in new input line format. Then do the new input line format parsing.

PR #2740, opened 2021-04-02
Downgrade timeoutpatch warning message to advice

Chick3nman:
Kernel times of >450ms are very uncommon and this warning is not a blocking, downgrading it to advice to allow for it to be hidden along with other advice messages.


jsteube:
Indeed. Good thinking!

PR #2749, opened 2021-04-11
KNX IP Secure - Device Authentication Code

robertguetzkow:
This plugin is for the KNX IP Secure protocol used for home and building automation. Due to the protocol's design it is susceptible to attacks on the server's authentication if a weak password has been used. 

When the client initiates a new session, it contacts the server with a `SESSION_REQUEST` that contains the public value X for ECDH. The server replies with a `SESSION_RESPONSE`. This frame contains a secure session identifier, the public value Y for ECDH and a MAC. The MAC is determined using AES-CCM with a custom block formatting function. It authenticates the frame header, secure session identifier, XOR of the public values and length of the authenticated data. Both the server and client have knowledge of the symmetric key that is used for AES-CCM. This key is called the device authentication code and is derived from a password using PBKDF2-HMAC-SHA256 with 65536 iterations and a constants salt `device-authentication-code.1.secure.ip.knx.org`. The plugin tries to determine this password that the server uses to authenticate itself to the client. It requires the secure session identifier, the XOR of the public values and the expected MAC as input. Since the `SESSION_REQUEST` and `SESSION_RESPONSE` are sent unencrypted, as the session key has not been established at this point, an attacker may be able to observe the content of these frames and thus gain the required information necessary to attack the authentication.

The plugin was developed as part of the security research about the KNX IP Secure protocol for my master's thesis.

NOTE: While this pull request also contains an attempt at an optimized kernel, it seems that the reduced operations in the SHA256 calculations have not significantly improved the performance. I'm open to suggestions for further improvements.


jsteube:
Many thanks for your contribution. This is really good research.

---

There's an implementation specific problem which make this scheme a bit harder to control in hashcat than other hash-mode implementation. While I did not know anything about this protocol before this PR, it obvious this scheme suffers from the same vulnerability which we know that WPA1/WPA2 has, too and which forces us to treat the plugin in a special way. 

The problem is that the salt for the PBKDF2 is fixed. Beside the problems it creates for us in handling the plugin properly, this could be seen as a major flaw because it allows precomputing of the compute intense parts. The strongest protection against brute-forcing is gone. Note that all the subsequent steps after the KDF are based on random data, so we can't rainbow table it, but precompute it.

From hashcat perspective this means that all the uniqueness for each hash line is stored somewhere in the esalt structure and not in the salt structure (as it should if this would be a normal scheme). 

To reproduce, let's start with this command:

```
$ tools/test.sh -t multi -a all -m 25900 -P -V1              
[ test_1618152460 ] > Init test for hash type 25900.
[ test_1618152460 ] [ Type 25900, Attack 0, Mode multi,  Device-Type Gpu, Vector-Width 1 ] > Error : 1/1 not found, 0/1 not matched, 0/1 timeout
```

The overview logging is a bit misleading, but if you take a look into the logfile (in my case test_1618152460/test_report.log) you see this:

```
Recovered........: 1/8 (12.50%) Digests
```

So there's one hash that was cracked, but the other 7 were not. That's exactly the problem we expect for this scheme. The reason is this:

```
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[DIGESTS_OFFSET].b1, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[DIGESTS_OFFSET].b2, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[DIGESTS_OFFSET].b3, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
```

Hashcat expects to iterate through the multiple hashes using the salt structure, not the esalt structure. The consequence is that the DIGESTS_OFFSET value is always zero, because we expect that have only one esalt entry per unique salt entry. That one cracked hash (from 1/8) is the one that is the first in the memory structure. The structure is generated by hashcat after sorting it with some logic, see src/hashes.c if interested but that's not relevant here.

What we need to do is inform hashcat that this plugin can have multiple esalt entries per unique salt entry. A good example for this is -m 22000 (WPA) because it suffers from the same problem (ESSID can be same). We need to set OPTS_TYPE_DEEP_COMP_KERNEL in the module. For a more detailed description, see the section "Data Structures: salt_t vs esalt" in the plugin developer guide. That's the first change.

```
static const u64   OPTS_TYPE       = OPTS_TYPE_PT_GENERATE_LE
                                   | OPTS_TYPE_DEEP_COMP_KERNEL;
```

Now that we have this special loop mode activated, we also need to update the _comp kernel slightly to make use of the different loop strategy hashcat follows. This means the DIGESTS_OFFSET alone is not enough anymore, we need to update it with:

```
  const u32 digest_pos = loop_pos;
  const u32 digest_cur = DIGESTS_OFFSET + digest_pos;
```

And then update the code which uses the variable:

```
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[digest_cur].b1, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[digest_cur].b2, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, esalt_bufs[digest_cur].b3, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
```

At this point, there's another problem. The function aes128_encrypt_cbc () expects an array input that is stored in the device register. However, we provide an array that resides in the global memory instead. CUDA doesn't report this as an error, but OpenCL does. There's two ways of fixing. Either you change the function declaration of aes128_encrypt_cbc () and add the macro GLOBAL_AS, or what I think is better, is to use something like this:

```
  u32 b1[4];

  b1[0] = esalt_bufs[digest_cur].b1[0];
  b1[1] = esalt_bufs[digest_cur].b1[1];
  b1[2] = esalt_bufs[digest_cur].b1[2];
  b1[3] = esalt_bufs[digest_cur].b1[3];

  u32 b2[4];

  b2[0] = esalt_bufs[digest_cur].b2[0];
  b2[1] = esalt_bufs[digest_cur].b2[1];
  b2[2] = esalt_bufs[digest_cur].b2[2];
  b2[3] = esalt_bufs[digest_cur].b2[3];

  u32 b3[4];

  b3[0] = esalt_bufs[digest_cur].b3[0];
  b3[1] = esalt_bufs[digest_cur].b3[1];
  b3[2] = esalt_bufs[digest_cur].b3[2];
  b3[3] = esalt_bufs[digest_cur].b3[3];

  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, b0, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, b1, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, b2, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
  aes128_encrypt_cbc (aes_ks, aes_cbc_iv, b3, yn, s_te0, s_te1, s_te2, s_te3, s_te4);
```

Now data is copied to registers first and it's good to use with aes128_encrypt_cbc (). As a bonus it aligns the code and makes it more beautiful. I am pretty sure you know already, but do not forget to remove cached kernels from $HOME/.nv $HOME/.cache and kernels after you change the code. You may also need to backport the changes to the optimized kernel.

There's one last thing to do. By activating the DEEP_KERNEL mode, hashcat expect us to define the _comp kernel name. In our case this is always the same, so it's pretty simple to fix:

```
u32 module_deep_comp_kernel (MAYBE_UNUSED const hashes_t *hashes, MAYBE_UNUSED const u32 salt_pos, MAYBE_UNUSED const u32 digest_pos)
{
  return KERN_RUN_3;
}
```


And in the module_init we register it:

```
  module_ctx->module_deep_comp_kernel         = module_deep_comp_kernel;
```

After these small changes, I was able to crack them all:

```
Recovered........: 8/8 (100.00%) Digests
```

---

The performance of the optimized kernel should be close to the pure kernel. The code from _init (and also _comp) does not affect much the overall performance. It's all about the most inner section of the _loop kernel. In this section the compilers' optimizer automatically generates code similar to your optimization because its clever enough to derive it from these hardcoded zero values and SHA256 constants:

```
  w2[0] = 0x80000000;
  w2[1] = 0;
  w2[2] = 0;
  w2[3] = 0;
  w3[0] = 0;
  w3[1] = 0;
  w3[2] = 0;
  w3[3] = (64 + 32) * 8;
```

A good way to find out about the compiler optimizers' changes is to look into the generated ptx code. Instead of using the CUDA backend, use the OpenCL backend on your NV GPU. You can then see the ptx code in human readable form in the kernels/ cache. Your optimized kernel demonstrates a good understanding of the underlaying crypto primitive SHA256 and a clever way how to optimize it. However, I would suggest to remove it from the PR because from production perspective its easier to maintain a single kernel than two kernels.

---

Great work!

robertguetzkow:
@jsteube Thank you very much for the fast and very detailed reply! The constant salt is indeed one of the many flaws in the protocol's design. A similar attack is also possible for the client's authentication when the attacker is able to impersonate a server or has one under their control.

Your explanation of `OPTS_TYPE_DEEP_COMP_KERNEL` clears up the misunderstandings I had. I also suspected that the compiler is clever enough to perform similar optimizations. This shows my lack of experience profiling and optimizing with OpenCL / CUDA, as this is the first kernel I wrote. It didn't hurt as an exercise, next time I will check the ptx code first though. 

Thank you for the tips, I will update the PR within the next week.

jsteube:
That was fast, many thanks!

PR #2762, opened 2021-04-17
Add mozilla2hashcat.py

Banaanhangwagen:


jsteube:
I can't merge this. While MIT licensed code can be used within GPL licensed project, the opposite direction is not. GPL is too strict. Please license your code under a MIT compatible license.

Banaanhangwagen:
Ah, licensing :-)
If I change my code to MIT-license, is that ok/enough? 
Since I re-use some parts of the code of @lclevy (GPL) is that ok also, or need I to rewrite it?

jsteube:
For hashcat it's enough if you change the license to MIT. I don't know about the reuse of some of code from @lclevy (GPL). We're all just coders not lawyers :)

If you think it's not possible we can't merge the code. In this case what people do is to create a separate repository for the project, in this case a repository mozilla2hashcat. See https://github.com/philsmd/7z2hashcat

lclevy:
Hi, I'm ok for MIT since your rewrote most of the code and citing me at the beginning of the code

jsteube:
OK, great. Thank for making this a success to @Banaanhangwagen and @lclevy

PR #2775, opened 2021-04-27
Change typedef hashconfig_t

ventaquil:
This `typedef` looks different than any other in `include/types.h`.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -935,7 +935,7 @@ typedef struct hashes
 
 } hashes_t;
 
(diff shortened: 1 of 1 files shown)


jsteube:
I think is from a time when we had parse_func inside hashconfig_t which references on struct hashconfig itself. But we have this disabled now so it's fine to merge.

ventaquil:
I quickly checked for any disadvantages but I didn't find any so here you have my little commit.

PR #2777, opened 2021-04-29
Added support for IBM KDFAES passwords, only modes 0 and 3

ecostin:


jsteube:
Hey there. First of all many thanks for the contribution. I didn't do an in-depth review of the algorithm but I can already see a few general problems. 

---

My first impression that this plugin was written a long time ago, where hashcat had no real plugin interface. The idea is because of this:

```
  if ((line_len < (7 + 1 + 1 + 1 + 16 + 1 + 8 + 24 + 32 + 16)) || (line_len > (7 + 1 + 8 + 1 + 16 + 1 + 8 + 24 + 32 + 16))) return (PARSER_GLOBAL_LENGTH);
```

This pattern looks familar to me, is from a very old hashcat style module. It's from a time when there was no tokenizer it seems. I would really appreciate if you can rewrite the module to a regular module. This makes it a ton more easy to understand whats going on here. Please refer to https://github.com/hashcat/hashcat/blob/master/src/modules/module_08500.c as an example.

---

Next thing is the following snippet from module_hash_encode():

```
int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
{
  const int line_len = snprintf (line_buf, line_size, "%s", hash_info->orighash);
  return line_len;
}
```

This is perfectly valid to do in hashcat, but we should not. Please refer to the plugin developer guide for the reason behind, but basically I find it better to encode the hash with the same effort we do to decode it. 

---

The lack of unit test in tools/test_modules/ disables me from supporting you with rewriting parts of the kernel. Also without this test I can't generate artificial hashes with known passwords easily on my local system. Beside the testing functionality, it also is a nice high level view on how the hash is calculated. If you could add it I can help you with rewriting the main kernel because then I have a better understanding.

From my first impression it looks like it basically continues what it does in -m 8500 and then uses the output hash as input password for the PBKDF2. Please feel free to correct me here. This is a typical way of how password hashes are upgraded without the need for the user to enter the password. Please confirm.

---

From kernel perspective I don't understand why you have written a perfectly implemented slow hash type kernel into a non matching fast hash kernel. If you had this implemented as slow hash kernel, you get all the attack mode functionalities for free. Plus you get the watchdog timer protection (because of split iterations). There's also a perfectly optimized PBKDF2-HMAC-SHA256 kernel.

Typically I'd have this kernel implemented by using -m 10900 type kernel as template and then add the DES phase based one -m 8500 and copy it into the _init kernel.

---

In the kernel there's multiple checks for REAL_SHM support, for instance here:

```
DECLSPEC void racf_kdfaes_aes_256(u32 *key, u32 *input, u32 *output
  // the aes implementation wants these buffers if REAL_SHM is defined.
  #ifdef REAL_SHM
    , LOCAL_AS u32 s_te0[256], LOCAL_AS u32 s_te1[256], LOCAL_AS u32 s_te2[256], LOCAL_AS u32 s_te3[256], LOCAL_AS u32 s_te4[256]
  #endif
)
```

Having this test is fine if you also provide an else {} solution. This makes the kernel useable only on GPU, but not on CPU. If you run hashcat with -D1, it will fail to compile. Is another reason to have the unit test.

---

So far I didn't went further in reviewing but I think it's enough to seriously think of rewriting the kernel into a slow hash kernel.

ecostin:
Hi, thanks for the feedback. This was indeed written a very long time ago for an older hashcat and ported to the plugin model in a hurry. I'll try to apply your suggestions, starting with rewriting it into a slow hash kernel.

Regarding your questions, it starts indeed with DES (as implemented in 8500), but the key derivation continues with a custom IBM PBKDF2 implementation, so we cannot use the already existing PBKDF2 algos. At the end we have a perfectly standard AES with the username as salt.

I will align the custom PBKDF2 to the code you've referenced.

jsteube:
Many thanks! Feel free to come to IRC if there's any questions.

PR #2804, opened 2021-06-04
Added new feature: autodetect hash-mode

matrix:
Hi, to make hashcat more user-friendly I have implemented a new feature: autodetect the hash-mode.

In essence, this new feature greatly simplifies the hashcat workflow :)

Specifically, if more than one compatible algorithm is found, a list of hash-modes to use will be shown.
In case the hash-mode is compatible with only one format supported by hashcat, the attack will be performed.

This new feature can be used by omitting the "-m" parameter, here are some examples:

**Multiple matches**
```
$ head -1 example0.hash > unknown_hash 
$ ./hashcat unknown_hash example.dict 
hashcat (v6.2.1-117-gdf6cafeba) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

Match: 0 - MD5
Match: 70 - md5(utf16le($pass))
Match: 900 - MD4
Match: 1000 - NTLM
Match: 2600 - md5(md5($pass))
Match: 3000 - LM
Match: 3500 - md5(md5(md5($pass)))
Match: 4300 - md5(strtoupper(md5($pass)))
Match: 4400 - md5(sha1($pass))
Match: 8600 - Lotus Notes/Domino 5
Match: 9900 - Radmin2
Match: 20900 - md5(sha1($pass).md5($pass).sha1($pass))

Multiple match(s), please specify the hash-mode by argument (-m).

Started: Fri Jun  4 15:46:04 2021
Stopped: Fri Jun  4 15:46:05 2021
```

**Single match**

```
$ head -1 example400.hash > unknown_hash
$ ./hashcat unknown_hash example.dict 
hashcat (v6.2.1-117-gdf6cafeba) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

Match: 400 - phpass

Rules: 1

Optimizers applied:
* Zero-Byte
* Single-Hash
* Single-Salt

Using pure kernels enables cracking longer passwords but for the price of drastically reduced performance.
If you want to switch to optimized backend kernels, append -O to your commandline.

Watchdog: Hardware monitoring interface not found on your system.
Watchdog: Temperature abort trigger disabled.

Host memory required for this attack: 1 MB

Dictionary cache built:
* Filename..: example.dict
* Passwords.: 128416
* Bytes.....: 1069601
* Keyspace..: 128416
* Runtime...: 0 secs

$H$9y5boZ2wsUlgl2tI6b5PrRoADzYfXD1:hash234                
                                                          
Session..........: hashcat
Status...........: Cracked
Hash.Name........: phpass
Hash.Target......: $H$9y5boZ2wsUlgl2tI6b5PrRoADzYfXD1
Time.Started.....: Fri Jun  4 15:47:02 2021 (6 secs)
Time.Estimated...: Fri Jun  4 15:47:08 2021 (0 secs)
Guess.Base.......: File (example.dict)
Speed.#1.........:    12584 H/s (9.93ms) @ Accel:512 Loops:128 Thr:1 Vec:4
Progress.........: 71680/128416 (55.82%)
Rejected.........: 0/71680 (0.00%)
Restore.Point....: 69632/128416 (54.22%)
Restore.Sub.#1...: Salt:0 Amplifier:0-1 Iteration:1920-2048
Candidates.#1....: h0n07u7u -> hhhpimp

Started: Fri Jun  4 15:46:54 2021
Stopped: Fri Jun  4 15:47:08 2021
```

Thanks :)

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -22,6 +22,12 @@
 - Added hash-mode: sha384(utf16le($pass).$salt)
 
 ##
+## Features
+##
+
+- Autodetect hash-type: performs an automatic analysis of the input hash(es), associating compatible algorithms, or executing the attack if only one compatible format is found.
+
+##
 ## Bugs
 ##
 
@@ -41,7 +47,6 @@
 - RC4 Kernels: Improved performance by 20%+ for hash-modes Kerberos 5 (etype 23), MS Office (<= 2003) and PDF (<= 1.6) by using new RC4 code
 - Status Screen: Show currently running kernel type (pure, optimized) and generator type (host, device)
 - UTF8-to-UTF16: Replaced naive UTF8 to UTF16 conversion with true conversion for RAR3, AES Crypt, MultiBit HD (scrypt) and Umbraco HMAC-SHA1
-- Autodetect hash-type: performs an automatic analysis of the hashlist, associating compatible algorithms, or executing the attack if only one compatible format is found.
 
 ##
 ## Technical
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -41,6 +41,7 @@
 - RC4 Kernels: Improved performance by 20%+ for hash-modes Kerberos 5 (etype 23), MS Office (<= 2003) and PDF (<= 1.6) by using new RC4 code
 - Status Screen: Show currently running kernel type (pure, optimized) and generator type (host, device)
 - UTF8-to-UTF16: Replaced naive UTF8 to UTF16 conversion with true conversion for RAR3, AES Crypt, MultiBit HD (scrypt) and Umbraco HMAC-SHA1
+- Autodetect hash-type: performs an automatic analysis of the hashlist, associating compatible algorithms, or executing the attack if only one compatible format is found.
 
 ##
 ## Technical
--- a/example0.cmd
+++ b/example0.cmd
@@ -1,2 +1,2 @@
-hashcat.exe -t 32 -a 7 example0.hash ?a?a?a?a example.dict
+hashcat.exe -m 0 -t 32 -a 7 example0.hash ?a?a?a?a example.dict
 pause
--- a/example0.sh
+++ b/example0.sh
@@ -1,1 +1,1 @@
-./hashcat -t 32 -a 7 example0.hash ?a?a?a?a example.dict
+./hashcat -m 0 -t 32 -a 7 example0.hash ?a?a?a?a example.dict
--- a/include/types.h
+++ b/include/types.h
@@ -210,6 +210,7 @@ typedef enum status_rc
   STATUS_ABORTED_RUNTIME    = 11,
   STATUS_ERROR              = 13,
   STATUS_ABORTED_FINISH     = 14,
+  STATUS_AUTODETECT         = 16,
 
 } status_rc_t;
 
@@ -443,6 +444,7 @@ typedef enum opts_type
   OPTS_TYPE_MP_MULTI_DISABLE  = (1ULL << 52), // do not multiply the kernel-accel with the multiprocessor count per device to allow more fine-tuned workload settings
   OPTS_TYPE_NATIVE_THREADS    = (1ULL << 53), // forces "native" thread count: CPU=1, GPU-Intel=8, GPU-AMD=64 (wavefront), GPU-NV=32 (warps)
   OPTS_TYPE_POST_AMP_UTF16LE  = (1ULL << 54), // run the utf8 to utf16le conversion kernel after they have been processed from amplifiers
+  OPTS_TYPE_AUTODETECT_DISABLE = (1ULL << 55), // skip autodetect engine
 
 } opts_type_t;
 
@@ -594,6 +596,7 @@ typedef enum user_options_defaults
 {
   ADVICE_DISABLE           = false,
   ATTACK_MODE              = ATTACK_MODE_STRAIGHT,
+  AUTODETECT               = false,
   BENCHMARK_ALL            = false,
   BENCHMARK                = false,
   BITMAP_MAX               = 18,
@@ -1939,6 +1942,7 @@ typedef struct user_options
   char       **hc_argv;
 
   bool         attack_mode_chgd;
+  bool         autodetect;
   #ifdef WITH_BRAIN
   bool         brain_host_chgd;
   bool         brain_port_chgd;
--- a/include/usage.h
+++ b/include/usage.h
@@ -9,7 +9,16 @@
 #include <stdio.h>
 #include <string.h>
 
+typedef struct usage_sort
+{
+  u32   hash_mode;
+  char *hash_name;
+  u32   hash_category;
+
+} usage_sort_t;
+
 void usage_mini_print (const char *progname);
 void usage_big_print  (hashcat_ctx_t *hashcat_ctx);
+int sort_by_usage (const void *p1, const void *p2);
 
 #endif // _USAGE_H
--- a/src/Makefile
+++ b/src/Makefile
@@ -189,7 +189,6 @@ endif
 ## because LZMA SDK
 ifeq ($(CC),clang)
 CFLAGS                  += -Wno-enum-conversion
-CFLAGS                  += -Wno-typedef-redefinition
 endif
 
 ## because ZLIB
--- a/src/Makefile
+++ b/src/Makefile
@@ -189,6 +189,7 @@ endif
 ## because LZMA SDK
 ifeq ($(CC),clang)
 CFLAGS                  += -Wno-enum-conversion
+CFLAGS                  += -Wno-typedef-redefinition
 endif
 
 ## because ZLIB
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -1444,11 +1444,18 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
 
     int *modes_buf = (int *) hccalloc (MODULE_HASH_MODES_MAXIMUM, sizeof (int));
 
-    if (!modes_buf) return -1;
+    if (modes_buf == NULL) return -1;
 
     const int modes_cnt = autodetect_hashmodes (hashcat_ctx, modes_buf);
 
-    if (modes_cnt > 0)
+    if (modes_cnt <= 0)
+    {
+      if (user_options->show == false) event_log_error (hashcat_ctx, "No hash-mode matches the structure of the input hash.");
+
+      return -1;
+    }
+
+    if (modes_cnt > 1)
     {
       event_log_info (hashcat_ctx, "The following %d hash-mode match the structure of your input hash:", modes_cnt);
       event_log_info (hashcat_ctx, NULL);
@@ -1463,35 +1470,33 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
         {
           event_log_info (hashcat_ctx, "%7u | %-51s | %s", hashcat_ctx->hashconfig->hash_mode, hashcat_ctx->hashconfig->hash_name, strhashcategory (hashcat_ctx->hashconfig->hash_category));
 
-          if (modes_cnt != 1) hashconfig_destroy (hashcat_ctx); // keep for later
+          hashconfig_destroy (hashcat_ctx);
         }
       }
 
       event_log_info (hashcat_ctx, NULL);
 
-      if (modes_cnt > 1)
-      {
-        event_log_error (hashcat_ctx, "Please specify the hash-mode by argument (-m).");
-
-        return -1;
-      }
-      else // 1
-      {
-        event_log_warning (hashcat_ctx, "You have not specified -m to select the correct hash-mode.");
-        event_log_warning (hashcat_ctx, "It was automatically selected by hashcat because it was the only hash-mode matching your input hash.");
-        event_log_warning (hashcat_ctx, "Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.");
-        event_log_warning (hashcat_ctx, "Do not report hashcat issues if you do not know exactly how the hash was extracted.");
-        event_log_warning (hashcat_ctx, NULL);
-
-        user_options->autodetect = false;
-      }
-    }
-    else
-    {
-      if (user_options->show == false) event_log_error (hashcat_ctx, "No hash-mode matches the structure of the input hash.");
+      event_log_error (hashcat_ctx, "Please specify the hash-mode by argument (-m).");
 
       return -1;
     }
+
+    // modes_cnt == 1
+
+    user_options->hash_mode = modes_buf[0];
+
+    if (hashconfig_init (hashcat_ctx) != 0) return -1;
+
+    event_log_warning (hashcat_ctx, "You have not specified -m to select the correct hash-mode.");
+    event_log_warning (hashcat_ctx, "It was automatically selected by hashcat because it was the only hash-mode matching your input hash:");
+    event_log_warning (hashcat_ctx, "\n%u | %s | %s\n", hashcat_ctx->hashconfig->hash_mode, hashcat_ctx->hashconfig->hash_name, strhashcategory (hashcat_ctx->hashconfig->hash_category));
+    event_log_warning (hashcat_ctx, "Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.");
+    event_log_warning (hashcat_ctx, "Do not report hashcat issues if you do not know exactly how the hash was extracted.");
+    event_log_warning (hashcat_ctx, NULL);
+
+    hashconfig_destroy (hashcat_ctx);
+
+    user_options->autodetect = false;
   }
 
   /**
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -28,6 +28,7 @@
 #include "event.h"
 #include "hashes.h"
 #include "hwmon.h"
+#include "hlfmt.h"
 #include "induct.h"
 #include "interface.h"
 #include "logfile.h"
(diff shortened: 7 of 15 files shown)


Xanadrel:
I would still make the user select the mode in case there is a single match.
(It can also come with a note to say that even if there is a match it may not be an actual hash)

matrix:
Hi @Xanadrel, if user need to set "-m" also with single match ... it is useless :)
In the hashcat output you can see `Match: 400 - phpass` and `Hash.Name........: phpass`
Is not enough?

Xanadrel:
@matrix no, not enough for me, I just want to avoid users reporting issues cause their "hash" is not cracking while they use the autodetect feature, so a strong emphasis on the fact that the "hash" MAY be of a given type is needed.

Also, what happens if there is not a single match ? I can be wrong but maybe the case is not handled.

jsteube:
From my first review I can tell it's a very good implementation, is very seamless.

---

But I agree to @Xanadrel that the automatic selection is not clear enough/visible enough. At least make it a warning so it appears in yellow.

---

There's another problem and that is if the hash does not match any hash-mode. In that case, hashcat just finishes after any type of error message. 

```
root@ht:~/hashcat# cat > hash.txt
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaax
root@ht:~/hashcat# ./hashcat hash.txt
hashcat (v6.2.1-116-g918a62150+) starting...

CUDA API (CUDA 11.3)
====================
* Device #1: NVIDIA GeForce GTX 980, 3963/4043 MB, 16MCU

OpenCL API (OpenCL 3.0 CUDA 11.3.55) - Platform #2 [NVIDIA Corporation]
* Device #3: NVIDIA GeForce GTX 980, skipped

Started: Fri Jun  4 19:24:03 2021
Stopped: Fri Jun  4 19:24:03 2021
```

That clearly should not happen. Instead it should output some error that no hash mode matched.

---

I think we can also make the selection table more like a real table. Most easy way and probably most informative is to copy the code from how the --help menu is structured. It will look like this:

```
The following hash-mode match the hash structure of your input hash:

      # | Name                                                | Category
  ======+=====================================================+======================================
    900 | MD4                                                 | Raw Hash
      0 | MD5                                                 | Raw Hash
     70 | md5(utf16le($pass))                                 | Raw Hash
   2600 | md5(md5($pass))                                     | Raw Hash, Salted and/or Iterated
   3500 | md5(md5(md5($pass)))                                | Raw Hash, Salted and/or Iterated
   4400 | md5(sha1($pass))                                    | Raw Hash, Salted and/or Iterated
   4300 | md5(strtoupper(md5($pass)))                         | Raw Hash, Salted and/or Iterated
  20900 | md5(sha1($pass).md5($pass).sha1($pass))             | Raw Hash, Salted and/or Iterated
   3000 | LM                                                  | Operating System
   1000 | NTLM                                                | Operating System
   9900 | Radmin2                                             | Operating System
   8600 | Lotus Notes/Domino 5                                | Enterprise Application Software (EAS)

```

The ordering is relevant. Please refer to src/usage.c how it is been done. I think you can more or less copy the code.

---

This I found by accident and did not follow what's happening, but it looks wrong:

```
root@ht:~/hashcat# ./hashcat 1
hashcat (v6.2.1-116-g918a62150+) starting...

CUDA API (CUDA 11.3)
====================
* Device #1: NVIDIA GeForce GTX 980, 3963/4043 MB, 16MCU

OpenCL API (OpenCL 3.0 CUDA 11.3.55) - Platform #2 [NVIDIA Corporation]
* Device #3: NVIDIA GeForce GTX 980, skipped

Hash '1': Token length exception
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched
Hash '1': Separator unmatched

....
100 more lines
....

Started: Fri Jun  4 19:35:02 2021
Stopped: Fri Jun  4 19:35:03 2021

```

---

jsteube:
More remarks:

---

For the case when there's only one matching algorithm we should also add a message like we have in case the user uses --force:

```
You have not specified -m to select the correct hash-mode. 
It was automatically selected by hashcat because it was the only hash-mode matching your input hash.
Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.
Do not report hashcat issues if you do not know exactly how the hash was extracted.
```

---

I didn't try, but what is happening if the user specifies multiple hashes of which multiple match multiple hash-modes. 

---

Similar to the above, what is happening in case there's multiple hashes in the hashlist, but one matches and the other does not. 

---

dropdeadfu:
I think the feature is a good idea in general but will not solve the underlying issue that new users might be mislead by hashtypes.
At the end of the day it is better being already presented with a list of feasable modes then not having it and forcing the user to a round-trip on the hash-example-page. As long as it is made clear that he presented possible matches may not contain the actual target algo in newbie-friendly language it should be fine.

Chick3nman:
How does this handle some of the more complicated formats? Such as the file formats like Truecrypt, the shadow file formats we accept for some of the crypt hashes, or the pwdump format we accept for LM and NTLM. The parser for pwdump format handles 2 different hash types and relies on the user to specify the mode to know which one to actually run. For hashes like LM, where it can come in 2 different formats, either as a half or combined into the full length most people recognize, it could be difficult for the user to understand why multiple formats return the same hash type. There are also several modes that take identical hash formats, such as the collider modes for old office documents and pdf files. I assume we would want to default to the non-collider modes, but we would need to consider this for the special modes like that.

I think this can be somewhat convenient, but there are quite a lot of formats and many that take special considerations in their parsing or mode selection. I worry that maintaining this across hundreds of algorithms could be a fairly sizable amount of work for that little convenience, especially if we are trying to keep it's behavior consistent across the board.

matrix:
Hi, I have reviewed the code, following the directions you wrote. The idea is to just try to avoid the manual work of identifying a hash with a hash-mode, as also wrote @dropdeadfu

Obviously if you feed it a mix of hashes, I expect to get different algorithms in output, so if it is used incorrectly, it could produce unexpected results.

matrix:
```
$ ./hashcat 1
hashcat (v6.2.1-117-gdf6cafeba+) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

You have not specified -m to select the correct hash-mode.
It was automatically selected by hashcat because it was the only hash-mode matching your input hash.
Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.
Do not report hashcat issues if you do not know exactly how the hash was extracted.

No match found.

Started: Fri Jun  4 23:51:44 2021
Stopped: Fri Jun  4 23:51:44 2021
```

```
$ head -1 example0.hash > hash
$ ./hashcat hash
hashcat (v6.2.1-117-gdf6cafeba+) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

You have not specified -m to select the correct hash-mode.
It was automatically selected by hashcat because it was the only hash-mode matching your input hash.
Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.
Do not report hashcat issues if you do not know exactly how the hash was extracted.

Match: 0 - MD5
Match: 70 - md5(utf16le($pass))
Match: 900 - MD4
Match: 1000 - NTLM
Match: 2600 - md5(md5($pass))
Match: 3000 - LM
Match: 3500 - md5(md5(md5($pass)))
Match: 4300 - md5(strtoupper(md5($pass)))
Match: 4400 - md5(sha1($pass))
Match: 8600 - Lotus Notes/Domino 5
Match: 9900 - Radmin2
Match: 20900 - md5(sha1($pass).md5($pass).sha1($pass))

Multiple match(s), please specify the hash-mode by argument (-m).

Started: Fri Jun  4 23:52:49 2021
Stopped: Fri Jun  4 23:52:50 2021
```

```
$ head -1 example400.hash > hash
$ ./hashcat hash
hashcat (v6.2.1-117-gdf6cafeba+) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

You have not specified -m to select the correct hash-mode.
It was automatically selected by hashcat because it was the only hash-mode matching your input hash.
Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.
Do not report hashcat issues if you do not know exactly how the hash was extracted.

Match: 400 - phpass

Rules: 1

Optimizers applied:
* Zero-Byte
* Single-Hash
* Single-Salt
[...]

**Comment by jsteube** [comment]
Maybe you had understood me incorrectly, but that's not what I was saying:

```
$ ./hashcat 1
hashcat (v6.2.1-117-gdf6cafeba+) starting...

* Device #1: Intel(R) Core(TM) i7-4578U CPU @ 3.00GHz, 8128/8192 MB (2048 MB allocatable), 4MCU
* Device #2: Iris, skipped

You have not specified -m to select the correct hash-mode.
It was automatically selected by hashcat because it was the only hash-mode matching your input hash.
Under no circumstances it is not to be understood as a guarantee this is the right hash-mode.
Do not report hashcat issues if you do not know exactly how the hash was extracted.

No match found.
```

If hashcat doesn't select a mode, we can't print the opposite by saying 'It was automatically selected by hashcat'. So that message is only in case hashcat is actually doing a selection because it actually found just one mode and then it makes sense.

The same applies for this scenario with the table being printed. Hashcat did not select a mode.

---

> Obviously if you feed it a mix of hashes, I expect to get different algorithms in output, so if it is used incorrectly, it could produce unexpected results.

That's too easy. There are scenarios that can come up and we can't just say shit in - shit out. We have some responsibility.

I think hashcat has to collect some information first before it can inform the user about the possible modes. The reason is because there's different situations and hashcat has to react accordingly. 

1. There is a hashlist in which all hashes match with hash-mode X, Y and Z. In this case it should print X, Y and Z as possible modes. 
2. There's one hash only matches X and Y, but all other hashes match X, Y and Z. In this case it should instead print that there's probably a different set of pattern in the hashlist. If we want to do it right, it needs to print all matching modes for all hashes inside the hashlist. But what do we do if there's millions of hashes in the hashlist? The output would be millions of lines, so that's not the way to go. So the reaction should be that it is not printing any possible modes. Better none than false.
3. There's only hash mode X matching for all hash entries, in this case it should automatically select this mode and proceed with cracking.
4. possible other scenarios ?

---

This table:

```
Match: 0 - MD5
Match: 70 - md5(utf16le($pass))
Match: 900 - MD4
Match: 1000 - NTLM
Match: 2600 - md5(md5($pass))
Match: 3000 - LM
Match: 3500 - md5(md5(md5($pass)))
Match: 4300 - md5(strtoupper(md5($pass)))
Match: 4400 - md5(sha1($pass))
Match: 8600 - Lotus Notes/Domino 5
Match: 9900 - Radmin2
Match: 20900 - md5(sha1($pass).md5($pass).sha1($pass))
```

You didn't mention it explicitly if you are going or if you are not going to transform this into the format I had suggested. However, you send in some patches afterwards. Because you did send in patches and at the same time you did not give any feedback to this suggestion I have to assume you accidentally overread it. 

I wouldn't care if that point is one of the most important ones in order to reduce the number of invalid issues which we will receive for sure after we add this feature. If we order it in a nice way and also add some minimal documentation (hash category) it's much more likely that the user are willing to activate the brains rather than just brute forcing through all of them and then find out that none of them match the hash mode they need. 

---

I think @Chick3nman had excellent points:

> How does this handle some of the more complicated formats? Such as the file formats like Truecrypt

For Truecrypt or any binary formats that is indeed a good question. Some decoders expect to receive a filename rather than a hash, in particular the ones with OPTS_TYPE_BINARY_HASHFILE. I think the brute force loop needs to handle this case separately in the same way as it is written in src/hashes.c

> the shadow file formats we accept for some of the crypt hashes, or the pwdump format we accept for LM and NTLM. The parser for pwdump format handles 2 different hash types and relies on the user to specify the mode to know which one to actually run.

Basically the brute force loop also needs to handle this case the same way as in truecrypt, see src/hashes.c

> For hashes like LM, where it can come in 2 different formats, either as a half or combined into the full length most people recognize, it could be difficult for the user to understand why multiple formats return the same hash type. 

I think LM is the only hash with this split 'feature' and I think we can ignore it. People who crack LM nowadays know about that fact.

> There are also several modes that take identical hash formats, such as the collider modes for old office documents and pdf files. I assume we would want to default to the non-collider modes, but we would need to consider this for the special modes like that.

I don't think we need special treatment here. We will list both modes in the table

> I think this can be somewhat convenient, but there are quite a lot of formats and many that take special considerations in their parsing or mode selection. I worry that maintaining this across hundreds of algorithms could be a fairly sizable amount of work for that little convenience, especially if we are trying to keep it's behavior consistent across the board.

That is indeed true but by looking into src/hashes.c the special cases a limited in number because they identify themself using special OPTS_TYPE_* flags that the brute force loop can use to decide which branch to go to. That was a requirement when I designed the plugin interface that we will never name a specific hash-mode in the core code.

---

We also discussed on IRC about having some sort of blacklisting in the plugin a developer can set to disable the plugin from being used in the autodetection. We can add some macros like OPTS_TYPE_AUTODETECT_DISABLE and then check for this flag in the brute force loop and skip it. We can also do the opposite that a plugin needs to be white listed but I think black listing fits better.

**Comment by matrix** [comment]
Hi, I think the best solution for now is to agree on IRC, so we avoid misunderstanding.

**Comment by matrix** [comment]
Pushed new version, feat. @jsteube ;)

**Comment by funkymonkey5150** [comment]
Nice

**Comment by jsteube** [comment]
OK after review new version, following things:

---

```
./hashcat example0.hash 
```

Ordering of table entries is lost. It was working with old version. Please restore.

---

```
./hashcat example400.hash 
```

1. Automatic selection is lost. If only 1 mode matches, we can use it. I think best way to do it is to set user_options->hash_mode to the new mode and to set user_options->hash_mode_chgd = true and then continue with hashcat_session_execute() regularly.
2. If only 1 mode match do not display table. It will confuse user.

---

I know I'm a bit lazy on commenting code. This is the result:

```
  if (user_options->hash_mode == 0 && user_options->hash_mode_chgd == false)
  {
    user_options->autodetect = true;
  }
```

You had placed this code into user_options_sanity(), but that's the wrong place. If you take a close look at user_options_sanity() there's really only tests being done and no variables modified (written to). You couldn't know because it's not documented.

The way to derive special variables depening on user input is user_options_preprocess() or user_options_postprocess(). I'd say place it into user_options_preprocess().

But, be carefull, because:

```
  if (user_options->stdout_flag == true)
  {
    user_options->force                 = true;
    user_options->hash_mode             = 2000;
```

This sets the mode to 2000. So we need to move the autodetect = true block below that section! I'd suggest to move it to the very bottom of the user_options_preprocess() function. This will also conveniently disable the autodetect in case the user uses --stdout.

---

After thinking about what needs to happen after we found only 1 matching hash_mode, then I realized we completely forgot about that at this point in time the many other initialization functions (like hashcat_session_init() and many other) are already executed at this point while user_options->hash_mode = 0.

So I had to find out if there's anything relevant that depends on user_options->hash_mode before we autodetect it and set it afterwards. 

First I did a manual check across all function code that ran before and then just to double check I ran a debugging session and added some breakpoints to trigger when user_options->hash_mode is accessed. Here's the results:

So this was made with commit 72d1937fe79b1a7499add71a22342cf697358b4e after applying this PR code and moving the autodetect = true block to the bottom line as discussed above.

write access to hash_mode:

```
root@ht:~/hashcat# gdb ./hashcat

(gdb) break src/main.c:1119
Breakpoint 1 at 0x9dc1: file src/main.c, line 1122.

(gdb) r example400.hash

Breakpoint 1, main (argc=2, argv=0x7fffffffe448) at src/main.c:1122
1122      if (user_options_getopt (hashcat_ctx, argc, argv) == -1)

(gdb) p &hashcat_ctx->user_options->hash_mode 
$1 = (int *) 0x5555557464b0

(gdb) watch *(int*)0x5555557464b0
Hardware watchpoint 2: *(int*)0x5555557464b0

(gdb) c
Continuing.

hashcat (v6.2.1-117-g72d1937fe+) starting...

CUDA API (CUDA 11.3)
====================
* Device #1: NVIDIA GeForce GTX 980, 3085/4043 MB, 16MCU

OpenCL API (OpenCL 3.0 CUDA 11.3.55) - Platform #2 [NVIDIA Corporation]
* Device #3: NVIDIA GeForce GTX 980, skipped


Thread 1 "hashcat" hit Hardware watchpoint 2: *(int*)0x5555557464b0

Old value = 0
New value = 1
autodetect_hashmodes (hashcat_ctx=0x55555572d2a0, modes_buf=0x555555d3f610) at src/hashcat.c:1375
1375        module_filename (folder_config, i, modulefile, HCBUFSIZ_TINY);
(gdb) 
```

Perfect! No problems. Now read access to hash_mode:

```
root@ht:~/hashcat# gdb ./hashcat

(gdb) break src/main.c:1119
Breakpoint 1 at 0x9dc1: file src/main.c, line 1122.

(gdb) r example400.hash

Breakpoint 1, main (argc=2, argv=0x7fffffffe448) at src/main.c:1122
1122      if (user_options_getopt (hashcat_ctx, argc, argv) == -1)

(gdb) p &hashcat_ctx->user_options->hash_mode
$1 = (int *) 0x5555557464b0

(gdb) rwatch *(int*)0x5555557464b0           
Hardware read watchpoint 2: *(int*)0x5555557464b0

(gdb) c
Continuing.

Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
user_options_sanity (hashcat_ctx=0x55555572d2a0) at src/user_options.c:661
661       if (user_options->hash_mode >= MODULE_HASH_MODES_MAXIMUM)
(gdb) c
Continuing.
hashcat (v6.2.1-117-g72d1937fe+) starting...


Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
user_options_preprocess (hashcat_ctx=0x55555572d2a0) at src/user_options.c:1897
1897      if (user_options->hash_mode == 0 && user_options->hash_mode_chgd == false)
(gdb) c
Continuing.

Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
default_hash_mode (hashconfig=0x5555557435f0, user_options=0x555555746350, user_options_extra=0x555555746320) at src/interface.c:533
533       const u32 hash_mode = user_options->hash_mode;
(gdb) bt
#0  default_hash_mode (hashconfig=0x5555557435f0, user_options=0x555555746350, user_options_extra=0x555555746320) at src/interface.c:533
#1  0x0000555555572ee3 in hashconfig_init (hashcat_ctx=0x55555572d2a0) at src/interface.c:90
#2  0x00005555555a0675 in user_options_check_files (hashcat_ctx=0x55555572d2a0) at src/user_options.c:2844
#3  0x0000555555560d75 in hashcat_session_init (hashcat_ctx=0x55555572d2a0, install_folder=0x555555702300 "/usr/local/bin", 
    shared_folder=0x55555570230f "/usr/local/share/hashcat", argc=2, argv=0x7fffffffe448, comptime=1622964656) at src/hashcat.c:1117
#4  0x000055555555def1 in main (argc=2, argv=0x7fffffffe448) at src/main.c:1166
```
OK, here's some relevant access, but it's the section where hashcat tries to find out if it can access kernel files, which is later relevant ot work around -I bugs in OpenCL drivers. We can ignore that section.

```
(gdb) c
Continuing.

Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
hashconfig_init (hashcat_ctx=0x55555572d2a0) at src/interface.c:107
107       const bool rc_load = module_load (hashcat_ctx, module_ctx, user_options->hash_mode);
(gdb) bt
#0  hashconfig_init (hashcat_ctx=0x55555572d2a0) at src/interface.c:107
#1  0x00005555555a0675 in user_options_check_files (hashcat_ctx=0x55555572d2a0) at src/user_options.c:2844
#2  0x0000555555560d75 in hashcat_session_init (hashcat_ctx=0x55555572d2a0, install_folder=0x555555702300 "/usr/local/bin", 
    shared_folder=0x55555570230f "/usr/local/share/hashcat", argc=2, argv=0x7fffffffe448, comptime=1622964656) at src/hashcat.c:1117
#3  0x000055555555def1 in main (argc=2, argv=0x7fffffffe448) at src/main.c:1166
```

Still same section, ignore

```
(gdb) c
Continuing.
[New Thread 0x7fffe5620700 (LWP 176818)]
[New Thread 0x7fffe3172700 (LWP 176819)]
[New Thread 0x7fffe2971700 (LWP 176820)]
[Thread 0x7fffe2971700 (LWP 176820) exited]
[Thread 0x7fffe3172700 (LWP 176819) exited]
CUDA API (CUDA 11.3)
====================
* Device #1: NVIDIA GeForce GTX 980, 3085/4043 MB, 16MCU

OpenCL API (OpenCL 3.0 CUDA 11.3.55) - Platform #2 [NVIDIA Corporation]
* Device #3: NVIDIA GeForce GTX 980, skipped


Thread 1 "hashcat" hit Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
0x00005555555a13c2 in user_options_logger (hashcat_ctx=0x55555572d2a0) at src/user_options.c:3024
3024      logfile_top_uint   (user_options->hash_mode);
(gdb) c
Continuing.

Thread 1 "hashcat" hit Hardware read watchpoint 2: *(int*)0x5555557464b0

Value = 0
autodetect_hashmodes (hashcat_ctx=0x55555572d2a0, modes_buf=0x555555d3f610) at src/hashcat.c:1375
1375        module_filename (folder_config, i, modulefile, HCBUFSIZ_TINY);
```

OK, all good. We don't need to worry about any previous access to hash_mode. We are save to modify user_options->hash_mode from inside hashcat_session_execute().

---

```
CFLAGS                  += -Wno-typedef-redefinition
```

Is this related to the PR? What did we do wrong to cause it?

---

```
if (!modes_buf) return -1;
```

Please stick to hashcat coding guidelines and use the positive syntax:

```
if (modes_buf == NULL) return -1;
```

---

```
if (modes_cnt != 1) hashconfig_destroy (hashcat_ctx); // keep for later
```

Described as comment, but then not actually called destroy() later. What's the idea here?

---

jsteube:
Almost there, great results. Only the table sorting is missing.

jsteube:
Excellent work everyone!

PR #2806, opened 2021-06-06
Restyling hwmon (2021)

matrix:
Hi,

I reworked the patch I sent two years ago, PR #2098
We have to test it on windows, on linux it seems to work correctly.

Let me know :)


jsteube:
Since hardware monitoring is not the most crucial section in a password cracking and since it was doing its job, it has only grown over time. I think you are right, a refactorization of that section can't hurt. 

Before going into all the detail:

1. Please especially what was the reason for you to rewrite it in the first place. There's lot's of code changed and it's simply too much to find out the logic all by myself

For instance, I can see most of the code from hwmon.c was moved into the API specific source files. I think I have an Idea why you did it, but:

2. Please explain what type of code you have moved and what type of code remained and why

---

Just a log. I have tested this change with:

- NVML Linux: OK
- SYSFS Linux: OK
- NVML Windows: OK
- NVAPI Windows: OK
- ADL Windows: Not tested (no system available)

---

```
-void event_call (const u32 id, hashcat_ctx_t *hashcat_ctx, const void *buf, const size_t len);
+void event_call (const u32 id, void *hashcat_ctx, const void *buf, const size_t len);
```

This comes very unexpected in this PR. Please explain why it was necessary.

jsteube:
I found a way to read utilization for sysfs, please see commit https://github.com/hashcat/hashcat/commit/e8ad343153a47ad2b9318e3a6e16ee5a3180041a
I think it needs to be added to your PR.

matrix:
> Since hardware monitoring is not the most crucial section in a password cracking and since it was doing its job, it has only grown over time. I think you are right, a refactorization of that section can't hurt.
> 
> Before going into all the detail:
> 
> 1. Please especially what was the reason for you to rewrite it in the first place. There's lot's of code changed and it's simply too much to find out the logic all by myself
> 
> For instance, I can see most of the code from hwmon.c was moved into the API specific source files. I think I have an Idea why you did it, but:
> 
> 1. Please explain what type of code you have moved and what type of code remained and why
> 
> Just a log. I have tested this change with:
> 
> * NVML Linux: OK
> * SYSFS Linux: OK
> * NVML Windows: OK
> * NVAPI Windows: OK
> * ADL Windows: Not tested (no system available)
> 
> ```
> -void event_call (const u32 id, hashcat_ctx_t *hashcat_ctx, const void *buf, const size_t len);
> +void event_call (const u32 id, void *hashcat_ctx, const void *buf, const size_t len);
> ```
> 
> This comes very unexpected in this PR. Please explain why it was necessary.

Hi, I simply did cut and paste from hwmon.c, moving the various features into their respective .c
The only relevant changes are done in event.c/h, because of some issue with includes.

Basically inside hwmon I would like to leave only generic functionalities :)

matrix:
git didn't like the change ... if we revert your patch to hwmon and put it back later? otherwise I have to do another PR :)

jsteube:
Please create a new PR

PR #2809, opened 2021-06-07
Enabled "unused" pw-max configurations.

pellekuiters:
Hey, 

While developing our own kernels we found a bug where we forgot to configure a module property while the method for this property was defined. 

To prevent these kind of mistakes in the future we made a small bash script to check all modules for such missing configurations. These are the modules that had unused module configuration methods. 

Kind regards

Pelle & Ewald


jsteube:
Nice find, many thanks for the fixes! 

Does your script check the inconsistencies of all hooks or did you check only for 'pw_max'? I made the same error couple of times for instance module_unstable_warning.

pellekuiters:
The  script checks all inconsistencies and we've attached the script to this issue.  

Run it in the modules folder `../module_check.sh`

[module_check.zip](https://github.com/hashcat/hashcat/files/6617313/module_check.zip)

We created this as part of a small hackathon and it's just a proof-of-concept but hopefully it will be useful while developing.

jsteube:
OK, thanks!

Issue #1277, opened 2017-06-16
DCC2 can't crack a length 16 password


magnumripper:
```
$ echo > test.in '$DCC2$10240#january#26b5495b21f9ad58255d99b5e117abe2'
```
The password is "verylongpassword". Hashcat produce a false negative for this. Is it too long? Then why is it not rejected? And BTW where can I see the max. password length for any given format without dry running the source?

jsteube:
Should be fixed now, thanks! There's no commandline option to show the max password length supported, but a relative good wiki article where the max lengths are listed.

magnumripper:
> There's no commandline option to show the max password length supported, but a relative good wiki article where the max lengths are listed.

Where? Google let me down.

magnumripper:
OK, `git grep 'hashconfig->pw.max'` is good enough for me 😆

jsteube:
Latest version on github can crack passwords up to length 256 and salts (in this case the domain name) up to length 256, also in combination: https://pastebin.com/8HkttUpy

PR #2814, opened 2021-06-09
Allow users to ignore the password length limit.

gorgiaxx:
In the straight attack mode "-a 0", we use weak password dictionaries to crack hashes. 
If a password length not matched to the hash mode, the password will not be add into the Hashcat wordlist

The Unix standard password valid length is 8.

```
$ openssl passwd -crypt -salt C9 oelinux123
Warning: truncating password to 8 characters
C98ULvDZe7zQ2

$ openssl passwd -crypt -salt C9 oelinux1
Warning: truncating password to 8 characters
C98ULvDZe7zQ2
```

Use hashcat to crack, 17 passwords were rejected.
```
./hashcat C98ULvDZe7zQ2 -m 1500 -w 3 -a 0 -O /tmp/test.txt

Dictionary cache hit:
* Filename..: /tmp/test.txt
* Passwords.: 76
* Bytes.....: 641
* Keyspace..: 76

Recovered........: 0/1 (0.00%) Digests
Progress.........: 76/76 (100.00%)
Rejected.........: 17/76 (22.37%)
```

Add ```--len-limit-ignore``` option to the cmdline, all passwords were used in the hashcat context.

```
./hashcat C98ULvDZe7zQ2 -m 1500 -w 3 -a 0 -O /tmp/test.txt --len-limit-ignore --force
C98ULvDZe7zQ2:oelinux1                            
Hash.Name........: descrypt, DES (Unix), Traditional DES
Hash.Target......: C98ULvDZe7zQ2
Progress.........: 76/76 (100.00%)
Candidates.#4....: 000000 -> oelinux1
```

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -629,6 +629,7 @@ typedef enum user_options_defaults
   KEYSPACE                 = false,
   LEFT                     = false,
   LIMIT                    = 0,
+  LEN_LIMIT_IGNORE         = false,
   LOGFILE_DISABLE          = false,
   LOOPBACK                 = false,
   MACHINE_READABLE         = false,
@@ -737,6 +738,7 @@ typedef enum user_options_map
   IDX_KEYSPACE                  = 0xff22,
   IDX_LEFT                      = 0xff23,
   IDX_LIMIT                     = 'l',
+  IDX_LEN_LIMIT_IGNORE          = 0xfe00,
   IDX_LOGFILE_DISABLE           = 0xff24,
   IDX_LOOPBACK                  = 0xff25,
   IDX_MACHINE_READABLE          = 0xff26,
@@ -1986,6 +1988,7 @@ typedef struct user_options
   bool         keep_guessing;
   bool         keyspace;
   bool         left;
+  bool         len_limit_ignore;
   bool         logfile_disable;
   bool         loopback;
   bool         machine_readable;
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -1415,7 +1415,7 @@ static int calc (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param)
             }
 
             if (attack_kern == ATTACK_KERN_STRAIGHT)
-            {        
+            {
               if (((line_len < hashconfig->pw_min) && (user_options->len_limit_ignore == false)) || ((line_len > hashconfig->pw_max) && (user_options->len_limit_ignore == false)))
               {
                 words_extra++;
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -258,7 +258,7 @@ static int calc_stdin (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_par
 
       if (attack_kern == ATTACK_KERN_STRAIGHT)
       {
-        if ((line_len < hashconfig->pw_min) || (line_len > hashconfig->pw_max))
+        if (((line_len < hashconfig->pw_min) && (user_options->len_limit_ignore == false)) || ((line_len > hashconfig->pw_max) && (user_options->len_limit_ignore == false)))
         {
           words_extra_total++;
 
@@ -1415,8 +1415,8 @@ static int calc (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param)
             }
 
             if (attack_kern == ATTACK_KERN_STRAIGHT)
-            {
-              if ((line_len < hashconfig->pw_min) || (line_len > hashconfig->pw_max))
+            {        
+              if (((line_len < hashconfig->pw_min) && (user_options->len_limit_ignore == false)) || ((line_len > hashconfig->pw_max) && (user_options->len_limit_ignore == false)))
               {
                 words_extra++;
 
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -51,6 +51,8 @@
 - RC4 Kernels: Improved performance by 20%+ for hash-modes Kerberos 5 (etype 23), MS Office (<= 2003) and PDF (<= 1.6) by using new RC4 code
 - Status Screen: Show currently running kernel type (pure, optimized) and generator type (host, device)
 - UTF8-to-UTF16: Replaced naive UTF8 to UTF16 conversion with true conversion for RAR3, AES Crypt, MultiBit HD (scrypt) and Umbraco HMAC-SHA1
+- Use --len-limit-ignore option to solve the length missmatch problem in -a 0 mode.
+
 
 ##
 ## Technical
--- a/src/usage.c
+++ b/src/usage.c
@@ -43,6 +43,7 @@ static const char *const USAGE_BIG_PRE_HASHMODES[] =
   "     --keep-guessing            |      | Keep guessing the hash after it has been cracked     |",
   "     --self-test-disable        |      | Disable self-test functionality on startup           |",
   "     --loopback                 |      | Add new plains to induct directory                   |",
+  "     --len-limit-ignore         |      | Ignore the password length limit in hash module      |",
   "     --markov-hcstat2           | File | Specify hcstat2 file to use                          | --markov-hcstat2=my.hcstat2",
   "     --markov-disable           |      | Disables markov-chains, emulates classic brute-force |",
   "     --markov-classic           |      | Enables classic markov-chains, no per-position       |",
--- a/src/user_options.c
+++ b/src/user_options.c
@@ -63,6 +63,7 @@ static const struct option long_options[] =
   {"hex-salt",                  no_argument,       NULL, IDX_HEX_SALT},
   {"hex-wordlist",              no_argument,       NULL, IDX_HEX_WORDLIST},
   {"hook-threads",              required_argument, NULL, IDX_HOOK_THREADS},
+  {"len-limit-ignore",          no_argument,       NULL, IDX_LEN_LIMIT_IGNORE},
   {"increment-max",             required_argument, NULL, IDX_INCREMENT_MAX},
   {"increment-min",             required_argument, NULL, IDX_INCREMENT_MIN},
   {"increment",                 no_argument,       NULL, IDX_INCREMENT},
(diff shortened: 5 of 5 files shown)


src/dispatch.c
```diff
@@ -1415,8 +1415,8 @@ static int calc (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param)
             }
 
             if (attack_kern == ATTACK_KERN_STRAIGHT)
-            {
-              if ((line_len < hashconfig->pw_min) || (line_len > hashconfig->pw_max))
+            {
```
ventaquil: Extra spaces?

gorgiaxx: Done

include/types.h
```diff
@@ -737,6 +738,7 @@ typedef enum user_options_map
   IDX_KEYSPACE                  = 0xff22,
   IDX_LEFT                      = 0xff23,
   IDX_LIMIT                     = 'l',
+  IDX_LEN_LIMIT_IGNORE          = 0xfe00,
```
gorgiaxx, line 741: I don't know how to specify the number. @ventaquil

jsteube:
Thanks for the PR. 

To reproduce, I had created:

```
root@ht:~/hashcat# openssl passwd -crypt -salt C9 oelinux123 > hash
Warning: truncating password to 8 characters
root@ht:~/hashcat# cat hash
C98ULvDZe7zQ2
root@ht:~/hashcat# cat word
oelinux123
root@ht:~/hashcat# ./hashcat -m 1500 hash word
hashcat (v6.2.1-146-g733f5e82e) starting...

...

Session..........: hashcat                                
Status...........: Exhausted
Hash.Name........: descrypt, DES (Unix), Traditional DES
Hash.Target......: C98ULvDZe7zQ2
Time.Started.....: Wed Jun  9 10:41:58 2021 (0 secs)
Time.Estimated...: Wed Jun  9 10:41:58 2021 (0 secs)
Guess.Base.......: File (word)
Speed.#1.........:        0 H/s (0.00ms) @ Accel:512 Loops:1 Thr:64 Vec:1
Recovered........: 0/1 (0.00%) Digests
Progress.........: 1/1 (100.00%)
Rejected.........: 1/1 (100.00%)
Restore.Point....: 1/1 (100.00%)
Restore.Sub.#1...: Salt:0 Amplifier:0-0 Iteration:0-1
Candidates.#1....: [Copying]
Hardware.Mon.#1..: Temp: 66c Fan: 39% Util: 42% Core:1215MHz Mem:3004MHz Bus:16
```

It created correctly the reject:

```
Rejected.........: 1/1 (100.00%)
```

Your code change from this PR wants to override that rejections by allowing any sort of password length which was set in the plugins' module section on module_pw_max().

The implementation to do so in this PR is more complicated than it needs to be. In addition, by not limiting the password length this can have unpredictable influence when it comes to overwrite certain buffers, leading to all sort of program crashes. There are some length limits in place, but since this is a check we do only one time on startup and also on the host is a cheap check so we should stick to it.

Therefore a better implementation would be to modify hashconfig->pw_max after it was set somewhere in the bottom section of hashconfig_init() in src/interface.c. The implementation should also respect if we are going to attack using a pure or an optimized kernel. As a bonus you can save modifying the branches in src/dispatch.c completely. The decisions made in these slightly complicated clauses would become even more complicated. Whenever we do changes in the source we try to do it as seamless as possible to keep the code easy to read. That should be top priority.

But then, there's another alternative and I think we need to discuss this first. You can use the existing features of hashcat to solve this "problem" without any code change. The question is, should we stick to this already existing feature or do we really want to add this new option. 

Here's how to do it:

```
root@ht:~/hashcat# cat hash
C98ULvDZe7zQ2
root@ht:~/hashcat# cat word
oelinux123
root@ht:~/hashcat# ./hashcat -m 1500 hash word -j "'8"
hashcat (v6.2.1-146-g733f5e82e) starting...

C98ULvDZe7zQ2:oelinux1                                    
                                                          
Session..........: hashcat
Status...........: Cracked
Hash.Name........: descrypt, DES (Unix), Traditional DES
Hash.Target......: C98ULvDZe7zQ2
Time.Started.....: Wed Jun  9 11:00:07 2021 (0 secs)
Time.Estimated...: Wed Jun  9 11:00:07 2021 (0 secs)
Guess.Base.......: File (word)
Speed.#1.........:     1969 H/s (0.07ms) @ Accel:512 Loops:1 Thr:64 Vec:1
Progress.........: 1/1 (100.00%)
Rejected.........: 0/1 (0.00%)
Restore.Point....: 0/1 (0.00%)
Restore.Sub.#1...: Salt:0 Amplifier:0-1 Iteration:0-1
Candidates.#1....: oelinux1 -> oelinux1
Hardware.Mon.#1..: Temp: 68c Fan: 38% Util:  5% Core:1126MHz Mem:3004MHz Bus:16
```

So I'm not sure if we want to use this. What's your opinion?

gorgiaxx (commented):
> Thanks for the PR.
> 
> To reproduce, I had created:
> 
> ```
> root@ht:~/hashcat# openssl passwd -crypt -salt C9 oelinux123 > hash
> Warning: truncating password to 8 characters
> root@ht:~/hashcat# cat hash
> C98ULvDZe7zQ2
> root@ht:~/hashcat# cat word
> oelinux123
> root@ht:~/hashcat# ./hashcat -m 1500 hash word
> hashcat (v6.2.1-146-g733f5e82e) starting...
> 
> ...
> 
> Session..........: hashcat                                
> Status...........: Exhausted
> Hash.Name........: descrypt, DES (Unix), Traditional DES
> Hash.Target......: C98ULvDZe7zQ2
> Time.Started.....: Wed Jun  9 10:41:58 2021 (0 secs)
> Time.Estimated...: Wed Jun  9 10:41:58 2021 (0 secs)
> Kernel.Feature...: Pure Kernel
> Guess.Base.......: File (word)
> Guess.Queue......: 1/1 (100.00%)
> Speed.#1.........:        0 H/s (0.00ms) @ Accel:512 Loops:1 Thr:64 Vec:1
> Recovered........: 0/1 (0.00%) Digests
> Progress.........: 1/1 (100.00%)
> Rejected.........: 1/1 (100.00%)
> Restore.Point....: 1/1 (100.00%)
> Restore.Sub.#1...: Salt:0 Amplifier:0-0 Iteration:0-1
> Candidate.Engine.: Device Generator
> Candidates.#1....: [Copying]
> Hardware.Mon.#1..: Temp: 66c Fan: 39% Util: 42% Core:1215MHz Mem:3004MHz Bus:16
> ```
> 
> It created correctly the reject:
> 
> ```
> Rejected.........: 1/1 (100.00%)
> ```
> 
> Your code change from this PR wants to override that rejections by allowing any sort of password length which was set in the plugins' module section on module_pw_max().
> 
> The implementation to do so in this PR is more complicated than it needs to be. In addition, by not limiting the password length this can have unpredictable influence when it comes to overwrite certain buffers, leading to all sort of program crashes. There are some length limits in place, but since this is a check we do only one time on startup and also on the host is a cheap check so we should stick to it.
> 
> Therefore a better implementation would be to modify hashconfig->pw_max after it was set somewhere in the bottom section of hashconfig_init() in src/interface.c. The implementation should also respect if we are going to attack using a pure or an optimized kernel. As a bonus you can save modifying the branches in src/dispatch.c completely. The decisions made in these slightly complicated clauses would become even more complicated. Whenever we do changes in the source we try to do it as seamless as possible to keep the code easy to read. That should be top priority.
> 
> But then, there's another alternative and I think we need to discuss this first. You can use the existing features of hashcat to solve this "problem" without any code change. The question is, should we stick to this already existing feature or do we really want to add this new option.
> 
> Here's how to do it:
> 
> ```
> root@ht:~/hashcat# cat hash
> C98ULvDZe7zQ2
> root@ht:~/hashcat# cat word
> oelinux123
> root@ht:~/hashcat# ./hashcat -m 1500 hash word -j "'8"
> hashcat (v6.2.1-146-g733f5e82e) starting...
> 
> C98ULvDZe7zQ2:oelinux1                                    
>                                                           
> Session..........: hashcat
> Status...........: Cracked
> Hash.Name........: descrypt, DES (Unix), Traditional DES
> Hash.Target......: C98ULvDZe7zQ2
> Time.Started.....: Wed Jun  9 11:00:07 2021 (0 secs)
> Time.Estimated...: Wed Jun  9 11:00:07 2021 (0 secs)
> Kernel.Feature...: Pure Kernel
> Guess.Base.......: File (word)
> Guess.Queue......: 1/1 (100.00%)
> Speed.#1.........:     1969 H/s (0.07ms) @ Accel:512 Loops:1 Thr:64 Vec:1
> Recovered........: 1/1 (100.00%) Digests
> Progress.........: 1/1 (100.00%)
> Rejected.........: 0/1 (0.00%)
> Restore.Point....: 0/1 (0.00%)
> Restore.Sub.#1...: Salt:0 Amplifier:0-1 Iteration:0-1
> Candidate.Engine.: Device Generator
> Candidates.#1....: oelinux1 -> oelinux1
> Hardware.Mon.#1..: Temp: 68c Fan: 38% Util:  5% Core:1126MHz Mem:3004MHz Bus:16
> ```
> 
> So I'm not sure if we want to use this. What's your opinion?

Thanks for your comment. It's much better than this PR. 
Because I didn't know about the rule-based attack before and never find a solution. 
I only found this #1277. So I added a switch to ignore the limit. In fact, I don't want to make the code more complicated.
Your solution is really helpful.

include/types.h
ventaquil: Honestly I don't know. It looks like next numbers starting from `0xFF00` so you can add your option at the end of the list and use latest value + 1 (currently I should be `0xFF4D`).

Am I correct @jsteube?

jsteube: This is most just following convention. You order the name of the macro alphabetically and then you update update all subsequent macros ID's + 1.

gorgiaxx: I understand. Thank you. @ventaquil @jsteube

Issue #2811, opened 2021-06-07
Feature Request: Add "auto detect only" functionality for hashes


sadreck:
Hi,

**Background**
This feature request is coming from this tweet: https://twitter.com/hashcat/status/1401612748115566603

**Current Functionality**
The way this has been implemented is that it's part of the cracking process, which means that it will kick off the session if it detects a single match, or otherwise it will prompt the user to select a type and then continue with the cracking.

**Feature Request**
It would be great if we could have an `--autodetect-only / --identify` feature where we could use it as a "hash-id" and simply identify a hash. It wouldn't even need to read it from a file, it could be as simple as:

```
hashcat --identify <hash>
```

Thanks,
Pavel

Chick3nman:
We currently have a flag, added somewhat recently, call `--hash-info` that prints out a great little set of information about a hash type, specified with `-m`. I'm going to look at implementing this such that if you do not specify `-m` but do specify a hash or hash file, the parser runs and then prints out the same hash information for the detected hash type(s).

Chick3nman:
An example of the `--hash-info` output looks like this:

```
Hash Info:
==========

Hash mode #2500
  Name................: WPA-EAPOL-PBKDF2
  Category............: Network Protocols
  Slow.Hash...........: Yes
  Password.Len.Min....: 8
  Password.Len.Max....: 63
  Salt.Type...........: Embedded
  Salt.Len.Min........: 0
  Salt.Len.Max........: 256
  Hashes.Count.Min....: 1
  Hashes.Count.Max....: 4294967295
  Kernel.Type(s)......: pure 
  Example.Hash.Format.: hex-encoded
[...]

PR #2821, opened 2021-06-10
add identify/autodetect-only option

matrix:
from #2811


include/types.h
```diff
@@ -685,6 +686,7 @@ typedef enum user_options_map
 {
   IDX_ADVICE_DISABLE            = 0xff00,
   IDX_ATTACK_MODE               = 'a',
+  IDX_AUTODETECT_ONLY           = 0xff4d,
```
jsteube: Is a bit annoying, but we try to keep the numbering synchronous with the order of the macro. That means that  from the point where you insert a new macro you need to update all subsequent macros +1.

src/user_options.c
```diff
@@ -63,6 +63,7 @@ static const struct option long_options[] =
   {"hex-salt",                  no_argument,       NULL, IDX_HEX_SALT},
   {"hex-wordlist",              no_argument,       NULL, IDX_HEX_WORDLIST},
   {"hook-threads",              required_argument, NULL, IDX_HOOK_THREADS},
+  {"identify",                  no_argument,       NULL, IDX_AUTODETECT_ONLY},
```
jsteube: please also keep the macro synchrouns with the variable name. so rename the identify to autodetect_only (or vice versa).

jsteube:
I just realized that the entries in `extra/tab_completion/hashcat.sh ` is missing. Please add the new --identify option so that we can tab auto-complete it.

jsteube:
I also do not see a reason to initialize the backend interface and devices in case we use --id. This creates a lot of startup time and requires the user to have either CUDA or OpenCL installed.

jsteube:
I'll merge and then do all the fixes, is too many files to modify and to name them all

jsteube:
Here's the fixes: https://github.com/hashcat/hashcat/commit/7fc0ac4ef17e3aed85bcdf782ed89a2af87f9ad3

- Entry in auto completion script
- Use --machine-readable instead of --quiet
- Skip backend initialization
- Fix session name

PR #2844, opened 2021-06-20
remove all "if (true)" and using one only "salt_repeats" var

matrix:
Hi,
reading the patch is hard, but the mods are very easy:
- removing all "if (true)", and realign the code
- use one only "salt_repeats" var


jsteube:
The if (true) cases are on purpose. Please stick to positive checks whenever positive. Also sometimes if (true) { } is just a way to put a new block. Please ignore this type warnings from cppcheck.

PR #2846, opened 2021-06-21
Fix access violation in hashconfig_destroy if hashcat_ctx_t is only p…

jtojanen:
Fix access violation in hashconfig_destroy if hashcat_ctx_t is only partially initialized; early return from hashconfig_init. Fix hashcat_ctx leak and refactor module and kernel existence checks. Now checks existence of requested module and kernel. In case of module_kern_type_dynamic checks existence of default kernel type.


jsteube:
Many thanks!

Issue #2838, opened 2021-06-19
tests: verify of -m 25400 = PDF edit password is broken


philsmd:
I tested to run our verification script (`tools/test.pl`) with the example hash:
```
$pdf$2*3*128*-3904*1*16*631ed33746e50fba5caf56bcc39e09c6*32*5f9d0e4f0b39835dace0d306c40cd6b700000000000000000000000000000000*32*842103b0a0dc886db9223b94afe2d7cd63389079b61986a4fcf70095ad630c24
```

(the password should be `hashcat`)

I've tested it like this:
```
perl tools/test.pl verify 25400 hash in out && cat out
```

the file `out` should now contain the `cracked  hashes`, but it does not. (the input was the above hash within the `hash` file and the hash and password - separated by : - within the `in` file) 

I've investigated this a little bit and it seems that the `user` vs `owner` password for this hash are different (but I don't think that is the reason or "fix" for this problem).

```
user password ($u variable in `tools/test_modules/m25400.pm`) different:
37747ccd83e3fceccdc1c06de41a0904 for password hashcat
5f9d0e4f0b39835dace0d306c40cd6b7 for empty password (zero length)
```

I think the main problem here is that whenever we want to verify one type of hash (`user` vs `owner`) , the other one needs to be kind of ignored/skipped and therefore send to the output/print AS-IS... i.e. whenever the `owner` password is correct, we shouldn't generate a "different"/"new" `user` hash with the same password... we should just marked the password as correct by passing the other part AS-IS (even if different password)...

Later on (different topic), we could think about changing the default example hash to a hash that uses `hashcat`  as the password for both `user` and `owner` part, but that's a different story !

cc:  @thatux

Thanks a lot guys

---

update: I also noted a strange/useless variable within the test script,  that we should remove:
https://github.com/hashcat/hashcat/blob/05125eb67c9da98e349dabea50f81f24080f6347/tools/test_modules/m25400.pm#L143

thatux:
Hi @philsmd,


Nice find, I'm not a 100% sure whether I can follow you compeletly but I think I agree.

Have you seen these comments: https://github.com/hashcat/hashcat/pull/2739#issuecomment-816998851
#2739 
In short the best idea as @jsteube suggested is to add the password required as a kind of salt, do you agree @philsmd?

Am I correct that these are the files you use?


```
cat hash
[...]

PR #2857, opened 2021-06-29
added note in test.pl potthrough: it only works with unsalted hashes

thatux:


jsteube:
I think there's a misunderstanding. What I meant with this only works with unsalted hashes is this:

```
echo $IN | perl tools/test.pl passthrough 10500 >> hash 
echo $IN | perl tools/test.pl potthrough 10500 >> in 
```

Of course the unit tests themselves support salted hashes and that's because it's not working.
Every time you "generate" new hashes it randomly chooses a new salt. This salt will create a different outcome than the salt produced in the previous "hash" output. So they don't match, and that's why verify does not create an output line.

I was expecting that you want to add a documentation note about that the user needs to keep in mind that with every call to passthrough and/or pottrough a new random salt will be generated.

thatux:
Ah, thanks for the explanation, that makes sense, sorry for my slow understanding..

Keeping the same example of 10500, I find in https://github.com/hashcat/hashcat/blob/master/tools/test_modules/m10500.pm#L151 that it unpacks the hash it's trying to compare, and uses the same salt values as the original hash to hash the input password, otherwise there the hash would of course be different (and never verify).

For future reference I encountered this in https://github.com/hashcat/hashcat/issues/2838

I've updated my comments in test.pl to be correct.

jsteube:
Great Update!

PR #2858, opened 2021-06-30
Fix several backend memory leaks

jtojanen:


jsteube:
This is too much code change at once without any description. Please explain the changes so that I can follow. Also we don't use static declarations because effects are sometimes confusing on different OS. We also do not use goto, please refactor that code as well.

jtojanen:
I can close this pull and split these to smaller ones with better description.

But I don't understand problem with "static". As the scope/need of those functions is local, best practice is to make them "static". This improves code stability, produces more efficient and smaller code. These are not the only "static" functions in this project. What would be the side effects?

Also I don't understand the "goto" hate as the use of "goto" would be highly beneficial here. There are numerous places in hashcat where initialized is interrupted after error with "return -1", and resources are not freed. All these returns mean that there is a lot of unnecessary copy-paste coding. With "goto", we can simplify the code by putting the intermediate resource cleanup code at the end of the initialization function and return with -1. As a result if we return 0 from any of _init functions, resources are known to be ready for use and afterwards we can safely call _destroy.

jsteube:
OK, please close this PR and create several smaller ones. But most importantly, please explain them. Sometimes changes (like this one) have scattered changes and they are hard to read. It would help to explain why did you come up with need for the change in the first place and then how you "solved" the problem.

For the static part, I think it's not absolutely necessary and I can see the benefits from it. The problem are actually more on the Kernel code, where static declarations create problems. In host code I just try to avoid it because of the reason I avoid it in the kernels, but that's it. OTOH what's the price for not having it static. A little performance increase in a place where it doesn't count anyways because it's executes only on startup.

There's no "goto hate", it more like a convention to which we agreed from the very beginning. Since then we spend quite some energy to avoid it, exactly as you described with the price of copy/paste code and accepting a few memory leaks (but only in error scenarios). I can understand the frustration but we are not going to change this just to handle error scenarios better.

PR #2860, opened 2021-06-30
try fix #2859

### LINKED ISSUE #2859: --hash-info does produce an error on certain modes
Used hashcat version is from the actual master.

When requesting mode information with --hash-info, it produces an error and quits for modes where only optimized kernels exist.

Example:
```
$ ./hashcat --hash-info -m 200
hashcat (v6.2.2-74-g56c2243df) starting in hash-info mode

.../hashcat/OpenCL/m00200_a0-pure.cl: No such file or directory
```

matrix:


jsteube:
I think the right fix is discussed here: https://github.com/hashcat/hashcat/commit/6967e706c2720f699ecd74682dac243ba7ea143a#commitcomment-52881004

I'll push a change later or tomorrow.

PR #2861, opened 2021-07-03
Fix iconv_ctx and iconv_tmp leaks in backend.c

jtojanen:
In run_cracker if iconv is needed for encoding, iconv_ctx and iconv_tmp both leak. Ignoring iconv_ctx leak after failure.


jsteube:
Nice find, many thanks

PR #2865, opened 2021-07-05
Fix several backend leaks

jtojanen:
There are numerous reasons why device_param->skipped can be true, but it is not indicating that resources should not be freed in backend_ctx_devices_destroy. Fixed obvious early return leaks in backend_ctx_init, backend_ctx_devices_init, backend_session_begin and load_kernel. Ensure build log is always null-terminated.

Locally defined IGNORE_DEVICE_NOT_FOUND and CHECK_BOARD_NAME_AMD are bit strange. Should we move these to Makefile and refactor code specially around IGNORE_DEVICE_NOT_FOUND? clGetDeviceIDs is called twice only to get the actual error code.


jsteube:
Good stuff. Many thanks.

PR #2867, opened 2021-07-06
Improve affinity mask handling

jtojanen:
Allow CPU affinity mask up to 64 processors in Windows. Remove call to SetThreadAffinityMask as SetProcessAffinityMask limits all threads in process. Report error code. pthread_setaffinity_np returns non-zero if failure; works also with OSX as it reports KERN_SUCCESS which is zero.

We could do even better by querying the current mask using pthread_getaffinity_np()/GetProcessAffinityMask(), filter the new mask and before setting new mask, warn the user if requested mask contained missing CPUs.

I tested with Windows, but not on other systems.

### DIFF
--- a/src/affinity.c
+++ b/src/affinity.c
@@ -56,8 +56,10 @@ int set_cpu_affinity (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
 
   #if defined (_WIN)
   DWORD_PTR aff_mask = 0;
+  const int cpu_id_max = 8 * sizeof (aff_mask);
   #else
   cpu_set_t cpuset;
+  const int cpu_id_max = 8 * sizeof (cpuset);
   CPU_ZERO (&cpuset);
   #endif
 
@@ -84,7 +86,7 @@ int set_cpu_affinity (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
       break;
     }
 
-    if (cpu_id > 32)
+    if (cpu_id > cpu_id_max)
     {
       event_log_error (hashcat_ctx, "Invalid cpu_id %d specified.", cpu_id);
 
@@ -94,7 +96,7 @@ int set_cpu_affinity (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
     }
 
     #if defined (_WIN)
-    aff_mask |= 1u << (cpu_id - 1);
+    aff_mask |= ((DWORD_PTR) 1) << (cpu_id - 1);
     #else
     CPU_SET ((cpu_id - 1), &cpuset);
     #endif
@@ -105,11 +107,9 @@ int set_cpu_affinity (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
 
   #if defined (_WIN)
 
-  SetProcessAffinityMask (GetCurrentProcess (), aff_mask);
-
-  if (SetThreadAffinityMask (GetCurrentThread (), aff_mask) == 0)
+  if (SetProcessAffinityMask (GetCurrentProcess (), aff_mask) == 0)
   {
-    event_log_error (hashcat_ctx, "%s", "SetThreadAffinityMask().");
+    event_log_error (hashcat_ctx, "SetProcessAffinityMask() failed with error: %d", (int) GetLastError ());
 
     return -1;
   }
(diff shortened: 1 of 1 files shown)


src/affinity.c
```diff
@@ -94,7 +96,7 @@ int set_cpu_affinity (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx)
     }
 
     #if defined (_WIN)
-    aff_mask |= 1u << (cpu_id - 1);
+    aff_mask |= ((DWORD_PTR) 1) << (cpu_id - 1);
```
jsteube, line 99: nice catch

jtojanen:
It's not perfect but good enough. At least user gets error message, but someone with ThreadRipper may encounter different issues..

jsteube:
Currently I am waiting for a multi GPU system from a beta tester to test it with multiple GPU. A nice way to produce some load for testing is to use --backend-ignore-cuda --spin-damp 0. This will remove the workaround on OpenCL to generate 100% CPU load per GPU. I'd like to see how it behaves in this case if there are multiple GPU. I should get access to such system in a matter of days.

jtojanen:
> Currently I am waiting for a multi GPU system from a beta tester to test it with multiple GPU. A nice way to produce some load for testing is to use --backend-ignore-cuda --spin-damp 0. This will remove the workaround on OpenCL to generate 100% CPU load per GPU. I'd like to see how it behaves in this case if there are multiple GPU. I should get access to such system in a matter of days.

My test setup has one AMD and one Nvidia card to quickly test CUDA and OpenCL. I have been trying to find a reason for performance drop and I have been looking at the spin damp section in backend. Yesterday I tested effect of disabling CL_QUEUE_PROFILING_ENABLE, As it disables all profiling, hc_clGetEventProfilingInfo cannot be used. I used host timer with, hc_timer_get/hc_timer_set to measure execution time. Oddly enough disabling profiling didn't improve performance.

jsteube:
The OpenCL driver strategy to synchronize is very annoying. So basically the host has to wait for an unknown time for the kernel to finish. To do so, the OpenCL driver runs a CPU burning loop are creates 100% on this particular thread. If you have multiple GPU, you have multiple threads and typically modern CPU have multiple cores and the OS distributes the threads to different cores. 

But if, due to the affinity, there's only one core being used, the both (or even more if more than 2 GPU) burning loop run on the same core and one effect of this could be to see a slight decrease in cracking performance. 

The workaround simple measures the time it took for the previous kernel to finish, removes a bit of time and sleeps for that time, before it enters the CPU burning loop again. This reduces the CPU load not completely to zero but to a very low percentage. So if we disable this workaround by setting --spin-damp 0, the effect should be that performance decreases slightly, but still have only one core being loaded 100% no matter how many GPU are in use.

jtojanen:
I have no idea how drivers implement callbacks, but think one option (not tested?) might be to use clSetEventCallback(CL_COMPLETE) and wait with standard mutex in backend.

jsteube:
Is a good idea, but is very likely the callback is just abstracted. So internally it waits for the burning loop to complete, then calls the callback function. Feel free to find out :)

I'm probably going to merge this PR anyway. I just want to test it on a multi GPU system first.

roycewilliams:
With this cmdline:

`hashcat -O -w 4 -a 3 -m 0 --backend-ignore-cuda --spin-damp 0 0d599f0ec05c3bda8c3b8a68c32a1b47 ?b?b?b?b?b?b?b`

... it still appears to consume 100% of a core per GPU.

jtojanen:
> With this cmdline:
> 
> `hashcat -O -w 4 -a 3 -m 0 --backend-ignore-cuda --spin-damp 0 0d599f0ec05c3bda8c3b8a68c32a1b47 ?b?b?b?b?b?b?b`
> 
> ... it still appears to consume 100% of a core per GPU.

Okey. I am able to reproduce this with my setup.
Windows running AMD Adrenalin 21.4.1 and NVIDIA 465.89 drivers. CUDA v11.3. 
Hashcat shows versions "Nvidia OpenCL 3.0 CUDA 11.3.55" and "OpenCL 2.1 AMD-APP (3240.6)".

It seems to be issue with at least Nvidia OpenCL implementation, not so with AMD.

If I use only one Nvidia card with OpenCL API, I get 100% on one core with:
```hashcat -O -w 4 -a 3 -d 1 -m 0 --backend-ignore-cuda --spin-damp 0 0d599f0ec05c3bda8c3b8a68c32a1b47 ?b?b?b?b?b?b?b```

CPU usage does not increase if I enable both cards, one AMD and one Nvidia and both use OpenCL API:
```hashcat -O -w 4 -a 3 -m 0 --backend-ignore-cuda --spin-damp 0 0d599f0ec05c3bda8c3b8a68c32a1b47 ?b?b?b?b?b?b?b```

But if I enable CUDA, CPU usage drops to zero with:
```hashcat -O -w 4 -a 3 -m 0 --spin-damp 0 0d599f0ec05c3bda8c3b8a68c32a1b47 ?b?b?b?b?b?b?b```

Only marginal difference in hash rates between Nvidia CUDA and Nvidia OpenCL. Don't use OpenCL with Nvidia?

jsteube:
The problem is known for NV OpenCL runtime, which is why this workaround was added in the first place. We should continue to keep an eye on this behavior in case they fix it at some point. I had reported this a few years ago to official NV bug tracker, but it was never fixed (and now the bug tracker is gone, too). For NV, it's best to stick to CUDA.

PR #2872, opened 2021-07-10
Use critical section instead of mutex in Windows

jtojanen:
Before 6092308324e3c1fd1d3f59f0d14a63442862aa55 critical section was used, but it was replaced with mutex. I cannot find any reasoning why the change was made.

It is well known that critical section gives better performance. As most of serialization spots are short, and some should be made even shorter, in most cases critical section don't need context switching and access to kernel space. Critical section is more compatible with pthread_mutex, and we can test if we need to spin/yield/sleep using TryEnterCriticalSection/pthread_mutex_trylock. If EnterCriticalSection takes long time, it will behave exactly like mutex right now.

As we have only single process, only benefit to use mutex would be to create named mutex to check if another instance of hashcat is already running in [pidfile.c](https://github.com/hashcat/hashcat/blob/master/src/pidfile.c), check_running_process().

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -22,17 +22,17 @@
 #define hc_thread_exit(t)           ExitThread (t)
 #define hc_thread_detach(t)         CloseHandle (t)
 
-/*
+#define hc_thread_mutex_init(m)     InitializeCriticalSection (&m)
 #define hc_thread_mutex_lock(m)     EnterCriticalSection      (&m)
 #define hc_thread_mutex_unlock(m)   LeaveCriticalSection      (&m)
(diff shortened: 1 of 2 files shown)


jsteube:
The change was made to align the technologies between Linux and Windows to make it easier to maintain them. From performance perspective, I can't see any advantages (but also no disadvantages) either using mutex or CriticalSection. However CriticalSection seems to fit much better for what we do, for simple thread handling. It would be nice if you have an example that shows the performance difference between the both versions. Since CriticalSection is more lightweight I have no objections switching it.

PR #2873, opened 2021-07-10
First draft HIP Version

reger-men:
Add HIP backend support. Things TODO:

- Grep the correct architecture for the hiprtc complier.
- Fix PDF segfault
- Optimize kernel for HIP
- Try dynamic shared memory
- And more ....


jsteube:
This is a fantastic contribution, many thanks!

Note: I had to drop all changes to src/backend.c for initial merge because the difference between v6.1.1 and current master are too numerous. I will backport the changes in a later step.

PR #2877, opened 2021-07-12
improve 25400: recover both owner and user; add user password to hash; better test

### LINKED ISSUE #2838: tests: verify of -m 25400 = PDF edit password is broken

thatux:
As we don't know the user password we cannot generate a new owner hash (as it's based on the user password).
Should fix https://github.com/hashcat/hashcat/issues/2838


jsteube:
@philsmd please review

tools/test_modules/m25400.pm
```diff
@@ -228,14 +237,11 @@ sub module_generate_hash
...
   #printf("\$u = %s\n", unpack ("H*", $u));
-
-  $u .= substr (pack ("H*", $u_save), 16, 16);
-
   #printf("\$o = %s\n", unpack ("H*", $o));
   #printf("\$u = %s\n", unpack ("H*", $u));
 
   my $hash = sprintf ('$pdf$%d*%d*128*%d*%d*16*%s*32*%s*32*%s', $V, $R, $P, $enc, $id, unpack ("H*", $u), unpack ("H*", $o));
-
+  #print("hash\n".$hash."\n");
```
philsmd: don't only comment it out, but please completely remove it

tools/test_modules/m25400.pm
```diff
@@ -171,32 +171,41 @@ sub module_generate_hash
 
 
   ################ USER PASSWORD #################
-  my $res = pdf_compute_encryption_key_user($word, $padding, $id, $u, $o, $P, $V, $R, $enc);
-
-  my $digest = md5 ($padding . pack ("H*", $id));
+  # do not change $u if it exists, keep this the same, as we don't know the user password,
+  #  we cannot calculate this part of the hash again
+  if("".$u eq "0000000000000000000000000000000000000000000000000000000000000000")
```
philsmd: why do you use this notation ?
I think
```
if ($u eq "0000000000000000000000000000000000000000000000000000000000000000")
```
would be much better

(also please use our code style with spaces after the "if". thx

philsmd (commented):
I think the strategy sounds correct.  i.e. use `$u` AS-IS if set, otherwise try to calculate it if "0000000000000000000000000000000000000000000000000000000000000000".

did I understand the strategy here ?
It's important that every password that hashcat accepts/cracks, can also be "verified" with test.pl. I think that is the case now and I hope you also tested it a little bit.

Yeah, I would generally give it a thumb up, just a few code style changes needed here. thx

thatux:
Thanks for the review, will do!

I'm actually rewriting the attack mode a bit such that it can use the user-password as input to verify whether the owner-password has been cracked correctly. Without a user-password we can check whether we decrypted the o-value to the correct padding.

Functionally it's almost working, I need to implement the check, and clean it up a bit more, this will be a bigger code change, so it will take a bit more time.

tools/test_modules/m25400.pm
thatux: I agree, fixed both

tools/test_modules/m25400.pm
thatux: Agree, removed

src/modules/module_25400.c
```diff
@@ -483,15 +478,81 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
...
+    u8OutBufPtr[i]=0x0;
+    remove_padding=true;
+    }
+  }
+
+  // if the password in tmp->out is equal to the password tried, then we recovered just the owner-password or just the user-password
+  //   we check whether we already have a user-password in the hash
+  //   TODO would be better to actually also verify the u-value whether we've retrieved the correct user-password,
+  //     however, we'd need to include a lot of code/complexity here to do so (or call into 10500 kernel).
+  //     this seems relevant: run_kernel (hashcat_ctx, device_param, KERN_RUN_3, 0, 1, false, 0)
```
thatux, line 529: is it possible to call into a kernel from a kernel? 

any other kernel already doing that?

src/modules/module_25400.c
```diff
@@ -483,15 +478,81 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
...
+      // we seem to only have recovered the user-password as we don't have one yet
+      return snprintf ((char *) dst_buf, dst_sz, "(user password=%s)", (char *) src_buf);
+    }
+  }
+  // we recovered both the user-password and the owner-password
+  return snprintf ((char *) dst_buf, dst_sz, "%s    (user password=%s)", (char *) src_buf, (char *) pdf_tmp->out);
+}
+
+
+// TODO how to add the recovered user-password to the hash?
```
thatux, line 543: Anybody a suggestion on how to add a recovered password back to the hash?

thatux:
I've finished the attack, if I'm not mistaken, this attack is a pretty cool addition to 10500.
It seems to me that attacking the o-value of a pdf is better target than attacking the u-value (as is done in 10500), by attacking the o-value you can both recover the user-password (if only a user-password is set), and recover both the owner-password AND user-password if both are set.

The only difference of this attack is that it recovers the password based on the o-value instead of the u-value.
We don't need the user-password as input for this attack (as we can just decrypt the o-value, instead of needing to hash a password).

I've added an extra (optional) user-password column to the hash as suggested, a cracked hash would look like:
```
./hashcat.exe -a3 -m25400 tools/pdf_tests/user-owner_userpw-in-hash.hash 'owner' --potfile-disable 
$pdf$4*4*128*-3392*1*16*f72f279e0db2ee4ca76b0db6884e85cd*32*665d9c5a3c2443ac2aa3a9b855843bf200000000000000000000000000000000*32*0ba3835f88f90388e74e54584125ce142be0de24c6b0
d37746e075b891756671*user:owner    (user password=user)

./hashcat.exe -a3 -m25400 tools/pdf_tests/owner.hash 'owner' --potfile-disable
$pdf$4*4*128*-3392*1*16*b438e5ce7c355548a28bf1408a36bbdf*32*63a1a7a95566adcf61a8c36002b6bfa900000000000000000000000000000000*32*566fa873ee33c797cd3b904fdadf814afa34df9a38f6
ed41b984e2c6da2aa6f5*:owner    (user password not set)

./hashcat.exe -a3 -m25400 tools/pdf_tests/user.hash 'user' --potfile-disable
$pdf$4*4*128*-1028*1*16*e77b55355f74d54895c93063ddbad299*32*61536f41f6bf35bce77a96eb45cebec500000000000000000000000000000000*32*ce69e44fd4717582076b810914829e577152c3110ee5
81d65e53c64a4e8f578d*:(user password=user)
```



**Adding user-password to hash difficulties**
However, this attack currently doesn't write the cracked user-password back to the hash.
It doesn't have to, but it think it'd look better, and make more sense to instead of 
```
$pdf$4*4*128*-1028*1*16*e77b55355f74d54895c93063ddbad299*32*61536f41f6bf35bce77a96eb45cebec500000000000000000000000000000000*32*ce69e44fd4717582076b810914829e577152c3110ee5
81d65e53c64a4e8f578d*:(user password=user)
```
have the output look like
```
$pdf$4*4*128*-1028*1*16*e77b55355f74d54895c93063ddbad299*32*61536f41f6bf35bce77a96eb45cebec500000000000000000000000000000000*32*ce69e44fd4717582076b810914829e577152c3110ee5
81d65e53c64a4e8f578d*user:
```
However, I've taken a quite extensive look tot he combination of module_hash_encode() and module_build_plain_postprocess() to do so. But wasn't able to work this out for now.

@jsteube would you have a suggestion on how to add a recovered password back to the hash?
https://github.com/hashcat/hashcat/pull/2877/commits/db2e7d1391f816307b388de56b59f20fd0fd3ede#diff-5fdcc91317d89c85a9dcf116e04a03fbc12af47c4c3af89212f075d2612d5b48R543


Finally is it possible to call into a kernel from a kernel? any other kernel already doing that?
https://github.com/hashcat/hashcat/pull/2877/commits/db2e7d1391f816307b388de56b59f20fd0fd3ede#diff-5fdcc91317d89c85a9dcf116e04a03fbc12af47c4c3af89212f075d2612d5b48R529

jsteube:
> Finally is it possible to call into a kernel from a kernel? any other kernel already doing that?
db2e7d1#diff-5fdcc91317d89c85a9dcf116e04a03fbc12af47c4c3af89212f075d2612d5b48R529

You can't run call a kernel from outside the backend interface, but hashcat can call kernels from modules as if they were regular C functions using the emulator mode. Please have a look here: https://github.com/hashcat/hashcat/blob/v6.2.3/src/modules/module_22000.c#L567-L606

src/modules/module_25400.c
jsteube: Please see comment below

PR #2879, opened 2021-07-14
Implement gpu_bzero

jtojanen:
As it seems that all memset's are for zeroing, I will implement kernel `KERNEL_FQ void gpu_bzero (GLOBAL_AS uint4 *buf, const u64 gid_max)` and make necessary modifications to use it.


jsteube:
Please wait until the HIP branch is getting merged in the next days, because the code changes done already also need to be applied to HIP.

jtojanen:
> Please wait until the HIP branch is getting merged in the next days, because the code changes done already also need to be applied to HIP.

Okey

jsteube:
OK, is merged

jtojanen:
Tested CUDA and OpenCL implementations. 

I was unable to test HIP. I renamed "fixme.dll" with "amdhip64.dll", and disabled HIP RTC version number testing. Correct implementation would use hipDriverGetVersion() and hipRuntimeGetVersion(). I stopped trying after an error message:

hiprtcCompileProgram(): HIPRTC_ERROR_COMPILATION
Invalid option: --gpu-max-threads-per-block=64

About OpenCL in general. OpenCL clFlush() just before clFinish() is not needed as clFinish() blocks until previously queued commands are issued and have completed. OpenCL specs is clear on this: "Call the clFlush (or clFinish) API to issue any outstanding commands from this command-queue."

As we have in-order command-queue, we can safely issue multiple commands and wait on last for synchronization. This is showcased to small extend in new implementation of OpenCL memset and bzero. There are many places where we can easily (and safely) minimize overhead because of in-order queue.

jsteube:
> I was unable to test HIP. I renamed "fixme.dll" with "amdhip64.dll", and disabled HIP RTC version number testing. Correct implementation would use hipDriverGetVersion() and hipRuntimeGetVersion(). I stopped trying after an error message:
>
> hiprtcCompileProgram(): HIPRTC_ERROR_COMPILATION
> Invalid option: --gpu-max-threads-per-block=64
> 

OK, that's good to know, because I didn't expect  HIP to work on Windows. I think it's worth digging more into this as it offers a chance to avoid the regular OpenCL JiT from the legacy AMD drivers.

> About OpenCL in general. OpenCL clFlush() just before clFinish() is not needed as clFinish() blocks until previously queued commands are issued and have completed. OpenCL specs is clear on this: "Call the clFlush (or clFinish) API to issue any outstanding commands from this command-queue."

Correct. In the very first versions of hashcat I also believed if we just follow the OpenCL specs everything will work. But there were many different OpenCL runtime which simply do not follow the OpenCL specs strictly enough. So the constructs with clFlush() and clFinish() is a result of a work around of OpenCL runtimes of the past. Of course no one could know because I did not comment it in the source.

> As we have in-order command-queue, we can safely issue multiple commands and wait on last for synchronization. This is showcased to small extend in new implementation of OpenCL memset and bzero. There are many places where we can easily (and safely) minimize overhead because of in-order queue.

The question on this, as well as the one with clFlush() before clFinish(), is what to we gain from the proposed change. If you have any use-case where we definitely some improvement, then I'm happy to merge. Maybe you have something in mind I can reproduce locally?

---

I have seen some inconsistencies in the implementation. For example you had added cuda_function_bzero, but there's no kernel_local_mem_size_memset. Yes it's not required to, but it's was not required for cuda_function_memset either. I added it because all kernel function have it, so cuda_function_bzero should be no different. It's just convention that we follow. Please add this missing part (and the other missing attributes, too).

src/backend.c
```diff
@@ -5211,52 +5253,77 @@ int run_opencl_kernel_memset (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *dev
 {
   const u64 num16d = size / 16;
   const u64 num16m = size % 16;
+  u32 tmp[4];
```
jsteube, line 5418: I think the level of this array is different than in the other functions for CUDA and HIP.

jtojanen: Reason is that hc_clEnqueueNDRangeKernel and hc_clEnqueueWriteBuffer both are now non-blocking and data needs to available until hc_clEnqueueWriteBuffer finishes at hc_clFinish.

jsteube: I see but then there's still an inconsistency with CUDA and HIP. Should it not be the same?

jsteube:
Update: I had updated my Windows Box and installed latest AMD driver: non-whql-radeon-software-adrenalin-2020-21.7.1-win10-64bit-july15.exe

After doing the exact changes @jtojanen was describing I also end up in  a non working situation, but a different one. The error message is:

```
hiprtcCompileProgram(): HIPRTC_ERROR_COMPILATION

error: invalid value '4.0.21025' in '--hip-version=4.0.21025'
error: invalid value '4.0.21025' in '--hip-version=4.0.21025'
error: cannot find HIP runtime. Provide its path via --rocm-path, or pass -nogpuinc to build without HIP runtime.
...
./OpenCL\inc_vendor.h:20:10: fatal error: 'hip/hip_runtime.h' file not found
#include <hip/hip_runtime.h>
         ^~~~~~~~~~~~~~~~~~~
4 errors generated when compiling for gfx900.
Error: Failed to compile opencl source (from CL or HIP source to LLVM IR).
```

I think there's some substantial elements missing in the Windows driver. Therefore we should leave the fixme.dll as it is right now, otherwise it will create confusion with the AMD GPU users in Windows. If we leave it as  it is, it will automatically select OpenCL.

src/backend.c
jtojanen: > I see but then there's still an inconsistency with CUDA and HIP. Should it not be the same?

At the moment CUDA (and HIP as it's copy of CUDA) implementation is using legacy mode. That is in-order fully synchronized where each OP is synchronous.  We need to replace cuMemcpy functions with cuMemcpyAsync versions, which take stream as parameter. Currently memory copying is synchronized with default stream, 0. 

OpenCL had everything ready so I started with it. In-order FIFO execution guarantees correctness and we need to synchronize only when we need results from previous calculations.

I use OpenCL style:
Currently: Enqueue->Flush->Finish->Enqueue->Flush->Finish->memcpy (blocking)->Enqueue->Flush->Finish
Future: Enqueue->Enqueue->memcpy (non-blocking)->Enqueue->(Flush if needed to workaround)->Finish

So host execution is overlapping with device execution with asynchronous OPs as much as possible. Only synchronizing when we need results.

jsteube: OK,  lets start with OpenCL changes and see if there's any reports from users. After a while we should backport the changes to CUDA and HIP, too. What you think?

jsteube:
Just for clarification, for merge, I am waiting for the updates regarding missing kernel_local_mem_size_memset and others.

src/backend.c
jtojanen: Already working on CUDA and HIP

jtojanen:
> Just for clarification, for merge, I am waiting for the updates regarding missing kernel_local_mem_size_memset and others.

I will make those additions in couple of hours when I get back.

jtojanen:
Now CUDA, HIP and OpenCL implementations are at the same level. Of course as we are saving only one single synchronization the performance gain is about zero, but by enlarging the asynchronous execution sections we will minimize synchronization overhead. Gain is bigger with shorter kernels or low latency workloads (-w 3 or less) and specially when execution is not GPU bound.

Other projects have reported around 10% speedups.

jsteube:
Excellent contribution, many thanks!

I also think GPU with low PCI bandwidth will benefit from it, those who are connected through riser cables and thus multiplier is reduced from x16 to x1.

jtojanen:
> Excellent contribution, many thanks!
> 
> I also think GPU with low PCI bandwidth will benefit from it, those who are connected through riser cables and thus multiplier is reduced from x16 to x1.

I totally forgot risers as I am not using those myself. I will have to buy one to play with.

PR #2882, opened 2021-07-14
Added hash-mode: VMware VMX (PBKDF2-HMAC-SHA1 + AES-256-CBC)

matrix:
Hi, with this commit we add support for cracking VMware encrypted VMs (vmx files), via Known-Plaintext Attack :)

### DIFF
--- a/OpenCL/m26700-pure.cl
+++ b/OpenCL/m26700-pure.cl
@@ -28,13 +28,13 @@ typedef struct pbkdf2_sha1_tmp
 
 } pbkdf2_sha1_tmp_t;
 
-typedef struct pbkdf2_sha1
+typedef struct vmware_vmx
 {
   u32 salt_buf[64];
   u32 iv_buf[4];
   u32 ct_buf[4];
 
-} pbkdf2_sha1_t;
+} vmware_vmx_t;
 
 DECLSPEC void hmac_sha1_run_V (u32x *w0, u32x *w1, u32x *w2, u32x *w3, u32x *ipad, u32x *opad, u32x *digest)
 {
@@ -72,7 +72,7 @@ DECLSPEC void hmac_sha1_run_V (u32x *w0, u32x *w1, u32x *w2, u32x *w3, u32x *ipa
   sha1_transform_vector (w0, w1, w2, w3, digest);
 }
 
-KERNEL_FQ void m26700_init (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1_t))
+KERNEL_FQ void m26700_init (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, vmware_vmx_t))
 {
   /**
    * base
@@ -144,7 +144,7 @@ KERNEL_FQ void m26700_init (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1
   }
 }
 
-KERNEL_FQ void m26700_loop (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1_t))
+KERNEL_FQ void m26700_loop (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, vmware_vmx_t))
 {
   const u64 gid = get_global_id (0);
 
--- a/OpenCL/m26700-pure.cl
+++ b/OpenCL/m26700-pure.cl
@@ -1,0 +1,351 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha1.cl"
+#include "inc_cipher_aes.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
+typedef struct pbkdf2_sha1_tmp
+{
+  u32  ipad[5];
+  u32  opad[5];
+
+  u32  dgst[32];
+  u32  out[32];
+
+} pbkdf2_sha1_tmp_t;
+
+typedef struct pbkdf2_sha1
+{
+  u32 salt_buf[64];
+  u32 iv_buf[4];
+  u32 ct_buf[4];
+
+} pbkdf2_sha1_t;
+
+DECLSPEC void hmac_sha1_run_V (u32x *w0, u32x *w1, u32x *w2, u32x *w3, u32x *ipad, u32x *opad, u32x *digest)
+{
+  digest[0] = ipad[0];
+  digest[1] = ipad[1];
+  digest[2] = ipad[2];
+  digest[3] = ipad[3];
+  digest[4] = ipad[4];
+
+  sha1_transform_vector (w0, w1, w2, w3, digest);
+
+  w0[0] = digest[0];
+  w0[1] = digest[1];
+  w0[2] = digest[2];
+  w0[3] = digest[3];
+  w1[0] = digest[4];
+  w1[1] = 0x80000000;
+  w1[2] = 0;
+  w1[3] = 0;
+  w2[0] = 0;
+  w2[1] = 0;
+  w2[2] = 0;
+  w2[3] = 0;
+  w3[0] = 0;
+  w3[1] = 0;
+  w3[2] = 0;
+  w3[3] = (64 + 20) * 8;
+
+  digest[0] = opad[0];
+  digest[1] = opad[1];
+  digest[2] = opad[2];
+  digest[3] = opad[3];
+  digest[4] = opad[4];
+
+  sha1_transform_vector (w0, w1, w2, w3, digest);
+}
+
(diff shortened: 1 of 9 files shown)


src/modules/module_26700.c
```diff
@@ -0,0 +1,270 @@
...
+#include "convert.h"
+#include "shared.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_4_4;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_RAW_CIPHER_KPA;
```
jsteube: Is not a RAW Cipher, I think FDE fits best

docs/readme.txt
```diff
@@ -389,6 +389,7 @@ NVIDIA GPUs require "NVIDIA Driver" (440.64 or later) and "CUDA Toolkit" (9.0 or
 - MetaMask Wallet
 - MultiBit Classic .key (MD5)
 - MultiBit HD (scrypt)
+- VMware VMX (PBKDF2-HMAC-SHA1 + AES-256-CBC)
```
jsteube: I think is the wrong position. Should be 1:1 as were it is if the user uses --help. Note that the category has an influence in it and needs also fix.

tools/test_modules/m26700.pm
```diff
@@ -0,0 +1,104 @@
...
+
+  my $hash = "";
+
+  if (defined $ct_str)
+  {
+    my $ct_bin = pack ("H*", $ct_str);
+    my $iv_bin = substr ($ct_bin, 0, 16);
+    my $encrypted = substr ($ct_bin, 16);
+
+    $hash = sprintf ("\$vmx\$0\$%s\$%s\$%s%s", $iterations, unpack ("H*", $salt), unpack ("H*", $iv_bin), unpack ("H*", $encrypted));
```
jsteube: Here's a fundamental error. To reproduce:

```
root@ht:~/hashcat# cat hash
$vmx$0$10000$41326347039781188312527423234988$60820f116b18406de6d2fe41d3b50caead2262d22b22924cbe3eb13a1ff4a1db8ad3fe403fb5f41f429c1c1456a426ce
root@ht:~/hashcat# cat crack 
$vmx$0$10000$41326347039781188312527423234988$60820f116b18406de6d2fe41d3b50caead2262d22b22924cbe3eb13a1ff4a1db8ad3fe403fb5f41f429c1c1456a426ce:hashcat
root@ht:~/hashcat# tools/test.pl verify 26700 hash crack x            
root@ht:~/hashcat# cat x
root@ht:~/hashcat# 
```

The x file should contain the same line as the crack file. Typically, inside the (defined) block, you decrypt, check if the decryption was successful or not. If it was successful, reuse the same data that you received as input with the idea that the script will produce the same hash line as the input line was. If it was unsuccessful, empty the data so that the resulting hash will be a different one.

tools/test_modules/m26700.pm
```diff
@@ -0,0 +1,104 @@
...
+use Crypt::PBKDF2;
+use Crypt::CBC;
+use Crypt::Mode::CBC;
+
+sub module_constraints { [[0, 256], [32, 32], [-1, -1], [-1, -1], [-1, -1]] }
+
+sub module_generate_hash
+{
+  my $word = shift;
+  my $salt_str = shift // random_hex_string (32);
```
jsteube: The // part is not required. It will be filled with random hex string of the length given in the module_contraints. So I'd set the contraint to 64, 64.

src/modules/module_26700.c
```diff
@@ -0,0 +1,270 @@
...
+static const char *HASH_NAME      = "VMware VMX (PBKDF2-HMAC-SHA1 + AES-256-CBC)";
+static const u64   KERN_TYPE      = 26700;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_SLOW_HASH_SIMD_LOOP;
+static const u64   OPTS_TYPE      = OPTS_TYPE_PT_GENERATE_LE
+                                  | OPTS_TYPE_ST_HEX
+                                  | OPTS_TYPE_HASH_COPY;
```
jsteube: Please try to avoid OPTS_TYPE_HASH_COPY. This is only for cases where there is known invalid encoding in the input buffer that encoder can not reproduce 1:1.

src/modules/module_26700.c
```diff
@@ -0,0 +1,270 @@
...
+
+  token.sep[2]     = '$';
+  token.len_min[2] = 32;
+  token.len_max[2] = 32;
+  token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_HEX;
+
+  token.sep[3]     = '$';
+  token.len_min[3] = 32;
+  token.len_max[3] = 512;
```
jsteube: Is ok, I'm just confused. Why is it not a fixed size block if the part we decrypt is only the first 16 byte? OTOH, if we have more structured data then it's probably more secure to run an entropy check instead because it;s possible they change the structure in the future and using entropy check is much better suited.

jsteube:
Thanks for the contribution, especially we a dedicated extraction tool. This is great!

Before we can merge, we need to fix some smaller things. I had commented inline for better visibility.

src/modules/module_26700.c
matrix: or EAS ?

jsteube: We can also make a new category. I just find it a bit lame when there's only one entry per category. But is for sure not EAS. As far as I understand this is a block device which is getting encrypted. And that's what we call a full disc encryption.

tools/test_modules/m26700.pm
matrix: fixed, the bug was in the verify

src/modules/module_26700.c
matrix: yes, vbox is coming bro ;) I will add a new category later

src/modules/module_26700.c
matrix: done, I also changed the tool to extract the hash, just iv + first block is enough to crack hashes

OpenCL/m26700-pure.cl
```diff
@@ -0,0 +1,351 @@
...
+
+    unpackv (tmps, out, gid, i + 0, out[0]);
+    unpackv (tmps, out, gid, i + 1, out[1]);
+    unpackv (tmps, out, gid, i + 2, out[2]);
+    unpackv (tmps, out, gid, i + 3, out[3]);
+    unpackv (tmps, out, gid, i + 4, out[4]);
+  }
+}
+
+KERNEL_FQ void m26700_comp (KERN_ATTR_TMPS_ESALT (pbkdf2_sha1_tmp_t, pbkdf2_sha1_t))
```
jsteube: I think change to vmware_vmx is missing here.

docs/changes.txt
```diff
@@ -12,6 +12,7 @@
 - Added hash-mode: AES-256-ECB NOKDF (PT = $salt, key = $pass)
 - Added hash-mode: iPhone passcode (UID key + System Keybag)
 - Added hash-mode: MetaMask Wallet
+- Added hash-mode: VMware VMX (PBKDF2-HMAC-SHA1 + AES-256-CBC)
```
jsteube: Needs update due to v6.2.3 release

tools/test_modules/m26700.pm
jsteube: I don't think this is the right way.

If the password in "crack" is invalid, the output will still be the correct hash line. So there's no true verification. This part is relevant for automated processing. Please update the unit test so that with a wrong password, there will be a wrong hash (which then will lead to an empty line in the output file automatically).

src/modules/module_26700.c
jsteube: OK, the change works, but I'd have preferred to go for the entropy direction, which makes a minimum 512. But of course this only works if the first 256 byte actually have a structure it follows. Could you please paste an example output in hex of the first 256 byte of a real archive? I think in the long run we will benefit from this because we do have less maintenance.

docs/changes.txt
matrix: could you add it after the merge? thanks :)

tools/test_modules/m26700.pm
matrix: for the ST_HASH, the full buffer including the IV is 

cbb368564d8dfb99f509d4922f4693413f3816af713f0e76bc2409ff9336935d069b1c0763c06ec30bd340d87849da7db5c946f6c2bbcc2cf47fcd01d704e4711310f2e09189e2b87ada80573cafade6f83c7a27a3ceca9c5722ba7c083b358151247bb9b13fb01ad5a98ec893d8a40cfb8db06c

The truncated one is:

cbb368564d8dfb99f509d4922f4693413f3816af713f0e76bc2409ff9336935d

matrix: > I don't think this is the right way.
> 
> If the password in "crack" is invalid, the output will still be the correct hash line. So there's no true verification. This part is relevant for automated processing. Please update the unit test so that with a wrong password, there will be a wrong hash (which then will lead to an empty line in the output file automatically).

done

src/modules/module_26700.c
jsteube: OK, lets not forget to add it at a later point in time when we have more plugins for that new category.

PR #2884, opened 2021-07-17
Added hash-modes: VirtualBox (PBKDF2-HMAC-SHA256 & AES-128-XTS/AES-256-XTS)

matrix:
Hi, with this commit we add support for cracking VirtualBox encrypted VMs (vbox files).

Test modules are missing, will be added asap.


tools/test_modules/m27500.pm
```diff
@@ -0,0 +1,134 @@
...
+  my $salt2_bin = pack ("H*", $salt2);
+
+  my $hash_buf = $pbkdf2->PBKDF2 ($salt2_bin, $dec_pass);
+
+  my $hash = sprintf ("\$vbox\$0\$%s\$%s\$8\$%s\$%s\$%s\$%s", $iter1, $salt1, $enc_pass, $iter2, $salt2, unpack ("H*", $hash_buf));
+
+  return $hash;
+}
+
+sub module_verify_hash
```
jsteube, line 100: So there's verify parsing code, but no verify functionality? Why is that? Same for 27600, too.

PR #2885, opened 2021-07-18
const and char fixes

neheb:
If compiling in C++ mode, the first two commits are bugfixes.

The last one is not.


jtojanen (commented):
You should know that AUTO is GCC only, but GCC is only one compiler.. Please follow contributing guidelines and GNU99 standard. Revert adae3ad and I think this should be accepted.

neheb:
Works with clang too.

neheb:
Note that hashcat is built with gnu99. Extensions should be fine.

jtojanen:
> Note that hashcat is built with gnu99. Extensions should be fine.

I agree with you about const correctness, but you could have resolved these as easily by replacing them with "const char *const" instead of using auto extension.

neheb:
right. My reasoning is that auto is easier to read as it's shorter.

jtojanen:
> right. My reasoning is that auto is easier to read as it's shorter.

True. Counterargument is that "const char *const" retains all type information, constness of data and pointer.

jsteube:
Could we replace the "const AUTO" with "const char *const" instead?

neheb:
Redone. Removed AUTO.

jsteube:
Many thanks!

PR #2886, opened 2021-07-19
Auto-tuning outside kernel can cause unpredictable behavior for tmp_t structs on CUDA

pellekuiters:
Proposed solution is to run init() and loop_prepare() during autotune to prevent possible corruptions during loop().


jsteube:
Good find,  thanks for the contribution!

PR #2888, opened 2021-07-20
Calculate kernel dynamic memory size based on CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN

jtojanen:
get_cuda_kernel_dynamic_local_mem_size() tried all sizes linearly, which is "slow". It seems that we can calculate this as difference with device maximum minus kernel local size, CU_FUNC_ATTRIBUTE_SHARED_SIZE_BYTES. I didn't find any differences but I think this needs to verified by others too.


jsteube:
This is a super nice approach, but it's not going to work :(

The problem is for example on NVIDIA, with GPU that have a larger shared memory size than 48kb, this memory is not getting returned correctly from the driver. It will return only 48kb for backward compatibility reasons. We really have to brute force this value.

I had tested your patch on my 2080Ti, which has 64kb shared memory, and the results are as expected:

```
Speed.#1.........:    39940 H/s (79.73ms) @ Accel:8 Loops:16 Thr:12 Vec:1
```

Compared to brute-force variant:

```
Speed.#1.........:    54086 H/s (77.61ms) @ Accel:8 Loops:16 Thr:16 Vec:1
```

jsteube:
Forgot to mention, this is -m 3200 and each thread requires 4kb of shared memory, thus you see Thr:12 and Thr:16.

jtojanen:
Darn! Sadly I didn't find documentation for CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN.
Well it's not totally broken idea. I think we can safely assume that this would be minimum to start search?

jsteube:
Yes, we can use it as minimum, that is true. In general the loop can be optimized to perform a bisect search in case we want to do that. I just thought that brute force this value is very fast and does not need optimization. But yes, that would be possible. Please update the code accordingly then we can use this method.

jtojanen:
> Yes, we can use it as minimum, that is true. In general the loop can be optimized to perform a bisect search in case we want to do that. I just thought that brute force this value is very fast and does not need optimization. But yes, that would be possible. Please update the code accordingly then we can use this method.

I will revert these changes and make the necessary changes.

jtojanen:
I though I saw something somewhere talking about this issue and here it is:
cudaFuncSetAttribute() documentation says following:
cudaFuncAttributeMaxDynamicSharedMemorySize - The requested maximum size in bytes of dynamically-allocated shared memory. The sum of this value and the function attribute sharedSizeBytes cannot exceed the device attribute cudaDevAttrMaxSharedMemoryPerBlockOptin. The maximal size of requestable dynamic shared memory may differ by GPU architecture.

Could you check what is the maximum size reported for you with CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN? I am getting 48k, but that's my old hardware..

jsteube:
Yes, it works! Using CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK I am getting only 49152, but with CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN I am getting 65536 :) So we can use this!

jtojanen:
> Yes, it works! Using CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK I am getting only 49152, but with CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN I am getting 65536 :) So we can use this!

Nice! At least once, the documentation is correct as soon as I found the right one. What do you think? Can we swap CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK with CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN in backend.c hc_cuDeviceGetAttribute() ? That value is used as device_param->device_local_mem_size. Some modules use that same value to calculate jit_build_options, but most if not all are AMD related limits. Increased limit would help some modules, but might break others.

jsteube:
Yeah we should switch to CU_DEVICE_ATTRIBUTE_MAX_SHARED_MEMORY_PER_BLOCK_OPTIN  and then use the non brute-force approach from this PR instead. Please add the change to the PR, then we can merge it.

jsteube:
Great contribution, thanks!

PR #2889, opened 2021-07-20
Added hash-mode: SNMPv3 HMAC-SHA1-96

matrix:
Hi, it should now be compliant with the review you did here #2662 :D
Thanks

### DIFF
--- a/OpenCL/m25200-pure.cl
+++ b/OpenCL/m25200-pure.cl
@@ -53,7 +53,7 @@ KERNEL_FQ void m25200_init (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
    * base
    */
 
-  const global u8 *pw_buf = (global u8 *) pws[gid].i;
+  const GLOBAL_AS u8 *pw_buf = (GLOBAL_AS u8 *) pws[gid].i;
 
   const u32 pw_len = pws[gid].pw_len;
 
@@ -63,7 +63,7 @@ KERNEL_FQ void m25200_init (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
 
   u32 idx = 0;
 
-  u32 buf[32] = { 0 };
+  u32 buf[16] = { 0 };
 
   u8 *tmp_buf = (u8 *) buf;
 
@@ -94,13 +94,13 @@ KERNEL_FQ void m25200_loop (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
 
   if (gid >= gid_max) return;
 
-  const global u8 *pw_buf = (global u8 *) pws[gid].i;
+  const GLOBAL_AS u8 *pw_buf = (GLOBAL_AS u8 *) pws[gid].i;
 
   const u32 pw_len = pws[gid].pw_len;
 
   u32 idx = tmps[gid].idx;
 
-  u32 buf[32] = { 0 };
+  u32 buf[16] = { 0 };
 
   u8 *tmp_buf = (u8 *) buf;
 
@@ -132,7 +132,7 @@ KERNEL_FQ void m25200_comp (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
 
   if (gid >= gid_max) return;
 
-  const global u8 *engineID_buf = esalt_bufs[DIGESTS_OFFSET].engineID_buf;
+  const GLOBAL_AS u8 *engineID_buf = esalt_bufs[DIGESTS_OFFSET].engineID_buf;
 
   u32 engineID_len = esalt_bufs[DIGESTS_OFFSET].engineID_len;
 
--- a/OpenCL/m25200-pure.cl
+++ b/OpenCL/m25200-pure.cl
@@ -174,7 +174,6 @@ KERNEL_FQ void m25200_comp (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
   sha1_init (&ctx);
 
   sha1_update_swap (&ctx, tmp_buf, i);
-  //sha1_update_swap (&ctx, (u32 *)buf, i);
 
   sha1_final (&ctx);
 
@@ -186,22 +185,11 @@ KERNEL_FQ void m25200_comp (KERN_ATTR_TMPS_ESALT (hmac_sha1_tmp_t, snmpv3_t))
   key[3] = ctx.h[3];
   key[4] = ctx.h[4];
 
-  u32 s[SNMPV3_SALT_MAX] = { 0 };
-
-  const u32 salt_len = esalt_bufs[DIGESTS_OFFSET].salt_len;
-
-  u32 idx;
-
-  for (i = 0, idx = 0; i < salt_len; i += 4, idx += 1)
-  {
-    s[idx] = esalt_bufs[DIGESTS_OFFSET].salt_buf[idx];
-  }
-
   sha1_hmac_ctx_t hmac_ctx;
 
   sha1_hmac_init (&hmac_ctx, key, 20);
 
-  sha1_hmac_update (&hmac_ctx, s, salt_len);
+  sha1_hmac_update_global (&hmac_ctx, esalt_bufs[DIGESTS_OFFSET].salt_buf, esalt_bufs[DIGESTS_OFFSET].salt_len);
 
   sha1_hmac_final (&hmac_ctx);
 
--- a/OpenCL/m25200-pure.cl
+++ b/OpenCL/m25200-pure.cl
@@ -1,0 +1,218 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+//#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include "inc_vendor.h"
+#include "inc_types.h"
+#include "inc_platform.cl"
+#include "inc_common.cl"
+#include "inc_simd.cl"
+#include "inc_hash_sha1.cl"
+#endif
+
+#define COMPARE_S "inc_comp_single.cl"
+#define COMPARE_M "inc_comp_multi.cl"
+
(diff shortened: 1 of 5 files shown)


OpenCL/m25200-pure.cl
```diff
@@ -0,0 +1,218 @@
...
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= gid_max) return;
+
+  /**
+   * base
+   */
+
+  const global u8 *pw_buf = (global u8 *) pws[gid].i;
```
jsteube: This is not going to work on non OpenCL backends. Please use GLOBAL_AS instead.

OpenCL/m25200-pure.cl
```diff
@@ -0,0 +1,218 @@
...
+{
+  /**
+   * base
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= gid_max) return;
+
+  const global u8 *pw_buf = (global u8 *) pws[gid].i;
```
jsteube: This is not going to work on non OpenCL backends. Please use GLOBAL_AS instead.

OpenCL/m25200-pure.cl
```diff
@@ -0,0 +1,218 @@
...
+{
+  /**
+   * modifier
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= gid_max) return;
+
+  const global u8 *engineID_buf = esalt_bufs[DIGESTS_OFFSET].engineID_buf;
```
jsteube: This is not going to work on non OpenCL backends. Please use GLOBAL_AS instead.

src/modules/module_25200.c
```diff
@@ -0,0 +1,318 @@
...
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_4_4; // 4_3
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_NETWORK_PROTOCOL;
+static const char *HASH_NAME      = "SNMPv3 HMAC-SHA1-96";
+static const u64   KERN_TYPE      = 25200;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_SLOW_HASH_SIMD_LOOP;
```
jsteube: There's no SIMD loop compatbility in the kernel. Please remove this flag

src/modules/module_25200.c
```diff
@@ -0,0 +1,318 @@
...
+
+  token.len[0]     = 10;
+  token.attr[0]    = TOKEN_ATTR_FIXED_LENGTH
+                   | TOKEN_ATTR_VERIFY_SIGNATURE;
+
+  // packet number
+  token.len_min[1] = 1;
+  token.len_max[1] = 8;
+  token.sep[1]     = '$';
+  token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH;
```
jsteube: maybe VERIFY_DIGIT missing?

OpenCL/m25200-pure.cl
```diff
@@ -0,0 +1,218 @@
...
+
+  const u32 pw_len = pws[gid].pw_len;
+
+  /**
+   * authkey
+   */
+
+  u32 idx = 0;
+
+  u32 buf[32] = { 0 };
```
jsteube: I think can be [16]

OpenCL/m25200-pure.cl
```diff
@@ -0,0 +1,218 @@
...
+
+  u8 *tmp_buf = (u8 *) buf;
+
+  sha1_ctx_t ctx = tmps[gid].ctx;
+
+  for (u32 j = 0; j < loop_cnt; j++)
+  {
+    for (int i = 0; i < 64; i++)
```
jsteube, line 111: This solution works, but I think we can do much better and get more speed.

All that is needed is that we allocate u32 buf[16*64]. The idea is that no matter the password length, after 64 times append of the password, the result buffer length % 64  is 0 :) 

To make this possible you need to do also: 

- fix the kernel loops to 1024 fixed in the module min and max
- for (u32 j = 0; j < loop_cnt; j++) becomes  for (u32 j = 0; j < loop_cnt; j+=64)

Additionally, we can calculate this one time in _init, then store in tmps buffer. the both variable currently in tmps get replaced with this buffer. then you can also use super fast sha1_update_64() and you do not need the swap in the loop any longer. Gutfeeling tells me this will improve performance massively.

jsteube: You also need limit password length to 64

src/modules/module_25200.c
```diff
@@ -0,0 +1,318 @@
...
+  for (i = 0; i < snmpv3->salt_len; i++)
+  {
+    line_len += snprintf (line_buf+line_len, 3, "%02x", salt_buf_ptr[i]);
+  }
+
+  line_len += snprintf (line_buf+line_len, 2, "$");
+
+  for (i = 0; i < snmpv3->engineID_len; i++)
+  {
+    line_len += snprintf (line_buf+line_len, 3, "%02x", snmpv3->engineID_buf[i]);
```
jsteube: We have hex_decode(), why not use it?

matrix: Hi, this is hex_encode() but yes, I updated the code to use both hex_encode and hex_decode. Thank you!

src/modules/module_25200.c
```diff
@@ -0,0 +1,299 @@
...
+
+  u32 salt_buf_32[SNMPV3_SALT_MAX_BIN] = { 0 };
+
+  for (i = 0; i < SNMPV3_SALT_MAX_BIN; i++) salt_buf_32[i] = byte_swap_32 (snmpv3->salt_buf[i]);
+
+  const u8 *salt_buf_ptr = (u8 *) salt_buf_32;
+
+  line_len += hex_encode (salt_buf_ptr, snmpv3->salt_len, (u8 *) line_buf+line_len);
+
+  line_len += snprintf (line_buf+line_len, 2, "$");
```
jsteube: Typically what we do in this situation is

```
  out_buf[out_len] = '$';

  out_len += 1;
```

See src/modules/module_21600.c

src/modules/module_25200.c
```diff
@@ -0,0 +1,299 @@
...
+
+  line_len += hex_encode (snmpv3->engineID_buf, snmpv3->engineID_len, (u8 *) line_buf+line_len);
+
+  line_len += snprintf (line_buf+line_len, 2, "$");
+
+  line_len += snprintf (line_buf+line_len, 9, "%08x", digest[0]);
```
jsteube: Typically what we do:

```
  u32_to_hex (tmp[0], out_buf + out_len); out_len += 8;
```

See src/modules/modules_00010.c

src/modules/module_25200.c
```diff
@@ -0,0 +1,299 @@
...
+  sha1_ctx_t sha1_ctx;
+
+  sha1_init   (&sha1_ctx);
+  sha1_update (&sha1_ctx, snmpv3->salt_buf, snmpv3->salt_len);
+  sha1_final  (&sha1_ctx);
+
+  // store sha1(snmpv3->salt_buf) in salt_buf
+
+  memcpy (salt->salt_buf, sha1_ctx.h, 20);
+
```
jsteube, line 169: salt->salt_len = 20;  missing ?

src/modules/module_25200.c
```diff
@@ -0,0 +1,299 @@
...
+  // salt
+
+  const u8 *salt_pos = token.buf[2];
+  const int salt_len = token.len[2];
+
+  u8 *salt_ptr = (u8 *) snmpv3->salt_buf;
+
+  snmpv3->salt_len = hex_decode (salt_pos, salt_len, salt_ptr);
+
+  for (int i = 0; i < salt_len / 4; i++)
```
jsteube: I think needs to be snmpv3->salt_len not salt_len

src/modules/module_25200.c
```diff
@@ -0,0 +1,299 @@
...
+  return (PARSER_OK);
+}
+
+int module_hash_encode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const void *digest_buf, MAYBE_UNUSED const salt_t *salt, MAYBE_UNUSED const void *esalt_buf, MAYBE_UNUSED const void *hook_salt_buf, MAYBE_UNUSED const hashinfo_t *hash_info, char *line_buf, MAYBE_UNUSED const int line_size)
+{
+  const u32 *digest = (const u32 *) digest_buf;
+
+  snmpv3_t *snmpv3 = (snmpv3_t *) esalt_buf;
+
+  int line_len = snprintf (line_buf, 10 + strlen ((char *) snmpv3->packet_number) + 1 + 1, "%s%s$", SIGNATURE_SNMPV3, snmpv3->packet_number);
```
jsteube, line 202: This is invalid:

```
10 + strlen ((char *) snmpv3->packet_number) + 1 + 1
```

Correct is just to use 'line_size'

src/modules/module_25200.c
matrix: Hi, in this case will be:

u32_to_hex (byte_swap_32 (digest[0]), (u8 *) line_buf+line_len); line_len += 8;

instead of

line_len += snprintf (line_buf+line_len, 9, "%08x", digest[0]);

src/modules/module_25200.c
matrix: line_size is not the same as "10 + strlen ((char *) snmpv3->packet_number) + 1 + 1". it does not work

src/modules/module_25200.c
matrix: updated but github don't see the changes here :)

```
[...]
  // store sha1(snmpv3->salt_buf) in salt_buf

  salt->salt_len = 20;

  memcpy (salt->salt_buf, sha1_ctx.h, salt->salt_len);

  // engineid
[...]
```

jsteube:
I'll merge, then update with optimized code.

PR #2666, opened 2020-12-25
Added full AES-GCM cipher & hash-mode 27000 - Stargazer Stellar Wallet XML

matrix:
Hi,
inspired by the issue #2524, I have implemented the **AES-GCM cipher** and the Stargazer Stellar Wallet XML algo.

Following the first AES-GCM benchmark in hashcat:

```
$ ./hashcat -b -m 27000

OpenCL API (OpenCL 1.2 CUDA 10.0.206) - Platform #1 [NVIDIA Corporation]
========================================================================
* Device #1: GeForce GTX 1080 Ti, 10880/11178 MB (2794 MB allocatable), 28MCU

Benchmark relevant options:
===========================
* --optimized-kernel-enable

Hashmode: 27000 - Stargazer Stellar Wallet XLM, PBKDF2-HMAC-SHA256 + AES-256-GCM (Iterations: 4095)

Speed.#1.........:   416.6 kH/s (67.50ms) @ Accel:8 Loops:512 Thr:1024 Vec:1
```

Merry Christmas, and happy cracking :)
(ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

PR #2894, opened 2021-07-22
Allow async execution

jtojanen:
There is actually a lot of cases where async execution will give performance boost. Currently limiting factor will be single stream/queue per device. There is no overlap between GPU memory transfers and kernel executions. To fully utilize all resources, we will have to split task to multiple streams/queues.


jtojanen:
Still work in progress..

jtojanen:
CUDA, HIP and OpenCL are now using non-default queue/stream and asynchronous operations. All synchronization happens in queue/stream level, never in device level.

Thanks to in-order queue, we don't have to worry about the execution order of host->device operations. We can safely use MemcpyHtoDAsync()/clEnqueueWriteBuffer(), run_kernel() and run_\*\_kernel_*. Same applies to device->device operations, MemcpyDtoDAsync()/clEnqueueCopyBuffer(). But as the current overall design is based on old synchronous blocking operations, we have to synchronize device->host operations as host is waiting for results and work continues only after receiving those.

At the moment there is only one spot where host and device have overlapping execution. That is in check_hash(), where we possible make a temporary copy.

AMD ROCm OpenCL documentation gives recommendation: "It is best to use non-blocking commands to allow multiple commands to be queued before the command queue is flushed to the GPU. This sends larger batches of commands, which amortizes the cost of preparing and submitting work to the GPU." 

Any OpenCL synchronous operation is implicit flush, and I have added flushes where needed. I have removed CUDA and HIP synchronous memcpy operations as those operate on default/null stream. Note that synchronous memcpy is easily implemented using MemcpyAsync()+StreamSynchronize() when needed, and it is blocking only the current stream.

There was no simply way of converting synchronous call to gpu_memset() to asynchronous. I have replaced kernel based gpu_memset with API calls MemsetD8()/MemsetD32/clEnqueueFillBuffer(). And there is a workaround if clEnqueueFillBuffer() is missing from OpenCL implementation. Also OpenCL has a limitation with run_opencl_kernel_bzero(), but clEnqueueFillBuffer() doesn't.

Small improvements:
- Spin damper section: If clGetEventInfo() tells that event status is CL_COMPLETE, we don't have to clWaitForEvents()
- OPTS_TYPE_PT_NEVERCRACK: Instead of having zero buffer and memcpyHtoD, simply use kernel_bzero(), or kernel_memset32() to zero

I was trying to remove unused gpu_memset() and its references. But it seems that module_03200.c, module_25600.c and module_25800.c are using device_param->kernel_dynamic_local_mem_size_memset to determinate dynamic memory available.

jtojanen:
I am seeing interesting, but mixed results. I cannot explain why calculated hash rate is lower when kernel execution time is lower, which should result just the opposite, higher hash rate.

Old: 2777.3 MH/s (72.26ms) @ Accel:128 Loops:256 Thr:1024 Vec:2
New: 2749.9 MH/s (72.21ms) @ Accel:128 Loops:256 Thr:1024 Vec:2

Visual Profiler shows lower latency and single benchmark execution completes faster..

jsteube:
Again great work, many thanks! Before we are going to merge we need to do some intensive testing. Please let me know which kind of tests you already did. Ideally you could run some tools/test.sh runs by yourself in parallel because they can take a few days to complete.

jtojanen:
> Again great work, many thanks! Before we are going to merge we need to do some intensive testing. Please let me know which kind of tests you already did. Ideally you could run some tools/test.sh runs by yourself in parallel because they can take a few days to complete.

I totally agree with you. This needs a lot testing, because there are so many use cases and these changes effect all of them. Due to HIP changes, last merge with your changes wasn't the easiest but tried my best to sort things out. Until now I have done only basic tests by running benchmarks and couple of my normal use cases. I will start doing those tests from tools/test.sh

jsteube:
Started to run the first tests. So far works great, also with new -T finding auto-tune engine.

About kernel_dynamic_local_mem_size_memset - We can replace it with kernel_dynamic_local_mem_size_bzero. I think bzero is even better suited because the smaller the kernel, the better. However it affects 3200, 9000, 18600, 25600 and 25800.

I will continue with testing.

src/backend.c
```diff
@@ -4926,52 +4890,23 @@ int run_cuda_kernel_utf8toutf16le (hashcat_ctx_t *hashcat_ctx, hc_device_param_t
...
-  }
-
-  if (hc_cuStreamSynchronize (hashcat_ctx, device_param->cuda_stream) == -1) return -1;
+  return hc_cuMemsetD8Async (hashcat_ctx, buf + offset, value, size, device_param->cuda_stream);
+}
 
-  return 0;
+int run_cuda_kernel_memset32 (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u64 offset, const u32 value, const u64 size)
+{
+  return hc_cuMemsetD32Async (hashcat_ctx, buf + offset * sizeof (u32), value, size, device_param->cuda_stream);
```
jsteube: The * sizeof (u32) feels unintuitive, because an offset is supposed to be absolute (in bytes). Same applies to the sibling functions for HIP and OpenCL. What does the SDK documentation say about?

include/common.h
```diff
@@ -63,6 +63,14 @@
 #define HC_API_CALL
 #endif
 
+#if defined (__GNUC__)
+#define HC_ALIGN(x) __attribute__((aligned(x)))
+#elif defined (_MSC_VER)
+#define HC_ALIGN(x) __declspec(align(x))
+#else
+#define HC_ALIGN(x)
```
jsteube, line 71: Macros are fine, we can leave them. But do we really need them and why? Btw. I don't know if _MSC_VER is really supported, never tried. But will it work on clang?

src/backend.c
```diff
@@ -3558,6 +3504,7 @@ int ocl_init (hashcat_ctx_t *hashcat_ctx)
   HC_LOAD_FUNC (ocl, clCreateProgramWithBinary, OCL_CLCREATEPROGRAMWITHBINARY,  OpenCL, 1);
   HC_LOAD_FUNC (ocl, clCreateProgramWithSource, OCL_CLCREATEPROGRAMWITHSOURCE,  OpenCL, 1);
   HC_LOAD_FUNC (ocl, clEnqueueCopyBuffer,       OCL_CLENQUEUECOPYBUFFER,        OpenCL, 1);
+  HC_LOAD_FUNC (ocl, clEnqueueFillBuffer,       OCL_CLENQUEUEFILLBUFFER,        OpenCL, -1);
```
jsteube, line 3507: Is it really -1? I think we used 0 for expected unavailability cases.

include/backend.h
```diff
@@ -178,17 +177,20 @@ void rebuild_pws_compressed_append (hc_device_param_t *device_param, const u64 p
 
 int run_cuda_kernel_atinit          (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u64 num);
 int run_cuda_kernel_utf8toutf16le   (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u64 num);
-int run_cuda_kernel_memset          (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u32 value, const u64 size);
+int run_cuda_kernel_memset          (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u64 offset, const u8  value, const u64 size);
+int run_cuda_kernel_memset32        (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, CUdeviceptr buf, const u64 offset, const u32 value, const u64 size);
```
jsteube, line 181: Did you try if they are faster than run_cuda_kernel_bzero()?

src/autotune.c
```diff
@@ -164,29 +164,19 @@ static int autotune (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param
...
-    int CU_rc;
-    int HIP_rc;
-    int CL_rc;
-
     if (device_param->is_cuda == true)
     {
-      CU_rc = run_cuda_kernel_atinit (hashcat_ctx, device_param, device_param->cuda_d_pws_buf, kernel_power_max);
-
-      if (CU_rc == -1) return -1;
+      if (run_cuda_kernel_atinit (hashcat_ctx, device_param, device_param->cuda_d_pws_buf, kernel_power_max) == -1) return -1;
```
jsteube, line 203: Nice I think these were left-overs. If you see more of them, please continue to rewrite the same way.

src/backend.c
```diff
@@ -4490,22 +4460,28 @@ int choose_kernel (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param,
...
         if (device_param->is_opencl == true)
         {
+          /* blocking */
           if (hc_clEnqueueReadBuffer (hashcat_ctx, device_param->opencl_command_queue, device_param->opencl_d_hooks, CL_TRUE, 0, pws_cnt * hashconfig->hook_size, device_param->hooks_buf, 0, NULL, NULL) == -1) return -1;
         }
 
         const int hook_threads = (int) user_options->hook_threads;
 
-        hook_thread_param_t *hook_threads_param = (hook_thread_param_t *) hccalloc (hook_threads, sizeof (hook_thread_param_t));
+        hook_thread_param_t *hook_threads_param = (hook_thread_param_t *) hcmalloc (hook_threads * sizeof (hook_thread_param_t));
```
jsteube, line 4481: Where's the advantage? I find it easier to read if there's no math involved.

src/backend.c
```diff
@@ -5160,59 +5044,74 @@ int run_opencl_kernel_utf8toutf16le (hashcat_ctx_t *hashcat_ctx, hc_device_param
 
   if (hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof (cl_ulong), device_param->kernel_params_utf8toutf16le[1]) == -1) return -1;
 
-  if (hc_clEnqueueNDRangeKernel (hashcat_ctx, device_param->opencl_command_queue, kernel, 1, NULL, global_work_size, local_work_size, 0, NULL, NULL) == -1) return -1;
+  return hc_clEnqueueNDRangeKernel (hashcat_ctx, device_param->opencl_command_queue, kernel, 1, NULL, global_work_size, local_work_size, 0, NULL, NULL);
```
jsteube: This breaks our convention, please undo.

src/backend.c
```diff
@@ -5160,59 +5044,74 @@ int run_opencl_kernel_utf8toutf16le (hashcat_ctx_t *hashcat_ctx, hc_device_param
...
+  const OCL_PTR       *ocl         = backend_ctx->ocl;
 
-  if (hc_clFinish (hashcat_ctx, device_param->opencl_command_queue) == -1) return -1;
+  int rc;
 
-  return 0;
+  /* workaround if missing clEnqueueFillBuffer() */
+  if (ocl->clEnqueueFillBuffer == NULL)
+  {
+    const u64 len = offset + size;
```
jsteube: Why offset?

src/backend.c
```diff
@@ -5160,59 +5044,74 @@ int run_opencl_kernel_utf8toutf16le (hashcat_ctx_t *hashcat_ctx, hc_device_param
...
+
+    char *tmp = hcmalloc (len * sizeof (u8));
+
+    memset(tmp, value, len);
+
+    /* blocking */
+    rc = hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_TRUE, offset, size, tmp, 0, NULL, NULL);
+
+    free(tmp);
```
jsteube: I think there's also hcfree()

src/hashes.c
```diff
@@ -313,39 +313,77 @@ void check_hash (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, pl
...
+      rc = hc_cuMemcpyDtoHAsync (hashcat_ctx, tmps, device_param->cuda_d_tmps + (plain->gidvid * hashconfig->tmp_size), hashconfig->tmp_size, device_param->cuda_stream);
+
+      if (rc == 0)
+      {
+        rc = hc_cuEventRecord (hashcat_ctx, device_param->cuda_event3, device_param->cuda_stream);
+      }
+
+      if (rc == -1)
+      {
+        free(tmps);
```
jsteube: hcfree()

src/backend.c
jtojanen: > The * sizeof (u32) feels unintuitive, because an offset is supposed to be absolute (in bytes). Same applies to the sibling functions for HIP and OpenCL. What does the SDK documentation say about?

Right. Offset param should be offset_elem, as SDK says that: 
> The dstDevice pointer must be four byte aligned.

And as buf is allocated by CUDA, I think it's guaranteed to have 256 byte alignment so with current implementation it's impossible not to have four byte alignedment.. I see run_cuda_kernel_memset32 as helper routine, as we are not hiding the actual library call hc_cuMemsetD32Async. One option would be to make a single run_cuda_kernel_memset, and check if we can call hc_cuMemsetD32Async or if we need to call hc_cuMemsetD8Async depending on offset and size, and in this case offset would be bytes, not elements.

jtojanen: Also from clEnqueueFillBuffer() API. It has

> offset
The location in bytes of the region being filled in buffer and must be a multiple of pattern_size.

include/common.h
jtojanen: > Macros are fine, we can leave them. But do we really need them and why? Btw. I don't know if _MSC_VER is really supported, never tried. But will it work on clang?

I have used this same macro before with GCC, clang, Intel and MSVC.

include/backend.h
jtojanen: > Did you try if they are faster than run_cuda_kernel_bzero()?

Not yet, but it will test because I have a feeling..

src/backend.c
jtojanen: If value is -1, we are ignoring the problem. But if it's 1, we would report error "missing from shared library" and return with -1. Then again if it's not -1 or 1, for example 0, we would report error "missing from shared library" and return with 0. Early return would result that the rest of symbols would not be loaded. Zero would work if this would be the last and only optional function ever. I decided to silently ignore it and solve the problem with a workaround.

Workaround should only be needed with OpenCL before version 1.2, but then again I didn't want to leave a possible new bug to be fixed later as workaround is quite small and easy. Same idea was used before.

src/backend.c
jtojanen: Now that I am thinking this again, you are right and I should have implemented this as a smart helper that automatically picks hc_cuMemsetD32Async if it can and fallback to hc_cuMemsetD8Async. I will "fix" this tomorrow.

Edit: Actually it's not that simply but I will rethink this.

src/backend.c
jtojanen: > This breaks our convention, please undo.

Ok

src/backend.c
jtojanen: This comes from old implementation, which used this. It may have been a bug or workaround for some buggy OpenCL implementation calculating offset to ptr. It should be sufficient to have buffer length equal to size as that is what the specs says.

src/backend.c
jtojanen: Ups,  I will fix.

src/hashes.c
jtojanen: Ups,  I will fix.

src/backend.c
jtojanen: > Where's the advantage? I find it easier to read if there's no math involved.

True. Normally calloc is safer when we are allocating long before actually using and when we want to be sure that it's zeroed, but here buffer is fully assigned in the next for-loop. Only in this context I wrote it.

src/hashes.c
jtojanen: Fixed 1ee222d

src/backend.c
jtojanen: Fixed 1ee222d

src/backend.c
jtojanen: Fixed dbe2bad

src/backend.c
jtojanen: I removed the offset from buffer sizing, fdbfae9

src/backend.c
jtojanen: Now run_cuda_kernel_memset32, run_hip_kernel_memset32, run_opencl_kernel_memset32 offset and size are assumed to be bytes and both must be a multiple of element size.  62a06f7

jtojanen:
I am now continuing testing with up-to-date version.

jsteube:
I did extensive testing on a large set of different configurations on Linux. I found some issues with the large test.sh runs, but they are not related to your changes and I will handle them in a separate commit. For the changes from this PR, everything worked fine. I have to note that I started them two days before, that was before your last set of commits. But I believe they will not have a big impact on the test result.

Testing on windows and OSX is harder, because getting test.sh and test.pl to run is not so easy. On there I did only manual checks and full benchmark checks. But with CMIYC 2021 coming up, it's a great opportunity to have a large set of people testing it.

From my perspective we are fine to merge. Pretty sure we did not think of any special corner case that's possible, but if we encounter problems we just do another commit. Waiting too long with big changes like this too often block other commits.

Thanks again for this effort, is good to see more people going really deep into hashcat core.

jsteube:
In case you were running test.sh, ignore the following results

```
grep cmdline test_*/test_report.log | grep -v -e 18600 -e 25700 -e 23700 -e 12500 -e 25500
```

These have been fixed with https://github.com/hashcat/hashcat/commit/b2d1f42905604e5fa093141dff31e2938e18c3ba and https://github.com/hashcat/hashcat/pull/2666

PR #2896, opened 2021-07-24
Added hash-mode: SNMPv3 HMAC-MD5-96

matrix:
with very good @jsteube optimizations, but fail with osx Iris.

```
hashcat (v6.2.3-44-g3a31b669b) starting in benchmark mode

* Device #2: Apple's OpenCL drivers (GPU) are known to be unreliable.
             You have been warned.

* Device #2: Iris, 1472/1536 MB (384 MB allocatable), 40MCU

Benchmark relevant options:
===========================
* --opencl-device-types=2
* --optimized-kernel-enable

Hashmode: 25100 - SNMPv3 HMAC-MD5-96 (Iterations: 1048576)

* Device #2: Not enough allocatable device memory for this attack.

Started: Sat Jul 24 13:58:39 2021
Stopped: Sat Jul 24 13:58:51 2021
```

Same issue with sha1 version / https://github.com/hashcat/hashcat/commit/3a31b669b52d8bba5d61573c24f3d1593c160af8

### DIFF
--- a/tools/test_modules/m25100.pm
+++ b/tools/test_modules/m25100.pm
@@ -1,0 +1,79 @@
+#!/usr/bin/env perl
+
+##
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::MD5 qw (md5 md5_hex);
+use Digest::HMAC qw (hmac hmac_hex);
+
+sub module_constraints { [[1, 256], [24, 3000], [-1, -1], [-1, -1], [-1, -1]] }
+
+sub module_generate_hash
+{
+  my $word = shift;
+  my $salt = shift;
+  my $pkt_num = shift // int(rand(99999999));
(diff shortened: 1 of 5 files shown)


tools/test_modules/m25100.pm
```diff
@@ -0,0 +1,79 @@
...
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::MD5 qw (md5 md5_hex);
+use Digest::HMAC qw (hmac hmac_hex);
+
+sub module_constraints { [[1, 256], [24, 3000], [-1, -1], [-1, -1], [-1, -1]] }
```
jsteube, line 14: Not a requirement, but I think you want to use 16 instead of 24

matrix: Hi, 24 is the hex len, it works (same as digest len in hex for md5-96 and sha1-96)

PR #2897, opened 2021-07-24
Added hash-mode: SNMPv3 HMAC-MD5-96/HMAC-SHA1-96

matrix:
the last SNMPv3 missing mode


jsteube:
On my GTX980 I'm getting this:

```
cuLaunchKernel(): out of memory
```

Reason is likely this:

```
  u32 tmp_md5[SNMPV3_TMP_ELEMS_OPT];
  u32 tmp_sha1[SNMPV3_TMP_ELEMS_OPT];
```

Can you please send in an update where you only have one of these buffers, like in 25100 and 25200, and then inside the loop you can reuse this buffer but need to add a swap(). I think in general swap() is much cheaper than double the register pressure.

jsteube:
Patch worked, thanks!

PR #2914, opened 2021-07-30
resolves hashcat/hashcat#2913

### LINKED ISSUE #2913: SHA1DASH / sha1(--$salt--$pass--)
**Is your request based on a publicly known cryptographic schema and where can we find information about?**
For instance:
SHA1DASH is based on salted SHA1 with a prepended salt and two dashes ("--") surrounding and separating the salt and password.

**Describe alternatives you've considered**
The only alternative is to manually process your entire hashlist to add the dashes before and after the salt, then use hash mode 120 (sha1($salt:$pass)) in hashcat with the "-j '$-$-'" option, then manually process your potfile to remove the extra dashes at the end of every password.

**Provide an example hash/database/file and a known correct password**
```
0002b80997b9d9cc39f7895eebb762ce0a360bbd:224f9e85d91171f9b86c2d087f0fcc842e3f467f:Darwin97
00024a5845fcc9e5e069a0e8debf9a6ef0715e8b:3a0a8e8aec50cb0455b259ef6441e23e3e14675b:sammie13
00018fa5405406f024ef447a4d6d59afafbeae65:8ccb43ef51daf67244ca3ec13ff66df54aa79a64:Nathan08
3999d08db95797891ec77f07223ca81bf43e1be2:5dcc47b04c49d3c8e1b9e4ec367fddeed21b7b85:hashcat
```

**Explain how the cryptographic schema is implemented**
```sha1(--$salt--$pass--)```

**Describe the known limitations of the algorithm**
Limitations are equivalent to salted SHA1.

**Where do you typically find the algorithm**
Zynga, AnimalJam, 8tracks, Avvo, JobandTalent, IndabaMusic, and Thefaveapp, as seen on hashmob.net

**Additional context**
I'm reviving Issue https://github.com/hashcat/hashcat/issues/1987 because there are now several sites using it, and having a dedicated kernel would be a huge to those desiring to recover their lost passwords.

### PR DESCRIPTION
**Pull request by therealartifex**

### CODE REVIEW

**Comment by Chick3nman** [comment]
This hash algorithm, "SHA1DASH" as its known colloquially, is actually simply a slightly older variant of the default hashing scheme used by "Ruby on Rails Restful-Authentication", which you will see we have currently implemented as 19500. I would name and classify this mode following the actual origin of the algorithm and not what it's typically called if you'd like to add it to the official plugins. I would also suggest taking a look at the code we are currently using for 19500 as it is extremely similar, though it takes in a static "sitekey" as well, whereas this variant leaves that field blank. It also employs a slightly different number of rounds if my memory is correct. Getting most of the kernels, module, and tests from the existing code is generally a pretty safe bet when creating a plugin for something that's so similar to an already existing mode.

**Comment by therealartifex** [comment]
> 
> 
> This hash algorithm, "SHA1DASH" as its known colloquially, is actually simply a slightly older variant of the default hashing scheme used by "Ruby on Rails Restful-Authentication", which you will see we have currently implemented as 19500. I would name and classify this mode following the actual origin of the algorithm and not what it's typically called if you'd like to add it to the official plugins. I would also suggest taking a look at the code we are currently using for 19500 as it is extremely similar, though it takes in a static "sitekey" as well, whereas this variant leaves that field blank. It also employs a slightly different number of rounds if my memory is correct. Getting most of the kernels, module, and tests from the existing code is generally a pretty safe bet when creating a plugin for something that's so similar to an already existing mode.

Thank you for the suggestion. After looking at the code for 19500, I noticed that the primary difference seems to be when the dashes get inserted. 19500 adds all the dashes in the kernel, whereas my code adds only the dashes that come after the password in at the kernel, and the rest in the module. I'm not sure what impact this would have on performance, if any. The other differences are as you mentioned, with there being no sitekey and only a single round of SHA1 instead of ten.

**Comment by jsteube** [comment]
Although there is actually a certain similarity to 19500, the lack of an iteration number and the lack of the site key (and the dash) is too big a difference to simply modify 19500 so that this kind of hash can also be cracked with it. So I think it's okay to have it as a hash-mode of its own.

Where did the name SHA1DASH comes from?

Please renumber 28200 to 26900 and, if possible, please add a unit test. You can use tools/test_modules/m19500.pm as template. Even if you're not a Perl coder, the modifications should be easy.

**Comment by jsteube** [comment]
Update: renumber to 27100

**Comment by Chick3nman** [comment]
@jsteube This should also be named after the origin, Ruby on Rails Restful Auth. We will need a way to denote the difference, which may mean changing the name on the existing one to include a version number or similar. "SHA1 Dash" comes from members of the community who didn't know or figure out the real origin of the algorithm i believe, as they were only exposed to it as a part of the large leaks/breaches.

**Comment by 0xVavaldi** [comment]
It might be good to add a sidenote to it that it is 'sha1dash' due to the popularity of the reference within the community. Even if it's officially referred to as "Ruby on Rails Restful AUTH"

**Comment by Chick3nman** [comment]
> It might be good to add a sidenote to it that it is 'sha1dash' due to the popularity of the reference within the community. Even if it's officially referred to as "Ruby on Rails Restful AUTH"

This can be done on the Wiki or on the Example hashes page, I don't think it's required in the plugin itself.

**Comment by jsteube** [comment]
I'm sorry, could you please renumber again to 27200? Thanks

**Comment by therealartifex** [comment]
Currently working on the optimized kernels as well. I just need a way to statically append two '0x2d' bytes to the candidate after the current rule has been applied. My current thought process is to create a method similar to append_0x80_4x4_VV to accomplish this.

**Comment by jsteube** [comment]
Let me know when you want me to start with the review

**Comment by therealartifex** [comment]
> 
> 
> Let me know when you want me to start with the review

I'm ready.

File: src/modules/module_27200.c [hunk 1]
```diff
@@ -0,0 +1,255 @@
...
+
+  token.sep[0]     = hashconfig->separator;
+  token.len_min[0] = 40;
+  token.len_max[0] = 40;
+  token.attr[0]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_HEX;
+
+  token.len_min[1] = 40;
+  token.len_max[1] = 40;
+  token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH;
```
jsteube: I think TOKEN_ATTR_VERIFY_HEX matches here, too?

OpenCL/m27200_a3-pure.cl
```diff
@@ -0,0 +1,159 @@
...
+  for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+  {
+    w[idx] = pws[gid].i[idx];
+  }
+
+  sha1_ctx_t ctx0;
+
+  sha1_init (&ctx0);
+
+  sha1_update_global (&ctx0, dash, 2);
```
jsteube: const u32x dash[16] is not global memory, so this will fail on many runtimes. Just use sha1_update() instead

jsteube:
I can't continue with review because the one with the dash[] I commented is blocking from further review. Please fix it first. Is very likely this kind of error can be found on all the kernel files.

OpenCL/m27200_a0-pure.cl
```diff
@@ -0,0 +1,134 @@
...
+   * base
+   */
+
+  COPY_PW (pws[gid]);
+
+  sha1_ctx_t ctx0;
+
+  sha1_init (&ctx0);
+
+  sha1_update (&ctx0, dash, 2);
```
jsteube: The first entry, where len is guaranteed to by zero, should be initialized with this, since it's static:

```
  ctx0.w0[0] = dash[0];
  ctx0.w0[1] = dash[1];

  ctx.len = 2;
```

OpenCL/m27200_a3-pure.cl
```diff
@@ -0,0 +1,159 @@
...
+    w[idx] = pws[gid].i[idx];
+  }
+
+  sha1_ctx_t ctx0;
+
+  sha1_init (&ctx0);
+
+  sha1_update (&ctx0, dash, 2);
+  sha1_update_global_swap (&ctx0, salt_bufs[SALT_POS].salt_buf, salt_bufs[SALT_POS].salt_len);
+  sha1_update (&ctx0, dash, 2);
```
jsteube: This will break the kernel for compute devices with native vector datatype support, such as CPU. The dash[] is declared as u32x, but applied on a scalar function. So for the outer loop, you need to have another variable like

```
const u32 dash2[16] = { 0x2d2d0000, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0 };
...
sha1_update (&ctx0, dash2, 2);
```

OpenCL/m27200_a3-optimized.cl
```diff
@@ -0,0 +1,828 @@
...
+   * reverse
+   */
+
+  const u32 e_rev = hc_rotl32_S (search[1], 2u);
+
+  /**
+   * salt
+   */
+
+  u32 salt_buf0[4] = { 0x2d2d, 0, 0, 0 };
```
jsteube: I think this should be 0x2d2d0000

src/modules/module_27200.c
```diff
@@ -0,0 +1,256 @@
...
+
+u32 module_pw_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  const bool optimized_kernel = (hashconfig->opti_type & OPTI_TYPE_OPTIMIZED_KERNEL);
+
+  u32 pw_max = PW_MAX;
+
+  if (optimized_kernel == true)
+  {
+    pw_max = 29;
```
jsteube: I think the maximum supported password length in optimized mode will be 9, not 29:

64 - 8 - 1 - 2 - 2 - 40 - 2 = 9

To have longer passwords supported it requires another sha1_transform() call.

tools/test_modules/m27200.pm
```diff
@@ -0,0 +1,44 @@
...
+## Author......: See docs/credits.txt
+## License.....: MIT
+##
+
+use strict;
+use warnings;
+
+use Digest::SHA qw (sha1_hex);
+
+sub module_constraints { [[0, 256], [40, 40], [0, 29], [40, 40], [-1, -1]] }
```
jsteube: Will be 9, not 29, too

OpenCL/m27200_a3-optimized.cl
```diff
@@ -0,0 +1,828 @@
...
+  salt_buf0[2] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 1] >> 16 | salt_bufs[SALT_POS].salt_buf[ 2] << 16);
+  salt_buf0[3] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 2] >> 16 | salt_bufs[SALT_POS].salt_buf[ 3] << 16);
+  salt_buf1[0] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 3] >> 16 | salt_bufs[SALT_POS].salt_buf[ 4] << 16);
+  salt_buf1[1] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 4] >> 16 | salt_bufs[SALT_POS].salt_buf[ 5] << 16);
+  salt_buf1[2] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 5] >> 16 | salt_bufs[SALT_POS].salt_buf[ 6] << 16);
+  salt_buf1[3] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 6] >> 16 | salt_bufs[SALT_POS].salt_buf[ 7] << 16);
+  salt_buf2[0] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 7] >> 16 | salt_bufs[SALT_POS].salt_buf[ 8] << 16);
+  salt_buf2[1] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 8] >> 16 | salt_bufs[SALT_POS].salt_buf[ 9] << 16);
+  salt_buf2[2] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[ 9] >> 16 | 0x2d2d0000);
+  salt_buf2[3] = hc_swap32_S (salt_bufs[SALT_POS].salt_buf[10]);
```
jsteube: I think this is wrong and should be just 0

jsteube:
OK, some progress, only few copy/paste errors and vector datatype problems that all are easily fixed.

Except one major problem. With a fixed salt length 40 plus 6 extra bytes for the dashes, the optimized kernel in it's current form can only support passwords of length 9. That's pretty small. For hashes with 64 byte blocks, you need a second transform after the data length > 55 (64 - 8 - 1). However, you always have to full 64 byte block you need to fill, which makes the logic a bit annoying to implement. So if you decide to continue with optimized kernel and to support > password length 9, then it's probably best to have a u32 tmp[32] buffer instead. It will make the implementation more easy.

OpenCL/m27200_a0-pure.cl
therealartifex: Does this apply to all the pure kernels in both single-hash and multi-hash modes?

jsteube: Exactly

src/modules/module_27200.c
therealartifex: Would that be sha1_transform(), or sha1_transform_vector()? Is there another similar kernel I can examine to get an idea of how this should work?

therealartifex:
> 
> 
> OK, some progress, only few copy/paste errors and vector datatype problems that all are easily fixed.
> 
> Except one major problem. With a fixed salt length 40 plus 6 extra bytes for the dashes, the optimized kernel in it's current form can only support passwords of length 9. That's pretty small. For hashes with 64 byte blocks, you need a second transform after the data length > 55 (64 - 8 - 1). However, you always have to full 64 byte block you need to fill, which makes the logic a bit annoying to implement. So if you decide to continue with optimized kernel and to support > password length 9, then it's probably best to have a u32 tmp[32] buffer instead. It will make the implementation more easy.

I fully intend to support passwords lengths > 9 once I figure out how to use the sha1_transform.

OpenCL/m27200_a3-optimized.cl
jsteube: Problem still exists in _a0 and _a1 optimized

jsteube: I have it fixed for you so that I can merge

jsteube:
Cool. All of the hashing algorithms we deal with they are limited to some block size. Mostly it's 64 byte for the 32 bit data types based and 128 byte for the 64 bit data types based. For SHA1 and many others there's, beside the actual data, they also put the number of input data bytes and the end of the buffer (the last buffer) and that is a requirement. That means, if the buffers size minus the space required to put the length (8 byte on sha1), then there's no other way to follow this requirement than to have another call to sha1_transform(). And you have the stop bit, which is actually a stop byte. 

For example, if you have a data block of 60, you can't reserve 8 byte to set the length. You will need another transform. If your data length % 64 is > 55 (64 - 8 - 1), then you always need another transform. Maybe OpenCL/m04510_a0-optimized.cl is a good reference, but if you search through the kernels with a search string like '>= 56' you will find a bunch.

I'll merge the mode as it is now because it actually works. Feel free to send in another PR with an improved optimized kernel.

PR #2915, opened 2021-07-30
fix buffer sizes for -m 13600 = WinZip

philsmd:
This minor change just corrects the buffer sizes of the WinZip algorithm:
the correct maximum data length is 16 MB in hexadecimal, but "only" 8 MB raw data.

related to : https://github.com/hashcat/hashcat/commit/4730cf6e79709912882636d77e0867b0be954f0e

thank you very much.

PS: I don't think that we need to update `docs/changes.txt` here, since as far as I see the file correctly states that 16 MB long hashes are supported... the only problem here is that the buffer was too large within the code

### DIFF
--- a/OpenCL/m13600-pure.cl
+++ b/OpenCL/m13600-pure.cl
@@ -37,7 +37,7 @@ typedef struct zip2
   u32 verify_bytes;
   u32 compress_length;
   u32 data_len;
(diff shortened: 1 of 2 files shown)


jsteube:
Nice find, many thanks!

PR #2926, opened 2021-08-09
Code cleanup and small fixes

jtojanen:
In is_same_device_type() device name is compared twice, I removed the first one as boolean comparisons are quicker. In run_kernel() inside "if (device_param->is_cuda == true)" we have another "if (device_param->is_cuda == true)" and same repeats with other APIs. Finally hc_cuEventSynchronize() seems to be a leftover from previous merge that I did, and not needed as we only need to synchronize with the end, not with the beginning.

Also could we replace all separate "if (device_param->is_cuda == true)", "if (device_param->is_hip == true)" and "if (device_param->is_opencl == true)" with a if-else if-else (or if-else if-else if ) structure? Is there a reason to this? As device can only be one of these types at any moment.

Added some flushes needed by OpenCL per device.

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -6519,7 +6519,7 @@ int run_cracker (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, co
 
               if (device_param->is_opencl == true)
               {
-                if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, device_param->opencl_d_combs_c, CL_TRUE, 0, innerloop_left * sizeof (pw_t), device_param->combs_buf, 0, NULL, NULL) == -1) return -1;
+                if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, device_param->opencl_d_combs_c, CL_FALSE, 0, innerloop_left * sizeof (pw_t), device_param->combs_buf, 0, NULL, NULL) == -1) return -1;
               }
             }
             else if (user_options->attack_mode == ATTACK_MODE_HYBRID1)
--- a/src/backend.c
+++ b/src/backend.c
@@ -5348,8 +5348,6 @@ int run_kernel (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, con
 
     if (hc_cuEventSynchronize (hashcat_ctx, device_param->cuda_event2) == -1) return -1;
 
-    if (hc_cuEventSynchronize (hashcat_ctx, device_param->cuda_event1) == -1) return -1;
-
     float exec_ms;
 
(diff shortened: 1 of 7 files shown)


jtojanen:
potfile_destroy() leaks filename always, out_buf and tmp_buf if potfile_disable is set to true by modules like 5200 etc..

jtojanen:
Do not open potfile if potfile_disable is true.

src/backend.c
```diff
@@ -161,8 +161,6 @@ static int backend_ctx_find_alias_devices (hashcat_ctx_t *hashcat_ctx)
 
 static bool is_same_device_type (const hc_device_param_t *src, const hc_device_param_t *dst)
 {
-  if (strcmp (src->device_name, dst->device_name) != 0) return false;
```
jsteube, line 164: Please explain the reason to remove this.

jtojanen: You can find exactly the same comparison 6 lines later after equivalent API tests.

jtojanen:
I have totally missed the fact that Hashcat can handle gzip/zip files. But now it seems that there are places where it is assumed that file is plain when it might not be.  Pot file should not be one, but it is technically possible. When hc_fclose checks for NULL, it is safe call without prechecks.

src/filehandling.c
```diff
@@ -596,7 +596,7 @@ void hc_fclose (HCFILE *fp)
 
     unzClose (fp->ufp);
   }
-  else
+  else if (fp->pfp)
```
jsteube, line 607: We should do the same change we do here:

```
@@ -596,7 +596,7 @@ void hc_fclose (HCFILE *fp)
 
     unzClose (fp->ufp);
   }
-  else
+  else if (fp->pfp)
   {
     fclose (fp->pfp);
   }
```

To all functions which follow the same code structure/convention. For instance:

```
@@ -576,7 +576,7 @@ void hc_fflush (HCFILE *fp)
   else if (fp->is_zip)
   {
   }
-  else
+  else if (fp->pfp)
   {
     fflush (fp->pfp);
   }
```

jtojanen: True. I was only looking at these from resource leak point of view.

jsteube: Are you going to fix them or want me to merge it like this?

jtojanen:
Interesting problem AppVeyor as I cannot find problem with build.

jtojanen:
Instead of copying pws_comp and pws_idx in rebuild_pws_compressed_append(), we can simply swap pointers and free old buffers. I came across this when I was looking at pinned host memory usage. I dropped is_gzip and is_zip, and used handles instead. I did a small refactoring in hc_fopen() and hc_fopen_raw() to make sure structure is always cleared before use. In old implementation zip file left fp->fd handle open. Now all file handling functions check if we have valid gzip, zip or file handle before calling. It should be safe to call any without preconditions. Good to go!

jsteube:
Looks good, many thanks!

Issue #2927, opened 2021-08-10
Metamask (26600)  Token length exception


ghost:
Try to submit hash 1532 word lenght, hashcat error 
`Hashfile 'C:\hashcat\task.txt' on line 1 ($metam...***): Token length exception
No hashes loaded.`

ghost:
If the metamask vault contain more than 5 wallets, it has long hash, hashcat can`t read it

Hashfile 'C:\work\hashcat\task.txt' on line 1 ($metam...==): Token length exception
No hashes loaded.

jsteube:
@kipilk0 Did you try using @matrix new extraction tool?

matrix:
@kipilk0 please include files that can be used to test

ghost:
> 
> 
> @kipilk0 Did you try using @matrix new extraction tool?

Yes.

> 
> 
> @kipilk0 please include files that can be used to test

Sorry, i can`t send you hash due , because it have private data.
$metamask$ hash has 1300+ symbols length

123Serg:
@matrix I have same error, some of my hashes have big length, can you increase hash length limit to 2000-3000 sybmols please?

ghost:
Metamask hash contain sensetive info , you have private e-mail where i can send hash ?

ghost:
Hashes >1000 length with error
https://i.imgur.com/RgUldAK.png

matrix:
Hi, could you please try this version? I doubled the length of the ciphertext

https://github.com/matrix/hashcat/tree/metamask_big

Let us know, thanks

jsteube:
@matrix Why no PR? Are you unsure if it will still work?

matrix:
> @matrix Why no PR? Are you unsure if it will still work?

I'm not sure if it's enough to double up. Maybe we need to triple the length, let's see what they say :)

its5Q:
> > @matrix Why no PR? Are you unsure if it will still work?
> 
> I'm not sure if it's enough to double up. Maybe we need to triple the length, let's see what they say :)

I've tested bruteforcing a wallet with 10 accounts with your branch, and the hash was pretty long, and it worked, so I think the length is good enough.

matrix:
> > > @matrix Why no PR? Are you unsure if it will still work?
> > 
> > 
> > I'm not sure if it's enough to double up. Maybe we need to triple the length, let's see what they say :)
> 
[...]

Issue #2818, opened 2021-06-10
new algorithm: metamask vault


123Serg:
Hello! Is this possible to add bruteforce on Hashcat for Metamask vault? Metamask store encrypted seed phrase (12 words) in json file, example: 

{data:wM+CSmCIIlaofXHPxCRFqbcPvEDZESBYM1XdAE8VFanedCIXvDEEt15nhzdv1pqzkW0MCQDBA5T1W1/Ltg7lKBszkoJJ1PtVHOpTn/afhvLyIp2VBBV1Tuj4s8LGo2GM9KOHn1Hn5BC1YhjP56opGrJlsIT9qnuGYHT8EtA5IiHP4zB4fwgGqDhtX+QUJi4rdm+apXZ1DpRvDq5J+bCca2GHXzU5H57Esn3THPXev3RdlkmjhuXVjV8=,iv:ejhZW94EsJZ049Sb/at0Xg==,salt:jgW3gi0eyPtXXguqMzi6uGyuR6m+lgQXw8onqivK5w0=}

Here is encrypted data, iv and salt.
Algorithm: PBKDF2-HMAC-SHA256 (10 000 iterations) + AES-GCM.

Wallet: https://metamask.io/
Official decryptor: https://github.com/MetaMask/vault-decryptor
Unofficial decryptor: https://github.com/miguelmota/metamask-vault-decrypter

P.S. if you'll want to test encrypted data that I provided, password is: mysecretpassword
P.S.S maybe there existing algorithm that I can use with some changes to decrypt metamask data until Metamask decription function is added?

philsmd:
do all metamask versions use this algorithm ?

can you give some details about when (which version etc) this algorithm was introduced ? Were there some variations in salt / iv length or restrictions in the password length / keyspace etc

123Serg:
> 
> 
> do all metamask versions use this algorithm ?
> 
> can you give some details about when (which version etc) this algorithm was introduced ? Were there some variations in salt / iv length or restrictions in the password length / keyspace etc

Metamask version: 9.2.0
I don't know for sure, but I think all metamask versions use this algorithm. 
There is password restriction: minimal length is 8 chars. 
Length of salt is 32 bytes, iv is 16 bytes, I think there no other variations of them.
Data that I provided for test is encdoded by base64.

philsmd:
[...]

PR #2930, opened 2021-08-12
add metamask extraction tool

matrix:
#2818 #2927


WiFik18 (commented):
C:/python/python.exe c:/hashcat-metamask_big/tools/metamask2hashcat.py
usage: metamask2hashcat.py [-h] --vault VAULT
metamask2hashcat.py: error: the following arguments are required: --vault

PR #2935, opened 2021-08-13
workaround to 'clEnqueueWriteBuffer(): CL_INVALID_VALUE' with apple gpu

matrix:
Hi,

the latest engine changes have broken gpu support on apple.

```
$ ./hashcat -b -m 0 -D2
hashcat (v6.2.3-220-gbeaa4e0da) starting in benchmark mode

* Device #2: Apple's OpenCL drivers (GPU) are known to be unreliable.
             You have been warned.

* Device #2: Iris, 1472/1536 MB (384 MB allocatable), 40MCU

Benchmark relevant options:
===========================
* --opencl-device-types=2
* --optimized-kernel-enable

Hashmode: 0 - MD5

clEnqueueWriteBuffer(): CL_INVALID_VALUE

Started: Fri Aug 13 15:59:40 2021
Stopped: Fri Aug 13 15:59:45 2021
```

@jtojanen please check this patch

Thanks

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -5156,6 +5156,15 @@ int run_opencl_kernel_bzero (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *devi
   const u64 num16d = size / 16;
   const u64 num16m = size % 16;
 
+  // with apple GPU clEnqueueWriteBuffer() return CL_INVALID_VALUE, workaround
+
+  if (device_param->opencl_platform_vendor_id == VENDOR_ID_APPLE && \
+     (device_param->opencl_device_vendor_id == VENDOR_ID_INTEL_SDK || device_param->opencl_device_vendor_id == VENDOR_ID_APPLE) && \
+     device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
+  {
+    return run_opencl_kernel_memset (hashcat_ctx, device_param, buf, 0, 0, size);
+  }
+
   if (num16d)
   {
     const u64 kernel_threads = device_param->kernel_wgs_bzero;
@@ -5175,23 +5184,7 @@ int run_opencl_kernel_bzero (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *devi
 
   if (num16m)
   {
-    // with apple GPU clEnqueueWriteBuffer() return CL_INVALID_VALUE
-    // following a workaround
-
-    if (device_param->opencl_platform_vendor_id == VENDOR_ID_APPLE && \
-       (device_param->opencl_device_vendor_id == VENDOR_ID_INTEL_SDK || device_param->opencl_device_vendor_id == VENDOR_ID_APPLE) && \
-       device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
-    {
-      u8 *bzeros_apple = (u8 *) hccalloc (num16m, sizeof (u8));
-
-      if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_TRUE, num16d * 16, num16m, bzeros_apple, 0, NULL, NULL) == -1) return -1;
-
-      hcfree (bzeros_apple);
-    }
-    else
-    {
-      if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_FALSE, num16d * 16, num16m, bzeros, 0, NULL, NULL) == -1) return -1;
-    }
+    if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_FALSE, num16d * 16, num16m, bzeros, 0, NULL, NULL) == -1) return -1;
   }
 
   return 0;
--- a/src/backend.c
+++ b/src/backend.c
@@ -5175,7 +5175,23 @@ int run_opencl_kernel_bzero (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *devi
 
   if (num16m)
   {
-    if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_FALSE, num16d * 16, num16m, bzeros, 0, NULL, NULL) == -1) return -1;
+    // with apple GPU clEnqueueWriteBuffer() return CL_INVALID_VALUE
+    // following a workaround
+
+    if (device_param->opencl_platform_vendor_id == VENDOR_ID_APPLE && \
+       (device_param->opencl_device_vendor_id == VENDOR_ID_INTEL_SDK || device_param->opencl_device_vendor_id == VENDOR_ID_APPLE) && \
+       device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
+    {
+      u8 *bzeros_apple = (u8 *) hccalloc (num16m, sizeof (u8));
+
+      if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_TRUE, num16d * 16, num16m, bzeros_apple, 0, NULL, NULL) == -1) return -1;
+
+      hcfree (bzeros_apple);
+    }
+    else
+    {
+      if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_FALSE, num16d * 16, num16m, bzeros, 0, NULL, NULL) == -1) return -1;
+    }
   }
 
   return 0;


jsteube:
I can't reproduce on AMD GPU. Probably Intel GPU only?

```
$ ./hashcat -b -m 0 -D2
hashcat (v6.2.3-220-gbeaa4e0da) starting in benchmark mode

* Device #2: Apple's OpenCL drivers (GPU) are known to be unreliable.
             You have been warned.

OpenCL API (OpenCL 1.2 (Feb 28 2021 03:51:16)) - Platform #1 [Apple]
* Device #1: Intel(R) Xeon(R) W-3223 CPU @ 3.50GHz, skipped
* Device #2: AMD Radeon Pro W5700X Compute Engine, 16304/16368 MB (4092 MB allocatable), 40MCU

Benchmark relevant options:
===========================
* --opencl-device-types=2
* --optimized-kernel-enable

Hashmode: 0 - MD5

Speed.#2.........: 23302.5 MH/s (56.94ms) @ Accel:512 Loops:1024 Thr:64 Vec:1

Started: Fri Aug 13 16:16:47 2021
Stopped: Fri Aug 13 16:16:54 2021
```

jtojanen:
You could try if calling run_opencl_kernel_memset() works with Apple:
```
if (device_param->opencl_platform_vendor_id == VENDOR_ID_APPLE && \
       (device_param->opencl_device_vendor_id == VENDOR_ID_INTEL_SDK || device_param->opencl_device_vendor_id == VENDOR_ID_APPLE) && \
       device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
    {

return run_opencl_kernel_memset(hashcat_ctx, device_param, buf, 0, 0, size);
}

if (num16d)
{
```

matrix:
> I can't reproduce on AMD GPU. Probably Intel GPU only?
> 
> ```
> $ ./hashcat -b -m 0 -D2
> hashcat (v6.2.3-220-gbeaa4e0da) starting in benchmark mode
> 
> Benchmarking uses hand-optimized kernel code by default.
> You can use it in your cracking session by setting the -O option.
> Note: Using optimized kernel code limits the maximum supported password length.
> To disable the optimized kernel code in benchmark mode, use the -w option.
> 
> * Device #2: Apple's OpenCL drivers (GPU) are known to be unreliable.
>              You have been warned.
> 
> OpenCL API (OpenCL 1.2 (Feb 28 2021 03:51:16)) - Platform #1 [Apple]
> ====================================================================
> * Device #1: Intel(R) Xeon(R) W-3223 CPU @ 3.50GHz, skipped
> * Device #2: AMD Radeon Pro W5700X Compute Engine, 16304/16368 MB (4092 MB allocatable), 40MCU
> 
> Benchmark relevant options:
> ===========================
> * --opencl-device-types=2
> * --optimized-kernel-enable
> 
> Hashmode: 0 - MD5
> 
> Speed.#2.........: 23302.5 MH/s (56.94ms) @ Accel:512 Loops:1024 Thr:64 Vec:1
> 
> Started: Fri Aug 13 16:16:47 2021
> Stopped: Fri Aug 13 16:16:54 2021
> ```

probably yes :)

matrix:
> You could try if calling run_opencl_kernel_memset() works with Apple:
> 
> ```
> if (device_param->opencl_platform_vendor_id == VENDOR_ID_APPLE && \
>        (device_param->opencl_device_vendor_id == VENDOR_ID_INTEL_SDK || device_param->opencl_device_vendor_id == VENDOR_ID_APPLE) && \
>        device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
>     {
> 
> return run_opencl_kernel_memset(hashcat_ctx, device_param, buf, 0, 0, size);
> }
> 
> if (num16d)
> {
> ```

Works :) @jsteube if ok I will push this

jsteube:
Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?

matrix:
> Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?

```
HC_ALIGN(16)
static const u32 bzeros[4] = { 0, 0, 0, 0 };
```

Shortly, with current version, I see too more bytes written to "buf" in case of "num16m" is not 0.
All hc_clEnqueueWriteBuffer() call with size < 16 cause this error (only with Apple gpu), like these

https://github.com/hashcat/hashcat/blob/master/src/backend.c#L14372-L14373

jtojanen:
> Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?

Agree. clEnqueueWriteBuffer() is used in many places. I have no access to that type Apple, I would assume that one of parameters is zero or out-of-range as the error is CL_INVALID_VALUE. 

> CL_INVALID_VALUE if the region being written specified by (offset, size) is out of bounds or if ptr is a NULL value or if size is 0

jtojanen:
> > Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?
> 
> ```
> HC_ALIGN(16)
> static const u32 bzeros[4] = { 0, 0, 0, 0 };
> ```
> 
> Shortly, with current version, I see too more bytes written to "buf" in case of "num16m" is not 0.
> All hc_clEnqueueWriteBuffer() call with size < 16 cause this error (only with Apple gpu), like these
> 
> https://github.com/hashcat/hashcat/blob/master/src/backend.c#L14372-L14373

This is one of the clEnqueueWriteBuffer() calls to have non-zero offset. We know that 1 <= num16m <= 15, buffer must be valid and it's size must be equal to 'size' argument. bzeros buffer is valid and contains 16 zero bytes.

What happens if you make it blocking call? By switching CL_FALSE to CL_TRUE in the original clEnqueueWriteBuffer()

matrix:
> > > Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?
> > 
> > 
> > ```
> > HC_ALIGN(16)
> > static const u32 bzeros[4] = { 0, 0, 0, 0 };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Shortly, with current version, I see too more bytes written to "buf" in case of "num16m" is not 0.
> > All hc_clEnqueueWriteBuffer() call with size < 16 cause this error (only with Apple gpu), like these
> > https://github.com/hashcat/hashcat/blob/master/src/backend.c#L14372-L14373
> 
> This is one of the clEnqueueWriteBuffer() calls to have non-zero offset. We know that 1 <= num16m <= 15, buffer must be valid and it's size must be equal to 'size' argument. bzeros buffer is valid and contains 16 zero bytes.
> 
> What happens if you make it blocking call? By switching CL_FALSE to CL_TRUE in the original clEnqueueWriteBuffer()

I've already tried, it wasn't enough

jtojanen:
> > > > Typically when it comes to workarounds it makes sense to check all other code if the same factors exists that could break hashcat the same way and fix them at the same time. Do we have other spots? And what do you believe is the root cause of the problem?
> > > 
> > > 
> > > ```
> > > HC_ALIGN(16)
> > > static const u32 bzeros[4] = { 0, 0, 0, 0 };
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Shortly, with current version, I see too more bytes written to "buf" in case of "num16m" is not 0.
> > > All hc_clEnqueueWriteBuffer() call with size < 16 cause this error (only with Apple gpu), like these
> > > https://github.com/hashcat/hashcat/blob/master/src/backend.c#L14372-L14373
> > 
> > 
> > This is one of the clEnqueueWriteBuffer() calls to have non-zero offset. We know that 1 <= num16m <= 15, buffer must be valid and it's size must be equal to 'size' argument. bzeros buffer is valid and contains 16 zero bytes.
> > What happens if you make it blocking call? By switching CL_FALSE to CL_TRUE in the original clEnqueueWriteBuffer()
> 
> I've already tried, it wasn't enough

Ok, thanks! We know that the workaround to call run_opencl_kernel_memset() works. Repeat the workaround, but modify `if (ocl->clEnqueueFillBuffer == NULL)` to `if (ocl->clEnqueueFillBuffer != NULL)` in run_opencl_kernel_memset(). I have a feeling..

matrix:
Works. What's the feeling?

jtojanen:
Could you try this? Performance is bad, but I need to know why we get CL_INVALID_VALUE. I found an old discussion where Intel employee didn't understand in-order queue, and for that reason I have added extra event to force the execution order. Also clEnqueueWriteBuffer() is blocking and we are using stack temp buffer for zeros.

```
int run_opencl_kernel_bzero (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, cl_mem buf, const u64 size)
{
  const u64 num16d = size / 16;
  const u64 num16m = size % 16;

  const u32 tmp[4] = { 0 };

  if (num16d)
  {
    const u64 kernel_threads = device_param->kernel_wgs_bzero;

    u64 num_elements = round_up_multiple_64(num16d, kernel_threads);

    cl_kernel kernel = device_param->opencl_kernel_bzero;

    if (hc_clSetKernelArg (hashcat_ctx, kernel, 0, sizeof(cl_mem),   (void *) &buf)    == -1) return -1;
    if (hc_clSetKernelArg (hashcat_ctx, kernel, 1, sizeof(cl_ulong), (void *) &num16d) == -1) return -1;

    const size_t global_work_size[3] = { num_elements,   1, 1 };
    const size_t local_work_size[3]  = { kernel_threads, 1, 1 };

    cl_event opencl_event;

    if (hc_clEnqueueNDRangeKernel (hashcat_ctx, device_param->opencl_command_queue, kernel, 1, NULL, global_work_size, local_work_size, 0, NULL, &opencl_event) == -1) return -1;

    if (hc_clFlush (hashcat_ctx, device_param->opencl_command_queue) == -1) return -1;

    if (num16m)
    {
      if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_TRUE, num16d * 16, num16m, tmp, 1, &opencl_event, NULL) == -1) return -1;
    }

    if (hc_clReleaseEvent (hashcat_ctx, opencl_event) == -1) return -1;
  }
  else if (num16m)
  {
    if (hc_clEnqueueWriteBuffer (hashcat_ctx, device_param->opencl_command_queue, buf, CL_TRUE, num16d * 16, num16m, tmp, 0, NULL, NULL) == -1) return -1;
  }

  return 0;
}
```

matrix:
I think I've already explained why it probably happens, however here is the output:

```
Hashmode: 0 - MD5

clEnqueueWriteBuffer(): CL_INVALID_VALUE
```

jtojanen:
> I think I've already explained why it probably happens, however here is the output:
> 
> ```
> Hashmode: 0 - MD5
> 
> clEnqueueWriteBuffer(): CL_INVALID_VALUE
> ```

Ok, good. I agree, we should go with the workaround. If we encounter another Apple OpenCL + Iris related bugs then we should revisit this.

PR #2937, opened 2021-08-15
Use pinned host buffers for device <-> host transfers

jtojanen:
Bandwidth is about double, and we avoid paging and/or memory copy overhead from staging buffer. Also we avoid allocating temporary buffers and runtime memory usage is more static. Now only gidd_to_pw_t() has couple of unpinned buffers.  

Next step is to use pinned buffers for host->device transfers which will have more impact of performance because they make up the majority. Already device_param->hooks_buf shows that it will be easy to implement.

CUDA uses hc_cuMemHostAlloc()/hc_cuMemFreeHost() and HIP uses hc_hipHostMalloc()/hc_hipHostFree() to allocate paged memory buffers. But OpenCL is different. hc_clCreateBuffer(CL_MEM_ALLOC_HOST_PTR)+hc_clEnqueueMapBuffer() to allocate and map paged memory once, and hc_clEnqueueUnmapMemObject()+hc_clReleaseMemObject() to free.

### DIFF
--- a/include/backend.h
+++ b/include/backend.h
@@ -73,6 +73,8 @@ int hc_cuMemcpyDtoDAsync         (hashcat_ctx_t *hashcat_ctx, CUdeviceptr dstDev
 int hc_cuMemcpyDtoHAsync         (hashcat_ctx_t *hashcat_ctx, void *dstHost, CUdeviceptr srcDevice, size_t ByteCount, CUstream hStream);
 int hc_cuMemcpyHtoDAsync         (hashcat_ctx_t *hashcat_ctx, CUdeviceptr dstDevice, const void *srcHost, size_t ByteCount, CUstream hStream);
 int hc_cuMemFree                 (hashcat_ctx_t *hashcat_ctx, CUdeviceptr dptr);
+int hc_cuMemFreeHost             (hashcat_ctx_t *hashcat_ctx, void *p);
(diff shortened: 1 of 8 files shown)


jsteube:
The same as with the async execution, older hashcat had pinned memory, too. I decided to drop it for a) more easy code and b) Many OpenCL runtimes are so unreliable and error prone the more strict we stick to the most common function the less is the chance of triggering a bug inside the runtime. That compared to the fact that I didn't see any performance advantage in a real hash modes. Sure we can go back to this model now that many OpenCL runtimes have improved in stability, but still I don't see any performance advantage. Did you see any? which mode?

jtojanen:
> The same as with the async execution, older hashcat had pinned memory, too. I decided to drop it for a) more easy code and b) Many OpenCL runtimes are so unreliable and error prone the more strict we stick to the most common function the less is the chance of triggering a bug inside the runtime. That compared to the fact that I didn't see any performance advantage in a real hash modes. Sure we can go back to this model now that many OpenCL runtimes have improved in stability, but still I don't see any performance advantage. Did you see any? which mode?

I am not only interested in OpenCL performance, but also CUDA and HIP in future when it's stable. There is no or only very small performance advantage when we are transfering few bytes back to host. I will report back after converting host->device transfers, converted to draft now.

PR #2943, opened 2021-08-21
Avoid directly accessing file handle

jtojanen:
I think we should avoid directly accessing file handle. Instead we should use accessors, for example we could use native Windows file APIs to access files. Dropped hc_fileno(). Implemented hc_fstat() and hc_fsync(). Use fd directly in locking. Call close() for fd in hc_fclose(). Fixed a bug caused by #2926.


jsteube:
I am wondering if we should call fsync also for potfile and outfile writes.

PR #2956, opened 2021-09-01
Set GZIP internal buffer size to 256k (default 8k)

jtojanen:
..to increase the speed of decompression. After benchmarking and testing I choose 256k as buffer size. This will also be the xz stream buffer size, so I defined/renamed it as HCFILE_BUFFER_SIZE.


jsteube:
@jtojanen please resolve conflict

PR #2960, opened 2021-09-02
Refactor file reading/writing

jtojanen:
There are many known limits and problems with Windows file reading and writing. Old implementation had a 3 GB size limit and it failed to avoid known problems. To avoid most of known problems I added 4 MB chunked reading or writing and limited 32 bit builds to total block size of 2 GB. Bigger request would be a indicator of problem elsewhere. New implementation removes the maximum size limit.

Function count_lines() reads file in HCBUFSIZ_LARGE chunks, but we should use HCFILE_CHUNK_SIZE to avoid use of workaround section. 4 MB vs 16 MB doesn't make any difference in performance with other systems. I didn't make this change yet so we can easily test the workaround.

Also I changed the hc_fread and hc_fwrite parameter tests. If buffer is NULL, it's an error.

### DIFF
--- a/src/filehandling.c
+++ b/src/filehandling.c
@@ -412,10 +412,25 @@ size_t hc_fread (void *ptr, size_t size, size_t nmemb, HCFILE *fp)
   else if (fp->ufp)
   {
     u64 len = (u64) size * nmemb;
-    if (len == (unsigned) len)
+    u64 pos = 0;
+
+    /* assume success */
+    n = nmemb;
+
+    do
     {
-      n = unzReadCurrentFile (fp->ufp, ptr, (unsigned) len);
-    }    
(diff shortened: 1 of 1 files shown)


jtojanen:
It seems that zip file reported bytes instead of elements as it should have, we had forgotten to divide by element size. We also ignored the possible errors here. unzReadCurrentFile() takes `unsigned len` but returns `int`. So we should be limiting request length to INT_MAX chunks. But to avoid the same problems with Windows I limited chunks to HCFILE_CHUNK_SIZE, 4 MB.

jtojanen:
Xz file also reported bytes instead of elements. Now hc_fread() returns `(size_t) -1`  if an error occurs and we should always check that or we will have buffer overrun. Or we could return zero..

jtojanen:
unzip is internally using UNZ_BUFSIZE read buffer, which is 16K by default. Others have increased that to 64K or 256K. I will test if that improves speed, but we don't need chunk size limit.

PR #2964, opened 2021-09-06
Fix fgetl() logic

jtojanen:
Old implementation has some problems. If line is too long and it gets truncated, fgetl() reads until EOF, not until newline. If line length is zero, it's not NULL terminated. And as it's impossible to have extra newline characters in the end, trim only carriage return(s).

New implementation fixes all these and if line gets truncated, we don't look for any '\n' or '\r' from the end of line. We always NULL terminate. 

Also there is no need to check feof() as fgetc() reports EOF. And as the return value is used instead of strlen(), check for NULL terminators.

### DIFF
--- a/src/filehandling.c
+++ b/src/filehandling.c
@@ -982,54 +982,38 @@ void hc_fclose (HCFILE *fp)
 
 size_t fgetl (HCFILE *fp, char *line_buf, const size_t line_sz)
 {
-  size_t line_truncated = 0;
+  int c;
 
   size_t line_len = 0;
 
-  while (!hc_feof (fp))
-  {
-    const int c = hc_fgetc (fp);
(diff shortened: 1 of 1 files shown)


jtojanen:
My last commit where fgetl() interrupts on NULL terminator might cause problem, but without it we must be clear that the fgetl() return value is the length of line, not the strlen() of line. Of course we can have two versions of this function.

In most calls to fgetl() we have too large buffer, mostly heap buffer with size of HCBUFSIZ_LARGE, which is 16M. I think stack buffer with size of HCBUFSIZ_TINY, whick is 4K should be sufficient as there are other limits to actual line length like PW_MAX and RP_PASSWORD_SIZE etc..

jtojanen:
I am closing as this got too messy.

PR #2970, opened 2021-09-08
Avoid false positives in pkzip formats

magnumripper:
After inflating data decrypted with an incorrect guess, you may get Z_STREAM_END (weakly indicating a correct guess) although the output length doesn't match the expected.  If the CRC-32 randomly match as well, we had a false positive before these tests were added.

Crazy yes, but I did see it happen!


jsteube:
Cool, thanks!

PR #2981, opened 2021-09-13
Fix fgetl() logic

jtojanen:
Second attempt, this is just a bugfix. 

Old implementation has some problems. If line is too long and it gets truncated, fgetl() reads until EOF, not until newline. If line length is zero, it's not NULL terminated. And as it's impossible to have extra newline characters in the end, trim only carriage returns. New implementation fixes all these and if line gets truncated, we don't look for any '\n' or '\r' from the end of line. We always NULL terminate. Also there is no need to check feof() as fgetc() reports EOF. And as the return value is used instead of strlen(), check for NULL terminators.

Some might argue that we should only trim single carriage return '\r', not all as extras might be part of actual line.

### DIFF
--- a/src/filehandling.c
+++ b/src/filehandling.c
@@ -982,54 +982,38 @@ void hc_fclose (HCFILE *fp)
 
 size_t fgetl (HCFILE *fp, char *line_buf, const size_t line_sz)
 {
-  size_t line_truncated = 0;
+  int c;
 
   size_t line_len = 0;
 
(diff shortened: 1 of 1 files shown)


jsteube:
> Some might argue that we should only trim single carriage return '\r', not all as extras might be part of actual line.

It's fine, in_superchop() works the same way. So far, no one complained. Many thanks for contribution!

PR #2999, opened 2021-09-27
Add hash-mode MurmurHash3

Slattz:
This pull request adds a new hash mode, `27800`, for cracking MurmurHash3 hashes.
It implements an optimised kernel for attack modes 0, 1 and 3.

An example use-case for this module is cracking the hashes of Yaml field names that Nintendo uses within Animal Crossing: New Horizons.

Let me know if any changes are required! :)


jsteube:
Many thanks! Ideally there's also the unit test so that we can test it under the different scenarios, especially with vector datatypes and such, but it's not a requirement.

PR #3017, opened 2021-11-02
Add Signature matching for mac2john format

Chick3nman:
Adds signature match for mac2john's $pbkdf2-hmac-sha512$ hash format, which currently function identically in the kernel. The extra fields have variable lengths and the maximum is not well defined so the limits were chosen arbitrarily and can be raised in the future without issue.

### DIFF
--- a/src/modules/module_07100.c
+++ b/src/modules/module_07100.c
@@ -61,6 +61,8 @@ typedef struct pbkdf2_sha512_tmp
 
 static const char *SIGNATURE_SHA512MACOS = "$ml$";
 
+static const char *SIGNATURE_SHA512MACOS_JOHN = "$pbkdf2-hmac-sha512$";
+
 u64 module_tmp_size (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
 {
   const u64 tmp_size = (const u64) sizeof (pbkdf2_sha512_tmp_t);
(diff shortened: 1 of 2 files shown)


src/modules/module_07100.c
```diff
@@ -97,89 +99,199 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
...
+
+    const int rc_tokenizer = input_tokenizer ((const u8 *) line_buf, line_len, &token);
+
+    if (rc_tokenizer != PARSER_OK) return (rc_tokenizer);
+
+    const int hash_len = token.len[3];
+
+    if ((hash_len != 128) && (hash_len != 256)) return (PARSER_HASH_LENGTH);
+
+    const u8 *hash_pos = token.buf[3];
```
jsteube, line 144: Is there a reason, from  this point in the code, to not continue outside the branches? They seem equal in both branches.

jsteube:
I think there's one thing missing, and that is that hashcat is supposed to output the same hash as it was loading. But with this change, the output format will always be the old format and only the old format. There needs to be added a flag which you can set from the decoder which format was used, and then read in the encoder to decide which output format to use.

src/modules/module_07100.c
Chick3nman: I ran into an issue with referencing token.buf outside of the branches that was resolved by moving the code into the branches. There may be a better way, this simply resolved the issue and worked so I did not try it a different way.

Chick3nman:
The module as it is above will output both original formatted hashes AND the new format because module_hash_encode simply does `  const int line_len = snprintf (line_buf, line_size, "%s", hash_info->orighash);`. When cracking I received the correct hash format for the input hash.

src/modules/module_07100.c
jsteube: OK, that's good enough.

PR #3022, opened 2021-11-06
Update combinator.rule to use an up-to-date year suffix

nbeguier:
I've noticed that these year suffix: 2007,2008 and 2009 are not relevant today. Maybe use year from 2017 to 2021 is better.
This is my small contribution to this great project.


jsteube:
Thanks :)

PR #3062, opened 2021-12-17
Updated Apple Silicon support

matrix:
Hi,

the following patch introduces two changes:
- with Apple M1*, optimized kernel doesn't work due to unknown OpenCL bug. Pure kernels will be forced. Using Optimized kernels, hashcat execution is blocked when the clCreateKernel () function is called at this location: https://github.com/hashcat/hashcat/blob/50e28ff3064db43eb3b10490974ec1e9c4910745/src/backend.c#L13888-L13894

- token_t has been renamed to hc_token_t, due to the following error:

Screen Shot 2021-12-17 at 01 57 40

Following the hashcat output with Apple M1 (using: ./hashcat -m 0 -t 32 -a 7 example0.hash ?a?a?a?a example.dict -O)

Screen Shot 2021-12-17 at 02 04 16 1

ends with

Screen Shot 2021-12-17 at 02 06 29

Thanks :)

PR #3066, opened 2021-12-17
Renamed struct token_t to hc_token_t, due to typedef redefinition with different types on Apple

matrix:
info in #3062


jsteube:
Double PR, as discussed in https://github.com/hashcat/hashcat/pull/3062

Issue #3070, opened 2021-12-17
Change position of Platform number in benchmark test result


Bogdan107:
**Describe the issue**
Platform number placed at the end of line.
I want to place platform number at start of line.
This allows to quick visual navigation of platform number in the environments, where many platform drivers installed.

**To Reproduce**
Example of output, wehere platform number is placed at the end of line (standard for hascat 6.2.5):
```
...
CUDA API (CUDA 11.4)
====================
* Device #1: NVIDIA GeForce GTX 1650, skipped

OpenCL API (OpenCL 1.2 LINUX) - Platform #1 [Intel(R) Corporation]
==================================================================
* Device #2: AMD Ryzen 5 4600H with Radeon Graphics, 3527/7119 MB (889 MB allocatable), 12MCU

OpenCL API (OpenCL 3.0 CUDA 11.4.158) - Platform #2 [NVIDIA Corporation]
========================================================================
* Device #3: NVIDIA GeForce GTX 1650, skipped

OpenCL API (OpenCL 2.0 pocl 1.8  Linux, Gentoo+Asserts, RELOC, SPIR, LLVM 13.0.0, SLEEF, CUDA, LTTNG) - Platform #3 [The pocl project]
======================================================================================================================================
* Device #4: pthread-AMD Ryzen 5 4600H with Radeon Graphics, skipped
* Device #5: NVIDIA GeForce GTX 1650, skipped

Benchmark relevant options:
...
```
**Expected behavior**
Example of output, wehere platform number is placed at the start of line (my proposal variant):
```
...
CUDA API (CUDA 11.4)
====================
  * Device #1: NVIDIA GeForce GTX 1650, skipped

Platform #1: OpenCL API (OpenCL 1.2 LINUX) -- [Intel(R) Corporation]
==================================================================
  * Device #2: AMD Ryzen 5 4600H with Radeon Graphics, 3527/7119 MB (889 MB allocatable), 12MCU

Platform #2: OpenCL API (OpenCL 3.0 CUDA 11.4.158) -- [NVIDIA Corporation]
========================================================================
  * Device #3: NVIDIA GeForce GTX 1650, skipped

Platform #3: OpenCL API (OpenCL 2.0 pocl 1.8  Linux, Gentoo+Asserts, RELOC, SPIR, LLVM 13.0.0, SLEEF, CUDA, LTTNG) -- [The pocl project]
======================================================================================================================================
[...]

PR #3074, opened 2021-12-22
Fix for #3070

b0lek:
Changed format of the output line in list of devices in `backend_info_compact` for OpenCL.
Platform number moved to front of the line for clarity. 
For me it makes sense to have Platform # at the beginning of the line.


jsteube:
Not sure if this makes sense. Platforms only exists below the OpenCL API. CUDA and HIP doesn't have something they call a Platform. For that reason the API type should be listed first.

b0lek:
I don't have strong opinion so I close this PR.

PR #3076, opened 2021-12-24
OpenCL Backend: added workaround to make optimized kernels work on Apple Silicon

matrix:
I hope it's the final version :)


jsteube:
Great work, thanks!

PR #3078, opened 2021-12-24
Create LICENSE

roycewilliams:
Just a suggestion - presence of correct LICENSE will assist other projects with license dependency-graph automation, etc.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2021 hashcat
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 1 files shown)


Chick3nman:
I think the LICENSE file should mirror exactly the license we have at https://github.com/hashcat/hashcat/blob/master/docs/license.txt but otherwise, good idea.

jsteube:
I also like the idea but I don't like the idea to have the same content in two files. Is there maybe a third alternative, maybe by using some GitHub internal configurations?

b0lek:
Wouldn't it work with just symbolic link LICENSE to docs/license.txt ?

roycewilliams:
The best compromise may be something _somewhat_ like this:

https://softwareengineering.stackexchange.com/a/234526/54447

* make LICENSE.TXT strictly match MIT
* move the third-party language to a LICENSE-3RD-PARTY.txt or something similar

jsteube:
I don't think we're the first to follow a docs / folder logic. How do other projects deal with using a docs/ folder logic? Moving the license.txt to the root folder would break this convention and I'm not a huge fan of it just because GitHub prefers it. There is no GitHub specific option where we can specify the (custom) path of LICENSE.TXT?

roycewilliams:
The GitHub doc [here](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license) describes how they autodetect license informaiton, and recommends:

_If your repository is using a license that is listed on the Choose a License website and it's not displaying clearly at the top of the repository page, it may contain multiple licenses or other complexity. To have your license detected, simplify your LICENSE file and note the complexity somewhere else, such as your repository's README file._

jsteube:
I tried that, but it will just create a new file in the root folder, which is not what we want.

roycewilliams:
Yeah, I think there's no way to get the automatic detection without having the file in the top level. We can dig this up later if a better way emerges. Closing for now.

PR #3103, opened 2022-01-03
Removed hc_clSetKernelArg() call from backend_session_update_mp_rl()

matrix:
I think these hc_clSetKernelArg() calls are redundant (and also wrong), because of this:

https://github.com/hashcat/hashcat/blob/master/src/backend.c#L3756-L3780


jsteube:
I think they are not, but because of code change over time, the link you posted is not valid anymore. Could you please update the line number, then I can review again.

matrix:
> I think they are not, but because of code change over time, the link you posted is not valid anymore. Could you please update the line number, then I can review again.

look the patch, it is still valid :)

PR #3104, opened 2022-01-03
Removed hc_clSetKernelArg() call from backend_session_update_mp()

matrix:
same logic of #3103


jsteube:
I think they are not, but because of code change over time, the link you posted is not valid anymore. Could you please update the line number, then I can review again.

matrix:
look the patch, it is still valid :)

jsteube:
I think we need this when used with mask files.

PR #3108, opened 2022-01-04
Renamed GID_MAX to GID_CNT, due to macro redefined warning on MacOS

matrix:
GID_MAX is used on MacOS

Screen Shot 2022-01-04 at 20 45 05


jsteube:
This PR is obsolete

PR #3109, opened 2022-01-05
Fix the build for NetBSD

khorben:
This commit lets me build hashcat on NetBSD. I believe it should not harm support for other systems.

With these changes, starting hashcat still fails with ": No such file or directory"; I haven't investigated this issue yet.


khorben:
I will re-issue with fewer changes to the Makefile, which is currently causing the CI to fail.

PR #3117, opened 2022-01-09
Fix the build for NetBSD

khorben:
This fixes the build for NetBSD. It should not influence other platforms.


jsteube:
Cool, many thanks! I don't have any NetBSD installation, but all the changes look good. I trust that you have tested it.

Chick3nman:
I believe this PR has a logic flaw on line 81 of the Makefile, filtering out BSD instead of filtering for it. I have opened https://github.com/hashcat/hashcat/pull/3124 to fix this issue.

PR #3124, opened 2022-01-11
Fix logic flaw in Makefile

Chick3nman:
Line 81 contains inverted logic introduced by PR#3117, this should fix that logic to detect BSD and correctly select `gsed` on BSD and not on linux


roycewilliams:
Fix confirmed working for me

PR #3140, opened 2022-01-18
Refactored standard kernel includes in order to support Apple Metal runtime, updated backend, test units and status code

matrix:
Hi,

this is the first prepare patch in order to add the Apple Metal runtime support on hashcat.
Due to an apple limitation I need to refactor the kernel build options. The limitation consists in not being able to specify options other than "-D", therefore it is not possible to specify the path for the includes in the classic way (using -I).

The apple way to do this implies to use the "offline compiler", the details of which are defined on the following links:
https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf (paragraph 1.5)
https://developer.apple.com/documentation/metal/libraries/building_a_library_with_metal_s_command-line_tools

I didn't like the apple way, so I created a functional workaround using an additional macro preprocessor named "INCLUDE_PATH", which works generically on both MacOS, Linux and Windows. We should take a ride with HIP (win / linux) and NVIDIA on windows (linux is ok).

The only bug currently concerns the use of POCL, in case there is a space in the path. I hope to find the time to contribute to this external project as well to fix the small bug.

Thanks :)


matrix:
for the bug on POCL I sent a PR: https://github.com/pocl/pocl/pull/1028

matrix:
The bug on POCL has been fixed. While waiting for the merge on their repository, I updated our PR. Thanks

matrix:
Full benchmark on Apple M1 / OpenCL using also this PR: https://gist.github.com/matrix/54a3830db3f686ee9596e21c8f819384

matrix:
> Full benchmark on Apple M1 / OpenCL using also this PR: https://gist.github.com/matrix/54a3830db3f686ee9596e21c8f819384

The same, but with 1080 Ti / CUDA
https://gist.github.com/matrix/01e24d45fffa18a6fef65bff1bf76621

jsteube:
So we have:

```
#define XSTR(x) #x
#define STR(x) XSTR(x)

#include STR(INCLUDE_PATH/inc_vendor.h)
```

Why not just:

```
#include $INCLUDE_PATH/inc_vendor.h
```

matrix:
> So we have:
> 
> ```
> #define XSTR(x) #x
> #define STR(x) XSTR(x)
> 
> #include STR(INCLUDE_PATH/inc_vendor.h)
> ```
> 
> Why not just:
> 
> ```
> #include $INCLUDE_PATH/inc_vendor.h
> ```

Hi :)
It doesn't work for me, I'm missing something?

PR #3153, opened 2022-02-05
Kernels: Set the default Address Space Qualifier for any pointer, refactored / updated KERN_ATTR macros and rc4 cipher functions, in order to support Apple Metal runtime

matrix:
Hi,

this is the second prepare patch in order to add the Apple Metal runtime support on hashcat. The major changes introduced are the followings:

**1. Set the default Address Space Qualifier for any pointer**

To compile the kernel sources successfully it was necessary to specify the "address space" (the region of memory from where buffer memory objects are allocated) for any pointers. Example:

```
diff --git a/OpenCL/m25800-pure.cl b/OpenCL/m25800-pure.cl
index 6624bcc62..95e1d34c2 100644
--- a/OpenCL/m25800-pure.cl
+++ b/OpenCL/m25800-pure.cl
[...]
@@ -421,10 +417,10 @@ DECLSPEC inline void SET_KEY32 (LOCAL_AS u32 *S, const u64 key, const u32 val)
 extern __shared__ u32 S[];
 #endif
 
-DECLSPEC void expand_key (u32 *E, u32 *W, const int len)
+DECLSPEC void expand_key (PRIVATE_AS u32 *E, PRIVATE_AS u32 *W, const int len)
 {
-  u8 *E_ptr = (u8 *) E;
-  u8 *W_ptr = (u8 *) W;
+  PRIVATE_AS u8 *E_ptr = (PRIVATE_AS u8 *) E;
+  PRIVATE_AS u8 *W_ptr = (PRIVATE_AS u8 *) W;
 
   for (int pos = 0; pos < 72; pos++) // pos++ is not a bug, we actually want that zero byte here
   {
[...]
```
As you see, it was necessary to change the parameters of the "expand_key" function and the variables used within it.
Obviously this procedure will have to be followed during the development of new modules, in order to make them compatible also with Metal.
Also, the new "PRIVATE_AS" macro is only valued when Metal is used, so these changes have no impact for all other runtimes (OpenCL, CUDA and HIP).

**2. Refactored / updated KERN_ATTR macros**

On this part, I think I have documented the kernel source extensively through comments (see the new "OpenCL/inc_common.h" file), but I try to summarise here.

In Metal there are no functions to identify the global work-item ID (gid), the local work-item ID (lid), etc.
To access this information you must modify the kernel functions by adding the necessary attributes.
For a complete list, please refer to section "5.2.3.6". (Kernel Function Input Attributes) of [Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)

In order to make the current code compatible and also to keep the complexity of developing a new module almost unchanged, changes have been made to the KERN_ATTR macros.
Since the current macros were used both in "kernel functions" definitions and for "not-kernel functions", it was necessary to split the macros into two, for example "KERN_ATTR_BASIC" and "KERN_ATTR_FUNC_BASIC".

For "kernel function" I mean all the functions with "KERNEL_FQ" macro as prefix, and for "not-kernel function" all the functions with "DECLSPEC" macro as prefix.
For less experienced users, only the "kernel function" are "callable"/"visible" from the host-code.

Below is a concrete example of their use in "OpenCL/m00020_a3-optimized.cl":

The m00020_m04 "KERNEL_FQ" function declaration remained unchanged, so we still using "KERN_ATTR_BASIC", but we need to retrieve lid/gid/lsize by using get_local_id(), get_global_id() and get_local_size() because they will be used later to call "m00020m" function. 

```
[...]
KERNEL_FQ void m00020_m04 (KERN_ATTR_BASIC ())
{
  /**
   * base
   */

  const u64 lid = get_local_id (0);
  const u64 gid = get_global_id (0);
  const u64 lsz = get_local_size (0);
[...]
  m00020m (w0, w1, w2, w3, pw_len, pws, rules_buf, combs_buf, bfs_buf, tmps, hooks, bitmaps_buf_s1_a, bitmaps_buf_s1_b, bitmaps_buf_s1_c, bitmaps_buf_s1_d, bitmaps_buf_s2_a, bitmaps_buf_s2_b, bitmaps_buf_s2_c, bitmaps_buf_s2_d, plains_buf, digests_buf, hashes_shown, salt_bufs, esalt_bufs, d_return_buf, d_extra0_buf, d_extra1_buf, d_extra2_buf, d_extra3_buf, kernel_param, gid, lid, lsz);
}
```
In the "m00020m" we need use the new macro "KERN_ATTR_FUNC_BASIC" instead of "KERN_ATTR_BASIC".

```
[...]
DECLSPEC void m00020m (PRIVATE_AS u32 *w0, PRIVATE_AS u32 *w1, PRIVATE_AS u32 *w2, PRIVATE_AS u32 *w3, const u32 pw_len, KERN_ATTR_FUNC_BASIC ())
{
  /**
   * modifiers are taken from args
   */
[...]
```

I would like to specify that these macros are only used in optimized kernels.
@jsteube let me know if you want to rename these new macros, at the moment I have simply added "_FUNC_" to differentiate them from the originals.  

**3. refactored / updated rc4 cipher**
based on the explanation in the previous point, I had to modify some functions regarding the rc4 cipher, as they made use of calls such as "get_local_id" inside not-kernel functions, a practice not manageable with Metal.

**For developers**: as you can see, there are no really major changes to be made in order to write code that can also be used with Metal. However, as soon as there is an "approved" version, the guide for plugin development on hashcat will also be updated.

if there are any doubts I am available both here and on [our official discord server](https://hashcat.net/discord)
Sorry for any mistakes in writing, english is not my native language.

Thanks!


matrix:
Actually there is one more thing to explain, which concerns a change on "OpenCL/inc_types.h": with Metal it is not possible to use "unsigned long long", so the only way I found as a workaround was the following:

```
diff --git a/OpenCL/inc_types.h b/OpenCL/inc_types.h
index ef76cf8ab..918e3a7f9 100644
--- a/OpenCL/inc_types.h
+++ b/OpenCL/inc_types.h
@@ -45,6 +45,14 @@ typedef unsigned long       ulong;
 typedef unsigned long long  ullong;
 #endif
 
+#ifdef IS_METAL
+typedef unsigned char  uchar;
+typedef unsigned short ushort;
+typedef unsigned int   uint;
+typedef unsigned long  ulong;
+#define ullong ulong
+#endif
+
 #ifdef IS_OPENCL
 typedef ulong   ullong;
 typedef ulong2  ullong2;
@@ -57,7 +65,11 @@ typedef ulong16 ullong16;
 typedef uchar  u8;
 typedef ushort u16;
 typedef uint   u32;
+#ifdef IS_METAL
+typedef ulong  u64;
+#else
 typedef ullong u64;
+#endif
 #else
 typedef uint8_t  u8;
 typedef uint16_t u16;
```

PR #3166, opened 2022-02-10
Metal: added support for vectors up to 4

matrix:
Hi,

with this PR we add support for vectors on Metal.
As described in section 2.2 - Vector Data Types, Metal supports 2, 3 or 4 component vectors.

In order to avoid changing the source code of the algorithms again, it was necessary to redefine s0/s1/s2/s3 in x/y/z/w in OpenCL/inc_vendor.h file. 

As a bonus, the "rotate" function has been enabled, as it is natively supported by Metal (section 6.3 - Integer Functions).

Reference:
[Metal Shading Language Specification](https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf)

Thanks ;)


matrix:
Hashcat v6.2.5-209-g431c46f43+, Full Benchmark on Apple M1 (arm64 binary) using Metal runtime, Vector size 2

https://gist.github.com/matrix/8f24ca610213af93a89b9e44c3a2e98b

jsteube:
Nice and clean patch, thanks

PR #3204, opened 2022-03-02
Added kernel, module and unit-test for hash-mode 28400: bcryptsha512

Lars-Saetaberget:
Based on the bcryptsha1 plugin (25800), with minimal changes to use sha512 instead of sha1.

This is used for NodeBB user passwords if passwordShaWrapped is set, and probably a bunch of other things.

### DIFF
--- a/OpenCL/m28400-pure.cl
+++ b/OpenCL/m28400-pure.cl
@@ -1,0 +1,1029 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#ifdef KERNEL_STATIC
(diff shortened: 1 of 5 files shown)


jsteube:
Hey there, thanks for the PR. I'll merge this as-is, but I have one comment:

bcrypt supports only passwords up to length 72 due to what's going on in the expand_key() function, but the input here has the length 128. So it means, the input is truncated. You could totally remove the call to expand_key() function (as well as the function itself) and write the uint_to_hex_lower8() results directly into E[]. This might not give a speed bonus or only minimal because this is not what makes the algorithm slow, but it might reduce code complexity so the JIT will be slightly faster to compile the kernel on runtime.

Issue #3213, opened 2022-03-09
Masterkey v1 Context 3 (-m 15310) CUDA Error


virusvfv:
Masterkey bruteforce with CUDA or OpenCL gives CUDA errors on Windows or Linux:

```
./hashcat -m 15310 -b
hashcat (v6.2.5) starting in benchmark mode
* Device #1: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
* Device #2: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
nvmlDeviceGetFanSpeed(): Not Supported
CUDA API (CUDA 11.6)
====================
* Device #1: Quadro P4200, 8011/8119 MB, 18MCU
OpenCL API (OpenCL 3.0 CUDA 11.6.110) - Platform #1 [NVIDIA Corporation]
========================================================================
* Device #2: Quadro P4200, skipped
Benchmark relevant options:
===========================
* --optimized-kernel-enable
--------------------------------------------------------------------------
* Hash-Mode 15310 (DPAPI masterkey file v1 (context 3)) [Iterations: 9999]
--------------------------------------------------------------------------

cuEventSynchronize(): an illegal memory access was encountered

cuMemcpyHtoDAsync(): an illegal memory access was encountered

cuMemcpyHtoDAsync(): an illegal memory access was encountered

cuMemcpyHtoDAsync(): an illegal memory access was encountered

....

cuLaunchKernel(): an illegal memory access was encountered

* Device #1: skipped, due to kernel autotune failure (-5).

Aborting session due to kernel autotune failures, for all active devices.

You can use --force to override this, but do not report related errors.

Started: Wed Mar  9 19:36:51 2022
Stopped: Wed Mar  9 19:37:04 2022
```

With open CL it gives CL_UNKNOWN_ERROR:

```
./hashcat -m 15310 -b -d 2
hashcat (v6.2.5) starting in benchmark mode

Benchmarking uses hand-optimized kernel code by default.
You can use it in your cracking session by setting the -O option.
Note: Using optimized kernel code limits the maximum supported password length.
To disable the optimized kernel code in benchmark mode, use the -w option.

* Device #2: WARNING! Kernel exec timeout is not disabled.
[...]

PR #3219, opened 2022-03-11
fix dpapimk_tmp_v1_t on module_15310.c

matrix:
Details here: https://github.com/hashcat/hashcat/issues/3213#issuecomment-1064168292
CC @virusvfv


jsteube:
Nice find, thanks!

PR #3224, opened 2022-03-13
sanity check for u32 cast for combs_cnt, bfs_cnt

philsmd:
This is just a very harsh sanity check that makes sure that the `(u32) combinator_ctx->combs_cnt` and `(u32) mask_ctx->bfs_cnt`  (see https://github.com/hashcat/hashcat/blob/8bc4a920892185ab52e5b05b92378a4658fc0864/src/backend.c#L3602-L3603) can never result in wrong amount of amplifier iterations due to a wrong / unsafe cast (if upper bits are set).

This is just a hash / critical preventative fix for another problem reported in #3144 (basically the first part of the fix).
We still need to restict how large the values of `combs_cnt` and `bfs_cnt` can be by also patching `mpsp.c` and `combinator.c` (and maybe others).... i.e. the root of the problem needs to be fixed too.

Thx

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -3595,6 +3595,17 @@ int run_cracker (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param, co
     }
     else
     {
+      // sanity check: do NOT cast to an u32 integer type without checking that it is safe (upper bits must NOT be set)
+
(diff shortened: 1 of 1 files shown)


jsteube:
Nice find, thanks :)

PR #3229, opened 2022-03-18
status-json: also print device name and type

davidbolvansky:
The patch adds device name to the (json) output of hashcat.

```
"devices": [{
		"device_id": 1,
		"device_name": "NVIDIA GeForce GTX 960M",
		"device_type": "GPU",
		"speed": 1027377466,
		"temp": 91,
		"util": 96
	}],
```

This info is useful for distributed password cracking systems* to show more details about cracking nodes (speed, temp, util is currently provided, but need to somehow get device name and this way looks simple and effective).

*As one of Fitcrack developers, we have a immediate use case for this new extension.

### DIFF
--- a/include/status.h
+++ b/include/status.h
@@ -88,6 +88,7 @@ int         status_get_innerloop_left_dev             (const hashcat_ctx_t *hash
 int         status_get_iteration_pos_dev              (const hashcat_ctx_t *hashcat_ctx, const int backend_devices_idx);
 int         status_get_iteration_left_dev             (const hashcat_ctx_t *hashcat_ctx, const int backend_devices_idx);
 char       *status_get_device_name                    (const hashcat_ctx_t *hashcat_ctx, const int backend_devices_idx);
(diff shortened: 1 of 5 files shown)


matrix:
Hi @davidbolvansky, at this point I would also take the device-type, could you add it?

davidbolvansky:
Ok, I will check it.

davidbolvansky:
@matrix added

```
"devices": [{
		"device_id": 1,
		"device_name": "NVIDIA GeForce GTX 960M",
		"device_type": "GPU",
		"speed": 1027377466,
		"temp": 91,
		"util": 96
	}],
```

s3inlc:
Just a small thing to consider. I think in general it could be possible (correct me if I'm wrong) that the device name contains characters that would need to be escaped in JSON.
When the device name is just printed between the quotes, it may produce broken JSON.

davidbolvansky:
Yes, possibly, so we can go with safe approach and escape them. There is a logic to do so for (target -> target_json_encoded) so I just need to generalize it a bit.

PR #3230, opened 2022-03-18
Do not use term colors if output is not terminal

davidbolvansky:
Do not emit ANSI escape control characters if hashcat's output is not terminal - as some tools may read it via pipe, or user may redirect it to file, etc..


jsteube:
Yep, this is cool, thanks!

PR #3240, opened 2022-03-22
add json output for hash info

davidbolvansky:
--hash-info provides nice text overview of supported hashes, but not in the form which is very friendly for machine processing.
I decided to provide json version of hash info, which can be easily used by third party tools and cracking systems.

Motivation: In Fitcrack, we have a old solution which takes output of --help table and converts it to dict - https://github.com/nesfit/fitcrack/blob/dev/webadmin/fitcrackAPI/src/src/api/fitcrack/attacks/hashcat_modes_to_fitcrack_dict.py - not great, not terrible., but if there is a possibility to have some more general solution, we are interested in, as we plan to use more details about hashes in the future (eg. " is slow hash" , " is salted" sounds very useful to improve cracking algorithms)

Attached json for current master. https://jsonlint.com/ says it is OK. Produced with: ./hashcat --hash-info --machine-readable --quiet &> hashinfo.json

[Uploading hashinfo.zip…]()

One thing I am not sure: I decided to put it under "--machine-readable" to keep code changes as small as possible, but alternative solution could be new option --hash-info-json - as we already have --status-json. (but this solution requires a bit more code). What do you think?

### DIFF
--- a/src/terminal.c
+++ b/src/terminal.c
@@ -632,6 +632,147 @@ void compress_terminal_line_length (char *out_buf, const size_t keep_from_beginn
   *ptr1 = 0;
 }
 
+void hash_info_single_json (hashcat_ctx_t *hashcat_ctx, user_options_extra_t *user_options_extra)
+{
+  if (hashconfig_init (hashcat_ctx) == 0)
+  {
+    hashconfig_t *hashconfig = hashcat_ctx->hashconfig;
+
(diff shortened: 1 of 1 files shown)


davidbolvansky:
cc @s3inlc from Hashtopolis as well.

jsteube:
I think is fine the way you have it implemented, so we can leave it like this, but you can add --hash-info-json if you want.

PR #3260, opened 2022-04-20
fixes #3253: memory leak in CPU rule engine fixed

### LINKED ISSUE #3253: Memory Leak when using specific rules (-j)
Setting -j">A" (reject candidates less than length 10) in a normal -a 0 wordlist attack appears to create some sort of runaway in memory. As candidates are processed, the memory usage grows indefinitely across a single or many wordlists during the attack. Pausing the attack also pauses the increase in memory so it appear related to candidate processing and not some sort of unrelated loop that is stuck.

Where large_example_dict is 7gb and contains ~255million passwords that are valid for >10:

`hashcat.exe -a 0 -m 0 8743b52063cd84097a65d1633f5c74f5 large_example_dict -j">A"`

In my testing, this command will consume memory in excess of 8gb by the time the dictionary cache is built and just under 14gb by the time the attack process completes. This is on a single computing device, Nvidia RTX 3090, on Windows.

philsmd:
This problem was first mentioned here https://github.com/hashcat/hashcat/issues/3253 by @Chick3nman  and I was now able to reproduce it. The main cause/culprit of the problem is clear if you debug this a little bit: every rule that "fails" to do its job to the password will/could return early from the CPU rule engine function and there was no clean memory cleanup in place.

The problem could also be worked around with other altnernatives like using stack (instead of heap) or using very ugly "gotos" (please dont!) etc... but I think this is quite a nice to read and elegant fix of the main problem. What do you think ?

As mentioned in the issue the problem can be easily reproduce with a rule that "fails" quite a lot of time to do anything to the plain, something like:
`./hashcat -a 0 -m 0 -j">A" 8743b52063cd84097a65d1633f5c74f5 large_example_dict`

kudos again to @Chick3nman for finding this ugly memory leak. Thx

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -44,6 +44,7 @@
 - Fixed handling of devices in benchmark mode for "kernel build error". Instead of canceling, skip the device and move on to the next
 - Fixed handling of password candidates that are shorter than the minimum password length in Association Attack
 - Fixed --hash-info example password output: force uppercase if OPTS_TYPE_PT_UPPER is set
+- Fixed memory leak in CPU rule engine
 - Fixed method of how OPTS_TYPE_AUX* kernels are called in an association attack, for example in WPA/WPA2 kernel
 - Fixed missing option flag OPTS_TYPE_SUGGEST_KG for hash-mode 11600 to inform the user about possible false positives in this mode
 - Fixed optimized (-O) candidate generation with --stdout and -a 7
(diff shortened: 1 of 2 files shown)


jsteube:
Indeed I think I'd use stack allocation instead, but it's also fine this way. Thanks for the patch and also thanks to @Chick3nman for reporting.

PR #3265, opened 2022-04-25
Tune scrypt config for RTX 3090 Ti

rarecoil:
RTX 3090 Ti appeared today and hashcat was the first benchmark I ran. Was getting about ~3800 H/s on master, so I ran the tuning steps in `src/modules/08900.c`. Best result in tuning was 4156.

```
hashcat (v5.1.0-3437-g6aa0d2082) starting in benchmark mode

Benchmarking uses hand-optimized kernel code by default.
You can use it in your cracking session by setting the -O option.
Note: Using optimized kernel code limits the maximum supported password length.
To disable the optimized kernel code in benchmark mode, use the -w option.

* Device #1: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
* Device #2: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
CUDA API (CUDA 11.6)
====================
* Device #1: NVIDIA GeForce RTX 3090 Ti, 23351/24233 MB, 84MCU

OpenCL API (OpenCL 3.0 CUDA 11.6.127) - Platform #1 [NVIDIA Corporation]
========================================================================
* Device #2: NVIDIA GeForce RTX 3090 Ti, skipped

Benchmark relevant options:
===========================
* --optimized-kernel-enable

---------------------------------------------
* Hash-Mode 8900 (scrypt) [Iterations: 16384]
---------------------------------------------

Speed.#1.........:     4059 H/s (29.47ms) @ Accel:84 Loops:1024 Thr:32 Vec:1
```

PR #3268, opened 2022-04-27
Port scrypt tuning for RTX 3090Ti to other scrypt-based hash types

rarecoil:
Per discussion on #3265, @jsteube said we could probably roll forward the same tuning to other scrypt-based hash types. I tested most on my platform but cut the lowest and highest ends of the iterations to between 50-120, as there did not appear to be returns outside of this range with 84 being the local maximum in most cases anyhow.

This pull request ports forward the tuned values for the RTX 3090 Ti to these other hash types.

Fastest speeds during testing:

15700: 32 H/s 
22700: 4315 H/s
27700: 4324 H/s
28200: 4134 H/s

I warn that the usual `hashcat -b -m 15700` shows 0 H/s for the `15700 (Ethereum Wallet, SCRYPT)` type but it appeared fine when running the extended tests. It does that with or without this patch.


rarecoil:
This is extremely easy work overall, so if there are other hash types that can/should be tuned in this manner, I can try to pick them up for this hardware as I have time.

PR #3272, opened 2022-05-03
fix DES (14000) for more 16 or more passwords

Freax13:
This pr fixes cracking DES keys in cases where there are 16 or more different passwords with identical plain texts.

To be honest I don't understand the code around this enough to be confident to say this is the correct fix, someone else should definitively double check! l basically just looked for similar patterns elsewhere in the code and found slightly different two versions:
https://github.com/hashcat/hashcat/blob/39b768a620d9446018c5bc5658ab0a445bfe3548/OpenCL/m01500_a3-pure.cl#L2264-L2275
https://github.com/hashcat/hashcat/blob/39b768a620d9446018c5bc5658ab0a445bfe3548/OpenCL/m03000_a3-pure.cl#L2105-L2116

### DIFF
--- a/OpenCL/m14000_a3-pure.cl
+++ b/OpenCL/m14000_a3-pure.cl
@@ -2171,8 +2171,8 @@ KERNEL_FQ void m14000_mxx (KERN_ATTR_BITSLICE ())
       #endif
       for (int i = 0; i < 32; i++)
       {
-        out0[i] = out[ 0 + i];
-        out1[i] = out[32 + i];
+        out0[i] = out[ 0 + 31 - i];
+        out1[i] = out[32 + 31 - i];
       }
 
       transpose32c (out0);
@@ -2183,8 +2183,8 @@ KERNEL_FQ void m14000_mxx (KERN_ATTR_BITSLICE ())
       #endif
       for (int slice = 0; slice < 32; slice++)
       {
-        const u32 r0 = out0[slice];
-        const u32 r1 = out1[slice];
+        const u32 r0 = out0[31 - slice];
+        const u32 r1 = out1[31 - slice];
         const u32 r2 = 0;
         const u32 r3 = 0;
 


bigfella237:
It sounds like you're searching an odd-parity 56-bit keyspace using a 64-bit charset, which is why you're getting sixteen positive results. If you ignore parity then all 16 resulting keys are correct, but only one of those results is the true 56-bit key. You need to check for odd-parity to determine which key it is (not to mention that your keyspace searches must be taking far longer than they should).

Freax13:
> It sounds like you're searching an odd-parity 56-bit keyspace using a 64-bit charset, which is why you're getting sixteen positive results. If you ignore parity then all 16 resulting keys are correct, but only one of those results is the true 56-bit key. You need to check for odd-parity to determine which key it is (not to mention that your keyspace searches must be taking far longer than they should).

No, that's not what I meant. The number 16 was not referring to the number of keys I cracked, but to the amount of ciphertext & plaintext pairs in my hashlist. If I put 16 or more ciphertext & plaintext pairs with the same plaintext for all pairs in my hashlist it wouldn't crack any of them. The number 16 stems from the fact that Hashcat has an optimization for less than 16 pairs (see line 2123 in the same file) and the code for 16 or more pairs seemed to be broken.

jsteube:
Looks good, thanks for the fix. Here's how I've been testing it:

```
diff --git a/tools/test.pl b/tools/test.pl
index 0c909d0ae..36826299c 100755
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -41,7 +41,7 @@ exists &{module_generate_hash} or die "Module function 'module_generate_hash' no
 exists &{module_verify_hash}   or die "Module function 'module_verify_hash' not found\n";
 
 my $giveup_at      = 1000000;
-my $single_outputs = 8;
+my $single_outputs = 32;
 
 my $constraints = get_module_constraints ();
 
diff --git a/tools/test_modules/m14000.pm b/tools/test_modules/m14000.pm
index e6edbbea5..a2db58c52 100644
--- a/tools/test_modules/m14000.pm
+++ b/tools/test_modules/m14000.pm
@@ -15,7 +15,7 @@ sub module_constraints { [[8, 8], [16, 16], [-1, -1], [-1, -1], [-1, -1]] }
 sub module_generate_hash
 {
   my $word = shift;
-  my $salt = shift;
+  my $salt = shift; $salt = "1111111111111111";
 
   my $cipher = new Crypt::DES ($word);
```
Generate files, but note this will attack -a 0 not -a 3:

```
root@ht:~/hashcat# tools/test.sh -t multi -a 0 -m 14000 -V1                                                  
[ test_1651838360 ] > Init test for hash type 14000.
[ test_1651838360 ] [ Type 14000, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

Then run it as -a 3 instead (this is with the patch in place - it was not cracking anything unpatched):

```
root@ht:~/hashcat# ./hashcat -m 14000 test_1651838360/14000_hashes.txt ?d?d?d?d?d?d?d?d  -a 3 --potfile-disable 
hashcat (v6.2.5-415-g5df7bed5a) starting

CUDA API (CUDA 11.6)
====================
* Device #1: NVIDIA GeForce RTX 2080 Ti, 10855/11019 MB, 68MCU

OpenCL API (OpenCL 3.0 CUDA 11.6.99) - Platform #2 [NVIDIA Corporation]
* Device #3: NVIDIA GeForce RTX 2080 Ti, skipped

Minimum password length supported by kernel: 8
Maximum password length supported by kernel: 8

Hashes: 32 digests; 32 unique digests, 1 unique salts

Optimizers applied:
* Zero-Byte
* Not-Iterated
* Single-Salt
* Brute-Force

Host memory required for this attack: 1474 MB

32b8a2ddd28ebf2c:1111111111111111:56053012                
92502a549e11c05d:1111111111111111:69917700                
909ce4bb4a644014:1111111111111111:22528423                
a721ff5a4da50fd8:1111111111111111:77266400                
57707f5b2cfbb774:1111111111111111:45425923                
a9f4bca6a53c9a85:1111111111111111:50934723                
e515b9f878b68b2c:1111111111111111:13629600                
2d10b749e65b1830:1111111111111111:63051855                
28a54a723b917e65:1111111111111111:07265855                
7ed5936d8e848a02:1111111111111111:87423388                
fc139176c254f3d7:1111111111111111:98085520                
020fbec302a7dc70:1111111111111111:07281320                
9d6fbf8bb8587c7a:1111111111111111:19960243                
c47dacd9fe07c9f2:1111111111111111:60901920                
31b614cd8ae3141b:1111111111111111:41685266                
4d667b42ab0f6a8c:1111111111111111:36047987                
f5133aaea4c0004e:1111111111111111:93087843                
946cabf555c974e4:1111111111111111:87788190                
ca2e37637703065c:1111111111111111:41227350                
4b653b7d7980108c:1111111111111111:19983650                
23f93547048d6c02:1111111111111111:92597719                
6e74938105c50446:1111111111111111:08774119                
222595f3c562846f:1111111111111111:77874071                
bd4ffbb812e681a9:1111111111111111:71033974                
9a516504a1e11a9f:1111111111111111:25956074                
196af67d0d6c5739:1111111111111111:48513562                
e23d54ba403fdd6b:1111111111111111:56251962                
89a298daff4be2a5:1111111111111111:50925636                
50bb373d03559dbd:1111111111111111:85909405                
d37876e376fa4cb9:1111111111111111:65020936                
df33d35d1756cf2c:1111111111111111:07806585                
dc1a55d2974c5f69:1111111111111111:28729358                
                                                          
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 14000 (DES (PT = $salt, key = $pass))
Hash.Target......: test_1651838360/14000_hashes.txt
Time.Started.....: Fri May  6 14:00:07 2022 (0 secs)
Time.Estimated...: Fri May  6 14:00:07 2022 (0 secs)
Guess.Mask.......: ?d?d?d?d?d?d?d?d [8]
Speed.#1.........:  2724.2 MH/s (0.60ms) @ Accel:128 Loops:1024 Thr:32 Vec:1
Recovered.Total..: 32/32 (100.00%) Digests
Progress.........: 78125000/100000000 (78.12%)
Rejected.........: 0/78125000 (0.00%)
Restore.Point....: 75000/100000 (75.00%)
Restore.Sub.#1...: Salt:0 Amplifier:0-1000 Iteration:0-1024
Candidates.#1....: 12345628 -> 68830165
Hardware.Mon.#1..: Temp: 56c Fan:  0% Util: 16% Core:1935MHz Mem:6800MHz Bus:16

Started: Fri May  6 14:00:03 2022
Stopped: Fri May  6 14:00:08 2022
```

PR #3276, opened 2022-05-07
Added sha1($salt.sha1(utf16le($username).':'.utf16le($pass), true)) (Radmin3 base)

piwvvo:
I'm not sure if that algo is really needed as a standalone, but it is a base for Radmin3 algorithm which was discussed in https://github.com/hashcat/hashcat/issues/3259 . Hash format is `hash:salt:hex_username`

I couldn't exactly figure out attack modes 1 and 3 for optimized kernels.


jsteube:
Hey there, many thanks for the contribution. The only Problem I see is the both missing kernels for optimized mode, not because hashcat couldn't work with it but it could be very confusing to have 2 missing kernels. So I'm going to remove the optimized -a0 kernel after merging. I will also rename the kernels from 9902 to 29000. If we want we can still add optimized kernels later.

PR #3295, opened 2022-05-28
Added Radmin3

piwvvo:
With huge help from @philsmd we finally got the version that seems to be ready for pull request. Algorithm is basically `sha1($salt.sha1_bin(utf16le($username).':'.utf16le($pass)))` with final sha1 digest used as an exponent for bignum operation. It's explained in detail in this issue https://github.com/hashcat/hashcat/issues/3259

The algorithm was reversed-engineered by synacktiv: https://www.synacktiv.com/publications/cracking-radmin-server-3-passwords.html 
The reason why we created lookup table is, because modulus and base seems to be hardcoded so it's always the same.


jsteube:
This looks like some very time intensive work, so thank you very much @piwvvo and @philsmd 

I wanted to test the PR using the regular GitHub feature to download it from https://github.com/hashcat/hashcat/pull/3295.diff and the use git apply, but I got the following error (which I never encountered before):

```
error: cannot apply binary patch to 'include/inc_radmin3_constants_pre.data' without full index line
error: include/inc_radmin3_constants_pre.data: patch does not apply
```

So I can't test the PR before merge. However from looking over the source it looks pretty clean so I'll merge it without test before. 

The only thing that looks slightly strange to me is this section:

```
#define INCLUDE_BIN(name, file) asm \
(                                   \
  ".section .rodata     \n"         \
  ".balign 4            \n"         \
  ".global "   name "   \n"         \
               name ":  \n"         \
  ".incbin \"" file "\" \n"         \
  ".section .text       \n"         \
)
```

This looks a bit over-engineered to me. It could also cause problems with compilers others than GCC/Clang. OTOH, I don't see too much of a performance loss by embedding the data into the module as u32 array directly. 

I'd prefer to see this without any use of inline ASM, please update this section or let me know if you want me to do it.

Other than that it's a great achievement to have the mod_4096() on GPU. Really great work, thanks!

jsteube:
OK after merge I was able to run some tests. There are some fails I think we need to debug this. Here's a small log in order to reproduce locally:

```
tools/test.sh -t single -m 29200 -a all
```

results:

```
[ test_1653845687 ] > Init test for hash type 29200.
[ test_1653845687 ] [ Type 29200, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > Error : 8/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
```

If you take a look into the logfile, this kind of error repeats:

```
password not found, cmdline : echo 47121184580876739260943492704846038866976155858 | ./hashcat --quiet --potfile-disable --hwmon-disable -O --runtime 400 -D 2 --backend-vector-width 1 -a 0 -m 29200 '$radmin3$44004c0067004c0057004e007a00500059004b00440066006500770057005200500071006a00520051006500630055007400450041004700440041006400510078006100680049004f0043006300650056006b004e004c00*3035373231313832363736353230353737313734303638383135343230373638*4a56b92363a297440bc79885efd8baec6550f0b9b3d4a440dcabd7562d543bca9a94b4a85570a1ae8c225675e9baab206ae47ec66620092991f7ac1dc8cd343dfa8c995edec94cbad4dd8362752263a924a50ed7335ed8b74e511fbfd8089367de7e785902f52caa1ce6e8550affcdc47d9f079951e4262d0c30ff70882645ec7ad7fb4b903940684a3e3185bd48a3ced4675bf7492d86639655005ee8ab382062912a8f706fba837deef8f5a3198914c5fa73792ec46a954a8c74190d282d85bc208ccfd9bb3d693771c8fa4c3bb369be8ab22ce1569bfb7bbe2f49fc8a09e559bcf5d7d98cb5be7c5a37c18da225c3f4c94bddd8202ae2a09f95bead9620fb'
```

jsteube:
Using test.sh in -P mode works fine. So probably  a problem with passwords lengths in optimized mode? Maybe the unit tests needs to be more restrictive?

jsteube:
I've just checking if there's actually a benefit from having the optimized kernel. It seems the performance is the same, caused by the bottleneck from the RSA operation. I can also see that you unset the the password and salt length options in the unit test. Maybe we just remove the optimizer kernels that would cause the unit tests to succeed?

piwvvo:
Hey @jsteube when I've been testing the kernels the optimzed ones seemed to have very simillar speeds (unless PRECOMP_BITS was set to higher value, but that obviously no the case here). I'm not sure how could those be restricted, because UTF-16 username concatenated with UTF-16 password can get easily pretty long. Adding transform in optimized kernels probably would make it slower (not sure about that) than pure kernels. So perhaps leaving only pure kernels is indeed the best solution

philsmd:
oh, this is already merged. Well done @piwvvo , I didn't expect this to be finished/merged this quickly....

I must also say that I'm sorry that it seems that I have (again) misunderstood/misunderpreted something that @piwvvo wrote here: https://github.com/hashcat/hashcat/issues/3259#issuecomment-1136246763 (about the unit tests). I was thinking that only the `test.sh` / `test.pl` was not working for them/him/her (I mean, that it's a `perl` module problem with the system they use, or that `test.sh` itself is not working for them etc, not related to this hash type), but now we see that something is not (yet) completely correct (this is/was only affecting the newly added new optimized kernels that we added later on). I'm sorry for this little mess up (completely my fault), but it's not a problem I will add some pull requests to fix this ... just stay tuned !

Thanks guys

--- 

and btw: we discussed the use of the `.incbin` a little bit here (https://github.com/hashcat/hashcat/issues/3259#issuecomment-1133587188) @jsteube . Please read after this "I was also thinking a little bit more about how and where we store the pre lookup table" . The problem is that a good alternative would only be to store it in the `src/` or `include/` directory.... and hexadecimal storage for 4 MiB will of course be much larger if we use the hexadecimal representation. so this was the only "alternative" that seems quite good, but we are of course willing to again use the old method if better compatible (but what we found out is that both `clang` and `gcc` work perfectly fine with this... the only thing that I am unsure is what happens with different CPUs with different endianess - big endian systems - etc, but they probably have all other problems with `hashcat` in addition to this data reading, right ?).

philsmd:
okay, fewer commits needed as I though to fix stuff (and only comment/cosmetic stuff actually, this is mainly because the optimized kernels were removed and therefore the unit tests were already working/fixed) https://github.com/hashcat/hashcat/pull/3298 and https://github.com/hashcat/hashcat/pull/3299. Anyway I have explained here how it could have worked also by restricting the unit tests for optimized kernels (only allow shorter user lengths), if anybody still wants to read this for reference/learning purposes:
https://github.com/hashcat/hashcat/issues/3259#issuecomment-1141286522

Well done guys, Thx

hmm, it seems that `appveyour` tests really fail because of the `.incbin` usage: https://ci.appveyor.com/project/jsteube/hashcat/builds/43701973/job/061embr0ldk9y78i ... damn, we probably need to use hexadecimal representation instead :(

okay, this should fix it, but it of course rolls back to the hexadecimal representation of the data (which comes with much more disk  space needed for the lookup table, since it's in the `include` folder, this shouldn't hurt too much, because only developers are affected): https://github.com/hashcat/hashcat/pull/3300

jsteube:
Small reproducible compiler error if using CPU for cracking (it seems GPU jit doesn't care but I think we should fix it):

```
$ ./hashcat -b -m 29200 -D1
hashcat (v6.2.5-473-gb5839644a) starting in benchmark mode

* Device #2: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz, 7816/15696 MB (1962 MB allocatable), 8MCU

Benchmark relevant options:
===========================
* --opencl-device-types=1
* --optimized-kernel-enable

---------------------------
* Hash-Mode 29200 (Radmin3)
---------------------------

clCompileProgram(): CL_COMPILE_PROGRAM_FAILURE

Compilation started
1:249:26: error: passing '__local u32 *' (aka '__local unsigned int *') to parameter of type 'const u32 *' (aka 'const unsigned int *') changes address space of pointer
      mul_mod (r_t, pre, m, fact); // r_t = (r_t * RADMIN3_PRE[n]) % m
                         ^
/root/git/hashcat/OpenCL/inc_bignum_operations.cl:849:90: note: passing argument to parameter 'm' here
DECLSPEC void mul_mod (PRIVATE_AS u32 *x, PRIVATE_AS const u32 *y, PRIVATE_AS const u32 *m, PRIVATE_AS const u32 *fact)
                                                                                         ^
1:249:29: error: passing '__local u32 *' (aka '__local unsigned int *') to parameter of type 'const u32 *' (aka 'const unsigned int *') changes address space of pointer
      mul_mod (r_t, pre, m, fact); // r_t = (r_t * RADMIN3_PRE[n]) % m
                            ^~~~
/root/git/hashcat/OpenCL/inc_bignum_operations.cl:849:115: note: passing argument to parameter 'fact' here
DECLSPEC void mul_mod (PRIVATE_AS u32 *x, PRIVATE_AS const u32 *y, PRIVATE_AS const u32 *m, PRIVATE_AS const u32 *fact)
                                                                                                                  ^
1:490:26: error: passing '__local u32 *' (aka '__local unsigned int *') to parameter of type 'const u32 *' (aka 'const unsigned int *') changes address space of pointer
      mul_mod (r_t, pre, m, fact); // r_t = (r_t * RADMIN3_PRE[n]) % m
                         ^
/root/git/hashcat/OpenCL/inc_bignum_operations.cl:849:90: note: passing argument to parameter 'm' here
DECLSPEC void mul_mod (PRIVATE_AS u32 *x, PRIVATE_AS const u32 *y, PRIVATE_AS const u32 *m, PRIVATE_AS const u32 *fact)
                                                                                         ^
1:490:29: error: passing '__local u32 *' (aka '__local unsigned int *') to parameter of type 'const u32 *' (aka 'const unsigned int *') changes address space of pointer
      mul_mod (r_t, pre, m, fact); // r_t = (r_t * RADMIN3_PRE[n]) % m
                            ^~~~
/root/git/hashcat/OpenCL/inc_bignum_operations.cl:849:115: note: passing argument to parameter 'fact' here
DECLSPEC void mul_mod (PRIVATE_AS u32 *x, PRIVATE_AS const u32 *y, PRIVATE_AS const u32 *m, PRIVATE_AS const u32 *fact)
                                                                                                                  ^
Compilation failed
```

If we can assume these both arrays always being read from local it's better to change the mul_mod() function, otherwise I think we need to copy the data to a local array before calling the function. Could you please fix this?

philsmd:
Wow, strange I didn't noticed this myself. Maybe because it only happens with Intel CPU and using -D 1. I have suggested a fix here: https://github.com/hashcat/hashcat/pull/3309
The problem with changing the `inc_bignum_operations.cl` file is that it doesn't really make much sense to use `LOCAL_AS` there (conceptually), furthermore we would need to have 2 altnernative functions for many other function calls, because we sometimes need `PRIVATE_AS` and sometimes we need `LOCAL_AS` as input... that would just be too confusing, in my opinion.
I tried that approach, but ended up with too many functions that have the same code, just another `prototype`, different variable type... but we would need to use those, because for instance `mul ()` etc are called with different inputs .
I think my suggested fix is overall better here (but I agree that normally changing the type would be better, in general). What do you think about my fix. It should work, I've tested it intensively.

Thanks guys. Great testing ! Nice find

ventaquil:
I have problem with compilation and installation (because of `src/emu_inc_bignum_operations.c` file).

```
$ sudo make install_{library{,_dev},hashcat,modules,kernels} SHARED=1
...
find OpenCL/   -mindepth 1 -type d -execdir install -m 755 -d     /usr/local/share/hashcat/OpenCL/{} \;
find OpenCL/   -mindepth 1 -type f -execdir install -m 644 {}     /usr/local/share/hashcat/OpenCL/{} \;
gcc                     obj/affinity.NATIVE.o obj/autotune.NATIVE.o obj/backend.NATIVE.o obj/benchmark.NATIVE.o obj/bitmap.NATIVE.o obj/bitops.NATIVE.o obj/combinator.NATIVE.o obj/common.NATIVE.o obj/convert.NATIVE.o obj/cpt.NATIVE.o obj/cpu_crc32.NATIVE.o obj/debugfile.NATIVE.o obj/dictstat.NATIVE.o obj/dispatch.NATIVE.o obj/dynloader.NATIVE.o obj/event.NATIVE.o obj/ext_ADL.NATIVE.o obj/ext_cuda.NATIVE.o obj/ext_hip.NATIVE.o obj/ext_nvapi.NATIVE.o obj/ext_nvml.NATIVE.o obj/ext_nvrtc.NATIVE.o obj/ext_hiprtc.NATIVE.o obj/ext_OpenCL.NATIVE.o obj/ext_sysfs_amdgpu.NATIVE.o obj/ext_sysfs_cpu.NATIVE.o obj/ext_iokit.NATIVE.o obj/ext_lzma.NATIVE.o obj/filehandling.NATIVE.o obj/folder.NATIVE.o obj/hashcat.NATIVE.o obj/hashes.NATIVE.o obj/hlfmt.NATIVE.o obj/hwmon.NATIVE.o obj/induct.NATIVE.o obj/interface.NATIVE.o obj/keyboard_layout.NATIVE.o obj/locking.NATIVE.o obj/logfile.NATIVE.o obj/loopback.NATIVE.o obj/memory.NATIVE.o obj/monitor.NATIVE.o obj/mpsp.NATIVE.o obj/outfile_check.NATIVE.o obj/outfile.NATIVE.o obj/pidfile.NATIVE.o obj/potfile.NATIVE.o obj/restore.NATIVE.o obj/rp.NATIVE.o obj/rp_cpu.NATIVE.o obj/selftest.NATIVE.o obj/slow_candidates.NATIVE.o obj/shared.NATIVE.o obj/status.NATIVE.o obj/stdout.NATIVE.o obj/straight.NATIVE.o obj/terminal.NATIVE.o obj/thread.NATIVE.o obj/timer.NATIVE.o obj/tuningdb.NATIVE.o obj/usage.NATIVE.o obj/user_options.NATIVE.o obj/wordlist.NATIVE.o obj/emu_general.NATIVE.o obj/emu_inc_common.NATIVE.o obj/emu_inc_platform.NATIVE.o obj/emu_inc_scalar.NATIVE.o obj/emu_inc_simd.NATIVE.o obj/emu_inc_rp.NATIVE.o obj/emu_inc_rp_optimized.NATIVE.o obj/emu_inc_hash_md4.NATIVE.o obj/emu_inc_hash_md5.NATIVE.o obj/emu_inc_hash_ripemd160.NATIVE.o obj/emu_inc_hash_sha1.NATIVE.o obj/emu_inc_hash_sha256.NATIVE.o obj/emu_inc_hash_sha384.NATIVE.o obj/emu_inc_hash_sha512.NATIVE.o obj/emu_inc_hash_streebog256.NATIVE.o obj/emu_inc_hash_streebog512.NATIVE.o obj/emu_inc_ecc_secp256k1.NATIVE.o obj/emu_inc_bignum_operations.NATIVE.o obj/emu_inc_cipher_aes.NATIVE.o obj/emu_inc_cipher_camellia.NATIVE.o obj/emu_inc_cipher_des.NATIVE.o obj/emu_inc_cipher_kuznyechik.NATIVE.o obj/emu_inc_cipher_serpent.NATIVE.o obj/emu_inc_cipher_twofish.NATIVE.o obj/emu_inc_hash_base58.NATIVE.o obj/brain.NATIVE.o obj/7zCrc.LZMA.NATIVE.o obj/7zCrcOpt.LZMA.NATIVE.o obj/7zFile.LZMA.NATIVE.o obj/7zStream.LZMA.NATIVE.o obj/Alloc.LZMA.NATIVE.o obj/Bra.LZMA.NATIVE.o obj/Bra86.LZMA.NATIVE.o obj/BraIA64.LZMA.NATIVE.o obj/CpuArch.LZMA.NATIVE.o obj/Delta.LZMA.NATIVE.o obj/LzmaDec.LZMA.NATIVE.o obj/Lzma2Dec.LZMA.NATIVE.o obj/Sha256.LZMA.NATIVE.o obj/Sha256Opt.LZMA.NATIVE.o obj/Xz.LZMA.NATIVE.o obj/XzCrc64.LZMA.NATIVE.o obj/XzCrc64Opt.LZMA.NATIVE.o obj/XzDec.LZMA.NATIVE.o obj/XzIn.LZMA.NATIVE.o obj/adler32.ZLIB.NATIVE.o obj/crc32.ZLIB.NATIVE.o obj/deflate.ZLIB.NATIVE.o obj/inflate.ZLIB.NATIVE.o obj/inffast.ZLIB.NATIVE.o obj/inftrees.ZLIB.NATIVE.o obj/trees.ZLIB.NATIVE.o obj/gzread.ZLIB.NATIVE.o obj/gzwrite.ZLIB.NATIVE.o obj/gzclose.ZLIB.NATIVE.o obj/zutil.ZLIB.NATIVE.o obj/gzlib.ZLIB.NATIVE.o obj/contrib/minizip/unzip.ZLIB.NATIVE.o obj/contrib/minizip/ioapi.ZLIB.NATIVE.o obj/xxhash.XXHASH.NATIVE.o obj/strlist.UNRAR.NATIVE.o obj/strfn.UNRAR.NATIVE.o obj/pathfn.UNRAR.NATIVE.o obj/smallfn.UNRAR.NATIVE.o obj/global.UNRAR.NATIVE.o obj/file.UNRAR.NATIVE.o obj/filefn.UNRAR.NATIVE.o obj/filcreat.UNRAR.NATIVE.o obj/archive.UNRAR.NATIVE.o obj/arcread.UNRAR.NATIVE.o obj/unicode.UNRAR.NATIVE.o obj/system.UNRAR.NATIVE.o obj/isnt.UNRAR.NATIVE.o obj/crypt.UNRAR.NATIVE.o obj/crc.UNRAR.NATIVE.o obj/rawread.UNRAR.NATIVE.o obj/encname.UNRAR.NATIVE.o obj/resource.UNRAR.NATIVE.o obj/match.UNRAR.NATIVE.o obj/timefn.UNRAR.NATIVE.o obj/rdwrfn.UNRAR.NATIVE.o obj/consio.UNRAR.NATIVE.o obj/options.UNRAR.NATIVE.o obj/errhnd.UNRAR.NATIVE.o obj/rarvm.UNRAR.NATIVE.o obj/secpassword.UNRAR.NATIVE.o obj/rijndael.UNRAR.NATIVE.o obj/getbits.UNRAR.NATIVE.o obj/sha1.UNRAR.NATIVE.o obj/sha256.UNRAR.NATIVE.o obj/blake2s.UNRAR.NATIVE.o obj/hash.UNRAR.NATIVE.o obj/extinfo.UNRAR.NATIVE.o obj/extract.UNRAR.NATIVE.o obj/volume.UNRAR.NATIVE.o obj/list.UNRAR.NATIVE.o obj/find.UNRAR.NATIVE.o obj/unpack.UNRAR.NATIVE.o obj/headers.UNRAR.NATIVE.o obj/threadpool.UNRAR.NATIVE.o obj/rs16.UNRAR.NATIVE.o obj/cmddata.UNRAR.NATIVE.o obj/ui.UNRAR.NATIVE.o obj/filestr.UNRAR.NATIVE.o obj/recvol.UNRAR.NATIVE.o obj/rs.UNRAR.NATIVE.o obj/scantree.UNRAR.NATIVE.o obj/qopen.UNRAR.NATIVE.o obj/hc_decompress_rar.UNRAR.NATIVE.o -o libhashcat.so.6.2.5                     -s -lstdc++ -lpthread -ldl -lrt -lm -shared -Wl,-soname,libhashcat.so.6.2.5
/usr/bin/ld: obj/emu_inc_bignum_operations.NATIVE.o: in function `mul_mod':
emu_inc_bignum_operations.c:(.text+0x36c0): multiple definition of `mul_mod'; obj/emu_inc_ecc_secp256k1.NATIVE.o:emu_inc_ecc_secp256k1.c:(.text+0x16b0): first defined here
collect2: error: ld returned 1 exit status
make: *** [src/Makefile:644: libhashcat.so.6.2.5] Error 1
make: *** Waiting for unfinished jobs....
```

It seems that it's impossible to build `libhashcat.so`...

PR #3300, opened 2022-05-30
radmin3: use hexadecimal instead of .incbin

philsmd:
Since we noticed that some compilers (old ones ?) indeed had some problem with our `.incbin` approach to include binary data, we need to roll back to the "hexadecimal" representation of the lookup table instead. This of course comes with a little bit more bytes that need to be stored in `include`, but it's needed and worth it, if we can't guarantee that the `.incbin` works across all our supported systems.

Thanks


jsteube:
Many thanks!

Issue #3256, opened 2022-04-10
Consider releasing a flatpak package


axtux:
Hello,

I am using an LTS OS for system stability purposes and the hashcat version shipped with it is quite old. Right now, I download source code, compile it and install manually. To make things easier for me but also for any potential user that wishes an up-to-date version, you could consider releasing a flatpack package. Already discussed in #1667 but the issue has been closed by the requestor while it was not resolved.

Some links:
- https://docs.flatpak.org/en/latest/first-build.html
- https://github.com/flathub/flathub/wiki/App-Submission

nilesh05apr:
is it done yet? I would love to work this issue.

mdawsonuk:
> is it done yet? I would love to work this issue.

If you're interested in working on this, we're migrating to GitHub Actions and there's an action for generating a flatpack package if you want to take a look at integration once that PR is merged

StayPirate:
Any update on that? Having a flatpak with all the CUDA/OpenCL/HIP runtimes setup by default would be very cool.

PR #3301, opened 2022-06-01
GitHub workflow for pushing or creating pull request

ventaquil:
Workflow to build application after pushing or creating pull request for `master` branch.

As you can see this workflow builds application on Windows, Ubuntu and MacOS.

Some extra steps (e.g. with checking code style) can be added.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -27,6 +27,7 @@ on:
 jobs:
   build:
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]  # todo: windows-latest - lack of iconv development files
         shared: [0, 1]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,6 +12,7 @@ env:
               rules/
               example*
               hashcat*
+              libhashcat*
 
 on:
   push:
@@ -44,17 +45,51 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest]  # todo: windows-latest - lack of iconv development files
+        os: [ubuntu-latest, macos-latest, windows-latest]  # todo: windows-latest - lack of iconv development files
         shared: [0, 1]
     runs-on: ${{ matrix.os }}
     steps:
-      - uses: actions/checkout@v3
-      - name: Build
+      - name: Checkout Hashcat
+        uses: actions/checkout@v3
+      - name: Checkout win-iconv
+        uses: actions/checkout@v3
+        if: ${{ matrix.os == 'windows-latest' }}
+        with:
+          repository: win-iconv/win-iconv
+          path: 'win-iconv/'
+          clean: false
+      - name: Install WSL
+        uses: Vampire/setup-wsl@v1
+        if: ${{ matrix.os == 'windows-latest' }}
+        with:
+          additional-packages:
+            gcc-mingw-w64-x86-64
+            g++-mingw-w64-x86-64
+            make
+            git
+            patch
+      - name: Setup WSL
+        if: ${{ matrix.os == 'windows-latest' }}
+        shell: wsl-bash {0}
+        run: cd win-iconv/ && patch < ../tools/win-iconv-64.diff && sudo make install
+      - name: Build (Win)
+        if: ${{ matrix.os == 'windows-latest' }}
+        shell: wsl-bash {0}
+        run: make win
+      - name: Build (Unix)
+        if: ${{ matrix.os != 'windows-latest' }}
         env:
           SHARED: ${{ matrix.shared }}
         run: make
-      - name: Generate artifacts (shared)
+      - name: Generate static artifacts
+        if: ${{ matrix.shared == 0 }}
         uses: actions/upload-artifact@v3
         with:
-          name: hashcat-${{ matrix.os }}-${{ matrix.shared }}
+          name: hashcat-${{ runner.os }}-static
+          path: ${{ env.include_paths }}
+      - name: Generate shared artifacts
+        if: ${{ matrix.shared == 1 }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: hashcat-${{ runner.os }}-shared
           path: ${{ env.include_paths }}
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -2,17 +2,17 @@ name: Build
 
 env:
   include_paths: |
-              charsets/
-              docs/
-              extra/
-              layouts/
-              masks/
-              modules/
-              OpenCL/
-              rules/
-              example*
-              hashcat*
-              libhashcat*
+    charsets/
+    docs/
+    extra/
+    layouts/
+    masks/
+    modules/
+    OpenCL/
+    rules/
+    example*
+    hashcat*
+    libhashcat*
 
 on:
   push:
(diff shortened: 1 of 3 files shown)


ventaquil:
Another one reason to move to GH Actions? 🙃 

https://ci.appveyor.com/project/jsteube/hashcat/builds/43762450/job/srnt2ryxj484q651

jsteube:
Let's try and see what happens :) Is there any way to see results before merging?

ventaquil:
You can check them in my repository - https://github.com/ventaquil/hashcat/pull/1.

When you enter into _Actions_ you will see also buggy commit (commited to `master`).

I added also `shared` strategy to build binary with `SHARED=0` and `SHARED=1` env variables.

Currently all actions will be stopped when error occurs - as far as I know it can be changed.

ventaquil:
Another cool feature - after creating a tag tests will be started. Currently all of jobs will fail because of GH-3337.

jsteube:
I understand. This is like a GitHub internal replacement for Appveyor? It supports Windows, Ubuntu and MacOS? From your repository I can see that Windows was canceled. Is it not working?

ventaquil:
That's it, GitHub Actions is GitHub builtin CI/CD system. It supports MacOS, Windows, Ubuntu and self-hosted runners.

I added option to do not stop other jobs when one fails - see https://github.com/ventaquil/hashcat/actions/runs/2549088867.

However I noticed that Windows runners don't have `iconv` library. I will try to fix it.

jsteube:
We had many problem with Appveyor in the past. I think @philsmd also would like to see it being replaced, right?

ventaquil:
Status of this workflow:
* The workflow will run when new commits will be pushed to `master` branch or when someone will create a PR.
* The workflow will run when new tag `v*` will appear.
* The workflow will run only when sources from `OpenCL/`, `include/` or `src/` are changed as well as tools from `tools/`.
* During the workflow code will be build with static and shared linking options.
* Only latest Ubuntu and MacOS images are used - Windows one is lack of iconv development files.

I opened discussion to add support for iconv on Windows image but will see the result - https://github.com/actions/virtual-environments/discussions/5814.

philsmd:
well, I have at least nothing against replacing it (`Appveyor`) !

If this works better for us, I'm very happy with this change.

That said, I am also not very familiar with `Workflow for GitHub Actions` so not sure if my opinion counts too much here.

We could/should at least try it, I guess.

ventaquil:
Issues like this GH-3256 can be done with extra workflows (possibly easy to do with [actions/flatpak-builder](https://github.com/marketplace/actions/flatpak-builder)).

mdawsonuk:
Throwing in my two cents here, GitHub actions is super super powerful. You could also add artifacts so that you can download the compiled build of hashcat for that commit after the tests are also done. 

`actions/upload-artifact@v3` is what you'd want to use to upload after zipping up the executables

ventaquil:
Hey @mdawsonuk

Thank you for your comment. I am not an expert in GitHub Actions so any suggestions/recommendations are welcome here! I will try to deal with [`actions/upload-artifact`](https://github.com/actions/upload-artifact) in my spare time and add proper commit.

mdawsonuk:
I can take a look shortly and suggest the extra job to add to the YAML file if you would like?

ventaquil:
If you have any experience and you can make proper changes I think that you can create PR to [`github-workflow`](https://github.com/ventaquil/hashcat/tree/github-workflow) branch in my repo [`ventaquil/hashcat`](https://github.com/ventaquil/hashcat).

As far as I know there is no other way to add your changes to this PR.

Thank you!

mdawsonuk:
Alright I've got the upload artifact functionality working, ~~my concern is that the hashcat executable itself doesn't appear to be collected correctly as an artifact~~, as it's not being included with the zipped artifact. I'll have the PR over to you shortly

Edit: I'm stupid, `hashcat` **obviously** won't be picked up by the wildcard `hashcat.*`

mdawsonuk:
@ventaquil ready for merging https://github.com/ventaquil/hashcat/pull/5

mdawsonuk:
Just thinking for Windows builds, as the windows-latest image comes with WSL pre-installed, it might be a good idea to use WSL to build Hashcat for Windows, rather than screwing around with MSYS2 or Cygwin

ventaquil:
@mdawsonuk I have tried to add `windows-latest` runner but there is lack of iconv library - see https://github.com/ventaquil/hashcat/runs/7075487788.

I don't know how to add it properly. I am not sure that WSL can be a solution but you can try to do it.

mdawsonuk:
I use WSL for my local Windows dev - at the very least it's a good work around for no iconv

mdawsonuk:
Can confirm that WSL building (in line with the hashcat documentation) works without issue 👍🏻

ventaquil:
Instead of WSL building I prepared configuration with MSYS2. It's building Hashcat and working very well.

Thanks @mdawsonuk for idea with artifacts 👍

ventaquil:
@jsteube @philsmd you can check example result here https://github.com/ventaquil/hashcat/actions/runs/2673637097.

mdawsonuk:
> Another cool feature - after creating a tag tests will be started. Currently all of jobs will fail because of [GH-3337](https://github.com/hashcat/hashcat/issues/3337).

Are we actually running tests? As far as I'm aware, we're not. 

I'm also a proponent of smaller PRs so perhaps we could get this PR merged and then add running tests in a separate PR (I'm more than happy to work on that myself)

ventaquil:
> > Another cool feature - after creating a tag tests will be started. Currently all of jobs will fail because of [GH-3337](https://github.com/hashcat/hashcat/issues/3337).
> 
> Are we actually running tests? As far as I'm aware, we're not.

Nope but shared binary building failed before (fixed in GH-3337).

> I'm also a proponent of smaller PRs so perhaps we could get this PR merged and then add running tests in a separate PR (I'm more than happy to work on that myself)

Totally agree, I am waiting for core team accept and merge. Then future work can be done.

However I am afraid that running tests will be impossible without self-hosted runners.

I am thinking about automatic issues/PRs labeling (or even closing), building release packages and adding archives to new releases and so on.

mdawsonuk:
> However I am afraid that running tests will be impossible without self-hosted runners.

How so? 

> I am thinking about automatic issues/PRs labeling (or even closing), building release packages and adding archives to new releases and so on.

Yes these are all great ideas

ventaquil:
> > However I am afraid that running tests will be impossible without self-hosted runners.
> 
> How so?

As far as I know provided runners do not have support for GPUs (so we cannot use CUDA or OpenCL) - see [GitHub Actions / Supported runners and hardware resources](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).

GPU support can be achieved in self-hosted runners - see [GitHub Actions / Using self-hosted runners in workflow](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-custom-labels-to-route-jobs). Without those runners the only possible solution is to use PoCL - it should be enough.

mdawsonuk:
> As far as I know provided runners do not have support for GPUs (so we cannot use CUDA or OpenCL) - see [GitHub Actions / Supported runners and hardware resources](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources).
> 
> GPU support can be achieved in self-hosted runners - see [GitHub Actions / Using self-hosted runners in workflow](https://docs.github.com/en/actions/hosting-your-own-runners/using-self-hosted-runners-in-a-workflow#using-custom-labels-to-route-jobs). Without those runners the only possible solution is to use PoCL - it should be enough.

I believe support for this has been very recently added https://github.com/github/roadmap/issues/505

mdawsonuk:
Just wondering as an aside, maybe the artifact names should be lowercase? A bit more in line with common naming conventions.

ventaquil:
@mdawsonuk by default `runner.os` returns `Windows`, `Linux` or `macOS` so I chose to use Pascal-case here.

It looks complicated to change case convention since there is no default `lowercase` function - github-community/community#10553.

mdawsonuk:
> @mdawsonuk by default `runner.os` returns `Windows`, `Linux` or `macOS` so I chose to use Pascal-case here.
> 
> 
> 
> It looks complicated to change case convention since there is no default `lowercase` function - github-community/community#10553.

I thought the same initially but then it's a different build step for each OS, so actually using runner.os is redundant?

ventaquil:
> I thought the same initially but then it's a different build step for each OS, so actually using runner.os is redundant?

Ah yeah, my fault. I made copy-paste and `runner.os` is not required anymore. I will prepare commit fix.

jsteube:
OK this is completely new territory for me, too. Let's see how it works out :)

ventaquil:
Thanks for merge, I hope that actions will be useful for everyone 👍

ventaquil:
I looks that it works quite okay - https://github.com/hashcat/hashcat/actions/runs/2719841951.

jsteube:
Looks good, yeah. Wondering if we could make windows work. The problem is only the iconv? Is there any way to trigger the test offline (on my dev system for instance)?

ventaquil:
Hey @jsteube

Windows jobs are looking fine imho (with big thanks for help to @mdawsonuk).

Jobs

ventaquil:
Also you can take a look at [nektos/act](https://github.com/nektos/act) to run workflow locally.

PR #3307, opened 2022-06-04
add make_utf16beN_S () and sha1_update_utf16beN ()

philsmd:
I think it makes sense for some kernels (e.g. `radmin3`, another pull request will follow soon), to have a non-SIMD / non-vector version of our `utf16beN` conversion. To do this, we need both a new function for the utf16 converstion and the missing `SHA1` function, that doesn't use SIMD/vector: `sha1_update_utf16beN ()`.

I think this is better than using the currently used approach (`sha1_init_vector_from_scalar ()` -> `sha1_update_vector_utf16beN ()` -> `sha1_final_vector ()` , even though we don't use the vectors at all here, so it's also very confusing).

What do you think ?

Thanks

(of course it doesn't, or at least shouldn't, make much speed difference... but it is still conceptually cleaner)

### DIFF
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -36773,6 +36773,44 @@ DECLSPEC void make_utf16be_S (PRIVATE_AS const u32 *in, PRIVATE_AS u32 *out1, PR
   #endif
 }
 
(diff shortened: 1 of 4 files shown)


jsteube:
Yes I think it makes sense, it's important to keep it conceptionally clean.

PR #3310, opened 2022-06-07
-m 27700: fix use of varied iteration counts

half-duplex:
Older versions of [Android "Bitcoin Wallet"](https://play.google.com/store/apps/details?id=de.schildbach.wallet) set the scrypt iterations [to 4096](https://github.com/bitcoin-wallet/bitcoin-wallet/blob/v4.63/wallet/src/de/schildbach/wallet/ui/EncryptKeysDialogFragment.java#L65), which conflicts with the minimum length of 5 and the hardcoded N=16384.

Example hash, pin=123456
```
$multibit$3*4096*8*1*976455c869487bac*67366a6ca0758389c524bb78f13a832f8826b7eba93ff4cc6842f669c7094835
```

From the documentation I've found:
N: bitcoinj once [considered 512 sufficient](https://github.com/bitcoinj/bitcoinj/blob/v0.12.1/core/src/main/java/org/bitcoinj/crypto/KeyCrypterScrypt.java#L95), and some implementations [cap at `2^0xff`](https://github.com/Tarsnap/scrypt/blob/83cda8b/lib/scryptenc/scryptenc.c#L220) (77 decimal digits).
r: "8 should still be optimal" ([2017](https://words.filippo.io/the-scrypt-parameters/))
p: All recommend 1

Filippo Valsorda [recommends](https://words.filippo.io/the-scrypt-parameters/) 2^20 (7 decimal digits) for file encryption, so [scrypt/8900](https://github.com/hashcat/hashcat/blob/f4e1bdf675d30dd01c9d2167f80090f55203faf7/src/modules/module_08900.c#L301) and other scrypt implementations may also need len_max raised.

### DIFF
--- a/src/modules/module_27700.c
+++ b/src/modules/module_27700.c
@@ -296,14 +296,14 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
   token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_DIGIT;
 
-  token.len_min[2] = 5;
-  token.len_max[2] = 5;
+  token.len_min[2] = 1;
+  token.len_max[2] = 10;
   token.sep[2]     = '*';
   token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks!

PR #3312, opened 2022-06-08
gid >= GID_CNT wasn't checked in some kernel functions

philsmd:
I just noticed, by accident, that I completely removed the check for `gid >= GID_CNT` in the -m 29200 = `Radmin 3` kernel with this fix: https://github.com/hashcat/hashcat/pull/3309.

of course this was a mistake...

but after that I've found out that some other kernel functions also for now don't have this check, which for sure is a bug/problem.

I've also noticed that all the `*_tm` kernels (`OPTS_TYPE_TM_KERNEL` on the host / `KERN_ATTR_TM` in OpenCL) do not check this. This of course could also be correct, since we have some fixed settings there, but I think it's better to have it and comment it out and maybe you (@jsteube ) could double-check if it is correct that `*_tm` kernels don't need the `gid >= GID_CNT` check.

Thank you very much
( btw some kernels even got a bit faster, also in `-b` / benchmarks, due to this check, for me, could this be correct/significant !? )

### DIFF
--- a/OpenCL/m01500_a3-pure.cl
+++ b/OpenCL/m01500_a3-pure.cl
@@ -1889,6 +1889,8 @@ KERNEL_FQ void m01500_tm (KERN_ATTR_TM)
 {
   const u64 gid = get_global_id (0);
 
+  // if (gid >= GID_CNT) return;
+
   const u32 block = gid / 32;
   const u32 slice = gid % 32;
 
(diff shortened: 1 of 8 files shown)


jsteube:
I think is a bug transported by copy/paste from one to the next kernel. We for sure need the check everywhere.

PR #3313, opened 2022-06-08
add unit tests for -m 29000

philsmd:
I've noticed that we currently do not have any unit tests (perl) for our new hash mode -m 29000 = `sha1($salt.sha1(utf16le($username).':'.utf16le($pass)))`, so I've just added the perl module.

I think it's better to have this too, just to be sure everything is working correctly and as expected (sometimes it even nice to have the perl - higher level code - in front of you, just to be quickly get an idea what the hash mode algorithm is = convenience).

Thanks

### DIFF
--- a/tools/test_modules/m29000.pm
+++ b/tools/test_modules/m29000.pm
@@ -1,0 +1,57 @@
+#!/usr/bin/env perl
+
+##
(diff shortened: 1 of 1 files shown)


jsteube:
Perfect code :)

PR #3317, opened 2022-06-11
fixes vector bug in -m 27800 = MurmurHash 3 with -a 3

philsmd:
After adding the perl unit tests for -m 27800 = `MurmurHash 3` I have discovered that some tests fail. The problem was actually just due to this line in code:
https://github.com/hashcat/hashcat/blob/15a0ad5903688a0192ac42893c7c6fa62f3c5fb3/OpenCL/m27800_a3-optimized.cl#L67

which converts a `u32x *` into just a `u8 *` and could results in many false positives.

This problem is restricted to `-a 3` and `--backend-vector-width` > 1, or `-V` > 1 in `test.sh`.

The following command with the current/not-fixed code does NOT work/crack for me (note that which passwords crack and which don't might also depend a little bit on the hardware too, kernel power / shaders etc, i.e. a typical false negative bug):
```
hashcat -a 3 -m 27800 -O --backend-vector-width 4 f7fdeae4:49095171 ?d?d
```

password is: `64`

(works without the `--backend-vector-width` parameter, or - depending on hardware - when using the default one i.e. set to 1)

another example: `8602c7d2:92900203:0` (using ?d as mask)

Note: do not wonder if this is "just" a optimized kernel bug (`-O`), because this hash mode currently only has optimized kernels. Meaning, this also happens if you do not use/set `-O` in your command line (so it could trigger whenever higher vector widths are used in mask attacks). The problem seems to only affect -a 3 attacks, which is also typical and not surprising, because vectorization is **NOT** used (similar to many other hash modes) in the `-a 0` and `-a 1` kernels.

Thank you very much

### DIFF
--- a/OpenCL/m27800_a0-optimized.cl
+++ b/OpenCL/m27800_a0-optimized.cl
@@ -15,22 +15,25 @@
 #include M2S(INCLUDE_PATH/inc_simd.cl)
 #endif
 
-DECLSPEC u32 Murmur32_Scramble(u32 k)
+DECLSPEC u32 Murmur32_Scramble (u32 k)
 {
   k = (k * 0x16A88000) | ((k * 0xCC9E2D51) >> 17);
+
   return (k * 0x1B873593);
 }
 
-DECLSPEC u32 MurmurHash3(const u32 seed, PRIVATE_AS const u32 *data, const u32 size)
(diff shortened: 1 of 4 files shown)


jsteube:
Nice find, thanks!

PR #3326, opened 2022-06-17
Increase salt length for PKCS#8

### LINKED ISSUE #3323: PKCS#8 modes accept only 8-byte salts
Modes 24410 and 24420 (PKCS#8 Private keys with PBKDF2-HMAC-SHA1 or PBKDF2-HMAC-SHA256) have a limitation: exactly 8 bytes of salt are required - no more, no less. This works fine for any PKCS#8 or PKCS#12 files generated from `openssl pkcs8` or `openssl pkcs12` as that tool is hardcoded to always use 8 bytes of random salt when generating those files: there's no options for changing the length of auto-generated salt or for specifing the exact salt value.
Java's `keytool` (as tested with version 18.0.1) can (and does by default) create PKCS#12 keystores. These stores use the same EncryptedPrivateKeyInfo as PKCS#8, so they should work with these modes. However, `keytool` uses 20-byte salts which cause a `Hashfile 'hashcat_pkcs8_from_jks.hash' on line 1 ($PEM$2...4d489ec7d92bf3ba81430ae14b88413e): Token length exception` error when run.
`keytool` is horrible to work with as it can't import certs and private keys separately: you still need to use something else to create a generic PKCS#12 file which `keytool` imports, re-encrypts, and adds some attributes to, resulting in a new PKCS#12 file (with 20-byte salts).
I've attached a source PKCS#12 file that was created with OpenSSL:
`>openssl pkcs12 -export -in hashcat.crt -inkey hashcat.key -name hashcat -out openssl.pkcs12`
This file uses 8-byte salts, so you could write a script or manually pull out the EncryptedPrivateKeyInfo (931 bytes starting at offset 53) and run it through pem2john.py (or manually pull out the correct values). However, lets continue with `keytool`:
`>keytool -importkeystore -srckeystore openssl.pkcs12 -destkeystore hashcat.jks -srcstoretype pkcs12 -alias hashcat`
It asks for both the new keystore password (twice) and the source keystore password, and then generates the JKS PKCS#12 file that includes attributes and uses 20-byte salts. Again, we can write a script or manually pull out the EncryptedPrivateKeyInfo (1344 bytes starting at offset 82) and go from there. I did it manually with a hex editor and then used pem2john.py:
`>python pem2john.py hashcat_pkcs8_from_jks.key`
and got the following:
`$PEM$2$4$854f4b91364c70859ac4d01f267eecb64f2b267d$10000$bfe88713bd8c9b4e8f6669c82c0794de$1232$7a0c79e9e9b456c5...`
This appears correct (2 = SHA256, 4 = AES-256-CBC, 40 hex chars salt, 10000 iterations, 32 hex chars IV, 1232 data length, and 2464 hex chars of data), but Hashcat won't take it since the salt field isn't exactly 16 hex chars. If you truncate the salt field to exactly 16 hex chars you can get Hashcat to start (though it will never find the correct password, obviously).
I'm happy to provide any additional examples or do some testing.
ZIP archive includes the original OpenSSL-generated PKCS#12 file, the `keytool` generated PKCS#12 file, the extracted PKCS#8 (EncryptedPrivateKeyInfo) file, and the corresponding hash file generated by pem2john.py. All passwords are `hashcat`
[PKCS12.zip](https://github.com/hashcat/hashcat/files/8910767/PKCS12.zip)

 - OS: Windows 10 Professional
 - Hashcat Version: 6.2.3

Hardware specifics should not be necessary as Hashcat _does_ work correctly with 8-byte salts, but for completeness:
```
>nvidia-smi
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 511.65       Driver Version: 511.65       CUDA Version: 11.6     |
|-------------------------------+----------------------+----------------------+
| GPU  Name            TCC/WDDM | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Quadro P2200       WDDM  | 00000000:01:00.0  On |                  N/A |
| 51%   44C    P0    22W /  75W |    461MiB /  5120MiB |      1%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+`
```

ventaquil:
Fixes GH-3323.

It should be relatively easy to increase salt length in the future (when it will be necessary) - just change `PKCS_MAX_SALT_LEN`.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -65,6 +65,7 @@
 - Fixed Unit Test false negative if there are spaces in the filesystem path to hashcat
 - Fixed Unit Test salt-max in case of optimized kernel, with hash-type 22 and 23
 - Fixed wordlist handling in -m 3000 when candidate passwords use the $HEX[...] syntax
+- Fixed accepted salt length by PKCS#8 Private Keys modules
 
 ##
 ## Technical
--- a/src/modules/module_24410.c
+++ b/src/modules/module_24410.c
@@ -43,6 +43,11 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig, 
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+#define PKCS_MIN_SALT_LEN     ( 8)
+#define PKCS_MAX_SALT_LEN     (32)
+#define PKCS_MIN_SALT_HEX_LEN (PKCS_MIN_SALT_LEN * 2)
+#define PKCS_MAX_SALT_HEX_LEN (PKCS_MAX_SALT_LEN * 2)
+
 typedef struct pkcs_sha1_tmp
 {
   u32  ipad[5];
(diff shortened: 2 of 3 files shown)


jsteube:
Thanks for the contribution. You may want to check hex_decode() and hex_encode() maybe it can be used to simplify the code.

PR #3320, opened 2022-06-11
fixes #3090 closes #3091: Added Bitcoin WIF/P2PKH private key algo

philsmd:
This PR should replace the one from @b0lek (#3091) and improves it in speed/performance, readability and consistency with other plugin implementations (e.g. using fast hashes for algos that do **NOT** use high iteration counts) etc.

The work of course is mostly of @b0lek with only minor changes from myself (the suggestions/changes are also discussed in #3091).

My changes primarily focus on separating the two algos into `28501` (for `compressed` addresses/keys) and `28502` (for `uncompressed`) and changing the original implementation of @b0lek from a slow hash algo to a fast `hashcat` kernel (which improved speed significantly, just test it if you want !).

@b0lek : thank you very much for the contribution

Thanks guys

BTW: #3091 can be closed whenever we merge this pull request

Issue #3090, opened 2021-12-29
Bitcoin private key recovery


b0lek:
Hi all,

I've a proposal for new algorithm that will recover BTC private keys. 

Use case:
- you have printed Bitcoin off-line wallet
- after years of keeping it hidden you want to spent your well kept BTCs
- you are so happy with new BTC value that you drink a bit :-)
- some wine or coffee gets on the paper and you loose some characters from your private key
- you want to recover some lost digits of your BTC private key

How it will work:
- your public P2PKH `1LhFXCGReUCuARYpqdj6x8kf1Pe1n2Ns18`
- your private key for this one is: `5KbeS8qxKHMYtYfb1iPUQwHJ7kjY2AVFvMNVRAStftqrP72H3ma`
- lost characters you replace with `?1` to create a private key mask
- so for example if you lost first 4 characters and last 2 characters, your mask would be: `?1?1?1?1S8qxKHMYtYfb1iPUQwHJ7kjY2AVFvMNVRAStftqrP72H3?1?1`
- run `hashcat -m 98700 -a 3 1LhFXCGReUCuARYpqdj6x8kf1Pe1n2Ns18 -1 123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz  ?1?1?1?1S8qxKHMYtYfb1iPUQwHJ7kjY2AVFvMNVRAStftqrP72H3?1?1`
- ... and wait :-)

I have prepared this algorithm and it is working :)

I will open PR with my proposal and wait for your comments. For sure it can be improved and optimized.

It's maybe not obvious way of using hashcat but I believe it can be and interesting addition to hashcat.

Chick3nman:
Just FYI to anyone eventually using this mode to recover a lost bitcoin private key, we always know the first letter of a standard mainnet private key will be 5, testnet will start with 9, and compressed keys, which you may need to deal with here, will start with L or K on the mainnet and c on the testnet.

ashraful1980:
Error massage as bellow:
└─$ ./hashcat -m 98700 -a 3 1LhFXCGReUCuARYpqdj6x8kf1Pe1n2Ns18 -1 123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz ?1?1?1?1S8qxKHMYtYfb1iPUQwHJ7kjY2AVFvMNVRAStftqrP72H3?1?1
[...]

PR #3336, opened 2022-06-20
Add -m 28503/28504 = Bitcoin WIF, P2WSH/bech32/segwit address

philsmd:
As recently discussed on this comment, https://github.com/hashcat/hashcat/issues/3090#issuecomment-1159411910 , it makes sense to also support this newer type of bitcoin addresses, because the support comes basically for free because we do not need any additional OpenCL/CUDA kernel for this type of addresses, but can re-use the kernels from -m 28501/-m 28502 (see https://github.com/hashcat/hashcat/pull/3320).

The interesting thing about this type of addresses is that they can be "cracked" the same way, it's just the encoding (and checksum) that is different, at least when it comes to comparing how a user needs to attack the WIF<->address cracking (of course the bitcoin transactions/network etc have a totally different and more significant changes, but it doesn't bother us here).

Many details about this are already discussed in https://github.com/hashcat/hashcat/issues/3090#issuecomment-1159411910 , for instance that users could even **externally** convert one address type into the other (it's just address/ripemd hash encoding + checksums).

I think it still makes sense to add support for this, the user could now easily find it in the `--help` output and see the address examples with `--hash-info` etc.

Thanks

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -6,6 +6,7 @@
 
 - Added hash-mode: Amazon AWS4-HMAC-SHA256
 - Added hash-mode: Bitcoin WIF private key (P2PKH)
+- Added hash-mode: Bitcoin WIF private key (P2WSH, Bech32)
 - Added hash-mode: BLAKE2b-512($pass.$salt)
 - Added hash-mode: BLAKE2b-512($salt.$pass)
 - Added hash-mode: DPAPI masterkey file v1 (context 3)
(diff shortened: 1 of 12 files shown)


jsteube:
Amazing, many thanks

ashraful1980:
Please give me full command where is adding  -m 28501/-m 28502...

philsmd:
in beta: https://hashcat.net/beta/

PR #3342, opened 2022-06-25
refactor/simplify code of MurmurHash3

philsmd:
This is just a little cleanup/refactor and simplification of the code within the `OpenCL` kernel functions of MurmurHash3 (-m 27800).

It seems that this simple changes also made the benchmark a "little bit" faster.

I think the better readability is also a great advantage of this change.

Thanks

### DIFF
--- a/OpenCL/m27800_a0-optimized.cl
+++ b/OpenCL/m27800_a0-optimized.cl
@@ -26,7 +26,7 @@ DECLSPEC u32 MurmurHash3 (const u32 seed, PRIVATE_AS const u32 *data, const u32 
 {
   u32 checksum = seed;
 
-  const u32 nBlocks = (size / 4);
+  const u32 nBlocks = size / 4; // or size >> 2
 
   if (size >= 4) // Hash blocks, sizes of 4
   {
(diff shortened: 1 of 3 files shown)


Miha22:
Why you were scared to shift 3 bits to the right here:   `const u32x val = data[nBlocks] & (0x00ffffff >> ((3 - (size & 3)) * 8));`

philsmd:
@Miha22 
> Why you were scared to shift 3 bits to the right here: `const u32x val = data[nBlocks] & (0x00ffffff >> ((3 - (size & 3)) * 8));`

What would be your **working** code ?

Also, changing this "constants" probably won't make the code run faster, the optimizers/compilers will find this out themself (if your comment was about just a little different shifts).... the major change was to avoid several branches and `switch` statements etc... I'm not sure if there is a better version which the compilers can't figure out and optimize themself, but I am happy to see what your suggestion of improvement of my change would be.

BTW: the `3 - x` just means that we don't **increase** (directly) with `x`, but just **decrease** the count (kind of a `anti-proportional` thingy, the greater the `x` value, the less we shift), maybe it's a misunderstanding ? not sure

--- 

**update**: of course you could re-write that specific calculation involving `3 - x` like this:
```
 (3 - (size & 3)) * 8 = 24 - (size & 3) * 8 = 24 - ((size & 3) << 3)
```

but as said this doesn't change anything (or at least it shouldn't, the compiler/optimizer will optimize this), also it is always 3 operations (and with the "24" it is also more confusing math, it is very confusing why the value 24 is used, that would not be something that looks easy to read and doesn't change performance)

philsmd:
Didn't get any new answer, I think this was just a misunderstanding and we can merge this without any hesitation.

PR #3343, opened 2022-06-25
fix code, add newline in 28503/28504

philsmd:
For some strange reason, it seems that I have missed to add a newline after the `input_tokenizer ()` call.
This of course was a mistake and this cleanup/code style fix is needed to make this code section more readable and fix this ugly code line.

It's just a minor fix to add a newline between `input_tokenizer ()` call and the `rc_tokenizer` variable (return code) check.
Thanks


Miha22:
honestly dude?

jsteube:
I think it's important this kind of fixes it makes it possible to "see" source instead of "reading" it.

PR #3347, opened 2022-06-29
fix test/verify for -m 28600

philsmd:
For me the `verify` code of the perl unit test for -m 28600 = `PostgreSQL SCRAM-SHA-256` was not working at all.

With this minor fix, the `verify` (but also `passthrough` and `single`, of course) will work again for this particular unit test.

Thanks

### DIFF
--- a/tools/test_modules/m28600.pm
+++ b/tools/test_modules/m28600.pm
@@ -15,9 +15,9 @@ use Crypt::PBKDF2;
 
 sub module_constraints { [[0, 256], [28, 28], [-1, -1], [-1, -1], [-1, -1]] }
 
-my $ITERATIONS = 4096;
-my $HMAC_SALT  = "Server Key";
-my $HMAC_SALT_2  = "Client Key";
+my $ITERATIONS  = 4096;
+my $HMAC_SALT   = "Server Key";
+my $HMAC_SALT_2 = "Client Key";
 
 sub module_generate_hash
 {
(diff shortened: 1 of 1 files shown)


philsmd:
Minor note to @b0lek (which PR'ed the code with https://github.com/hashcat/hashcat/commit/415e95ab386744f4503d7d25d59cb489a2a8f7f1):
I'm just curious how this "Client Key" is supposed to work... I understand that it is not needed to crack the hash (we have both `client` and `server` part within the hashes i.e. hash format).

So my analysis is this:
`hPciC1CcnBna3szR8Mf3MVc8t0W7QPbIHoMMrh4zRV0=` (base64 encoded, -m 1460):
`84f7220b509c9c19dadeccd1f0c7f731573cb745bb40f6c81e830cae1e33455d:c7ee7ee903f6ab77a277501450716dc3b3fb60676548cd5938b6b57a1ba4f338` with password "Server Key" works
while `iRw3qwTp18uaBnsTOEExbtgWdKeBMbSSnZvqD4sdqLQ=` (-m 1460):
`891c37ab04e9d7cb9a067b133841316ed81674a78131b4929d9bea0f8b1da8b4:c7ee7ee903f6ab77a277501450716dc3b3fb60676548cd5938b6b57a1ba4f338 with password` "Client Key" does **NOT** work

This is the example hash from https://github.com/hashcat/hashcat/blob/e0fae5c7ab0e94dd8c536c0909fef429f2628f5b/src/modules/module_28600.c#L30-L32

Could you please @b0lek answer how this is supposed to work with password "Client Key" (see https://github.com/hashcat/hashcat/blob/master/tools/test_modules/m28600.pm#L20) ?

BTW: this is just a note, this commit is **STILL NEEDED** to make verify work, so this is just an extra comment/curiosity @jsteube 

--- 

**update**: I think I figured it out, we need another call to `sha256 () ` for the `Client Key` !!! (just found this by accident: https://stackoverflow.com/questions/68400120/how-to-generate-scram-sha-256-to-create-postgres-13-user/71102510#71102510)

b0lek:
@philsmd 
Yes it requires one extra sha256(). I remember I had to dig into PostgreSQL source code.
The trick is done in line 216 in: 
https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/common/scram-common.c;h=126862592999b509ada293dea13d32123792eb63;hb=5513dc6a304d8bda114004a3b906cc6fde5d6274#l216
Thanks for fixing.

PR #3352, opened 2022-06-29
Add -m 28505/28506, Bitcoin WIF, P2SH(P2WPKH)/compat address

philsmd:
This is the last missing type of "Bitcoin address" that was not supported yet (the compatibility, `compat`, `P2SH` or rather `P2SH(P2WPKH)` address). The details again are already discussed over here: https://github.com/hashcat/hashcat/issues/3090#issuecomment-1159411910

These are the addresses starting with a "3" (we already added support for "1" and "bc1" addresses recently).

This type of address is very similar, but it needs an extra round of `RIPEMD(SHA256())`, so we need extra `OpenCL`/`CUDA` kernels for this type of bitcoin address (remember we didn't need them for just `P2WPKH` addresses, starting with "bc1").

This should now be the final address type that we could support easily after this was implemented https://github.com/hashcat/hashcat/issues/3090 with https://github.com/hashcat/hashcat/pull/3320 (so we have support of **3** / **3**, at least this is what I know about the most common addresses, for which the user could have a partially unknown `WIF`).

Thanks

### DIFF
--- a/OpenCL/m28505_a0-pure.cl
+++ b/OpenCL/m28505_a0-pure.cl
@@ -1,0 +1,365 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+//#define NEW_SIMD_CODE
+
+#define SECP256K1_TMPS_TYPE PRIVATE_AS
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_rp.h)
+#include M2S(INCLUDE_PATH/inc_rp.cl)
+#include M2S(INCLUDE_PATH/inc_scalar.cl)
+#include M2S(INCLUDE_PATH/inc_hash_base58.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha256.cl)
+#include M2S(INCLUDE_PATH/inc_hash_ripemd160.cl)
+#include M2S(INCLUDE_PATH/inc_ecc_secp256k1.cl)
+#endif
+
+KERNEL_FQ void m28505_mxx (KERN_ATTR_RULES ())
+{
+  /**
+   * modifier
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= GID_CNT) return;
+
+
+  /**
+   * base
+   */
+
+  secp256k1_t preG; // need to change SECP256K1_TMPS_TYPE above to: PRIVATE_AS
+
+  set_precomputed_basepoint_g (&preG);
+
+  COPY_PW (pws[gid]);
+
+
+  /**
+   * loop
+   */
+
+  for (u32 il_pos = 0; il_pos < IL_CNT; il_pos++)
+  {
+    pw_t p = PASTE_PW;
+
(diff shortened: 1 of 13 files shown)


b0lek:
I have tested it a bit and it looks very good.
You have done a lot of good optimizations :) Well done.

I see one small issue with BENCHMARK_MASK
- You use a mask with '?b?  which is not proper alphabet for base58.
- You have put a mask with ? at the beginning o of a mask

Due to that 
- I get  benchmark speed:
49692.0 MH/s (MEGA !!!)
- while in normal cracking:
26047.3 kH/s (KILO only)

That is of course because your optimizations for early rejections
- first characters optimization
- base58 alphabet rejects

That benchmark could be a bit misleading.

philsmd:
@b0lek I was also thinking about it, but it seems we need to have an extra way in the modules to not only specify the `BENCHMARK_MASK` variable (you could look at any module, they should all have this variable), but also something like `BENCHMARK_MASK_CUSTOM_CHARSET1` or (shorter variable name) `BENCHMARK_CUSTOM_CHARSET1` similar (this is currently not supported by the hashcat interface/"api" as far as I know). This "interface" variable is currently missing, but we could think of adding this. maybe also a good question for @jsteube .

Yeah, I agree that the speed is somehow misleading, the actual speed of course also depends on what the user specifies on the command line.

It's funny because I was actually thinking of exactly the same problems before pushing the first `WIF` hash mode, questions like this:
shouldn't we show a huge warning/error if a user doesn't "use charsets/base58.hcchr" or valid base58 in their mask/custom-charsets. It would be a waste of resources.

but on the other hand, this requires some (maybe a lot of) source code changes... and sometimes it's just a tradeoff that we think the user should actually just use their brains and do clever attacks (instead of dumb/invalid password recovery attacks). But this of course is debatable (warnings/errors are always good if needed/possible, to even catch some stupid mistakes or typos in the charset etc, even for advanced users).

Good points there. Thanks @b0lek 

--- 

**update**: suggested a change for a custom charset setting in benchmarks here: https://github.com/hashcat/hashcat/pull/3381

sinaimani83:
Hello <3
I have a question
I want to recover the lost characters of my wallet with the codes you are going to give, but I can't find the correct terminal command.

for example:
hashcat.exe -m 98700 -a 3 13mdm9HZikn8D6yFsHBbgwF26eEmywm29D -1 123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz L1DUMgcHYxBUkA29D3aSx7YCt87FP3iK2fkcoKjS1bwAxTjKs?1?1?1

terminal:
Cannot load module ./modules/module_98700.dll

and i can't find module_98700.dll

What is the correct terminal command and what should I do to fix my problem?

Thanks a lot <3

b0lek:
Hi @sinaimani83 

For your case you should use -m 28501.
Mode number 98700 was an old test one used. 
Finally there are 6 new modes 28501-28506 for different address types.

sinaimani83:
> 

Thank you for your quick reply, but I still have the following problem
Screenshot 2022-07-26 161426

How do I find the files required by the program?

sinaimani83:
The problem is that there are no files named module28501-28506.dll (and .so) :(

philsmd:
it's still in beta: ( https://hashcat.net/beta ), it's a new feature. we didn't have any release version since this change was committed.

LazyGeek-dev:
Screenshot 2024-05-04 124145
How to fix this ? I am running the latest hashcat 6.2.6 release file.
any help is appreciated !

philsmd:
@faiyaz-libscr  use smaller keyspace (and more clever attacks), e.g. by reducing the charset and using hashcat mask files etc

Issue #3205, opened 2022-03-04
Feature Request - Python Werkzeug Module/Parser


godylockz:
# Issue
Feature Request - Python Werkzeug Module/Parser being added to hashcat for the hash format: `method$salt$hash`

# Background
Werkzeug is a comprehensive [WSGI](https://wsgi.readthedocs.io/en/latest/) web application library. It began as a simple collection of various utilities for WSGI applications and has become one of the most advanced WSGI utility libraries.
This python package is can be used for authentication in various Python web frameworks (i.e. Flask)

# Example Hash Equivalent
This potentially would only be for automatic heuristic detection such as `method` = `sha256` would then reformat the hash and use hashcat mode `1460` for instance...
I ran into this trying to crack a similar hash to `sha256$1tNyTG7rtLMYuIos$fa8b1915621cc59f3967a9747e083ad9b59934806336a6a24afff926044334e1` plaintext: `test`
The default Werkzeug hash is `pbkdf2:sha256` however in this case `sha256` was used.
This would be equivalent to using hashcat mode 1460 `HMAC-SHA256 (key = $salt)` with some changes to the hash format: `fa8b1915621cc59f3967a9747e083ad9b59934806336a6a24afff926044334e1:1tNyTG7rtLMYuIos`

# Documentation
Full documentation of implementing this algorithm / hash format can be found here:
https://werkzeug.palletsprojects.com/en/2.0.x/utils/#werkzeug.security.check_password_hash
https://werkzeug.palletsprojects.com/en/2.0.x/utils/#werkzeug.security.generate_password_hash
https://github.com/pallets/werkzeug/blob/main/src/werkzeug/security.py

# Hash Generation
## Hashlib
Hashlib is used for selecting the specific hash method for generating a Werkzeug hash:
```python
import hashlib
print(hashlib.algorithms_available)
```
`{'md5-sha1', 'sha256', 'sha512_224', 'blake2s', 'shake_256', 'ripemd160', 'sha384', 'sha3_256', 'whirlpool', 'sha3_224', 'sha224', 'sm3', 'sha1', 'md5', 'blake2b', 'md4', 'sha512', 'shake_128', 'sha3_384', 'sha512_256', 'sha3_512'}`
## Werkzeug Generate Hashes
`generate_password_hash(password, method, salt_length)`
- password – the password to hash.
- method – the hash method to use (one that hashlib supports). Can optionally be in the format `pbkdf2:method:iterations` to enable PBKDF2.
[...]

PR #3367, opened 2022-07-04
Added Python Werkzeug MD5 module

mdawsonuk:
Continuing to iterate over hash formats raised in https://github.com/hashcat/hashcat/issues/3205 this one is MD5 (identical to -m 60)

Due to the hash being exclusive to Werkzeug, this is marked as HASH_CATEGORY_FRAMEWORK.


ventaquil:
> identical to -m 60

What is purpose to add identical module - just with different name?

mdawsonuk:
Different hash format, requested in the referenced issue

ventaquil:
Then maybe it's a good idea to make common part of signature like

```c
static const char *ST_HASH        = "werkzeug$md5$84143$7f51edecfa6fb401a0b5e63d33fc8c0e";
```

instead of simple

```c
static const char *ST_HASH        = "md5$84143$7f51edecfa6fb401a0b5e63d33fc8c0e";
```

mdawsonuk:
Perhaps I'm missing something but that's just not the format that the hash it in. It will always be in the format `md5$salt$hash`.

I've had discussions around whether this should be part of the same module as the kernel is (in this case mode 60) but it was agreed that it is more user-friendly to have it as a separate module.

PR #3362, opened 2022-07-02
fixed crossbuild for mingw/win64 issue #3361

enwony:

PR #3370, opened 2022-07-09
bech32/polymod: rename variable

philsmd:
This is a new problem introduced with this commit https://github.com/hashcat/hashcat/pull/3336.

The problem here is just that some compilers can't deal with this variable name and think the `CONST` is a ANSI C source code `keyword` and therefore it is a syntax error.

My suggestion is to just rename it to `TABLE` (because it is kind of a lookup too).

Thanks for the report (that comes from discord @ https://hashcat.net/discord)


jsteube:
This was already fixed with an older PR https://github.com/hashcat/hashcat/pull/3362 that I've merged yesterday. Thanks!

PR #3375, opened 2022-07-09
avoid OPTS_TYPE_HASH_COPY, just use digest_buf in -m 26200 = OpenEdge

philsmd:
In general, we try to avoid `OPTS_TYPE_HASH_COPY` wherever possible... in this case (in my opinion) it's wrong to use `OPTS_TYPE_HASH_COPY` because we normally only use this in exceptional cases where we need to copy some extra data that we can't fit into `salt` or `esalt` and don't want to "store" in other buffers.

but in this case everything we need is in the `digest_buf` variable, so we should just use this (as with almost every other hash mode).

Thanks


jsteube:
In general I tend to avoid OPTS_TYPE_HASH_COPY as much as possible. If we avoid it it's a good sign that all input data has been correctly parsed/stored since we can "restore" it by 100% in the encoder function.

Issue #3285, opened 2022-05-18
Terra Station (PBKDF2, SHA1, 100 rounds, 256 key size) + AES CBC


0xVavaldi:
**Is your request based on a publicly known cryptographic schema and where can we find information about?**
- The hashing function PBKDF2 and AES are well known algorithms for key derivation and encryption respectively
- The password protection for the Terra Station was reversed by me. Details about it can be found within the Terra Station extension files when searching for the string "PBKDF" in the main js file.

This is how it's implemented in the source (variable names are because it's minified):
```javascript
      var n = r.a.lib.WordArray.random(16);
      var a = r.a.PBKDF2(t, n, {keySize: 8, iterations: 100});
      var c = r.a.lib.WordArray.random(16);
      var i = r.a.AES.encrypt(e, a, {iv: c, padding: r.a.pad.Pkcs7, mode: r.a.mode.CBC});
      return n.toString() + c.toString() + i.toString();
```
Var n is the salt of PBKDF2, a random 16 Byte char array
Var a is the bytes resulting from the PBKDF2 function. It's important to note here that although the keySize is specified as 8, it's actually ignored and the default 256 is taken with the algorithm being SHA1.
Var c is the IV used in AES, a random 16 Byte char array
Var i is the bytes resulting from the AES CBC operation.

The value e, being encrypted is a 64 length hex string, more examples given below.

**Provide an example hash/database/file and a known correct password**
Example 1:
```
d2ea8d9efe17a1d910433d899bec35076941f8c63a1d080dfe4f909b2115621bkUDFGkI1dI/k19R6o2o17bSMNRnYIrSJ0h1fqxREwpgZoSVxokJj7yLqos3aZ4JBoyQkf3R5gLkS5B1w7ZLPRSXrA/tusY8h2NU3NnJeP0Q=

PBKDF2 salt: d2ea8d9efe17a1d910433d899bec3507
IV: 6941f8c63a1d080dfe4f909b2115621b
Encrypted AES in Base64: kUDFGkI1dI/k19R6o2o17bSMNRnYIrSJ0h1fqxREwpgZoSVxokJj7yLqos3aZ4JBoyQkf3R5gLkS5B1w7ZLPRSXrA/tusY8h2NU3NnJeP0Q=
```
Example 2:
```
[...]

PR #3385, opened 2022-07-18
Terra Station Wallet -m 29600

TomStokes14:
Hi there,

as per #3285 please see the relevant plugin. Doing that trick I mentioned in the issue chat where we only actually encrypt the penultimate block, as it will equal the final padding block. I haven't had the chance to learn perl yet (sorry...!) but I promise the next plugin I submit will have all the testing goodness ;)

All the best,
Tom

### DIFF
--- a/OpenCL/m29600-pure.cl
+++ b/OpenCL/m29600-pure.cl
@@ -1,0 +1,322 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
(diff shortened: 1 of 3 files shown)


jsteube:
Thanks for the contribution :) The Perl module is still optional, but adds massive to product quality.

Chick3nman:
Function names in the kernel may need to be updated, this may not be functional as-is.

PR #3394, opened 2022-07-25
Makefile improvements

ventaquil:
Done:
* New `Makefile` target which allows to install tools - check `make install_tools` (run alongside with `make install`),
* New `Makefile` target which allows to build package - check `make package` or `make hashcat.7z`,
* Changed artifacts steps in workflow to use `make package`,
* ~~Changed strategies in workflow to build with or without brain support~~.

Propositions:
* ~~Similar change like brain support can be added for `ENABLE_CUBIN` to test it during workflow~~,
* Script `tools/package_bin.sh` can be removed (use `make package` instead).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,19 +1,5 @@
 name: Build
 
-env:
-  include_paths: |
-    charsets/
-    docs/
-    extra/
-    layouts/
-    masks/
-    modules/
-    OpenCL/
-    rules/
-    example*
-    hashcat*
-    libhashcat*
-
 on:
   push:
     branches:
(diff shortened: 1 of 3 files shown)


ventaquil:
Maybe brain cross-compilation is not necessary - here are the limits for free tier [Included storage and minutes](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#included-storage-and-minutes). I am afraid that tons of concurrent jobs can quickly exhaust free limits.

However packaging artifacts with new `Makefile` target offers good compression which significantly reduces size of the artifacts.

jsteube:
Can we have these PR separted? Because for building the hashcat.7z we use the `tools/package_bin.sh`. Maybe you want to check this script and take over the missing parts in the Makefile, too.

ventaquil:
I will split it up and take a closer look to `tools/package_bin.sh` script 👍

ventaquil:
Tools installation in `install_docs` target was replaced by `install_tools` target.

Motivation - it's easier to install all tools in `/usr/local/bin` or any other path that is in `$PATH` variable and just call `...2hashcat.py` instead of providing full path to share/docs or adding share/docs path to `$PATH` variable.

ventaquil:
Hey @jsteube 

I am not quite sure about putting `dos2unix` and `unix2dos` calls inside `Makefile` - in my opinion it needs temporary directory (like `~/xy/hashcat-6.2.5` from `tools/package_bin.sh`) to store whole content there however it compliates packaging process.
Currently to make it easy to read and understand I am using files without any modifications - it can be changed on your request.

I also added `package-linux`, `package-win` and `package-all` targets - it allows to create packages by using cross-compilation targets.

In my opinion it's ready to review but any comments are welcome here.

jsteube:
We really need dos2unix and unix2dos calls to make sure .txt files are openable using notepad (default .txt editor in windows if not changed).

beschio:
Just want to add that the Windows 10 `notepad.exe` has had [support for line endings other than CRLF](https://devblogs.microsoft.com/commandline/extended-eol-in-notepad/) since stable release Windows 10 1809 (October 2018 Update).

PR #3410, opened 2022-07-30
no need for OPTS_TYPE_HASH_COPY in -m 29600

philsmd:
I think it makes sense to get rid of `OPTS_TYPE_HASH_COPY` in -m 29600 = `Terra Station Wallet` because we don't necessarily need it.

I only had to add the `iv` to the struct and increase the ciphertext `ct` length, but this is all negligible and I think it is better code at the end, i.e. our `module_hash_encode ()` and `module_hash_decode ()` are more consistent with other hash-modes and with this fix it doesn't fill anylonger that we kind of "cheat" by just copying the whole input hash line.

Thanks

### DIFF
--- a/src/modules/module_29600.c
+++ b/src/modules/module_29600.c
@@ -24,7 +24,6 @@ static const char *HASH_NAME      = "terra";
 static const u64   KERN_TYPE      = 29600;
 static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE;
 static const u64   OPTS_TYPE      = OPTS_TYPE_STOCK_MODULE
-                                  | OPTS_TYPE_HASH_COPY
                                   | OPTS_TYPE_PT_GENERATE_BE;
 static const u32   SALT_TYPE      = SALT_TYPE_EMBEDDED;
 static const char *ST_PASS        = "hashcathas";
@@ -60,7 +59,8 @@ typedef struct pbkdf_sha1_tmp
 typedef struct terra
 {
   u32 salt_buf[8];
-  u32 ct_block_a[4];
+  u32 ct[16]; // 16 * 4 = 64 bytes (we have extra 16 bytes in digest: 64 + 16 = 80)
+  u32 iv[4];
   
 } terra_t;
 
@@ -122,7 +122,8 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   const u8 *salt_pos = token.buf[0];
   const int salt_len = token.len[0];
-  // iv unused
+  const u8 *iv_pos   = token.buf[1];
+  const int iv_len   = token.len[1];
   const u8 *hash_pos = token.buf[2];
   const int hash_len = token.len[2];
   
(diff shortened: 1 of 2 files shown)


src/modules/module_29600.c
```diff
@@ -143,6 +144,13 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
   terra->salt_buf[2] = hex_to_u32 (salt_pos + 16);
   terra->salt_buf[3] = hex_to_u32 (salt_pos + 24);
 
+  // store IV
+  if (iv_len != 32) return (PARSER_SALT_LENGTH);
+  terra->iv[0] = hex_to_u32 (iv_pos +  0);
+  terra->iv[1] = hex_to_u32 (iv_pos +  8);
+  terra->iv[2] = hex_to_u32 (iv_pos + 16);
+  terra->iv[3] = hex_to_u32 (iv_pos + 24);
```
ventaquil, line 161: Cannot we use `hex_decode` here?

```suggestion
  hex_decode (iv_pos, iv_len, (u8 *) terra->iv);
```

src/modules/module_29600.c
```diff
@@ -152,13 +160,18 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
...
 
   // Penultimate block, i.e. IV, xored with a whole padding block
-  terra->ct_block_a[0] = byte_swap_32(whole_digest[0xc] ^ 0x10101010);
-  terra->ct_block_a[1] = byte_swap_32(whole_digest[0xd] ^ 0x10101010);
-  terra->ct_block_a[2] = byte_swap_32(whole_digest[0xe] ^ 0x10101010);
-  terra->ct_block_a[3] = byte_swap_32(whole_digest[0xf] ^ 0x10101010);
+  terra->ct[0] = byte_swap_32(whole_digest[0xc] ^ 0x10101010);
+  terra->ct[1] = byte_swap_32(whole_digest[0xd] ^ 0x10101010);
+  terra->ct[2] = byte_swap_32(whole_digest[0xe] ^ 0x10101010);
+  terra->ct[3] = byte_swap_32(whole_digest[0xf] ^ 0x10101010);
```
ventaquil, line 177: Referring to style recommendations there should be space between function name and bracket.

```suggestion
  terra->ct[0] = byte_swap_32 (whole_digest[0xc] ^ 0x10101010);
  terra->ct[1] = byte_swap_32 (whole_digest[0xd] ^ 0x10101010);
  terra->ct[2] = byte_swap_32 (whole_digest[0xe] ^ 0x10101010);
  terra->ct[3] = byte_swap_32 (whole_digest[0xf] ^ 0x10101010);
```

src/modules/module_29600.c
```diff
@@ -171,7 +184,52 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
...
+
+  hex_encode ((u8 *) terra->salt_buf, 16, (u8 *) salt_hex); // or use: generic_salt_encode ()
+
+  salt_hex[32] = 0;
+
+  // iv:
+
+  char iv_hex[32 + 1] = { 0 };
+
+  hex_encode ((u8 *) terra->iv, 16, (u8 *) iv_hex);
```
ventaquil, line 211: I am thinking about adding some `TERRA_SALT_LEN` and `TERRA_SALT_HEX_LEN` like here

https://github.com/hashcat/hashcat/blob/master/src/modules/module_29311.c#L49-L53

https://github.com/hashcat/hashcat/blob/master/src/modules/module_29311.c#L233-L237

src/modules/module_29600.c
jsteube: Yes, but with the exception that if the length is both fixed and limited in size we prefer hex_to_u32() because it makes it easier to read.

src/modules/module_29600.c
jsteube: Correct

src/modules/module_29600.c
jsteube: OK

src/modules/module_29600.c
ventaquil: Noted!

Issue #3440, opened 2022-09-05
Flask Session Cookie (Mode 29100) Token Length Too Small


IppSec:
The maximum payload size for this module is 27, which is the exact length used in the example:
```
{
  "username": "admin"
}
```
If the username is longer than admin or there are any more data in the payload than above hashcat says it is an invalid hash.  JWT is a very similar format and supports lengths up to 2047 for each token field.

Chick3nman:
Closing as fixed, https://github.com/hashcat/hashcat/pull/3441 was merged into master.

PR #3441, opened 2022-09-05
Updated flask session max length for payload to be 2047

IppSec:
Pull request for issue: #3440 copies the max length from JWT and applies to flask session (mode 29100)

### DIFF
--- a/src/modules/module_29100.c
+++ b/src/modules/module_29100.c
@@ -62,7 +62,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[0]     = '.';
   token.len_min[0] =  0;
-  token.len_max[0] = 27;
+  token.len_max[0] = 2047;
   token.attr[0]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_BASE64C;
 
--- a/src/modules/module_29100.c
+++ b/src/modules/module_29100.c
@@ -62,7 +62,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[0]     = '.';
   token.len_min[0] =  0;
-  token.len_max[0] = 2047;
+  token.len_max[0] = 56;
   token.attr[0]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_BASE64C;
 


ZoeS17 (commented):
Was there a reason to not update `token.len_max[2]` while this code([Line 77](https://github.com/hashcat/hashcat/blob/1687fd80f45b9ac7d72ea885e9f2d44c0e253007/src/modules/module_29100.c#L77)) is being updated?
```c
  token.len_max[2] = 27;
```
For example in the JWT module(16500) [lines 128 - 144](https://github.com/hashcat/hashcat/blob/1687fd80f45b9ac7d72ea885e9f2d44c0e253007/src/modules/module_16500.c#L128-L144):

```c
  token.sep[0]     = '.';
  token.len_min[0] = 1;
  token.len_max[0] = 2047;
  token.attr[0]    = TOKEN_ATTR_VERIFY_LENGTH
                   | TOKEN_ATTR_VERIFY_BASE64C;

  token.sep[1]     = '.';
  token.len_min[1] = 1;
  token.len_max[1] = 2047;
  token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH
                   | TOKEN_ATTR_VERIFY_BASE64C;

  token.sep[2]     = '.';
  token.len_min[2] = 43;
  token.len_max[2] = 86;
  token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH
                   | TOKEN_ATTR_VERIFY_BASE64C;
```

IppSec:
No reason other than I was trying to make as few changes as possible in case there was some optimization due to the small size. AFAIK, the only field that has a dynamic size in this format is the first one.
* Payload
* Timestamp
* Cryptographic signature

Timestamps and signatures shouldn't change in size, so left it alone.

Edit: To go on about the JWT size, the first field is dynamic in size as it says how the payload is protected and can include RSA keys. The second field in a JWT is the payload (first field in Flask Session).  So it makes sense for JWT to support larger key spaces for the first two fields.

ZoeS17:
Okay, I figured just the first field was dynamic but I wasn't sure of the spec myself. Minimal change needed seems like a good idea. Nothing further to add/ask here.

Chick3nman (changes requested):
From what I can tell, the max this token can safely be raised to is 57, not 2047. 

The kernel currently inits a salt buf that is capped at 64 bytes, an array of 16 u32x.
https://github.com/hashcat/hashcat/blob/master/OpenCL/m29100_a0-pure.cl#L35-L43

The module builds the salt buffer from the first 2 tokens, with a known length of token[0]+1+token[1]. 
https://github.com/hashcat/hashcat/blob/master/src/modules/module_29100.c#L90

With token[1] being capped at 6, plus the static added 1, we have only 57 bytes left of salt buffer in the kernel that the module could safely copy into. Anything greater than 57 will likely cause an overflow in the current kernel. 

My recommendation would be to raise to 56 or 57 at most for now and then if that's not enough we can revisit this and make modifications to the kernel to support longer lengths.

IppSec:
I believe I updated the PR to be 56.

Chick3nman (approved):

PR #3446, opened 2022-09-07
Add warning for potfile-disable in Assoc Attack mode

Chick3nman:
Adding a warning to alert users that the potfile is disabled in Association Attack mode.

https://hashcat.net/forum/thread-11007.html


Chick3nman:
Fixes https://github.com/hashcat/hashcat/issues/3447

Issue #3450, opened 2022-09-08
double setting of --status option


UnrelatedWorks:
**GitHub is for bugs and features - not for support**
For support, please use the hashcat forums https://hashcat.net/forum/

**Check the FAQ**
Some items that might appear to be issues are not issues. Please review the hashcat FAQ https://hashcat.net/wiki/doku.php?id=frequently_asked_questions before submitting a bug report.

**Describe the bug**
when setting the --status-timer=x option the --status also has to  be set 

**To Reproduce**
does not display status:
hashcat --status-timer=30 -a 3 -m 0 example0.hash ?a?a?a?a?a?a 
does display status:
hashcat --status --status-timer=30 -a 3 -m 0 example0.hash ?a?a?a?a?a?a 

**Expected behavior**
hashcat --status-timer=30 -a 3 -m 0 example0.hash ?a?a?a?a?a?a 
should display updated status after 30 seconds. If the user explicitly wants an output after x seconds it does not make sense to also require that a status is needed at all. 

**Hardware/Compute device (please complete the following information):**
Not important, on any hardware 

**Hashcat version (please complete the following information):**
Not important, any version

roycewilliams:
Looks like this should be added to the command-line sanity checking, so that hashcat quits early and explains that `--status-timer` requires `--status`.

Chick3nman:
The problem is that --status-timer doesn't set the value for --status. I can sorta see both sides to it, if you've set a status-timer value, you have implied that you want --status on. The question is if that is enough reason to set --status or if we still want explicit direction from the user. Is there ever a time when the timer would be set but --status would not? If not, I'm comfortable changing the logic to flip both if the timer value is changed.

UnrelatedWorks:
Before posting the issue I also went through a couple scenarios where a user might not want to see the status but still set a time for it. 

The only way this makes sense if we add a --pipe-temp, --pipe-progress ... option that feeds out the information. I use hashcat in headless environments where I cannot interact so the timer is essential to check for progress, temps etc. I don't mind "sifting" through the whole status screen but for some they might want to get their values "cleaner". Then it would make sense to keep --status-timer and --status in place. 

I don't see a sanity check like @roycewilliams pointed out working in the long term. It will just frustrate new users when defining their options.

Chick3nman:
[...]

PR #3451, opened 2022-09-08
Fixed missing status update when --status-json is used

davidbolvansky:
hashcat does not run periodic (defined by status-timer) status update in json format. Fixed.

### DIFF
--- a/src/monitor.c
+++ b/src/monitor.c
@@ -75,7 +75,7 @@ static int monitor (hashcat_ctx_t *hashcat_ctx)
     remove_check = true;
   }
 
-  if (user_options->status == true)
+  if (user_options->status == true || user_options->status_json == true)
   {
     status_check = true;
   }


Chick3nman:
This is actually being discussed as a part of a different issue here: https://github.com/hashcat/hashcat/issues/3450

We don't turn status on by default for _any_ of the alternative status options, including --status-timer. We may want to enable it for status-timer and status-json or potentially rethink the how we enable/disable status output.

davidbolvansky:
Well, current behaviour does not look as something intended.

—status —status-timer=3              prints update every 3 sec
—status-json —status-timer=3      prints nothing

I doubt this is a feature and not a bug.

Chick3nman:
Assuming you meant `—status-json —status-timer=3 prints nothing`, this is actually the intended behavior right now. `--status` enables the automatic status output, `--status-json` and `--status-timer` just modify the format or frequency. 

If you wanted to see json output every 3 seconds, first you would enable status output then you'd set your modifiers. `--status --status-json --status-timer=10`. The default values for `--status` and `--status-timer` are 0(false) and 10(seconds).

Also, given that it looks like you are also interested in adding device information to `--status-json` in your other PR, it's probably good to note that a frequency that short(3 seconds) can actually start to have noticeable performance impacts on larger machines when we need to reach out to the runtime for device information that often.

davidbolvansky:
Ooh, right.—status  —status-json —status-timer=3 does exactly what I want.

> it's probably good to note that a frequency that short(3 seconds) can actually start to have noticeable performance impacts on larger machines when we need to reach out to the runtime for device information that often.

3 was just an example value, we use 10, but yes, I agree, it is not free to fetch device information.

PR #3454, opened 2022-09-09
Fix for #3450

UnrelatedWorks:
sanity check for not displaying status while status-timer is set. More discussion in: #3450

### DIFF
--- a/src/user_options.c
+++ b/src/user_options.c
@@ -558,7 +558,7 @@ int user_options_sanity (hashcat_ctx_t *hashcat_ctx)
 
     return -1;
   }
-    if ((user_options->status_timer > 0) $$ (user_options->status == false))
+    if ((user_options->status_timer > 0) && (user_options->status == false))
     {     
       event_log_error (hashcat_ctx, "--status must the set to get an output");
       return -1;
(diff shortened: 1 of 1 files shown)


Chick3nman (changes requested):
The logic on line 562 is causing hard failures in the build tests because it fails to take into account that we set this value by default, even when the user has not asked for it. That means the error will always happen, even if you don't want a status at all. The default value for the status time is 10. This will need to be changed to detect when the user has added the status timer to their command vs when the default values is set or logic will need to be introduced to alter when the status timer default value is applied.

jsteube:
I can understand that from a user perspective these options are a little bit technical, but from a technical perspective they are perfectly fine. Both options can be used individually or in combination without breaking hashcat's functionality. In fact I find it to be an advantage that with hashcat the user can shoot them self in the foot. 

There's also the problem that there are hashcat wrappers that try to disable interactive functionality such as monitoring in order to reduce runtime overhead of the hashcat host application. They might set --status-timer 0 or 1, but at the same time not set --status. These wrapper would be broken by this PR.

I am thinking about to reject this PR or we rewrite it so that it only prints a warning instead.

UnrelatedWorks:
I don't mind if you reject this, I'll work on one which only outputs a warning if --status = true and --status-timer is not 10

PR #3456, opened 2022-09-10
Escape some fields of --status-json output

### LINKED ISSUE #3453: --status-json output is not properly escaped
**Describe the bug**
Some fields of `--status-json` output are not being escaped, this includes:
[session](https://github.com/hashcat/hashcat/blob/d04dec91d13f55d3800684666b22f8eb81f0bced/src/terminal.c#L1831)
[guess_base](https://github.com/hashcat/hashcat/blob/d04dec91d13f55d3800684666b22f8eb81f0bced/src/terminal.c#L1835)
[guess_mod](https://github.com/hashcat/hashcat/blob/d04dec91d13f55d3800684666b22f8eb81f0bced/src/terminal.c#L1849)

**To Reproduce**
Just run hashcat with `--status-json`
+ref: https://github.com/s77rt/hashcat.launcher/issues/101

**Expected behavior**
fields to be JSON escaped

**Hashcat version:**
 - Version: 6.2.6 and d04dec91d13f55d3800684666b22f8eb81f0bced
**UPDATE** Version 6.2.5 or even earlier are also effected (session field only since other fields did not exist)

davidbolvansky:
Should fix https://github.com/hashcat/hashcat/issues/3453

### DIFF
--- a/src/terminal.c
+++ b/src/terminal.c
@@ -1828,11 +1828,24 @@ void status_display_status_json (hashcat_ctx_t *hashcat_ctx)
     end = time_now + sec_etc;
   }
 
-  printf ("{ \"session\": \"%s\",", hashcat_status->session);
+  char *session_json_encoded = (char *) hcmalloc (strlen (hashcat_status->session) * 2);
+
+  json_encode (hashcat_status->session, session_json_encoded);
+
+  printf ("{ \"session\": \"%s\",", session_json_encoded);
+
+  hcfree (session_json_encoded);
+
   printf (" \"guess\": {");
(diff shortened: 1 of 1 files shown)


s77rt:
Thanks for taking the time for this.
Just minor notes for consistency:
rename the variables to be **var_json_encoded** instead of **var_encoded**
example: **guess_mod_json_encoded** instead of **guess_mod_encoded**
also remove the extra empty lines.

other than that, LGTM

davidbolvansky:
Oops, nice catch. Fixed.

PR #3321, opened 2022-06-15
LUKS strings

ventaquil:
Similar changes like in GH-3293 and GH-3302 but for LUKS modules.

TODO:

- [x] ~~Add tests for new modules.~~
- [ ] Add missing Whirlpool kernels.

```
- [ Hash modes ] -

      # | Name                                                       | Category
  ======+============================================================+======================================
  14600 | LUKS v1 (legacy)                                           | Full-Disk Encryption (FDE)
  29541 | LUKS v1 RIPEMD-160 + AES                                   | Full-Disk Encryption (FDE)
  29542 | LUKS v1 RIPEMD-160 + Serpent                               | Full-Disk Encryption (FDE)
  29543 | LUKS v1 RIPEMD-160 + Twofish                               | Full-Disk Encryption (FDE)
  29511 | LUKS v1 SHA-1 + AES                                        | Full-Disk Encryption (FDE)
  29512 | LUKS v1 SHA-1 + Serpent                                    | Full-Disk Encryption (FDE)
  29513 | LUKS v1 SHA-1 + Twofish                                    | Full-Disk Encryption (FDE)
  29521 | LUKS v1 SHA-256 + AES                                      | Full-Disk Encryption (FDE)
  29522 | LUKS v1 SHA-256 + Serpent                                  | Full-Disk Encryption (FDE)
  29523 | LUKS v1 SHA-256 + Twofish                                  | Full-Disk Encryption (FDE)
  29531 | LUKS v1 SHA-512 + AES                                      | Full-Disk Encryption (FDE)
  29532 | LUKS v1 SHA-512 + Serpent                                  | Full-Disk Encryption (FDE)
  29533 | LUKS v1 SHA-512 + Twofish                                  | Full-Disk Encryption (FDE)
```

PR #3459, opened 2022-09-14
Fix improper strings comparision in LUKS modules

ventaquil:
Bug introduced here - GH-3321.

When LUKS container is using `cbc-plain64` or `xts-plain64` then `cbc-plain`/`xts-plain` respectively were used.

There shouldn't be any impact on cracking.

https://github.com/hashcat/hashcat/blob/161a5a2a3c968504115471e352390e2099b542b1/OpenCL/inc_luks_aes.cl#L2692

However potfiles can grow up because hashes after encoding don't match to hashes before decoding.


ventaquil:
Tested by me, looks okay.

Issue #3462, opened 2022-09-16
Add new algorithm: BISQ .wallet (scrypt)


Banaanhangwagen:
This post is a placeholder to discuss the adding of a new algorithm "Bisq .wallet (scrypt)".

Bisq (https://bisq.network/) is a decentralized bitcoin for fiat exchange.
After installing the application, the user can use a built-in bitcoin/bisq-wallet.

The user is encouraged to protect his wallet with a password of minimum 8 characters.
Once set, this password is asked at startup or when doing a transaction.

Bisq creates two encrypted .wallet files.
On Windows, you will find these files in the folder `AppData\Roaming\Bisq\btc_mainnet\wallet`.

The wallet files have the following header: `x0A x16 org.bitcoin.production`.
After reading [the code on their Github](https://github.com/bisq-network/bisq/blob/master/core/src/main/java/bisq/core/crypto/ScryptUtil.java), it appears that they use the same `ScryptUtil.java` as Multibit-wallet, but with slightly modified N, r, p parameters.

In order to crack this wallet password, I created a new mode for Hashcat (-m 29800). Since there is a big resemblance with `Multibit wallet (scrypt)`-mode 27700, this could be easily done. (right ;-) )

Please use `bisq2hashcat.py` (found in tools) on the .wallet files for extracting the hash.

---

Example :
```
$ python bisq2hashcat.py bisq_BTC.wallet
bisq_BTC.wallet:$bisq$3*32768*8*6*31d838af87f99cb8*5cfb7bf3228d9e865881156e17b1866589ffa6b757011e25d1319083595236d2
```
```
$ ./hashcat -m 29800 '$bisq$3*32768*8*6*31d838af87f99cb8*5cfb7bf3228d9e865881156e17b1866589ffa6b757011e25d1319083595236d2' -w3 -a3 hashcat1 -d4
hashcat (v6.2.6) starting

CUDA API (CUDA 11.7)
====================
* Device #1: NVIDIA GeForce GTX 1080 Ti, skipped
* Device #2: NVIDIA GeForce GTX 1080, skipped
* Device #3: NVIDIA GeForce GTX 1080, skipped
* Device #4: NVIDIA GeForce GTX 1070 Ti, 8014/8117 MB, 19MCU

Minimum password length supported by kernel: 0
Maximum password length supported by kernel: 256

Hashes: 1 digests; 1 unique digests, 1 unique salts
[...]

PR #3463, opened 2022-09-16
Add new algorithm: BISQ .wallet (scrypt) (-m 29800)

Banaanhangwagen:
Please find here a pull request in order to implement cracking a Bisq-wallet (scrypt). See [this issue](https://github.com/hashcat/hashcat/issues/3462) for more details.

It is mainly a copy from -m 27700, so no big changes...
Tested on my Linux-environment, and it works like a charm. ;)

Feel free to change little errors.

Thank you


tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+        iv = binascii.hexlify(iv).decode("ascii")
+        block_iv = binascii.hexlify(block_iv).decode("ascii")
+        block_noiv = binascii.hexlify(block_noiv).decode("ascii")
+        sys.stdout.write("%s:$bisq$%d*%s*%s*%s\n" % (bname, version, iv, block_iv, block_noiv))
+        return
+
+    f.close()
+
+if __name__ == "__main__":
+    if len(sys.argv) < 2:
```
ventaquil, line 121: Cannot we use [`argparse.ArgumentParser`](https://docs.python.org/3/library/argparse.html) here?

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+        sys.stderr.write("Usage: %s [Bisq wallet file (.wallet)]\n" % sys.argv[0])
+ #       sys.stderr.write("\nMultiBit Classic -> for a wallet named 'xyz', we need the xyz-data/key-backup/xyz*.key OR xyz-data/wallet-backup/xyz*.wallet file\n")
+        sys.exit(1)
+
+    for j in range(1, len(sys.argv)):
+        process_file(sys.argv[j])
```
ventaquil, line 127:
```suggestion
    for arg in sys.argv[1:]:
        process_file(arg)
```

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+    except IOError:
+        e = sys.exc_info()[1]
+        sys.stderr.write("%s\n" % str(e))
+        return
+
+    if "wallet" in bname or b"org.bitcoin.production" in data:
+ #       sys.stderr.write("[WARNING] Cracking .wallet files is a very slow process, try cracking the associated .key file instead!\n")
+        version = 3  # MultiBit Classic .wallet file
+        # def is_wallet_file(wallet_file) from btcrecover
+        wallet_file = open(filename, "rb")
```
ventaquil, line 46: Are we opening this file twice (`f` handle in 35th line is not closed)?

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+        wallet_file.seek(0)
```
ventaquil, line 47: Do we need to do `seek(0)` when we open file?

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+        is_valid_bitcoinj_wallet = False
+        if wallet_file.read(1) == b"\x0a":  # protobuf field number 1 of type length-delimited
```
ventaquil, line 49: I prefer to use [`with` statement](https://docs.python.org/3/reference/compound_stmts.html#with).

```suggestion
        with open(filename, "rb") as wallet_file:
            wallet_file.seek(0)
            is_valid_bitcoinj_wallet = False
            if wallet_file.read(1) == b"\x0a":  # protobuf field number 1 of type length-delimited
```

See also https://stackoverflow.com/a/31334226.

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+            network_identifier_len = ord(wallet_file.read(1))
+            if 1 <= network_identifier_len < 128:
+                wallet_file.seek(2 + network_identifier_len)
+                c = wallet_file.read(1)
+                if c and c in b"\x12\x1a":   # field number 2 or 3 of type length-delimited
+                    is_valid_bitcoinj_wallet = True
+        if is_valid_bitcoinj_wallet:
+            pb_wallet = Wallet()
+            pb_wallet.ParseFromString(data)
```
ventaquil, line 58: It's better to check for `not is_valid_bitcoinj_wallet` and then return early - one indent level less at all.

```suggestion
        if not is_valid_bitcoinj_wallet:
            return

        pb_wallet = Wallet()
        pb_wallet.ParseFromString(data)
```

tools/bisq2hashcat.py
```diff
@@ -0,0 +1,127 @@
...
+from protobuf.wallet_pb2 import *
+
+
+def process_file(filename):
+    bname = os.path.basename(filename)
+    try:
+        f = open(filename, "rb")
+        data = f.read()
+    except IOError:
+        e = sys.exc_info()[1]
```
ventaquil, line 38:
```suggestion
    except IOError as e:
```

ventaquil (commented):
I am not sure which Python version is target version for `tools/bisq2hashcat.py` (it looks like to support both 2.7 and 3.x version).

However we should keep in mind that Python 2.7 is not supported anymore - [Sunsetting Python 2](https://www.python.org/doc/sunset-python-2/).

I left few comments but I could have many more. If it's working I can add improvements in different PR to make this one clean.

jsteube:
I'm merging this as-as, @ventaquil feel free to add another PR for the extraction tool changes.

Issue #3468, opened 2022-10-02
Mode 3710 - First MD5 in upper case


JohnTar22:
mode 3710 takes md5($salt.md5($pass)) as hash.

The first md5($pass) returns a md5 sum in lowercase, which in turn is hashed again with the addition of $salt.

Is it possible to make the first hash (md5($pass)) return an uppercase string, e.g. it is being hashed as $salt.UPPERCASE_MD5($pass)?

I tried editing the code by myself with no success.

Thanks.

JohnTar22:
I'll elaborate further, according to this: https://github.com/mcw0/PoC/blob/master/dahua-telnetd-json.py:
`def dahua_md5_hash(Dahua_random, Dahua_realm, username, password):

	PWDDB_HASH = hashlib.md5(username + ':' + Dahua_realm + ':' + password + '').hexdigest().upper()
	PASS = ''+ username + ':' + Dahua_random + ':' + PWDDB_HASH + ''
	RANDOM_HASH = hashlib.md5(PASS).hexdigest().upper()

#	print "[i] MD5 hash:",PWDDB_HASH
#	print "[i] Random value to encrypt with:",Dahua_random
#	print "[i] Built password:",PASS
#	print "[i] MD5 generated login hash:",RANDOM_HASH
	return RANDOM_HASH`
if i get the hash from the authentication process i'll be able to recover the password by having all the parameters:
random, username, realm (which is "Login to SN in md5).
i was able to change 3710 mode to make it create the "first" md5 hash in upper case and i'm able to crack the hash and recover my password but in a very "ugly" way. 
first i get the md5 hash from Wireshark, i also get the random (which is  a 10 digit number) and the Realm (which is Login to blablamd5string).
then i'm creating hashlist which looks like that: 
8743b52063cd84097a65d1633f5c74f5:$(HERE is the username and random in upper case hex)

after that i'm able to crack using this command for Ex: 
-a3 -w4 -m 3710 hash.file -1 ?l?d --hex-salt "admin:Login to $(SN in MD5):?1?1?1?1?1?1?1"
 
It doesn't work with wordlists and the entire process is a pain is the a$$.

any help is much appreciated.
John.

piwvvo:
Hey, could you clarify few things about this algorithm? SN is already hashed or you have to do it by yourself? 

[...]

PR #3482, opened 2022-10-20
Add hash-mode 3730 - Dahua NVR/DVR - md5($salt1.strtoupper(md5($salt2…

davidrozen76:
….$pass)))

Pure kernel on attack mode 0 doesn't crack when only a single hash is present in the hashfile (otherwise ok) - self-test fails as well.

In accordance to issue #3468


OpenCL/m03730_a0-pure.cl
```diff
@@ -0,0 +1,309 @@
...
+  for (u32 i = 0, idx = 0; i < salt2_len; i += 4, idx += 1)
+  {
+    s2[idx] = esalt_bufs[DIGESTS_OFFSET_HOST].salt2_buf[idx];
+  }
+
+  md5_ctx_t ctx0;
+
+  md5_init (&ctx0);
+
+  md5_update_global (&ctx0, esalt_bufs[DIGESTS_OFFSET_HOST].salt1_buf, esalt_bufs[DIGESTS_OFFSET_HOST].salt1_len);
```
piwvvo: It seems that single hash kernel doesn't work correctly because you used salt1_buf here instead of salt2_buf

jsteube:
Please provide unit-test in tools/test_modules for this plugin. Shouldn't be too hard to realize there's plenty of other modules to copy the code from and then just adapt. But much more important is to remove the changes to inc_common.h. If you have a fast salt with esalt, there's also plenty of other modules than can be used as template.

davidrozen76:
The unit-test fails on optimized kernels when salt length exceeds 16 characters.

Changes are to be made to support long salt lengths (a must for this hash-mode).

Any help will be appreciated :D

jsteube:
Any handling of data larger the block size of the underlaying primitive in question requires branching. This takes a lot of effort. Of course I can't merge an unfinished plugin. If you want me to merge this, please remove the optimized kernels, because the pure kernels should work. We can later add the optimized kernels.

jsteube:
Very cool, many thanks!

PR #3485, opened 2022-10-21
Add support for non-zero IV to lastpass (-m 6800)

Lars-Saetaberget:
New versions of Lastpass no longer use all zeroes for the IV, hence this change. They have also changed some other stuff like the default iterations, and also where the hashes can be found (as the desktop apps are discontinued). To help with that last part we also added ``lastpass2hashcat.py``.

The example hash has been updated to reflect a more current hash, but the old mode is still usable by specifying all zeroes for the IV. So the old example hash for example would be: ``82dbb8ccc9c7ead8c38a92a6b5740f94:500:pmix@trash-mail.com:00000000000000000000000000000000``

Also changed the test module to account for these changes.

I'm not sure what the "offline mode" in the kernel is referring to, or if it's even needed anymore. As far as I can see we're always done when we decrypt the ciphertext and get the salt as the plaintext. Perhaps someone can shed some light on that functionality/mode?

Let me know if any changes should be made! I'm a bit unsure if I did everything right with regards to the salt/esalt stuff in particular, as I've not had to use that stuff until now.

### DIFF
--- a/OpenCL/m06800-pure.cl
+++ b/OpenCL/m06800-pure.cl
@@ -18,6 +18,11 @@
 #define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
 #define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
 
+typedef struct lastpass
+{
+    u32 iv[4];
+} lastpass_t;
+
 typedef struct lastpass_tmp
 {
   u32 ipad[8];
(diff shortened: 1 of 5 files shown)


Lars-Saetaberget:
It strikes me that the bugfix is really a seperate issue, so I removed it from this PR.

Made new issue and PR for that bug and keeping this one clean with only one change: 
Issue https://github.com/hashcat/hashcat/issues/3488 
PR https://github.com/hashcat/hashcat/pull/3489

jsteube:
Fantastic, many thanks!

PR #3486, opened 2022-10-22
Update restore.c

PenguinKeeper7:
Increase `--restore` max args limit to 5,000 when `--force`'d, especially helpful when restoring an attack with many wordlists for caching dictstat info.

### DIFF
--- a/src/restore.c
+++ b/src/restore.c
@@ -67,17 +67,15 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
 
   // we only use these 2 checks to avoid "tainted string" warnings
 
-  if (rd->argc < 1)
+  // raise upper bound to 5000 when --force'd
+  if(rd ->argc > 5000)
   {
-    event_log_error (hashcat_ctx, "Unusually low number of arguments (argc) within restore file %s", eff_restore_file);
+    event_log_error (hashcat_ctx, "Unusually high number of arguments (argc) within restore file %s", eff_restore_file);
 
     hc_fclose (&fp);
 
     return -1;
-  }
-
-  if (rd->argc > 250) // some upper bound check is always good (with some dirs/dicts it could be a large string)
-  {
+  } else if (hashcat_ctx->user_options->force == false && rd->argc > 250) // some upper bound check is always good (with some dirs/dicts it could be a large string)
     event_log_error (hashcat_ctx, "Unusually high number of arguments (argc) within restore file %s", eff_restore_file);
 
     hc_fclose (&fp);
--- a/src/restore.c
+++ b/src/restore.c
@@ -66,7 +66,6 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
   }
 
   // we only use these 2 checks to avoid "tainted string" warnings
-
   // raise upper bound to 5000 when --force'd
   if(rd ->argc > 5000)
   {
@@ -76,6 +75,7 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
 
     return -1;
   } else if (hashcat_ctx->user_options->force == false && rd->argc > 250) // some upper bound check is always good (with some dirs/dicts it could be a large string)
+    {
     event_log_error (hashcat_ctx, "Unusually high number of arguments (argc) within restore file %s", eff_restore_file);
 
     hc_fclose (&fp);
--- a/src/restore.c
+++ b/src/restore.c
@@ -66,16 +66,18 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
   }
 
   // we only use these 2 checks to avoid "tainted string" warnings
-  // raise upper bound to 5000 when --force'd
-  if(rd ->argc > 5000)
+
+  if (rd->argc < 1)
   {
-    event_log_error (hashcat_ctx, "Unusually high number of arguments (argc) within restore file %s", eff_restore_file);
+    event_log_error (hashcat_ctx, "Unusually low number of arguments (argc) within restore file %s", eff_restore_file);
 
     hc_fclose (&fp);
 
     return -1;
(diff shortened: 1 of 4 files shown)


roycewilliams:
I suspect that jsteube's sense of what `--force` is for does not include extending basic functionality in this way. Maybe just proposing the change without `--force` would help, or maybe branching on detected OS family instead?

PenguinKeeper7:
It's difficult to find exact numbers on what each OS can support. Some places report Windows is limited to 8,192 characters or 10k args, but surpassing both doesn't trigger any warnings or errors with Hashcat's `--restore`. As for `--force`, the current limit is to avoid any issues, no matter what OS it is, in this case, it'd be potentially raising the limit to above what an older OS can support and removing the blocking "Too many args" error so it seems fitting. I'm up for removing the limit entirely with `--force` but was recommended by Chick3nman to raise it instead for safety

jsteube:
This check has been added by @philsmd in 2017 with commit https://github.com/hashcat/hashcat/commit/6eaa7e37418bcb064c522040b274ad93bc0ddbc8

I'm not 100% certain how he came up with 250, therefore it's better to ask @philsmd to comment on this first.

philsmd:
hey
@PenguinKeeper7 , @jsteube. yeah, I think I've already investigated something similar recently where a user requested (on discord) a change here. I think the old limit was choosen to be just a bit smaller than 256 (command itself + dict + hash list etc, so 256 - x, where I've choosen 6 as max "normal" args without command line flags/switches).
Yeah, I think we could change it if we respect (or if needed, update !) the restore file specification from : https://hashcat.net/wiki/doku.php?id=restore

It's also important to make sure that both writing and reading the `hashcat.restore` file still works (with "normal" and exaggerated many switches/flags/dicts).

I think the reason for choosing a relatively small number was that we just wanted the users to use folders/directories instead for very large file lists (because it also has the disadvantage that the restore file becomes larger with a lot of files)... but, again, I'm completely fine with increasing the limit, if everything else still continues to work (reading the file names from the restore file etc).

philsmd:
Hey @PenguinKeeper7 . Did you do some test if everything is still working with your patch ? and did you make sure that the file format (hashcat.restore) spec doesn't need any update ?

Maybe only the value of 
```
#define RESTORE_VERSION_MIN 600
#define RESTORE_VERSION_CUR 611
```
in `include/restore.h` need to be double checked (but should be good for next release because we've already updated these !?) ?

I also would say that we shouldn't use `--force` here, but just increase the limit and please also add a comment in the source (this time ! I know it wasn't that clear before, but at least we should do it now) to mention why the `5000` number was chosen.

Thanks

PenguinKeeper7:
@philsmd Yes, I checked (only on Windows with cmd/powershell) with up to 50,000 files and it worked perfectly fine, with the only change needed being the ones in the commit. As for the 5,000 number, it was just a completely random pick and just seemed right-ish. I doubt anyone has wordlist folders with over 5,000 files in but I wouldn't be against raising it higher/to infinity as when Hashcat expands the wildcard for a folder path, it can go as high as you like with the .restore file scaling to match the situation and I couldn't find any (reasonable) limit to it.

PenguinKeeper7:
Sorry for the close/reopen, Github being awful

src/restore.c
```diff
@@ -66,18 +66,16 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
   }
 
   // we only use these 2 checks to avoid "tainted string" warnings
-
-  if (rd->argc < 1)
```
philsmd: why remove the lower limit too ?

src/restore.c
```diff
@@ -66,18 +66,16 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
...
+  // raise upper bound to 5000 when --force'd
```
philsmd: also add a comment here why exactly this 5000 number was choosen, otherwise we will risk to not remember where exactly this value comes from in the future

src/restore.c
```diff
@@ -66,18 +66,16 @@ static int read_restore (hashcat_ctx_t *hashcat_ctx)
...
+    event_log_error (hashcat_ctx, "Unusually high number of arguments (argc) within restore file %s", eff_restore_file);
 
     hc_fclose (&fp);
 
     return -1;
-  }
-
-  if (rd->argc > 250) // some upper bound check is always good (with some dirs/dicts it could be a large string)
-  {
+  } else if (hashcat_ctx->user_options->force == false && rd->argc > 250) // some upper bound check is always good (with some dirs/dicts it could be a large string)
```
philsmd: didn't we say we always support 5000 as maximum, but **do NOT add** the `--force` restriction ?

philsmd (changes requested):
I've added some comments, please discuss and/or fix it directly. Thanks

src/restore.c
philsmd: (and do not forget to remove the "when --force'd" if you agree to get rid of the --force dependency)

PenguinKeeper7:
Requested changes made and a small change to main.c to quit early if the amount of args is too high, preventing a bad restore file from being naturally generated in the first place

philsmd:
hmm... I think we could ignore this `appveyor` test failure here.

thanks @PenguinKeeper7  for the update, looks good to me what do you think @jsteube ? I think the `RESTORE_VERSION_MIN` is already updated for the next release, right (so we don't need to make sure about `RESTORE_VERSION_MIN` / `RESTORE_VERSION_MAX` again) ? 

The only think a little bit confusing to me in the new code are the checks `argc >= 5000`  vs `rd->argc > 5001`.
Are these checks correct for a 5000 limit ? Shouldn't the check be `rd->argc > 5000` (so 5000 is still acceptable, but 5001 isn't) and the same for `argc > 5000` ?

PR #3493, opened 2022-10-27
Added error message describing workaround when detecting more than 64

fse-a:
… backend devices.


roycewilliams:
A good problem to have. ;)

PR #3510, opened 2022-11-16
GPG with SHA-512 as s2k digest algorithm

OneHumpCamel:


jsteube:
Works fine, many thanks!

jsteube:
Could you please double check this section? 

https://github.com/hashcat/hashcat/pull/3510/files#diff-0d1e4a64574038cbe202bb7fa01ba5f67148ceeeb7690e6d9fe8d8b3d8bf7d19R265-R272

I think this is a bug (out of boundary write) and we can remove it. If yes, please send a PR or let me know if you want me to do it.

Issue #3516, opened 2022-11-27
Module 25000 (SNMPv3 HMAC-MD5-96/HMAC-SHA1-96) should not restrict the minimum password length to 8


RLStrug:
The RFC 2571, 2574, and 3414 which describe SNMP do not restrict the password length to a minimum size. Therefore, the associated hashcat module should not do it either.

Chick3nman:
Fix Merged, Closing.

matrix:
@RLStrug I'm not convinced about this modification you made

```
> Testing hash type 25000 with attack mode 0, markov enabled, single hash, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1.
[ len 1 ] Session..........: hashcat
Status...........: Exhausted
Hash.Mode........: 25000 (SNMPv3 HMAC-MD5-96/HMAC-SHA1-96)
Hash.Target......: $SNMPv3$0$60014572$66819626813663360706917303945764...520d65
Time.Started.....: Sat Apr 22 17:01:19 2023, (0 secs)
Time.Estimated...: Sat Apr 22 17:01:19 2023, (0 secs; Runtime limited: 6 mins, 40 secs)
Kernel.Feature...: Pure Kernel
Guess.Base.......: Pipe
Speed.#2.........:       19 H/s (0.11ms) @ Accel:4 Loops:131072 Thr:128 Vec:1
Recovered........: 0/1 (0.00%) Digests (total), 0/1 (0.00%) Digests (new)
Progress.........: 1
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#2...: Salt:0 Amplifier:0-1 Iteration:917504-1048576
Candidate.Engine.: Device Generator
Candidates.#2....:  -> 
[...]

PR #3517, opened 2022-11-27
Corrected issue for module 25000 (SNMPv3 HMAC-MD5-96/HMAC-SHA1-96)

RLStrug:
Password length should not be restricted to a minimum length
Corrects [issue 3516](https://github.com/hashcat/hashcat/issues/3516#issue-1465603665)


matrix:
TODO: restore the code to set the minimum password length to 8 characters

PR #3522, opened 2022-11-28
IPMI v2 RAKP HMAC MD5 

rjancewicz:
adds support for IPMI v2 RAKP HMAC MD5 kernel (based off the hmac-md5 -m50) 
selected -m7350 to immediately follow -m7300 (IPMI RAKP HMAC SHA1)

Unrolling md5 loops with fixed salt and maximum key length (as defined by the IPMIv2 spec) we are able to achieve a 3-5x speed up over the hmac-md5 pure kernel.

### DIFF
--- a/OpenCL/m07350_a0-optimized.cl
+++ b/OpenCL/m07350_a0-optimized.cl
@@ -16,24 +16,28 @@
 #include M2S(INCLUDE_PATH/inc_hash_md5.cl)
 #endif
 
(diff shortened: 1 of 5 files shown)


jsteube:
Thanks for your contribution! There are just some small issues (missing -a 1 mode, missing unit test, wrong kernel names in -a 3 pure mode), but I'll add them in the upcoming days.

PR #3533, opened 2022-12-09
Promote tools/*2hashcat.p[ly] to be installed alongside hashcat

hlein:
These are small, useful, and have unique names; it should simplify scripts, docs and HOWTOs to have them in $PATH.


hlein:
Updated - I looked around and realized I was duplicating one part of https://github.com/hashcat/hashcat/pull/3394, but not as well: the uninstall target didn't clean up the tools from their new location. Fixed.

src/Makefile
```diff
@@ -602,9 +607,15 @@ install_hashcat: $(HASHCAT_FRONTEND)
...
 
+.PHONY: install_tools
+install_tools:
+	$(INSTALL) -m 755 -d                                                    $(DESTDIR)$(INSTALL_FOLDER)
+	$(INSTALL) -m 755 ${TOOLS}                                              $(DESTDIR)$(INSTALL_FOLDER)/
+
 .PHONY: uninstall
 uninstall:
 	$(RM) -f  $(DESTDIR)$(INSTALL_FOLDER)/$(HASHCAT_FRONTEND)
+	cd $(DESTDIR)$(INSTALL_FOLDER) && $(RM) -f $(notdir ${TOOLS})
```
ventaquil: Why are you `cd`-ing instead of something like that?

```Makefile
.PHONY: uninstall
uninstall
  ...
  $(RM) -f  $(DESTDIR)$(INSTALL_FOLDER)/*hashcat.p[ly]
```

ventaquil:
I don't have enough time to make my PR complete - hopefully winter vacations are coming so I will have more time to read about best practices about packaging and makefiles and finally finish my PR.

Feel free to make it complete, I will rebase my PR when this one will be merged.

src/Makefile
hlein: That would be the easy way, but: then we would be deleting everything that matches `*hashcat.p[ly]` in `/usr/local/bin/` or wherever. Which could be more than just the things we deployed there. So instead, I'm using the list of matching names from the `tools/` source dir. ...Which only works conveniently, if we are in `.../$(INSTALL_FOLDER)` at the time.

ventaquil: Hmm... right.

How about using `notdir` along with `addprefix`? I am really beware of using `cd` - it looks ugly.

```suggestion
	$(RM) -f  $(addprefix $(DESTDIR)$(INSTALL_FOLDER)/,$(notdir ${TOOLS}))
```

hlein: Great idea! Tested, and updated the PR.

ventaquil (approved):

jsteube:
Many thanks

PR #3163, opened 2022-02-10
stdout mode: batched password index/buffer transfers from device

willcrozi:
Related issue: https://github.com/hashcat/hashcat/issues/3162

This batches password index/buffer transfers from the GPU/device reducing the overhead when passing the `--stdout` to hashcat.

The first commit (https://github.com/hashcat/hashcat/commit/b2274c16c0080db984d6f65974c7f517932a82e2) is just a temporary fix I've included (ran into problems installing latest pocl and llvm on my system), which can be dropped if this gets merged.

I've made the assumption that within `process_stdin` both the indexes and compressed buffer have the same ordering, i.e. I'm relying on the device's indexes being in the same (sequential) order as the compressed pw data, are there any cases where this is not the case? (e.g. due to some kernel amplification doing clever tricks)

PR #3545, opened 2022-12-20
stdout mode: zero output buffer between rules (fixes #3537)

### LINKED ISSUE #3537: Duplicate rule appears to be broken
**Describe the bug**
The duplicate rule appears to be broken in the 6.2.6 version.

**To Reproduce**
 - ./hashcat  --stdout  -m 99999 -a 0 mini-wordlist.txt -r duplicate.txt

**Expected behavior**
Input from mini-wordlist:
Aap
Noot
Mies

Expected Duplication:
AapAap
NootNoot
MiesMies

Actual Output:
AapA
NootOoow
MiesMies

Tested on the release and on the current master (15 december 2022)

**Additional context**
[duplicate.txt](https://github.com/hashcat/hashcat/files/10238427/duplicate.txt)
[mini-wordlist.txt](https://github.com/hashcat/hashcat/files/10238446/mini-wordlist.txt)

Also tested previous versions but it appears to be broken from the 6.2.6 version.

willcrozi:
Hi, this is a proposed fix for #3537 which was introduced by https://github.com/hashcat/hashcat/pull/3163/commits/edf7365cdabd0c9b0ee4d9c29cc56c78d11720ab (PR: #3163)

The bug is due to not zeroing the output buffer between iterations when in stdout-mode and using 'straight' or 'association' attack modes, i.e. rule-processing functions in `inc_rp.cl` require the buffer to have adequate trailing zeroes. The fix is to use `memset` to clear the entire buffer between iterations (simply appending a single trailing `u32 0` does not reliably fix the problem).

In the process I've removed unnecessary copying past the password bytes, resulting in similar (slight improvement) in overall performance. I've also added a `#define` for the buffer size, and tidied up some unneeded local vars from the previous performance work.

I've not added the `memset` to other attack modes since it's only the rule-processor that relies on trailing zeroes in the buffer. It could be added for all modes to guard against future changes (with a slight performance cost).

Testing:
```
$ git log --oneline -n 1
6d3cf3689 (HEAD -> bugfix-stdout-rules-#3537) stdout mode: zero output buffer between rules (fixes #3537)

$ ./hashcat --stdout -m 99999 -a 0 bug-#3537/mini-wordlist.txt -r bug-#3537/duplicate.txt
AapAap
NootNoot
MiesMies

$ while IFS='' read -r line; do echo "${line}${line}"; done < example.dict | md5sum
8647554c52ac0baa17148983479df0af  -

$ ./hashcat --stdout -m 99999 -a 0 example.dict -r bug-#3537/duplicate.txt | md5sum
8647554c52ac0baa17148983479df0af  -
```
Performance:

without fix (master):
```
$ git log --oneline -n 1
1154272d6 (HEAD -> master, origin/master, origin/HEAD) Normalize initial out_buf line in encoder in -m 30700

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m12.529s
user	0m11.694s
sys	0m2.601s

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m8.555s
user	0m7.657s
sys	0m2.438s

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m7.814s
user	0m7.441s
sys	0m2.197s
```
with fix:
```
$ git log --oneline -n 1
6d3cf3689 (HEAD -> bugfix-stdout-rules-#3537) stdout mode: zero output buffer between rules (fixes #3537)

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m12.344s
user	0m11.292s
sys	0m2.733s

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m7.131s
user	0m6.751s
sys	0m2.363s

$ time for i in $(seq 1 100); do cat example.dict; done | ./hashcat --stdout > /dev/null

real	0m7.514s
user	0m6.749s
sys	0m2.368s
```

### DIFF
--- a/src/stdout.c
+++ b/src/stdout.c
@@ -96,9 +96,11 @@ int process_stdout (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param,
 
   out.len = 0;
 
-  u32 plain_buf[64] = { 0 };
+  #define BUF_SZ (PW_MAX / sizeof(u32))
 
-  u8 *plain_ptr = (u8 *) plain_buf;
+  u32 plain_buf[BUF_SZ] = { 0 };
+
+  u8 *const plain_ptr = (u8 *) plain_buf;
 
   u32 plain_len = 0;
 
@@ -197,10 +199,7 @@ int process_stdout (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param,
       {
         while (pw_idx <= pw_idx_last)
         {
-          u32 *pw     = pws_comp_blk + (pw_idx->off - off_blk);
-          u32  pw_len = pw_idx->len;
-
-          pw_idx++;
+          u32 *pw = pws_comp_blk + (pw_idx->off - off_blk);
 
           for (u32 il_pos = 0; il_pos < il_cnt; il_pos++)
           {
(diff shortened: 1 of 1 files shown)


jsteube:
The original Idea to use 64 iterations (fixed) is done on purpose, with the idea to not need to zero the buffer after the copy and  ignoring the performance impact because we never tried to optimize stdout originally. Ideally we could try to find the root cause why the source buffer was not zeroed in the first place and maybe fix any uncharted other bugs caused from it. But for now it should be fine with the current handling.

willcrozi:
> Ideally we could try to find the root cause why the source buffer was not zeroed in the first place and maybe fix any uncharted other bugs caused from it.

It's kind of by design. Previous to https://github.com/hashcat/hashcat/commit/edf7365cdabd0c9b0ee4d9c29cc56c78d11720ab the source buffer was zeroed. Each password was copied individually from the device into into the u32[64] buffer of `pw_t` with trailing data zeroed (happens in `gidd_to_pw_t()`).

Since https://github.com/hashcat/hashcat/commit/edf7365cdabd0c9b0ee4d9c29cc56c78d11720ab passwords are bulk copied along with the `pw_idx_t` indexes into 'block' buffers to reduce device->host copy overhead. I.e. the source password bytes are 'compressed', i.e. packed at 4-byte intervals with up to three bytes of zero padding. The upshot is that it's not possible to zero-pad these in place so it now needs to be done on the output buffer (`plain_buf`) before applying rules.

I figured the best way to do this was to zero the entire output buffer then copy only the meaningful password data (i.e. `memset` with a static count might allow compiler to better optimise).

Issue #3546, opened 2022-12-20
No pkzip hash larger than 32KB can be cracked with versions >= 6.2.5


s3inlc:
The changes from pull request #2970 makes it impossible to crack any larger pkzip hash.

@magnumripper back then when you provided this fix, did you test it with any hashes larger than ~32KB of data?

For some reason infstream.total_out in my case stops at 32'768 and does not increase further during the inflating process (even while uncompressing more data). And obviously, in that case the length check afterwards does not match.
Simply create an encrypted zip with a file larger than 32KB and smaller than 320KB to reproduce this. With version 6.2.4 (which was the last release before the change) it finds the password without any issue.

This makes hashcat never cracking any pkzip hash having uncompressed data larger than 32KB. 
Any idea why this counter is not increased further?

magnumripper:
Ouch, that's not good. I usually test with a decent set of samples but I can't recall how I tested this one.

Why did you close this issue though?

magnumripper:
I can reproduce. Very odd, just like you say the length doesn't increment beyond 32768. I have yet to spot why but it "has to" be some piece missing in `inc_zip_inflate.cl`... and I strongly believe my change in #2970 should stand.

s3inlc:
> Why did you close this issue though?

I was notified yesterday, that there was already an issue open regarding that bug (issue #3434), that's why I closed it. It seems that the total_out value does not seem to provide what the name would make you think. Apparently, it gets reset on certain cases, so it may not be the right value to use for the total output check.

magnumripper:
Oh, I see it now. I guess it differs from the normal library code because you are inflating into a smaller window. Easy fix then, just add a separate variable.

PR #3547, opened 2022-12-21
Fix a bug causing -m172xx (pkzip) to fail on data sizes > 32 KB

### LINKED ISSUE #3434: False negatives in (some) compressed PKZIP files
hashcat fails to crack certain compressed PKZIP files. I think the condition is if the uncompressed size is larger than 32768 bytes. More testing is certainly needed in that area!
This was [reported](https://discord.com/channels/566331281558732800/874753308953280613/1008755855681929268) by 64_nickel on Discord.

To reproduce I've attached the mentioned [list16-FL_kdIZUGpI.zip](https://github.com/hashcat/hashcat/files/9451947/list16-FL_kdIZUGpI.zip) file (and the with zip2john extracted hash [list16-FL_kdIZUGpI.txt](https://github.com/hashcat/hashcat/files/9451946/list16-FL_kdIZUGpI.txt)) from the CMIYC 2022 street hashes. The password is `Hackers`
```
$ echo Hackers | ./hashcat --username -m 17200 list16-FL_kdIZUGpI.txt
hashcat (v6.2.5-667-g7ca2627bc+) starting
<--snip-->
Session..........: hashcat
Status...........: Exhausted
Hash.Mode........: 17200 (PKZIP (Compressed))
Hash.Target......: $pkzip$1*1*2*0*6c59*b7ae*77d34619*0*50*8*6c59*48d4*...pkzip$
<--snip-->
```
This bug was introduced with #2970. While I think we do want to have that additional check, we can't rely on `infstream.total_out` because it gets [reset](https://github.com/hashcat/hashcat/blob/master/OpenCL/inc_zip_inflate.cl#L1210) in `hc_inflate()` if `pStream->avail_out` falls below `TINFL_LZ_DICT_SIZE`.

I've mainly opened this issue because I don't know what the best solution is. We could add a separate counter to track the uncompressed bytes but maybe there's a better/easier solution?

magnumripper:
We need a correct total_out size so a separate variable is introduced to maintain the window buffer.

Closes #3434 (see #3546).

### DIFF
--- a/OpenCL/inc_zip_inflate.cl
+++ b/OpenCL/inc_zip_inflate.cl
@@ -434,6 +434,7 @@ typedef struct mz_stream_s
     PRIVATE_AS unsigned char *next_out; /* pointer to next byte to write */
     unsigned int avail_out;  /* number of bytes that can be written to next_out */
     mz_ulong total_out;      /* total number of bytes produced so far */
+    mz_ulong window_out;     /* total number of bytes produced in window so far */
 
     PRIVATE_AS char *msg;                       /* error msg (unused) */
     PRIVATE_AS inflate_state *state; /* internal state, allocated by zalloc/zfree */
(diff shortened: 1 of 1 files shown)


s3inlc:
Thanks for the quick fix, it seems to work now for me.

jsteube:
Many thanks!

PR #3548, opened 2022-12-22
Allow up to 11 chars for the P value of PDFs

s3inlc:
According to  the PDF standard, the P value is a signed 32 bit integer, therefore the value can get longer than 6 chars (what hashcat currently accepts as limit).

A generated example hash (-m 10400) of with such a value like it can happen from some PDFs:
```
$pdf$1*2*40*-2147483647*0*16*65425250680715858027578013731348*32*7d586b1742bd978aa1834d14cbc5265920268faca3980014b1f50b064a6116c8*32*0000000000000000000000000000000000000000000000000000000000000000:hashcat
```

### DIFF
--- a/src/modules/module_10400.c
+++ b/src/modules/module_10400.c
@@ -150,7 +150,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
                    | TOKEN_ATTR_VERIFY_DIGIT;
 
   token.len_min[4] = 1;
-  token.len_max[4] = 6;
+  token.len_max[4] = 11;
   token.sep[4]     = '*';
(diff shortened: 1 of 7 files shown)


Chick3nman:
This is a patch I've been putting off adding, mostly because the P value is not even being used in some of the modes. There's quite a few things about the PDF modes that needs to overhauled imo but we'll discuss elsewhere.

That said, this has basically no implications for the usage due to P not being considered in the kernel, so easy change.

Chick3nman (approved):

s3inlc:
> That said, this has basically no implications for the usage due to P not being considered in the kernel, so easy change.

That's actually only true for the 10410/10420 kernels, in the other ones the P value is taken in as part of the hashing process.

PR #3561, opened 2023-01-04
fixes #3487: added -m 3090x = Bitcoin raw private key

### LINKED ISSUE #3487: Feature request:Bitcoin HEX private key
Hashcat has mode 28501 ( Bitcoin WIF private key (P2PKH), compressed ). Works great, great speed! But the problem is that my private key is in HEX format  with missing characters, so i can't use 28501 mode and i can't convert my Hex private key with missing characters to WIF format to use it with mode 28501.
For example - 28501 mode:
1PNsAQsneakL4CJyhDMoX7goq76MvXzAGm:L1ULo4htHVh1Vanw6YTNGWpjZrnXjTC5T1mJZFNozRZkzZEU2M6B
New Hash mode:
1PNsAQsneakL4CJyhDMoX7goq76MvXzAGm:7ee05e336a5303dee1df768c2520a46de16f4830eef4648447c3654c403d6f61

Thanks!

philsmd:
This commit adds the following hash modes (requested in #3487):
- -m  30901 = `Bitcoin raw private key (P2PKH), compressed`
- -m 30902 = `Bitcoin raw private key (P2PKH), uncompressed`
- -m 30903 = `Bitcoin raw private key (P2WPKH, Bech32), compressed`
- -m 30904 = `Bitcoin raw private key (P2WPKH, Bech32), uncompressed`
- -m 30905 = `Bitcoin raw private key (P2SH(P2WPKH)), compressed`
- -m 30906 = `Bitcoin raw private key (P2SH(P2WPKH)), uncompressed`

They are similar to the hash modes -m 28501 to -m 28506 that we already support (the `WIF` format of the private key), but these new hash modes support the raw 32 bytes (64 hex char) private keys as password.

This should be all straight forward, of course the speed is different to the WIF hash modes, because we have no more advantage of the quick/fast base58check (checksum) testing (early reject) here.

Thank you very much


philsmd:
There is maybe one tuning thing that @jsteube might want to double-check / correct: I had to introduce the `module_kernel_loops_max ()` (see for instance https://github.com/hashcat/hashcat/blob/3ef69f2919b5cfe0e0fb25c64dd1caf712285e39/src/modules/module_30901.c#L52-L57) with setting `-u = 16` otherwise the `--benchmark` took a little bit more time (and in general the cracking was slower for me). I think this could have to do with too aggressive autotuning, but please correct me (and the code) if that is a wrong assumption.

Thanks

jsteube:
What I find really strange that by offloading this section

```
    // convert password from hex to binary

    u32 tmp2[16] = { 0 };

    for (u32 i = 0, j = 0; i < 8; i += 1, j += 2)
    {
      tmp2[i] = hex_u32_to_u32 (w[j + 0], w[j + 1]);
    }
```

to one level above, there's almost no speed gain. I think the compiler was just clever enough to do that by itself. I also tried using different sections for the inner loop, but 16 seems to be a good one, therefore we leave it.

jsteube:
I've also found that limiting the thread count also improves the performance slightly on my 2080Ti. Could you please confirm it also improves the performance on your system? We can then apply the same limit to the remaining hash-modes.

philsmd:
hmm...
@jsteube for my 3080ti the default of the `--benchmark` with 32 threads `-T 32` is the fastest with about 3600 kH/s (`./hashcat -m 30901 -b`, same as `./hashcat -m 30901 -b --force -T 32`).
(**update** this was a misunderstanding, see update below)
Lower (or higher) values for `-T` give me lower speeds (again, this test is on a 3080ti, other GPUs might be slightly different).


I was also curious and tested with binary input (by removing the hex conversation completely) and the speed was about the same (again):
```diff
diff --git a/OpenCL/m30901_a3-pure.cl b/OpenCL/m30901_a3-pure.cl
index 6df675272..5e47e1329 100644
--- a/OpenCL/m30901_a3-pure.cl
+++ b/OpenCL/m30901_a3-pure.cl
@@ -60,28 +60,6 @@
 //   }
 // };
 
-DECLSPEC u32 hex_convert_u32 (PRIVATE_AS const u32 c)
-{
-  return (c & 15) + (c >> 6) * 9;
-}
-
-DECLSPEC u32 hex_u32_to_u32 (PRIVATE_AS const u32 hex0, PRIVATE_AS const u32 hex1)
-{
-  u32 v = 0;
-
-  v |= hex_convert_u32 ((hex0 >>  0) & 0xff) << 28;
-  v |= hex_convert_u32 ((hex0 >>  8) & 0xff) << 24;
-  v |= hex_convert_u32 ((hex0 >> 16) & 0xff) << 20;
-  v |= hex_convert_u32 ((hex0 >> 24) & 0xff) << 16;
-
-  v |= hex_convert_u32 ((hex1 >>  0) & 0xff) << 12;
-  v |= hex_convert_u32 ((hex1 >>  8) & 0xff) <<  8;
-  v |= hex_convert_u32 ((hex1 >> 16) & 0xff) <<  4;
-  v |= hex_convert_u32 ((hex1 >> 24) & 0xff) <<  0;
-
-  return (v);
-}
-
 KERNEL_FQ void m30901_mxx (KERN_ATTR_VECTOR ())
 {
   /**
@@ -99,23 +77,18 @@ KERNEL_FQ void m30901_mxx (KERN_ATTR_VECTOR ())
 
   const u32 pw_len = pws[gid].pw_len;
 
-  if (pw_len != 64) return;
+  if (pw_len != 32) return;
 
 
   // copy password to w
 
-  u32 w[16];
+  u32 w[8];
 
-  for (u32 i = 0; i < 16; i++) // pw_len / 4
+  for (u32 i = 0; i < 8; i++) // pw_len / 4
   {
     w[i] = pws[gid].i[i];
   }
 
-  for (u32 i = 1; i < 16; i++)
-  {
-    if (is_valid_hex_32 (w[i]) == 0) return;
-  }
-
   secp256k1_t preG; // need to change SECP256K1_TMPS_TYPE above to: PRIVATE_AS
 
   set_precomputed_basepoint_g (&preG);
@@ -135,28 +108,19 @@ KERNEL_FQ void m30901_mxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
-    if (is_valid_hex_32 (w[0]) == 0) continue;
-
 
-    // convert password from hex to binary
-
-    u32 tmp[16] = { 0 };
-
-    for (u32 i = 0, j = 0; i < 8; i += 1, j += 2)
-    {
-      tmp[i] = hex_u32_to_u32 (w[j + 0], w[j + 1]);
-    }
+    // private key
 
     u32 prv_key[9];
 
-    prv_key[0] = tmp[7];
-    prv_key[1] = tmp[6];
-    prv_key[2] = tmp[5];
-    prv_key[3] = tmp[4];
-    prv_key[4] = tmp[3];
-    prv_key[5] = tmp[2];
-    prv_key[6] = tmp[1];
-    prv_key[7] = tmp[0];
+    prv_key[0] = hc_swap32 (w[7]);
+    prv_key[1] = hc_swap32 (w[6]);
+    prv_key[2] = hc_swap32 (w[5]);
+    prv_key[3] = hc_swap32 (w[4]);
+    prv_key[4] = hc_swap32 (w[3]);
+    prv_key[5] = hc_swap32 (w[2]);
+    prv_key[6] = hc_swap32 (w[1]);
+    prv_key[7] = hc_swap32 (w[0]);
 
 
     // convert: pub_key = G * prv_key
@@ -192,6 +156,8 @@ KERNEL_FQ void m30901_mxx (KERN_ATTR_VECTOR ())
     sha256_update (&ctx, pub_key, 33); // length of public key: 33
     sha256_final  (&ctx);
 
+    u32 tmp[16] = { 0 };
+
     for (u32 i = 0; i < 8; i++) tmp[i] = ctx.h[i];
 
     // tmp[ 8] = 0; tmp[ 9] = 0; tmp[10] = 0; tmp[11] = 0;
@@ -246,23 +212,18 @@ KERNEL_FQ void m30901_sxx (KERN_ATTR_VECTOR ())
 
   const u32 pw_len = pws[gid].pw_len;
 
-  if (pw_len != 64) return;
+  if (pw_len != 32) return;
 
 
   // copy password to w
 
-  u32 w[16];
+  u32 w[8];
 
-  for (u32 i = 0; i < 16; i++) // pw_len / 4
+  for (u32 i = 0; i < 8; i++) // pw_len / 4
   {
     w[i] = pws[gid].i[i];
   }
 
-  for (u32 i = 1; i < 16; i++)
-  {
-    if (is_valid_hex_32 (w[i]) == 0) return;
-  }
-
   secp256k1_t preG; // need to change SECP256K1_TMPS_TYPE above to: PRIVATE_AS
 
   set_precomputed_basepoint_g (&preG);
@@ -282,28 +243,19 @@ KERNEL_FQ void m30901_sxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
-    if (is_valid_hex_32 (w[0]) == 0) continue;
 
-
-    // convert password from hex to binary
-
-    u32 tmp[16] = { 0 };
-
-    for (u32 i = 0, j = 0; i < 8; i += 1, j += 2)
-    {
-      tmp[i] = hex_u32_to_u32 (w[j + 0], w[j + 1]);
-    }
+    // private key
 
     u32 prv_key[9];
 
-    prv_key[0] = tmp[7];
-    prv_key[1] = tmp[6];
-    prv_key[2] = tmp[5];
-    prv_key[3] = tmp[4];
-    prv_key[4] = tmp[3];
-    prv_key[5] = tmp[2];
-    prv_key[6] = tmp[1];
-    prv_key[7] = tmp[0];
+    prv_key[0] = hc_swap32 (w[7]);
+    prv_key[1] = hc_swap32 (w[6]);
+    prv_key[2] = hc_swap32 (w[5]);
+    prv_key[3] = hc_swap32 (w[4]);
+    prv_key[4] = hc_swap32 (w[3]);
+    prv_key[5] = hc_swap32 (w[2]);
+    prv_key[6] = hc_swap32 (w[1]);
+    prv_key[7] = hc_swap32 (w[0]);
 
 
     // convert: pub_key = G * prv_key
@@ -339,6 +291,8 @@ KERNEL_FQ void m30901_sxx (KERN_ATTR_VECTOR ())
     sha256_update (&ctx, pub_key, 33); // length of public key: 33
     sha256_final  (&ctx);
 
+    u32 tmp[16] = { 0 };
+
     for (u32 i = 0; i < 8; i++) tmp[i] = ctx.h[i];
 
     // tmp[ 8] = 0; tmp[ 9] = 0; tmp[10] = 0; tmp[11] = 0;
diff --git a/src/modules/module_30901.c b/src/modules/module_30901.c
index af4021761..77ba585ed 100644
--- a/src/modules/module_30901.c
+++ b/src/modules/module_30901.c
@@ -26,11 +26,11 @@ static const u32   OPTI_TYPE         = OPTI_TYPE_NOT_SALTED;
 static const u64   OPTS_TYPE         = OPTS_TYPE_STOCK_MODULE
                                      | OPTS_TYPE_PT_GENERATE_LE;
 static const u32   SALT_TYPE         = SALT_TYPE_NONE;
-static const char *ST_PASS           = "59887ec9920239bd45b6a9f82b7c4e024f80beaf887e5ee6aac5de0a899d3068";
+static const char *ST_PASS           = "\x59\x88\x7e\xc9\x92\x02\x39\xbd\x45\xb6\xa9\xf8\x2b\x7c\x4e\x02\x4f\x80\xbe\xaf\x88\x7e\x5e\xe6\xaa\xc5\xde\x0a\x89\x9d\x30\x68";
 static const char *ST_HASH           = "14Fqy5AGRehazZ4NLzxFWy2E4BiNFdH9Ut";
-static const char *BENCHMARK_MASK    = "?h?h?h?h?h?h?h9920239bd45b6a9f82b7c4e024f80beaf887e5ee6aac5de0a899d3068";
+static const char *BENCHMARK_MASK    = "?b?b?b?b?b\x02\x39\xbd\x45\xb6\xa9\xf8\x2b\x7c\x4e\x02\x4f\x80\xbe\xaf\x88\x7e\x5e\xe6\xaa\xc5\xde\x0a\x89\x9d\x30\x68";
 static const u32   PUBKEY_MAXLEN     = 64; // our max is actually always 25 (21 + 4)
-static const u32   RAW_LEN           = 64;
+static const u32   RAW_LEN           = 32;
 
 
 u32         module_attack_exec       (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ATTACK_EXEC;     }

```
This was a note/suggestion I've already mentioned here (https://github.com/hashcat/hashcat/issues/3487#issuecomment-1371072259), but again, the speed gain is not significant, so I would suggest we keep (at least this hexadecimal password input) AS-IS for now.

Yeah, we could still keep trying to test a little bit around with speeds for this particular hash mode, but for me the only thing that worked to improve speed was the `-u 16` restriction (as already commited).
It's clear that the **bottleneck** for this hash mode is the ECC multiplication, but that is kind of obvious (and btw it is already really, really fast, because we do every computation on the GPU with our secp256k1 kernel support).

---

**update**: I think I didn't realize that @jsteube already pushed the `-T 32` restriction here (https://github.com/hashcat/hashcat/commit/d19882ff71ad14cb2c252b47b1329bb39ad65fdd)... .yeah this commit seems to be very good, because it uses the perfect tuning. I think we should push this for all 3090x hash modes. I agree !

---

**update 2**: I've now pushed the same `-T 32` improvement for the other hash modes (from -m 30902 up to -m 30906), see https://github.com/hashcat/hashcat/commit/d19882ff71ad14cb2c252b47b1329bb39ad65fdd 
Thanks, very good suggestion

PR #3566, opened 2023-01-07
QoL status hex fix

PenguinKeeper7:
The current behaviour is that both `Candidates` in the `[s]tatus` prompt are hexxed if only one needs it, this fix will only hex the candidates that need it, be it neither, one or both.

Current:
`Candidates.#1....: $HEX[313537393434] -> $HEX[042a0337c2a1]`

Fixed:
`Candidates.#1....: 157944 -> $HEX[042a0337c2a1]`

### DIFF
--- a/src/status.c
+++ b/src/status.c
@@ -865,23 +865,24 @@ char *status_get_guess_candidates_dev (const hashcat_ctx_t *hashcat_ctx, const i
   const bool need_hex1 = need_hexify (plain_ptr1, plain_len1, 0, always_ascii);
   const bool need_hex2 = need_hexify (plain_ptr2, plain_len2, 0, always_ascii);
 
-  if ((need_hex1 == true) || (need_hex2 == true))
-  {
+  // Left candidate needs to be $HEX-ed
+  if(need_hex1 == true){
     exec_hexify (plain_ptr1, plain_len1, plain_ptr1);
-    exec_hexify (plain_ptr2, plain_len2, plain_ptr2);
-
     plain_ptr1[plain_len1 * 2] = 0;
+  }
+  else{plain_ptr1[plain_len1] = 0;}
+
+  // Right candidate needs to be $HEX-ed
+  if(need_hex2 == true){
+    exec_hexify (plain_ptr2, plain_len2, plain_ptr2);
     plain_ptr2[plain_len2 * 2] = 0;
-
-    snprintf (display, HCBUFSIZ_TINY, "$HEX[%s] -> $HEX[%s]", plain_ptr1, plain_ptr2);
   }
-  else
-  {
-    plain_ptr1[plain_len1] = 0;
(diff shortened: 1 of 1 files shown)


Chick3nman (changes requested):
Looks like we can avoid some of the if/else comparisons here by changing the structure to actually branch on the earliest indication that its necessary. Here's a quick rewrite I just did, please test and update the PR if it's passing. I didn't add comments or test yet as I've just written it here so it may make sense to add some, though the comparisons are relatively easy to follow.

```
if(need_hex1 == true || need_hex2 == true) {
    if(need_hex1 == false) {
        exec_hexify (plain_ptr2, plain_len2, plain_ptr2);
        
        plain_ptr1[plain_len1] = 0;
        plain_ptr2[plain_len2 * 2] = 0;
        
        snprintf (display, HCBUFSIZ_TINY, "%s -> $HEX[%s]", plain_ptr1, plain_ptr2);
    }
    else if(need_hex2 == false) {
        exec_hexify (plain_ptr1, plain_len1, plain_ptr1);
        
        plain_ptr1[plain_len1 * 2] = 0;
        plain_ptr2[plain_len2] = 0;

        snprintf (display, HCBUFSIZ_TINY, "$HEX[%s] -> %s", plain_ptr1, plain_ptr2);
    }
    else {
        exec_hexify (plain_ptr1, plain_len1, plain_ptr1);
        exec_hexify (plain_ptr2, plain_len2, plain_ptr2);
        
        plain_ptr1[plain_len1 * 2] = 0;
        plain_ptr2[plain_len2 * 2] = 0;
        
        snprintf (display, HCBUFSIZ_TINY, "$HEX[%s] -> $HEX[%s]", plain_ptr1, plain_ptr2);
    }
}
else {
    plain_ptr1[plain_len1] = 0;
    plain_ptr2[plain_len2] = 0;
    
    snprintf (display, HCBUFSIZ_TINY, "%s -> %s", plain_ptr1, plain_ptr2);
}

```

PenguinKeeper7:
I mean, it doesn't really need to be *that* performant when [s]tatus calls happen so infrequently so imo having the readability will be a bigger benefit than saving < millisecond of binary 1=1 comparisons

Chick3nman:
Momentarily ignoring the readability differences, the suggested code change is also closer to being compliant with the projects standards and won't require as much cleanup to merge. I used 4 spaces instead of 2 by default so it needs to be respaced and I opened my braces on the same line as my control statements as well, so my suggestion isn't _exactly_ Allman style like it should be but it's much closer. The original PR is not in a good state to be merged style wise, hence the suggested rewrite, so redoing the branching for performance is an added bonus with the required cleanup.

PenguinKeeper7:
I still disagree but I've tested and updated your code with 2 spaces and curly braces on separate lines

jsteube:
Thanks for update

PR #3571, opened 2023-01-09
Removed check which aborted on empty passwords for mode 10700

### LINKED ISSUE #3562: Empty passwords can not be cracked in PDF 10700 mode
PDF 1.7 Level 8 Hashes (-m 10700) where the user password is empty are not cracked due to a check of the password length in the middle of the computation in the _loop() kernel call (see https://github.com/hashcat/hashcat/blob/master/OpenCL/m10700-pure.cl#L1259 and https://github.com/hashcat/hashcat/blob/master/OpenCL/m10700-pure.cl#L1259).

It is unclear to me why this is done and it causes hashes for such PDFs to never being cracked. Of course there may not be that many cases where a PDF with an empty password needs to be attacked, as it can be opened directly, but this is not detectable if only the hash is given. 

This check was introduced in 2b9715944fb047097f21916499ec032f5dc6d13c, was this a specific purpose to solve the mentioned issue? @philsmd 
If there is not a specific reason why this check must be there, I would propose to remove it, as in this case, the user cannot even be aware that a password was not checked completely as it does not go into the rejected count.

s3inlc:
the check for pw_len == 0 made pdf hashes of mode 10700 with empty user password to never crack. closes #3562

### DIFF
--- a/OpenCL/m10700-optimized.cl
+++ b/OpenCL/m10700-optimized.cl
@@ -647,8 +647,6 @@ KERNEL_FQ void m10700_loop (KERN_ATTR_TMPS_ESALT (pdf17l8_tmp_t, pdf_t))
 
   const u32 pw_len = pws[gid].pw_len & 31;
 
-  if (pw_len == 0) return;
-
   /**
    * digest
    */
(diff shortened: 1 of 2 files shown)


jsteube:
I think we had this added to avoid a never ending loop situation, but it was more a result for a problem existing only in optimized mode because of some very specific optimization. How did you test the change?

s3inlc:
I generated around 4.5k of these PDF hashes with passwords from example.dict, they were all successfully recovered. Additionally, I generated around 1k hashes with an empty password, these were also all successfully recovered.

jsteube:
OK, thanks!

PR #3583, opened 2023-01-19
Support for longer scrypt hashes in module 8900

jmvermeulen:
The is a open source tool that creates crypted file systems, the hash is normal scrypt but the module 8900 does not support the token size.

The example hash (resolves to the password "hashcat"):
`SCRYPT:65536:8:1:M/oaRDW94PYv/L0qjTpcFrViwtPJapZxe8UWIipf/n4=:3ntSy4wA+TNxmSlOzJdFLmqQOm5UqbDlgYXckF+MDmrmD5E+krpW9bkKMzaBFqqwkZIZV9nMHkTptcYOO0Imew==` results in `Hashfile './testhash.hash' on line 1 (SCRYPT...MzaBFqqwkZIZV9nMHkTptcYOO0Imew==): Token length exception`

The digest max length in the module was 44, but in this case the digest has a length of 88.
The token validation in the module accepts 88 length now and the digest buffer was doubled in size.

I could successfully crack the hash now.

### DIFF
--- a/src/modules/module_08900.c
+++ b/src/modules/module_08900.c
@@ -304,7 +304,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
                    | TOKEN_ATTR_VERIFY_BASE64A;
 
   token.len_min[5] = 44;
-  token.len_max[5] = 44;
+  token.len_max[5] = 88;
   token.sep[5]     = ':';
   token.attr[5]    = TOKEN_ATTR_VERIFY_LENGTH
                    | TOKEN_ATTR_VERIFY_BASE64A;
@@ -333,7 +333,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
   const u8 *salt_pos = token.buf[4];
   const int salt_len = token.len[4];
 
-  u8 tmp_buf[33] = { 0 };
+  u8 tmp_buf[63] = { 0 };
 
   const int tmp_len = base64_decode (base64_to_int, (const u8 *) salt_pos, salt_len, tmp_buf);
 


jmvermeulen:
I think I've made a mistake, the hash is accepted but not cracked.

jsteube:
I think the hash ```SCRYPT:65536:8:1:M/oaRDW94PYv/L0qjTpcFrViwtPJapZxe8UWIipf/n4=:3ntSy4wA+TNxmSlOzJdFLmqQOm5UqbDlgYXckF+MDmrmD5E+krpW9bkKMzaBFqqwkZIZV9nMHkTptcYOO0Imew==``` is just not ```hashcat```

Here's a valid hash for 'hashcat' with digest length 64: ```SCRYPT:65536:8:1:NTg1MjI5NjMwMjMxNDE1NDU1NTgwNTIxOTcyNzQ3MzE=:aPcHjWA6yHg/k9eADV9kZQqok+xc0GYssIh6NWQXGWBzZ11MEGmiYzYGFwPV7NBFc6yX/0z+nVGHR95qhAPHSg==```

However there are some other problems for instance the salt size, so I'm going to update your PR after merge slightly.

jmvermeulen:
Your are right, the user password is hashcat but looking at the source code there is an additional step that leads to decryption of the master key. Will look at that and post an update.

Thanks so far!

jmvermeulen:
Code from the application that generates the hash with user password 'hashcat' I've posted before `SCRYPT:65536:8:1:M/oaRDW94PYv/L0qjTpcFrViwtPJapZxe8UWIipf/n4=:3ntSy4wA+TNxmSlOzJdFLmqQOm5UqbDlgYXckF+MDmrmD5E+krpW9bkKMzaBFqqwkZIZV9nMHkTptcYOO0Imew==`.

```
       // Generate derived key from password
	scryptHash := cf.ScryptObject.DeriveKey(password)

	// Unlock master key using password-based key
	useHKDF := cf.IsFeatureFlagSet(FlagHKDF)
	ce := getKeyEncrypter(scryptHash, useHKDF)

	masterkey, err = ce.DecryptBlock(cf.EncryptedKey, 0, nil)
```

@jsteube I think this would need a custom module for this specific application right?
Now on the the first hash will be generated but there is no way to test if its valid. Additionally testing a AES decrypt would be required.

jsteube:
Exactly, there will be another plugin needed. Testing some AES decryption after SCRYPT is very common, so it's not something untested or rare. There's plenty of examples for that and you can also use 8900 as the base plugin template. Sometimes if test decrypted buffer is not a fixed or known one, often we can use padding data to check of decryption was succesfull. If GCM is used, the tag also works. In any case, you typically have to add some sort of IV and encrypted data blob.

PR #3609, opened 2023-02-03
Fixes hashcat#3580: bug in --skip/--limit w/ --stdout

### LINKED ISSUE #3580: update message: "Use of --skip/--limit is not supported with --increment or mask files." when --stdout is used
when running an attack with --skip, it does not allow the use of --stdout. 

While it does throw this error, can this message be updated to include that? 

Suggestion: "Use of --skip/--limit is not supported with --increment, mask files, or --stdout"

gemesa:
Fixes #3580 

Before fix:
```
$ ./hashcat --stdout -a 0 in1.txt in2.txt         
1
2
3
a
b
c
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt in2.txt 
Use of --skip/--limit is not supported with --increment or mask files.
```
```
$ ./hashcat --stdout --limit 1 -a 0 in1.txt in2.txt
Use of --skip/--limit is not supported with --increment or mask files.
```

After fix:
```
$ ./hashcat --stdout -a 0 in1.txt in2.txt
1
2
3
a
b
c
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt in2.txt
Use of --skip/--limit is not supported with --stdout.
```
```
$ ./hashcat --stdout --limit 1 -a 0 in1.txt in2.txt 
Use of --skip/--limit is not supported with --stdout.
```

### DIFF
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -787,7 +787,7 @@ static int outer_loop (hashcat_ctx_t *hashcat_ctx)
   {
     if ((mask_ctx->masks_cnt > 1) || (straight_ctx->dicts_cnt > 1))
     {
-      event_log_error (hashcat_ctx, "Use of --skip/--limit is not supported with --increment or mask files.");
+      event_log_error (hashcat_ctx, "Use of --skip/--limit is not supported with --increment, mask files, or --stdout.");
 
       return -1;
     }


s3inlc:
Attention, this change limits too much regarding the usage of --skip/--limit with --stdout.

In the examples only the cases with two dictionaries are shown. In these cases it is correct that --skip/--limit cannot be used. But if there is only one dictionary given, the usage of --skip/--limit is (and should remain) allowed.

This should be possible:
```
./hashcat --stdout -a 0 example.dict --limit 1
```

This should not be possible (and produce a proper error message):
```
./hashcat --stdout -a 0 dict1.txt dict2.txt --limit 1
```

Essentially, the only required change is to adapt the line 790 and update the message: https://github.com/hashcat/hashcat/blob/master/src/hashcat.c#L790

gemesa:
You are right @s3inlc. I did not think of testing it with only 1 dictionary because I always use multiple, that is my bad. I updated the commit accordingly.

Before fix:
```
$ ./hashcat --stdout -a 0 in1.txt                                                               
1
2
3
```
```
$ ./hashcat --stdout -a 0 in1.txt in2.txt
1
2
3
a
b
c
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt         
2
3
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt in2.txt
Use of --skip/--limit is not supported with --increment or mask files.
```
After fix:
```
$ ./hashcat --stdout -a 0 in1.txt                                                               
1
2
3
```
```
$ ./hashcat --stdout -a 0 in1.txt in2.txt
1
2
3
a
b
c
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt         
2
3
```
```
$ ./hashcat --stdout --skip 1 -a 0 in1.txt in2.txt
Use of --skip/--limit is not supported with --increment, mask files, or --stdout.
```

Issue #3340, opened 2022-06-24
adding support for SM3


junshunhu:
SM3 is a cryptographic hash function used in the Chinese National Standard.Is it possible to support this algorithm?

jsteube:
It looks a lot like SHA2 and I think is a GPU friendly algorithm. I can see there's some nice implementation in linux coreutils, too. So technically is not a problem is more like the number of requests. Where is it used in?

mdawsonuk:
For one, it is used in the HMAC-SM3 format in Werkzeug, which I'm working through implementing hash formats in #3368

loukabvn:
Hi, I would like to contribute for this issue, is it still available ?

mdawsonuk:
> Hi, I would like to contribute for this issue, is it still available ?

Doesn't look like anyone's made a PR for it, so go ahead!

Zgzorx:
Issue resolved by PR : https://github.com/hashcat/hashcat/pull/3620

Chick3nman:
Looks like this was added in #3620, closing as complete.

PR #3620, opened 2023-02-17
Adding support for SM3

Zgzorx:
Issue : https://github.com/hashcat/hashcat/issues/3340

We (@loukabvn and I) are adding SM3 a0 a1 a3 pure and optimized.

Perl is lacking SM3 implementation so we are discussing the way for the modules to be tested.


Zgzorx:
Has mentionned previously, Perl can't handle SM3, what kind of test can we give ?

jsteube:
Wow, that's really cool. Algorithm reminds me a lot on SHA256 :)
Regarding the perl module, it's optional. But maybe you can use SHA256 implementation in perl as template.

jsteube:
Let me know when you think it's ready for merge. Please rename to 31100.

Zgzorx:
I used Python to run my own tests and everything seems in order. Note that I don't have access to any CUDA capable chip and @loukabvn is unavailable for a week. Could you run a benchmark please ? If it passes, I think it's ready for merge.

jsteube:
I manually checked -a 0 and -a 3 and it worked fine.

jsteube:
Many thanks for contribution, this is great work!

jsteube:
@Zgzorx @loukabvn Please see commit: https://github.com/hashcat/hashcat/commit/9acfc26d818dc3e2f2a7a00a850e09ca0da67657

It's just a tiny optimization (on my 2080ti from 6.6GH/s to 6.9GH/s), but I think it can be replicated to other optimized kernels (single hash only).
Just in case you enjoy optimizing same way I do :)

Issue #3582, opened 2023-01-18
Make hashcat understand FIFOs (named pipes)


ghost:
**Describe the bug**

`-o/--outfile` doesn't work when the outfile is a FIFO. The cracking process doesn't even start. Hashcat only prints `hashcat (v6.2.6) starting` and then hangs indefinitely.

**To Reproduce**
```
mkfifo fifo
hashcat.bin -m 0 -o fifo -a 3 827ccb0eea8a706c4c34a16891f84e7b ?d?d?d?d?d
```

and in another console

```
cat fifo
```

**Expected behavior**

hashcat successfully cracks the hash (`12345`).

**Hashcat version (please complete the following information):**
 - OS: Linux
 - Distribution: Gentoo
 - Version: 6.2.6

PR #3622, opened 2023-02-18
Don't close and reopen a FIFO

ghost:
It doesn't really work beyond a single write if we're closing and reopening a FIFO outfile. Fix that.

Updates #3582.


ghost:
Since we don't close the FIFO outfile now until we quit, we want to flush the stream for the receiving side to receive the founds immediately.

ghost:
And now it seems to work famously.

PR #3626, opened 2023-02-22
Add tuning records for a number of algorithms for the AMD RX 7900XTX

ketlerd:
It's my first time building tunings for a card, so please correct me if I'm done anything incorrectly. 

I added an alias for the full name and gfx1100 for the 7900XTX, and added tunings for the majority of the algorithms with current tunings files.

The uplift is a mixed bag, averaging somewhere around 15%, with the outliers being 09300 (cisco-ios scrypt), and 15700 (eth, scrypt) with > 100% improvements.

It's a start at least.


jsteube:
Many thanks!

Issue #3618, opened 2023-02-13
Teach hashcat to exit gracefully on SIGINT


ghost:
hashcat should exit gracefully on interrupt, exactly as if `q` was pressed. The second SIGINT will still kill it as before.

mmtawous:
@jsteube 

I have looked into this for a while and can't think of a way to implement it without introducing a global variable into the equation. I know this project has been refactored in the past to remove all global variables for the sake of memory improvements so I am wondering if you could make an exception for this one case. 

The only global variable needed would be a pointer to the hashcat context; it could be stored in hashcat.c and accessed from the thread implementation which would contain the signal handlers. There seems to already be commented code there for signal handling.

roycewilliams:
@justpretending - based on the discussion in #3618, should this issue also be closed?

mmtawous:
I'd say it seems reasonable to close this issue. @jsteube was not fond of the idea of exiting gracefully on SIGINT regardless of the implementation complications.

PR #3633, opened 2023-03-02
Added graceful exiting on SIGINT: In response to issue #3618

mmtawous:
Implemented graceful exiting on SIGINT using already existing code and a newly added global variable holding a pointer to the hashcat context which is need for the signal handlers to be able to call the myquit and myabort functions.

### DIFF
--- a/include/thread.h
+++ b/include/thread.h
@@ -58,7 +58,7 @@
 
 #endif
 
-/*
+
 #if defined (_WIN)
 
 BOOL WINAPI sigHandler_default (DWORD sig);
@@ -72,7 +72,7 @@ void sigHandler_benchmark (int sig);
 void hc_signal (void (callback) (int));
 
 #endif
-*/
+
 
 int mycracked (hashcat_ctx_t *hashcat_ctx);
 int myabort_runtime (hashcat_ctx_t *hashcat_ctx);
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -54,6 +54,9 @@
 #include "brain.h"
 #endif
 
+// hashcat_ctx stored globally for signal handling
+hashcat_ctx_t *hc_ctx;
+
 // inner2_loop iterates through wordlists, then calls kernel execution
 
 static int inner2_loop (hashcat_ctx_t *hashcat_ctx)
@@ -1093,6 +1096,7 @@ int hashcat_init (hashcat_ctx_t *hashcat_ctx, void (*event) (const u32, struct h
   hashcat_ctx->user_options       = (user_options_t *)        hcmalloc (sizeof (user_options_t));
   hashcat_ctx->wl_data            = (wl_data_t *)             hcmalloc (sizeof (wl_data_t));
 
+  hc_ctx = hashcat_ctx;
   return 0;
 }
 
@@ -1742,6 +1746,19 @@ int hashcat_session_execute (hashcat_ctx_t *hashcat_ctx)
     if (user_options->identify == true) return 0;
 
     user_options->autodetect = false;
+  }
+
+  /**
+   * enable custom signal handler(s)
+   */
+
+  if (user_options->benchmark == false)
+  {
+    hc_signal (sigHandler_default);
+  }
+  else
+  {
+    hc_signal (sigHandler_benchmark);
   }
 
   /**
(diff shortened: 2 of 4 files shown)


BenBE:
This patch introduces a race condition on the signal handler when the main hashcat context is freed but there is a signal to be handled.

Also I don't think putting this hook in `hashcat_init` is clean, as functions should not cause such side effects to outside their scope. A better place IMHO would be managing that global pointer from https://github.com/hashcat/hashcat/blob/ad3dca2a90eb2214a79772b2c2b0f548a9520b12/src/main.c#L1203, probably best around https://github.com/hashcat/hashcat/blob/ad3dca2a90eb2214a79772b2c2b0f548a9520b12/src/main.c#L1287.

In the signal handler itself, check for the pointer to be non-NULL before passing it on and reset the global context pointer to NULL before `hashcat_destroy` is called (avoid UAF).

jsteube:
My main problem with this PR is not the technical details (because there will be plenty) and actually I'm not even there yet. I just think that if we go this route, we make it a lot harder for the user to kill hashcat ungracefully. That's a feature! There might be a good reason for the user why they want to kill it ungracefully. IOW. I find this is a power that we should leave to the user. 

From my own experience with programs which intercept ctrl-c is just that I even more press ctrl-c trying to overrule their own graceful shutdown logic. I feel annoyed by the program which took away that power from me.

Since compute APIs/drivers are still unstable hashcat runs into problems that look like a hashcat problem to the user, but that are actually problems of the compute API. If we then make it harder for the user to kill hashcat in such a case that could be very frustrating and also add a real reason for the user to blame hashcat.

I was thinking what's the reason for this PR. Maybe you could explain how did you came up with this PR in the first place.

mmtawous:
You make some excellent points! I was only responding to an issue I found (#3618). Maybe @justpretending could tell us what the original motive was.

I will say that a similar implementation was present in this project back in September of 2016, so that was my motive of bringing it back.

BenBE:
I think there's nothing completely wrong with performing a graceful shutdown on Ctrl+C, if you make sure the response to the shutdown request is performed reasonably fast. There are three camps for this:

1. Just instantly exit, skipping the usual "save settings" step normally performed on shutdown
2. Shutdown any current actions gracefully, but with a timeout not present otherwise
3. Take it as a hint that the user might want to shutdown the program sometime in the near future.

While from a user perspective solution 1 is the most intuitive (as it most resembles what not handling Ctrl+C does), IMHO solution 2 is more what I'd expect in the case of hashcat: Finish all current work on the GPU/CPU and do not upload any new tasks, but otherwise finish processing gracefully. If this was to hang for some reason beyond some timeout, the task should be forced to terminate.

If you really need a program killed instantly, there's a sledgehammer called SIGKILL …

ghost:
The problem I'm trying to solve is hashcat not flushing debug output (it's buffered) to the debug file when Ctrl+C is pressed. And you can't control hashcat when it's reading  words not from a wordlist but from stdin. You can't make it quit gracefully in this case by pressing `q`.

jsteube:
I think we're going to make Hashcat too hard to debug by adding support for unexpected external inputs like signals. You may be right that there are corner cases, but these are just corner cases. I think you can solve them by simply changing the way you use hashcat.

An example is that the PR (as it is at the time of writing) cause hashcat to run into segmentation fault in some cases. For example, just start hashcat in -b and immediately after the startup screen, when you see "Benchmark relevant options", press ctrl-c and a segfault will occur. The problem isn't this particular problem, which is certainly just some kind of null pointer dereference, but all the different cases that we don't currently see because the factors that lead to these circumstances are difficult to control. And I don't mean the Hashcat sources or their architecture but the influence of 3rd party libraries like JiT compilers and then the impact because Hashcat doesn't only run on Linux.

Also, it's important to note that hashcat is a library. It's just that users don't see that when they run hashcat because they only see the frontend to that library. But the Hashcat library currently allows multiple instances within the same thread. But with this change it won't work anymore because of the global variable. Maybe Hashcat overlays already exist that use the Hashcat library directly, and I'd really like to see more projects that do this instead of using the command line tool.

I end up having a bad feeling about this PR, so I better reject. Sorry for the time you spend.

PR #3628, opened 2023-02-26
Implement metamask 26610 short data 

thatux:
This PR enables cracking metamask wallets that contain a lot of data and thus solves https://github.com/hashcat/hashcat/issues/3429

26600 required all data as it compares the AES-GCM tag
26610 just tries to AES-GCM decrypt the first block of data and checks whether the plaintext is ascii.

I've added new reference data to generate the hash, as the original reference hash https://github.com/hashcat/hashcat/blob/master/src/modules/module_26600.c#L29
 decrypts to all 0xFF. That hash is probably generated by on https://github.com/hashcat/hashcat/blob/master/tools/test_modules/m26600.pm#L61
```
#!/usr/bin/env python3

from base64 import b64decode
from hashlib import pbkdf2_hmac

from Crypto.Cipher import AES 

[...]

PR #3637, opened 2023-03-08
Correct m25400 m26610 plaintextcheck

### LINKED ISSUE #3181: confusion between PDF mode 10500 and 25400
hi all, 

I am confused by the following two hashcat modes:
```
  10500 | PDF 1.4 - 1.6 (Acrobat 5 - 8)                       | Document
  25400 | PDF 1.4 - 1.6 (Acrobat 5 - 8) - user and owner pass | Document
```
I have a PDF 1.4 file:
```
user@host:~$ head -n1 test.pdf 
%PDF-1.4
```
from which I extracted the hash using `pdf2john.pl`, which comes out as:
```
$pdf$2*3*128*-1324*1*16*6adf3addaeb473206a2c780c706250f6*32*43f239ad53458a920d01393da38347e200000000000000000000000000000000*32*6a0549e56631c0b97b69723baca43b412f7ae6439a513d4906dfd11fa60a8afe
```
**the PDF file can be opened at will (no "user password", I believe) yet it is secured against modification ("owner password", I believe).**

the "hash examples" webpage does not help me detect which mode is more appropriate.
yet a bruteforce attack using mode 25400 **seems to find the solution**:
```
user@host:~$ ./hashcat.bin -a 3 -m 25400 -i "$HASH" "?a?a?a?a"
[...]
$pdf$2*3*128*-1324*1*16*6adf3addaeb473206a2c780c706250f6*32*43f239ad53458a920d01393da38347e200000000000000000000000000000000*32*6a0549e56631c0b97b69723baca43b412f7ae6439a513d4906dfd11fa60a8afe*:$HEX[592f6c2020202028757365722070617373776f72643d7e277a3a5c1a29192720706616356e6e29]
                                                          
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 25400 (PDF 1.4 - 1.6 (Acrobat 5 - 8) - user and owner pass)
Hash.Target......: $pdf$2*3*128*-1324*1*16*6adf3addaeb473206a2c780c706...a8afe*
[...]
```

on the other hand, attack mode 10500 exhausts all the possible passwords within the same attack space, in this case up to four `?a` characters.
I have tried the same steps multiple times and sometimes hashcat seems to crack the same hash with different solutions, even with different lengths.

anyway, this specific solution amounts to:
```
user@host:~$ echo 592f6c20202020287573657220706173737 | xxd -r -p
Y/l    (user password=~'z:\�)' pf5nn)
```
first of all, this seems to be providing a "user" password which is not actually needed, since anybody can open this PDF, but most importantly the "owner" password is definitely not `Y/l`.
furthermore, there seems to be a non-ASCII character in the reconstructed user password which, honestly, I wouldn't even know how to type into Acrobat Professional for modification.

other "cracked" passwords that this workflow produced, for the same hash, are:
```
user@host:~$ echo 6a5257362020202028757365722070617373776f72643d4f5c4473277e7a3d363b144f375a4a7929 | xxd -r -p
jRW6    (user password=O\Ds'~z=6;O7ZJy)
```
and again, the "owner" password is not `jRW6`.

so: how come hashcat thinks it is guessing the password?
how come by removing the pot file and retrying I am getting different results?
which mode is the appropriate mode for this type of solution?
I have tried using Acrobat Professional to create a brand new PDF file with no user password but with an owner password, e.g. 123456, and hashcat failed to crack it using attack mode 10500 yet successfully cracked it using attack mode 25400.

any suggestions are appreciated.

thatux:
Should fix https://github.com/hashcat/hashcat/issues/3181 and https://github.com/hashcat/hashcat/pull/3628#issuecomment-1458517407

I'd love to add hc_get_entropy() to m26610 as suggested per
https://github.com/hashcat/hashcat/pull/3628#issuecomment-1447869282
But I'm not sure how to make sure what MAX_ENTROPY should be. Some small test suggest that incorrect password guesses have a plaintext with entropy ranging from 4.578 to 5. But how to be sure?
```
  const float entropy = hc_get_entropy (pt, 8);
  printf("entropy=%f\n", entropy);
```

### DIFF
--- a/OpenCL/inc_cipher_aes-gcm.cl
+++ b/OpenCL/inc_cipher_aes-gcm.cl
@@ -303,3 +303,9 @@ DECLSPEC void AES_GCM_GHASH_GLOBAL (PRIVATE_AS const u32 *subkey, PRIVATE_AS con
 
   AES_GCM_ghash (subkey, len_buf, 16, out);
 }
+
+void AES_GCM_decrypt (PRIVATE_AS u32 *key, PRIVATE_AS u32 *J0, PRIVATE_AS const u32 *in, int in_len, PRIVATE_AS u32 *out, SHM_TYPE u32 *s_te0, SHM_TYPE u32 *s_te1, SHM_TYPE u32 *s_te2, SHM_TYPE u32 *s_te3, SHM_TYPE u32 *s_te4)
+{
+    AES_GCM_inc32(J0); // the first ctr is used to compute the tag, only the second is used for decryption: https://en.wikipedia.org/wiki/Galois/Counter_Mode#/media/File:GCM-Galois_Counter_Mode_with_IV.svg
+    AES_GCM_GCTR (key, J0, in, in_len, out, s_te0, s_te1, s_te2, s_te3, s_te4); // decrypt the first block of ciphertext
+}
--- a/OpenCL/inc_cipher_aes-gcm.h
+++ b/OpenCL/inc_cipher_aes-gcm.h
@@ -17,5 +17,6 @@ DECLSPEC void AES_GCM_gctr (PRIVATE_AS const u32 *key, PRIVATE_AS const u32 *iv,
 DECLSPEC void AES_GCM_GCTR (PRIVATE_AS u32 *key, PRIVATE_AS u32 *J0, PRIVATE_AS const u32 *in, int in_len, PRIVATE_AS u32 *out, SHM_TYPE u32 *s_te0, SHM_TYPE u32 *s_te1, SHM_TYPE u32 *s_te2, SHM_TYPE u32 *s_te3, SHM_TYPE u32 *s_te4);
 DECLSPEC void AES_GCM_GHASH (PRIVATE_AS const u32 *subkey, PRIVATE_AS const u32 *aad_buf, int aad_len, PRIVATE_AS const u32 *enc_buf, int enc_len, PRIVATE_AS u32 *out);
 DECLSPEC void AES_GCM_GHASH_GLOBAL (PRIVATE_AS const u32 *subkey, PRIVATE_AS const u32 *aad_buf, int aad_len, GLOBAL_AS const u32 *enc_buf, int enc_len, PRIVATE_AS u32 *out);
+DECLSPEC void AES_GCM_decrypt (PRIVATE_AS u32 *key, PRIVATE_AS u32 *J0, PRIVATE_AS const u32 *in, int in_len, PRIVATE_AS u32 *out, SHM_TYPE u32 *s_te0, SHM_TYPE u32 *s_te1, SHM_TYPE u32 *s_te2, SHM_TYPE u32 *s_te3, SHM_TYPE u32 *s_te4);
 
 #endif // INC_CIPHER_AES_GCM_H
--- a/OpenCL/m25400-pure.cl
+++ b/OpenCL/m25400-pure.cl
@@ -3,7 +3,7 @@
  * License.....: MIT
  */
 
-// https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf
+// https://web.archive.org/web/20220306152229/https://www.adobe.com/content/dam/acom/en/devnet/pdf/pdfs/pdf_reference_archives/PDFReference.pdf
 
 #ifdef KERNEL_STATIC
 #include M2S(INCLUDE_PATH/inc_vendor.h)
@@ -43,7 +43,7 @@ typedef struct pdf
 typedef struct pdf14_tmp
 {
   u32 digest[4];
-  u32 out[4];
+  u32 out[8];
 
 } pdf14_tmp_t;
 
@@ -183,6 +183,10 @@ KERNEL_FQ void m25400_init (KERN_ATTR_TMPS_ESALT (pdf14_tmp_t, pdf_t))
   tmps[gid].out[1] = rc4data[1];
   tmps[gid].out[2] = 0;
   tmps[gid].out[3] = 0;
+  tmps[gid].out[4] = 0; // we only need the size of out for the plaintext check
+  tmps[gid].out[5] = 0; // we only need the size of out for the plaintext check
+  tmps[gid].out[6] = 0; // we only need the size of out for the plaintext check
+  tmps[gid].out[7] = 0; // we only need the size of out for the plaintext check
 }
 
 KERNEL_FQ void m25400_loop (KERN_ATTR_TMPS_ESALT (pdf14_tmp_t, pdf_t))
@@ -211,12 +215,6 @@ KERNEL_FQ void m25400_loop (KERN_ATTR_TMPS_ESALT (pdf14_tmp_t, pdf_t))
   digest[1] = tmps[gid].digest[1];
   digest[2] = tmps[gid].digest[2];
   digest[3] = tmps[gid].digest[3];
-
-  u32 out[4];
-  out[0] = tmps[gid].out[0];
-  out[1] = tmps[gid].out[1];
-  out[2] = tmps[gid].out[2];
-  out[3] = tmps[gid].out[3];
 
   for (u32 i = 0, j = LOOP_POS; i < LOOP_CNT; i++, j++)
   {
@@ -255,10 +253,19 @@ KERNEL_FQ void m25400_loop (KERN_ATTR_TMPS_ESALT (pdf14_tmp_t, pdf_t))
     }
   }
 
+
+  u32 out[4];
   out[0] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[0]; // store original o-value in out (scratchpad)
   out[1] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[1];
   out[2] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[2];
   out[3] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[3];
+
+  u32 out2[4];
+  out2[0] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[4]; // store original o-value in out (scratchpad)
+  out2[1] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[5];
+  out2[2] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[6];
+  out2[3] = esalt_bufs[DIGESTS_OFFSET_HOST].o_buf[7];
+
   u32 o_rc4_decryption_key[4];
   o_rc4_decryption_key[0] = digest[0]; // store the owner-key
   o_rc4_decryption_key[1] = digest[1];
@@ -268,6 +275,7 @@ KERNEL_FQ void m25400_loop (KERN_ATTR_TMPS_ESALT (pdf14_tmp_t, pdf_t))
   // we decrypt the o-value to obtain either the owner-password (or user-password if no owner-password is set)
   //  see: "Algorithm 3.3 Computing the encryption dictionary’s O (owner password) value": "If there is no owner password, use the user password instead".
   u32 tmp[4];
+  u8 j;
   for (u32 i = 19; i>0; i--)
   {
     // xor the iterator into the rc4 key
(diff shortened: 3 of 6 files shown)


thatux:
hmm just now I see https://github.com/hashcat/hashcat/commit/d5a74b2536a32e19cd9e31941c5f735caa57a19c will try and make this mergeable in the future, shouldn't be too difficult

thatux:
merge conflict solved, now using the new `is_valid_printable()` and increased the plaintext check of both kernels to 32 bytes to prevent false positives.

perl tests are ok, should be fine to merge

jsteube:
Looks good. I'll see if entropy check is actually useful here and let you know

jsteube:
Just from gut feeling and a bit more mathematical approach, it should be something like this:

```
  const float entropy = hc_get_entropy (pt, 8);

  if (entropy > (8 / (256 / (0x7e - 0x20)))) return;
```

But this requires testing

jsteube:
I've did some more tests and it works as expected: https://github.com/hashcat/hashcat/commit/37923e79e95b83b9e6a835b552d221e1cf7dcb08

Feel free to run your own tests. We can always revert this change if needed.

thatux:
could you expand a bit on your reasoning? why choose this limit?
`8 / (256 / (0x7e - 0x20)) = 2,9375`

This'd show it doesn't ok all printables
```
#!/usr/bin/env python3

from base64 import b64decode, b64encode
from hashlib import pbkdf2_hmac
import random
import string

from Crypto.Cipher import AES 

vault = {"data":"R95fzGt4UQ0uwrcrVYnIi4UcSlWn9wlmer+//526ZDwYAp50K82F1u1oacYcdjjhuEvbZnWk/uBG00UkgLLlO3WbINljqmu2QWdDEwjTgo/qWR6MU9d/82rxNiONHQE8UrZ8SV+htVr6XIB0ze3aCV0E+fwI93EeP79ZeDxuOEhuHoiYT0bHWMv5nA48AdluG4DbOo7SrDAWBVCBsEdXsOfYsS3/TIh0a/iFCMX4uhxY2824JwcWp4H36SFWyBYMZCJ3/U4DYFbbjWZtGRthoJlIik5BJq4FLu3Y1jEgza0AWlAvu4MKTEqrYSpUIghfxf1a1f+kPvxsHNq0as0kRwCXu09DObbdsiggbmeoBkxMZiFq0d9ar/3Gon0r3hfc3c124Wlivzbzu1JcZ3wURhLSsUS7b5cfG86aXHJkxmQDA5urBz6lw3bsIvlEUB2ErkQy/zD+cPwCG1Rs/WKt7KNh45lppCUkHccbf+xlpdc8OfUwj01Xp7BdH8LMR7Vx1C4hZCvSdtURVl0VaAMxHDX0MjRkwmqS","iv":"h+BoIf2CQ5BEjaIOShFE7g==","salt":"jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s="}
password = "hashcat1"

salt = b64decode(vault["salt"])
key = pbkdf2_hmac("sha256", password.encode(), salt, 10_000)

iv = b64decode(vault["iv"])
for i in range(1, 10):
    s = ''.join(random.choice(string.digits + string.ascii_letters + string.punctuation) for _ in range(32))
    print(f"in={s}")
    sb = s.encode()
    cipher = AES.new(key, AES.MODE_GCM, nonce=iv)
    print('{"data":"'+b64encode(cipher.encrypt(sb)).decode()+'","iv":"h+BoIf2CQ5BEjaIOShFE7g==","salt":"jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s="}') 
```

```
python3 generate_random.py > tools/2hashcat_tests/metamask_random_vaults.json
cat metamask_random_vaults.json | grep -v 'in='| while read l; do echo $l > /tmp/l; ../metamask2hashcat.py --vault /tmp/l --shortdata ; done > metamask_random_vaults.hashes
```

with
```
diff --git a/OpenCL/m26610-pure.cl b/OpenCL/m26610-pure.cl
index 4bae5c1d0..92c878f53 100644
--- a/OpenCL/m26610-pure.cl
+++ b/OpenCL/m26610-pure.cl
@@ -367,6 +367,8 @@ KERNEL_FQ void m26610_comp (KERN_ATTR_TMPS_ESALT (pbkdf2_sha256_tmp_t, pbkdf2_sh
   // if ((gid == 0) && (lid == 0)) printf ("pt[0]=%08x%08x\n", pt[4], pt[5]); // should be 2054726565222c22 or  Tree","

   const float entropy = hc_get_entropy (pt, 8);
+  printf("\n\nentropy=%9.6f pt[4]pt[5]=%08x%08x\n\n\n",entropy, pt[4], pt[5]);
+  printf("\n\nMIN_entropy=%9.6f\n\n\n", float (8 / (256 / (0x7e - 0x20))));

   if (entropy > (8 / (256 / (0x7e - 0x20)))) return;
```
and
```
./hashcat.exe --quiet -d1 -a3 -m26610 '$metamask-short$jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s=$h+BoIf2CQ5BEjaIOShFE7g==$eNcE/X5oVh59hagvWKndov0WSHC6lxk4NZ+P8Nj+Pik=' 'hashcat1'
```
I get
```
entropy= 4.875000 pt[4]pt[5]=585e704078423c7c




MIN_entropy= 4.000000
```

Am I doing something wrong? MIN_entropy should be 2.9375.

And the hash its data decrypts to ```dryEl`b1ieW@-kpPX^p@xB<|+ADng~!7``` which would be all printables

jsteube:
It was a clever Idea to do the multi hashes to improve the rate to check for false results. 

I had to adapt the kernel and the module slightly to allow cracking them all and in that process I made the kernel a bit easier to read.

I've also removed the entropy check because my logic was flawed and it didn't work as I expected it to work. However, while I was in the process of reverting it, I realized that we actually have an even more nice way to check if a string is a valid UTF8 string and that is by using our UTF8 to UTF16 conversion library.

```
  const int pt_len = 28; // not using 32 byte but 28 because our UTF8 allows up to 4 byte per character and since we decrypt 32 byte
                         // only we can't garantee it is not in the middle of a UTF8 byte stream at that point

  if (hc_enc_scan (pt, pt_len))
  {
    hc_enc_t hc_enc;

    hc_enc_init (&hc_enc);

    while (hc_enc_has_next (&hc_enc, pt_len))
    {
      u32 enc_buf[16] = { 0 };

      const int enc_len = hc_enc_next (&hc_enc, pt, pt_len, 32, enc_buf, sizeof (enc_buf));

      if (enc_len == -1) return;
    }
  }
```

So that means this new method will also cover the corner case if the user is using high bits for user-defined fields (as long as they are valid UTF8 - but that's what metamask will take care of). But the problem is that the false-positive rate is a lot higher, because there's now a much larger allowed range of byte streams that count as "valid" than with only checking for 0x20 - 0xff.

So the solution still is not perfect. Maybe we can use a combination of UTF8 check + some well known substring that will appear somewhere in the first 32 byte?

thatux:
"I had to adapt the kernel and the module slightly to allow cracking them all and in that process I made the kernel a bit easier to read." for future reference https://github.com/hashcat/hashcat/commit/24d375073cc45f9f9b076719cb1136a0375711e5 implementing DEEP_COMP

Thanks, nice idea on using the UTF8 to UTF16 library!
Do I understand correctly that hc_enc_scan() checks if every eighth bit of every byte is set (as required by UTF8?)? and that in your attempt you try to see whether all plaintext is UTF8 using hc_enc_scan()?

A problem with that would be that UTF8 and ASCII can be used interchangeably e.g.
` $metamask-short$jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s=$h+BoIf2CQ5BEjaIOShFE7g==$7DrvEXpHXXlqjdcSTYefuPYvRB+CshJ1K/30t5mhf2kqJa83` 
would decrypt to 
```💩hOiV~m(}8E2JSg|/@g715#?)&!`w_IT.``` 
which in hex is 
`f09f92a9684f69567e6d287d3845324a53677c2f4067373135233f29262160775f49542e`
so interchangeable have first four bytes eighth bit set and the remaining bytes not.

generated using
```
#!/usr/bin/env python3

from base64 import b64decode, b64encode
from hashlib import pbkdf2_hmac
import random
import string

from Crypto.Cipher import AES 

vault = {"data":"R95fzGt4UQ0uwrcrVYnIi4UcSlWn9wlmer+//526ZDwYAp50K82F1u1oacYcdjjhuEvbZnWk/uBG00UkgLLlO3WbINljqmu2QWdDEwjTgo/qWR6MU9d/82rxNiONHQE8UrZ8SV+htVr6XIB0ze3aCV0E+fwI93EeP79ZeDxuOEhuHoiYT0bHWMv5nA48AdluG4DbOo7SrDAWBVCBsEdXsOfYsS3/TIh0a/iFCMX4uhxY2824JwcWp4H36SFWyBYMZCJ3/U4DYFbbjWZtGRthoJlIik5BJq4FLu3Y1jEgza0AWlAvu4MKTEqrYSpUIghfxf1a1f+kPvxsHNq0as0kRwCXu09DObbdsiggbmeoBkxMZiFq0d9ar/3Gon0r3hfc3c124Wlivzbzu1JcZ3wURhLSsUS7b5cfG86aXHJkxmQDA5urBz6lw3bsIvlEUB2ErkQy/zD+cPwCG1Rs/WKt7KNh45lppCUkHccbf+xlpdc8OfUwj01Xp7BdH8LMR7Vx1C4hZCvSdtURVl0VaAMxHDX0MjRkwmqS","iv":"h+BoIf2CQ5BEjaIOShFE7g==","salt":"jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s="}
password = "hashcat1"

salt = b64decode(vault["salt"])
key = pbkdf2_hmac("sha256", password.encode(), salt, 10_000)
iv = b64decode(vault["iv"])

for i in range(1, 10):
    s = "💩" + ''.join(random.choice(string.digits + string.ascii_letters + string.punctuation) for _ in range(32))
    print(f"in={s}")
    sb = s.encode()
    cipher = AES.new(key, AES.MODE_GCM, nonce=iv)
    print('{"data":"'+b64encode(cipher.encrypt(sb)).decode()+'","iv":"h+BoIf2CQ5BEjaIOShFE7g==","salt":"jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s="}') 
```

**Comment by thatux** [comment]
Some more considerations on the plaintext check. I really like the general idea of not relying on the plaintext check of ASCII as UTF8 could be used in the plaintext. But it might not be necessary in this specific case. The data we decrypt is: `[{"type":"HD Key Tree","data":{"` and I'm pretty sure no part of that is user input. so in this case 1 below. may be best. I'd opt not to check for this specific string, but I'd be surprised if the metamask devs add a 💩 wallet type. For now it seems to have these types: 
```
      "Simple Key Pair",
      "HD Key Tree",
      "Trezor Hardware",
      "Ledger Hardware",
      "Lattice Hardware"
```
https://github.com/MetaMask/metamask-extension/blob/develop/development/states/navigate-txs.json#L256
'keyringTypes - An array of unique keyring identifying strings, representing available strategies for creating accounts.'

Some strategies to deal with this
1. assume plaintext doesn't contain user input thus only ASCII, and that the structure won't change (IMHO the best option in this case, and this is what we're doing now). we could add a check for the first two character `[{` as I'm pretty sure that won't change.
2. lower the limit of bytes that need to be ASCII i.e. instead of checking 32 bytes check 28 bytes, would allow for a single UTF8 character (which a easy fix, but a bit stupid because it'd break on two UTF8 characters)
3. decrypt even more data (64 bytes?) and implement a check for utf8, this would be a bit slower, but more robust: no false negative on UTF8 characters, and no false positive due to accepting decrypted garbage e.g. we could implement something like 
 https://stackoverflow.com/questions/1031645/how-to-detect-utf-8-in-plain-c in OpenCL. However I'm not sure what the minimum amount of data we're provided is in this case, we need to be sure the encrypted data has a minimum of 64 bytes.

The last would be best in the general case, I'll make a issue for it if you agree.

magnumripper:
You already have functions such as `printable_utf8()` in convert.c or `hc_enc_validate_utf8()` in inc_common.cl. They're pretty fast and surprisingly effective: The probability of a random string of bytes ending up as valid is 3.9% for a two-byte sequence, and decreases ***exponentially*** for longer sequences. Oh, and pure ASCII is obviously valid too.

jsteube:
> Thanks, nice idea on using the UTF8 to UTF16 library! Do I understand correctly that hc_enc_scan() checks if every eighth bit of every byte is set (as required by UTF8?)? and that in your attempt you try to see whether all plaintext is UTF8 using hc_enc_scan()?

No, that's not the case. It checks every byte, but not to check if every of them make use of a high but, but any of them. In that case, it will start interpreting UTF8.

> A problem with that would be that UTF8 and ASCII can be used interchangeably e.g. ` $metamask-short$jfGI3TXguhb8GPnKSXFrMzRk2NCEc131Gt5G3kZr5+s=$h+BoIf2CQ5BEjaIOShFE7g==$7DrvEXpHXXlqjdcSTYefuPYvRB+CshJ1K/30t5mhf2kqJa83` would decrypt to `` 💩hOiV~m(}8E2JSg|/@g715#?)&!`w_IT. `` which in hex is `f09f92a9684f69567e6d287d3845324a53677c2f4067373135233f29262160775f49542e` so interchangeable have first four bytes eighth bit set and the remaining bytes not.

That's supported, no worries.

jsteube:
> Some strategies to deal with this
> 
> 1. assume plaintext doesn't contain user input thus only ASCII, and that the structure won't change (IMHO the best option in this case, and this is what we're doing now). we could add a check for the first two character `[{` as I'm pretty sure that won't change.

Agree, if the structure doesn't contain any user-defined strings I think we are fine with the current implementation

Issue #3635, opened 2023-03-07
Add status message when loading rules


Shooter3k:
When loading a large amount of rules (like 30million), Hashcat says "Generated bitmap tables" for the 5 or whatever minutes it's loading the rules. If the rules have errors and it shows those errors, you can sort of follow along because you can see line numbers from the errors. But, if there is no errors, it appears Hashcat has hung. 

Can a message be added at the start of loading the rules that says something along the lines of "Loading rules. Please be patient..." 

In an ideal state, a message would be added at the beginning of all the loading steps, similar to what it currently does already like: "Sorting hashes. Please be patient..." 
"Generated bitmap tables" 
"Parsing Hashes" 
"Removing duplicate hashes" 
"Generating bitmap tables..."

Perhaps the only one that's currently missing is "Loading rules"  ??

PR #3638, opened 2023-03-08
Prompt when loading rules

PenguinKeeper7:
See #3635


jsteube:
I think we also use "..." to indicate whenever the work is in front

PR #3645, opened 2023-03-15
SecureCRT master passphrase v2 module

YSaxon:
This module cracks the master password in SecureCRT config files from versions < 9.3

### DIFF
--- a/OpenCL/m31400_a0-pure.cl
+++ b/OpenCL/m31400_a0-pure.cl
@@ -121,10 +121,10 @@ DECLSPEC void aes256_scrt_format (PRIVATE_AS u32 *aes_ks, PRIVATE_AS u32 *pw, co
 
         shift_buffer_by_offset(hash,pw_len+4);
 
-        hash[0] = hc_swap32(pw_len);
-        hash[1] |= hc_swap32(pw[0]);
-        hash[2] |= hc_swap32(pw[1]);
-        hash[3] |= hc_swap32(pw[2]);
+        hash[0] = hc_swap32_S(pw_len);
+        hash[1] |= hc_swap32_S(pw[0]);
+        hash[2] |= hc_swap32_S(pw[1]);
+        hash[3] |= hc_swap32_S(pw[2]);
         
         AES256_encrypt (aes_ks, hash, out, s_te0, s_te1, s_te2, s_te3, s_te4);
 }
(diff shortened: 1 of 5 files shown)


jsteube:
Very cool, many thanks for contribution! 

There's a big issue with the vector datatypes (only relevant for CPU performance). We get the following compiler error:

```
1:137:33: error: member reference base type '__private u32x *__private' (aka '__private uint8 *__private') is not a structure or union
    aes256_scrt_format(aes_ks, w.s0, pw_len.s0, hash.s0, out.s0, s_te0, s_te1, s_te2, s_te3, s_te4);
```

CUDA/OpenCL don't allow us to access an array globally using the .sX identifier. It's not really a problem, all we need to do is to rewrite the code to something like this:

```
    u32 tmp_w[4];

    tmp_w[0] = w[0].s0;
    tmp_w[1] = w[1].s0;
    tmp_w[2] = w[2].s0;
    tmp_w[3] = w[3].s0;

    u32 tmp_pw_len = pw_len.s0;

    u32 tmp_out[4];

    aes256_scrt_format(aes_ks, tmp_w, tmp_pw_len, hash.s0, tmp_out, s_te0, s_te1, s_te2, s_te3, s_te4);

    out[0].s0 = tmp_out[0];
    out[1].s0 = tmp_out[1];
    out[2].s0 = tmp_out[2];
    out[3].s0 = tmp_out[3];
```

Then just repeat the same for all the elements, like .s1 etc. No worries, this will not cause any performance loss compared, since the compiler would have to do it that or the other way. It's just that we have to do this by hand.

You can reproduce locally if you install an OpenCL driver for your CPU. Ideally the Intel one, but I think pocl can do, too. Then use the following flags in hashcat:

```
 -D1 --backend-vector-width 8 
```

YSaxon:
@jsteube 
I'm having a lot of trouble installing the Intel OpenCL module (or at least, getting it recognized by hashcat), and the pocl one gives me Kernel Self-Test problems on -a 3, but I think I've fixed all the compile errors. If you wouldn't mind giving it a try and letting me know if it all works.

jsteube:
No worries, I'll fix that after merge later. Just FYI, the problem is that w[] is 64 not 4 and hash[] is 8 not 4. Other than that your update is fine.

jsteube:
Fixed with https://github.com/hashcat/hashcat/commit/074411418fa43e8f69a003419e5d84c9d8dcdd93

PR #3655, opened 2023-03-24
Add error message when mixing --show and --username

Chick3nman:
Mixing --show and --username can cause an exponential delay in output, added an error message to warn users who may not wish to wait. Message is log level ERROR to avoid polluting files that the user wishes to send the cracks to.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -55,6 +55,7 @@
 - Status Code: Add specific return code for self-test fail (-11)
 - SCRYPT: Increase buffer sizes in module for hash mode 8900 to allow longer SCRYPT digests
 - Unicode: Update UTF8 to UTF16 conversion to match RFC 3629
+- User Options: Added error message when mixing --username and --show to warn users of exponential delay 
 
 * changes v6.2.5 -> v6.2.6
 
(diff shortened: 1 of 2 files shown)


jsteube:
We might want to allow using --force?

Chick3nman:
> We might want to allow using --force?

That's a good point, I'll verify behavior before merge and make sure it can be bypassed. I'll ping you when done.

Chick3nman:
```
$ ./hashcat.exe -a0 -m 2100 --show --username example.hash
Mixing --show with --username can cause exponential delay in output.

tom:$DCC2$10240#6848#e2829c8af2232fa53797e2f0e35e4626:hashcat
```

Current behavior is to continue to operate and show the error on stderr, but not block the function. No need for --force and the use of stderr ensures that the message doesnt pollute the stdout/hashes. Only a visual difference to the user, not operational.

mohemiv:
@Chick3nman Could you please show in what cases mixing --show and --username can cause an exponential delay in output?

I cannot see any significant delay, even if the hashlist has more than 1 million hashes.

Also related: https://github.com/hashcat/hashcat/issues/3988

PR #3659, opened 2023-03-25
fixes #3653 test.sh parsing hashes with spaces

### LINKED ISSUE #3653: test.sh splitting
Something is going wrong in test.sh for hashes which include space characters.

To reproduce:

```
tools/test.sh -t multi -a 0 -m 31400 -V1
```

```
[ test_1679595827 ] > Init test for hash type 31400.
! unhandled return code, see test_1679595827/logfull.txt or test_1679595827/test_report.log for details.
```

The error log says for all hashes in the hashfile:

```
Hashfile 'test_1679595827/31400_hashes.txt' on line 1 (S:"Config): Token length exception
...
```

However, the test module and test.pl created them correctly:

```
$ head -1 test_1679595827/all.sh
echo 316                             | ./hashcat ${OPTS} -a 0 -m 31400 'S:"Config Passphrase"=02:f051ef12b9b0da6aeb1830344c619e3c86b9e66b49a0d9c1380a2e6530eb25a51f60392fb850226992f7d87670262210'
```

I guess it's just some splitting that went wrong.

@philsmd @matrix Could you have a look at the script please?

philsmd:
This seems to fix the problem reported in #3653 where a new problem with `test.sh` was discovered when hashes contained spaces, now we just use every field after the 10th field (with `awk`).

Thanks

### DIFF
--- a/tools/test.sh
+++ b/tools/test.sh
@@ -256,8 +256,8 @@ function init()
   grep " ${hash_type} '" "${OUTD}/all.sh" > "${cmd_file}" 2>/dev/null
 
   # create separate list of password and hashes
-  sed 's/^echo *|.*$//'       "${cmd_file}" | awk '{print $2}'                  > "${OUTD}/${hash_type}_passwords.txt"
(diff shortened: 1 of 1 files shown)


jsteube:
Fantastic, thanks!

PR #3683, opened 2023-04-11
Fixed bug in input_tokenizer when TOKEN_ATTR_FIXED_LENGTH is used and refactor modules

matrix:
Hi,

this has been a bug for quite some time, until now as a workaround the TOKEN_ATTR_VERIFY_LENGTH attribute has been used even where it was not needed.

To understand the problem just look at the modification done on shared.c, in a nutshell sometimes the separator would go into the buffer of the next token so the check that was done right after that would fail.

Since this problem no longer exists, I put the correct attribute back where it was needed ;)

I have currently tested the change only for modules where the self-test feature is not disabled (almost all of them).

Thanks :)


jsteube:
I think it was correct as it was and we do not need this change. 

To understand, check this invalid hash:

```
$ cat hash
3d83c8e717ff0e7ecfe187f088d69954x343141
```

To recognize this, hashcat module is configured like this:

```
  token.sep[0]     = hashconfig->separator;
  token.len_min[0] = 32;
  token.len_max[0] = 32;
  token.attr[0]    = TOKEN_ATTR_VERIFY_LENGTH
                   | TOKEN_ATTR_VERIFY_HEX;
```

If we set fixed length, then there's this extra character which is not checked for correctness.

So we use VERIFY to enable hashcat to check separator (if the separator matches correctly).

Also, using a value min/max equal to the fixed size might look unintuitional, but technically is not incorrect.

Old version complains and does not start:

```
$ ./hashcat -m 10 hash
Hashfile 'hash' on line 1 (3d83c8e717ff0e7ecfe187f088d69954x343141): Separator unmatched
No hashes loaded.
```

After PR, hashcat starts up even if the hash is incorrect:

```
$ cat hash
$ ./hashcat -m 10 hash
.... starting
```

matrix:
Hi :)

please retry, I should have solved it

Issue #3676, opened 2023-04-11
PDF 10500 hashes are not printed identically as they were in the input


s3inlc:
The PDF permission P value can be stored in different forms, some editors save them as signed integer (which makes them negative typically, as in most of the examples) and some save it as unsigned integer.
The issue now is, in some cases when this permission value is stored as unsigned integers and the resulting number is very large, hashcat prints the hash differently than it was given as input.

Example (Note, this is the example hash modified, so it is not crackable):
```
$pdf$2*3*128*4294963392*1*16*da42ee15d4b3e08fe5b9ecea0e02ad0f*32*c9b59d72c7c670c42eeb4fca1d2ca15000000000000000000000000000000000*32*c4ff3e868dc87604626c2b8c259297a14d58c6309c70b00afdfb1fbba10ee571 
```
When this hash is given to hashcat, the issue is already visible in the status output, where it renders the hash as: 
```
$pdf$2*3*128*-3904*1*16*da42ee15d4b3e08fe5b9ece...
```
This permission value is used for flags, so technically it does not really matter if it's printed as signed or unsigned, but for handling hashes it makes it really difficult to identify which hashes belong together, as they're not printed as they were given to hashcat. In order to preserve original structure of hashes, hashcat should print hashes always the same way they were given as input.

Note: This issue most likely also exists with the other PDF modes.

matrix:
fixed with https://github.com/hashcat/hashcat/pull/3684

PR #3684, opened 2023-04-11
Handle signed/unsigned PDF permission P value for all PDF hash-modes

matrix:
As requested in #3676 now PDF modes handle signed/unsigned permission P value.
@s3inlc could you also check that everything is okay now?

Thanks ;)


s3inlc:
Awesome! Thanks for the very quick fix, I checked and it works correctly for me now :)

PR #3703, opened 2023-04-25
Fix 31900 hash parsing error

matrix:
due to missing memset ...

```
$ ./hashcat -m 31900 '$metamaskMobile$NDgzNDYzNjkyMTkwMzY4NQ==$eb4ce9427d0ae89fdcd0f88a0be05309$8P/wQdD3KyOqwphMRUT6fg==' --force
hashcat (v6.2.6-470-gf847593a6+) starting

The device #1 has been disabled as it most likely also exists as an OpenCL device, but it is not possible to automatically map it.
You can use -d 1 to use Metal API instead of OpenCL API. In some rare cases this is more stable.

METAL API (Metal 263.8)
=======================
* Device #1: Apple M1, skipped

OpenCL API (OpenCL 1.2 (Aug  8 2022 21:29:55)) - Platform #1 [Apple]
* Device #2: Apple M1, GPU, 960/10922 MB (1024 MB allocatable), 8MCU

Minimum password length supported by kernel: 8

Hash '$metamaskMobile$NDgzNDYzNjkyMTkwMzY4NQ==$eb4ce9427d0ae89fdcd0f88a0be05309$8P/wQdD3KyOqwphMRUT6fg==': Separator unmatched
No hashes loaded.
```

### DIFF
--- a/src/modules/module_31900.c
+++ b/src/modules/module_31900.c
@@ -131,6 +131,8 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   hc_token_t token;
 
+  memset (&token, 0, sizeof (hc_token_t));
+
   token.token_cnt  = 4;
 
   token.signatures_cnt    = 1;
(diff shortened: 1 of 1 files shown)


jsteube:
Wondering if there's a way to do without memset(). I'm really concerned about private plugins that are not on this repository.

matrix:
Let's say if you just need an extra memset it's easy, but if we go around it gets complicated.

PR #3707, opened 2023-04-27
JSON-escape example_hash in hash info

davidbolvansky:
JSON output was broken after https://github.com/hashcat/hashcat/commit/6b28169c66dadbd884f5fc4a4dc62d8b734e02a1#diff-d734c437450f632dd416ce9001da3d82a9597b16c2fd20333c41d03e986de0a8


davidbolvansky:
@jsteube we need this fix for Fitcrack, as Fitcrack uses this JSON file, so it needs to be valid.

PR #3720, opened 2023-05-09
Added new changed info about 2500 and 22000

mj789-98:
Added new changed info about 2500 and 22000 wpa2 attacks

### DIFF
--- a/docs/limits.txt
+++ b/docs/limits.txt
@@ -85,5 +85,9 @@ This is a trade-off game.
 ## Position identifiers in rules are limited to 36
 ##
 
+##
+## When cracking wpa2.hccapx handshake file use of  "--deprecated-check-disable" is required in latest versions
+##
+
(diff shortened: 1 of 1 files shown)


Chick3nman:
I'm not sure we should be encouraging users to switch on `--deprecated-check-disable` as a part of the limitations. Instead, we should be encouraging users not to use the .hccapx format at all since it's deprecated and to instead produce the newer 22000 formatted hashes.

mj789-98:
> I'm not sure we should be encouraging users to switch on `--deprecated-check-disable` as a part of the limitations. Instead, we should be encouraging users not to use the .hccapx format at all since it's deprecated and to instead produce the newer 22000 formatted hashes.

mj789-98:
Okay, I get what you are saying  but most tutorials online teach new users to use 2500, till they understand hashcat and get their way around it, so it is great to know how this will work.

PR #3728, opened 2023-05-15
check if selected hash-mode cannot crack multiple hashes with the same salt only if salt_type is SALT_TYPE_GENERIC or SALT_TYPE_EMBEDDED

matrix:
Hi,

the title is also the description of the change ;)

Thanks


matrix:
fixed on master

Issue #2603, opened 2020-11-01
Hashtype Kerberos AS-REP with etype 17/18


rumiljonov:
Currently hashcat only supports etype 23 (NTLM) for AS-REP, but in case of Linux krb5kdc hash will be encrypted with AES-128 (etype 17) or AES-256 (etype 18).

Here is an example hash from my local krb5kdc test server
`$krb5asrep$18$roastable@rumiljonov.local:2a45894d63625b4cabc91a8fb8d7f2b7$118dec18ad6d7cdb3b46cd6234d6902b0ea4ba8278460a4b60aea8d4d2e8d24e9747bcf0277574604292bb0dcfcb81f975fe6587301503320e865c77f3d9617c5daf59a5a3c961ee0c38934fce6f88861c91a02850a2c787db89dea5b00b2a5d51d3f533cd9c13795d9542a46d6322e8cad8cad75c98e79c1589cc4b11dda84ce8a929178d384e7e0a8cddc93a8ef0b7c1ccb1b89c3c6712a4c6f7e9def82439c0440f5cfd1618239a3c9888b6e3a0b664a2fd46046e36859b83756b3c6ace4b456228b1496433fa6ab8c5ca8b0e8fed6028f6a222580ec4507e29803ef8fe322f0e`

password is `password`

lapolis:
Hi everyone, is this algorithm being implemented any time soon? or there is a long list ahead?

PR #3729, opened 2023-05-15
Added plugins for AES128 and AES256 AS-REPs

erasmusc:
Hi, I've added two new hash formats for Kerberos AS-REPs with AES encryption (addressing #2603):
* Hash-mode 33100: Kerberos 5, etype 17 (AES128), AS-REP
* Hash-mode 33200: Kerberos 5, etype 18 (AES256), AS-REP

I used hash-mode numbers that were a bit higher than the current highest ones as per the plugin development guide, so I'm assuming I'll need to change them once some official hash-mode numbers are reserved.

The plugins are basically just modified versions of the existing TGS-REP AES plugins (19600 and 19700) and some code was also used from the AS-REP RC4 plugin (18200), so most of the credit should still go to the authors of those.

I only implemented pure kernels for them, in line with what's been done for the other Kerberos AES plugins.

## Hash Format

I decided to make the hash format similar to the other Kerberos hash formats, e.g. 19600 and 19700.

Hash format for AS-REP type 18 (type 17 is obviously very similar):
```
$krb5asrep$18$user$realm$checksum$edata2
```

But I also added support for John the Ripper's hash format for ease of use, since its AS-REP plugin already supports AES encryption types. For comparison, John's format for type 18 is:
```
$krb5asrep$18$salt$edata2$checksum
```

## Example Hashes

Here are some example hashes, both have the password "hashcat".

Example AS-REP type 17 hash:
```
$krb5asrep$17$user$example.com$a419c4030e555734b06c2629$c09a1421f96eb126c757a4b87830381f142477d9a85b2beb3093dbfd44f38ddb6016a479537fb7b36e046315869fe79187217971ff6a12c1e0a2df3f68045e03814b21f756d8981f781803d65e8572823c88979581d93cf7d768f2efced16f3719b8d1004d9e73d798de255383476bced47d1982f16be77d0feb55a1f44f58bd013fa4caee58ac614caf0f1cf9101ec9623c5b8c2a1491b73f134f074790088fdb360b5ebce0d32a8145ed00a81ddf77188e150b92d8e8ddd0285d27f1514253e5546e6bba864b362bb1e6483b26d08fa4cc268bfbefe0f690039bcc524b774599df3680c1c3431d891bfa99514a877f964e
```

Example AS-REP type 18 hash:
```
$krb5asrep$18$user$example.com$aa4c494f520b27873a4de8f7$ebc9976a77f62e8ccca02d43d68bafcc66a81fcbb44a336b00ce401982f32975a5f9bcdc752643252185866685b0a30aaf50e449e392a5994e6979f23aba25f7704c90b2efa03b703c3c2f9e3617cc588ed226d0417e7742d45407878fd946d046b4a9732b9a203cb857811714b009c195b7c96b9bccb7e48832b11a4e92ecf24c49e54de8d0d5d5351445b5126db90bb7eebc7861db1e61de1175824b0a45023a6fa06c2a9d3035fdcf863bea922648e3dc28b48e39b1dec0869e7fe4de399cb52dfcf2596599da54a4bb0169c72d9496de2e137a4594e0e8a69082fc558ac9ace65d32eae5e260a65ca3f2f5871aaeee7a3b090b50f39321d120c144421e0abe7d
```

And the same Hashes in John's format if that's necessary for testing:

Example AS-REP type 17 hash (John format):
```
$krb5asrep$17$EXAMPLE.COMuser$c09a1421f96eb126c757a4b87830381f142477d9a85b2beb3093dbfd44f38ddb6016a479537fb7b36e046315869fe79187217971ff6a12c1e0a2df3f68045e03814b21f756d8981f781803d65e8572823c88979581d93cf7d768f2efced16f3719b8d1004d9e73d798de255383476bced47d1982f16be77d0feb55a1f44f58bd013fa4caee58ac614caf0f1cf9101ec9623c5b8c2a1491b73f134f074790088fdb360b5ebce0d32a8145ed00a81ddf77188e150b92d8e8ddd0285d27f1514253e5546e6bba864b362bb1e6483b26d08fa4cc268bfbefe0f690039bcc524b774599df3680c1c3431d891bfa99514a877f964e$a419c4030e555734b06c2629
```

Example AS-REP type 18 hash (John format):
```
$krb5asrep$18$EXAMPLE.COMuser$ebc9976a77f62e8ccca02d43d68bafcc66a81fcbb44a336b00ce401982f32975a5f9bcdc752643252185866685b0a30aaf50e449e392a5994e6979f23aba25f7704c90b2efa03b703c3c2f9e3617cc588ed226d0417e7742d45407878fd946d046b4a9732b9a203cb857811714b009c195b7c96b9bccb7e48832b11a4e92ecf24c49e54de8d0d5d5351445b5126db90bb7eebc7861db1e61de1175824b0a45023a6fa06c2a9d3035fdcf863bea922648e3dc28b48e39b1dec0869e7fe4de399cb52dfcf2596599da54a4bb0169c72d9496de2e137a4594e0e8a69082fc558ac9ace65d32eae5e260a65ca3f2f5871aaeee7a3b090b50f39321d120c144421e0abe7d$aa4c494f520b27873a4de8f7
```

## Testing and Benchmarks

Everything seems to work fine according to the testing I've done so far, and the benchmark results were basically the same as the TGS-REP AES plugins on a cracking rig with 3 RTX3080's. I'm hoping there are no issues given how similar these are to the TGS-REP plugins, but please let me know if I need to fix anything.

## Additions to Existing Checks for Valid ASN.1 Data

The kernel code for the AS-REP RC4 plugin has a check for valid ASN.1 structs in the first part of the decrypted data, which I used in my kernel code as well since I'd be looking for the same data. But I realised there was something that could be added to those checks, based on one of the test hashes from John the Ripper's AS-REP module (the `luser-18-12345678.pcap` hash in https://github.com/openwall/john/blob/bleeding-jumbo/src/krb5_asrep_fmt_plug.c, for reference).

The code checks if the first byte is `0x79`, which corresponds to an "APPLICATION 25" ASN.1 struct. The first byte from that example hash from John was `0x7a`, even though it looked like the right data from an AS-REP message, and cracked successfully. I found something about this in Section 5.4.2 of RFC4120:

> Compatibility note: Some implementations unconditionally send an encrypted EncTGSRepPart (application tag number 26) in this field regardless of whether the reply is a AS-REP or a TGS-REP. In the interest of compatibility, implementors MAY relax the check on the tag number of the decrypted ENC-PART.

It looks like that's what happened with the test hash from John, since `0x7a` would correspond to "APPLICATION 26" instead of the usual 25. So based on this, I modified the checks that look for `0x79` in the first byte to also match `0x7a`, and then it was able to crack the John test hash successfully.

This was the original if statement from the hash-mode 18200 kernel code:
```
if (((out0[2] & 0x00ff80ff) != 0x00300079) &&
	((out0[2] & 0xFF00FFFF) != 0x30008179) &&
	((out0[2] & 0x0000FFFF) != 0x00008279 || (out0[3] & 0x000000FF) != 0x00000030))
```

And this is the modified one in the kernel code for my plugins:
```
if (((decrypted_block[0] & 0x00ff80ff) == 0x00300079) ||
	((decrypted_block[0] & 0x00ff80ff) == 0x0030007a) ||
	((decrypted_block[0] & 0xFF00FFFF) == 0x30008179) || 
	((decrypted_block[0] & 0xFF00FFFF) == 0x3000817a) ||
	((decrypted_block[0] & 0x0000FFFF) == 0x00008279 && (decrypted_block[1] & 0x000000FF) == 0x00000030) ||
	((decrypted_block[0] & 0x0000FFFF) == 0x0000827a && (decrypted_block[1] & 0x000000FF) == 0x00000030))
```

The main reason I'm mentioning this here is that if this is okay, then we can probably add this modification to the 18200 hash format as well.

### DIFF
--- a/OpenCL/m33100-pure.cl
+++ b/OpenCL/m33100-pure.cl
@@ -1,0 +1,1038 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha1.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_aes.cl)
+#endif
+
+typedef struct krb5asrep_17
+{
+  u32 user[128];
+  u32 domain[128];
+  u32 account_info[512];
+  u32 account_info_len;
+
+  u32 checksum[3];
+  u32 edata2[5120];
+  u32 edata2_len;
+  u32 format;
+
+} krb5asrep_17_t;
+
+typedef struct krb5asrep_17_tmp
+{
+  u32 ipad[5];
+  u32 opad[5];
+  u32 dgst[10];
+  u32 out[10];
+
+} krb5asrep_17_tmp_t;
+
+DECLSPEC void aes128_encrypt_cbc (PRIVATE_AS const u32 *aes_ks, PRIVATE_AS u32 *aes_iv, PRIVATE_AS const u32 *in, PRIVATE_AS u32 *out, SHM_TYPE u32 *s_te0, SHM_TYPE u32 *s_te1, SHM_TYPE u32 *s_te2, SHM_TYPE u32 *s_te3, SHM_TYPE u32 *s_te4)
+{
+  u32 data[4];
+
+  data[0] = hc_swap32_S (in[0]);
+  data[1] = hc_swap32_S (in[1]);
+  data[2] = hc_swap32_S (in[2]);
+  data[3] = hc_swap32_S (in[3]);
+
+  data[0] ^= aes_iv[0];
+  data[1] ^= aes_iv[1];
+  data[2] ^= aes_iv[2];
+  data[3] ^= aes_iv[3];
+
+  aes128_encrypt (aes_ks, data, out, s_te0, s_te1, s_te2, s_te3, s_te4);
+
+  aes_iv[0] = out[0];
+  aes_iv[1] = out[1];
+  aes_iv[2] = out[2];
+  aes_iv[3] = out[3];
+
+  out[0] = hc_swap32_S (out[0]);
+  out[1] = hc_swap32_S (out[1]);
+  out[2] = hc_swap32_S (out[2]);
+  out[3] = hc_swap32_S (out[3]);
+}
+
+DECLSPEC void aes128_decrypt_cbc (PRIVATE_AS const u32 *ks1, PRIVATE_AS const u32 *in, PRIVATE_AS u32 *out, PRIVATE_AS u32 *essiv, SHM_TYPE u32 *s_td0, SHM_TYPE u32 *s_td1, SHM_TYPE u32 *s_td2, SHM_TYPE u32 *s_td3, SHM_TYPE u32 *s_td4)
+{
+  aes128_decrypt (ks1, in, out, s_td0, s_td1, s_td2, s_td3, s_td4);
+
+  out[0] ^= essiv[0];
+  out[1] ^= essiv[1];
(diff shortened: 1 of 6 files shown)


jsteube:
Thank you for your contribution and the comprehensive PR description.

I've done all the checks and it looks fine. All tests were also passed.

FYI, I will rename the hash modes to 32100 and 32200 after merging the PR.

erasmusc:
Great, thanks!

PR #3742, opened 2023-05-23
Add support for Kremlin Encrypt (-m 32700)

hans-vh:
Add support for Kremlin Encrypt 3.0 with NewDES


matrix:
Hi,

the first thing to do is to format the code you wrote in the same way the others were written.

In the module code I would also insert the function "module_jit_build_options" and set "-D SHA1_TRANSFORM_SWAP" as default. Also "token" should always be set to zero (memset) before using it.

Then I would remove all the sha1 implementation you put in .cl and put a couple of ifdef (using SHA1_TRANSFORM_SWAP) inside inc_hash_sha1.cl/sha1_transform to handle the additional input swap. Remember to include inc_hash_sha1.cl in the kernel code at the end.

In general, using "uchar" is not recommended, but the code can be fixed later.

More changes are needed to make it work with Apple Metal, but I can put those in later as well.

If you can write test module in perl that would be great, at least you/we can test the implementation you made.

Thanks ;)

hans-vh:
Thanks for the feedback! I'm sorry for the long delay.

I have moved the SHA1 code to _inc_hash_sha1.cl_ and included that in _m32000-pure.cl_ as you wanted.

Regarding code style: I have used GNU Indent as specified in https://github.com/hashcat/hashcat/blob/master/README.md to attempt to conform to your wishes; Feel free to make changes.

Regarding test module: I have not used Perl, but Python together with the original KremSDK.dll from the Kremlin Windows program during development to verify all steps. After the hashcat module was working I have cracked several passwords with it, e.g. the included self-test hash ($kgb$0ab30cf7a52dad93$82a7c454246fc7570224e9f24279791aa2a63bf4:**hashcat**) and

- $kgb$292bfe0448011ef6$c4f03ab1b3a81590b0dc3882b43403481af2d7ea:**Passord**
- $kgb$6ae944ac850a9605$44ef497e2a0bfea5bf1666b0c049e0dde0aae36e:**Tony**
- $kgb$86fc3e7e31410f36$c17fe58cf41adae07d7cab3bb77975a0645b5e4a:**$HEX[e6d8e5]** (æØå with one byte per Latin character)
- $kgb$2ed936a4394bbc30$11c44ae3d4d39114cc47a6efb5d6bb89cf2be943:**$HEX[626ce562e67273796c746574f879]** (blåbærsyltetøy with one byte per Latin character)

amongst others. All grabbed with tools/kremlin2hashcat.py from KGB files encrypted with the Kremlin Windows program.

I see that module 32000 is taken now. If the code ends up acceptable, can you please assign a new number for me?

jsteube:
Hey @hans-vh please move from hash-mode 32000 to 32600. The perl unit test is not a strict requirement, it's just a good way to ensure the algorithm is working as expected also in corner cases. We can also merge it without it.

@matrix LMK if it runs on Metal

hans-vh:
Moved to mode 32600 as requested.
I see now that this mode is taken by CubeCart; Please advise.

jsteube:
Please update to the next free one. You can use src/modules/ to find out

hans-vh:
> Please update to the next free one. You can use src/modules/ to find out

Alright! Moved to mode 32700.

jsteube:
Code works fine on CUDA, but fails on ROCM. I'll fix this in the next days.

Issue #2994, opened 2021-09-23
dogechain sha256+pbkdf2-hmac-sha256+aes256


NFTCryptoList:
**Is your request based on a publicly known cryptographic schema and where can we find information about it?**
For instance:

I would like to request PBKDF2-HMAC-SHA256 for Dogechain.info to support GPU. Below is the encryption and decryption method. None of the current modes support it, I have tried all the modes in various formats. Only BTCrecover supports it without GPU. Dogechain.info password recovery will be used by a ton of community members and is much needed.

**Encryption**
```
self.addEventListener('message', function(e) {
    try {
        // Benchmark start
        var startTime = (new Date).getTime();
        
        // Fix scripts that depend on window
        window = self;
        
        // Load scripts
        importScripts("../crypto-js-2.5.4.js?v=4");
        
        // SHA256 password and Base64 encode it afterwards
        var password = e.data.password;
        var passwordSHA256 = Crypto.SHA256(password);
        var passwordBase64 = Crypto.util.bytesToBase64(Crypto.util.hexToBytes(passwordSHA256));
        
        var salt = Crypto.util.base64ToBytes(e.data.salt);
        
        var data = JSON.stringify(e.data.payload);
        
        var key = Crypto.PBKDF2(passwordBase64, salt, 32, { hasher: Crypto.SHA256, asBytes: true, iterations: 5000 });
        var encrypted = Crypto.AES.encrypt(data, key, { mode: new Crypto.mode.CBC(Crypto.pad.iso10126), iterations: 20000 });
        
        // Benchmark stop + calculation
        var stopTime = (new Date).getTime();
        var totalTime = stopTime - startTime;
        
        // Result
        var result = {
            encrypted: encrypted,
            time: totalTime
        }
        
        self.postMessage(result);
    } catch (e) {
        var result = {
            decrypted: null
        };
        
        self.postMessage(result);
    }
}, false);
```

**Decryption**

```
self.addEventListener('message', function(e) {
    try {
        // Benchmark start
        var startTime = (new Date).getTime();
        
        // Fix scripts that depend on window
        window = self;
        
        // Load scripts
        importScripts("../crypto-js-2.5.4.js?v=4");
        
        // SHA256 password and Base64 encode it afterwards
        var password = e.data.password;
        var passwordSHA256 = Crypto.SHA256(password);
        var passwordBase64 = Crypto.util.bytesToBase64(Crypto.util.hexToBytes(passwordSHA256));
        
[...]

PR #3746, opened 2023-05-26
Plugin for issue #2994 - dogechain implementation

JoeMcGeever:
Created a hashcat plugin based off the proof of concept written in issue #2994
https://github.com/hashcat/hashcat/issues/2994
Current hash format:
signature$payload$salt

### DIFF
--- a/OpenCL/m98765-pure.cl
+++ b/OpenCL/m98765-pure.cl
@@ -94,6 +94,8 @@ CONSTANT_VK u32 base64_table[64] =
   '4', '5', '6', '7', '8', '9', '+', '/',
 };
 
+// Wow it's the right file
+
 u32 base64_encode_three_bytes_better (u32 in){ //in has 3 u8s in, first u8 is not set)
     u32 out;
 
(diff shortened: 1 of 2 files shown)


jsteube:
@philsmd I can see you did some research on the theories behind how to find out the decryption was successful. Do you think it's ok to merge this PR from that perspective?

I could also see that you suggested a format, which is different to the one used in this PR. @JoeMcGeever could you update it to match that suggestion?

JoeMcGeever:
Updated the format accordingly.

jsteube:
Many thanks. I'll merge and update hash-mode to 32500 in a later step

dalmatele:
@JoeMcGeever @jsteube  I cannot find your contact's info to make a request of this commit.
Currently, to decrypt data, you decode all the content of the hash (304 byte) but I think we can increase the speed of process by only decoding the first block (not IV).
The reason is the content of the first block always contains below content:
`{"guid":"`
Current script, when run with RTX 2060 card, can reaches 200kH/s but if we update code maybe it can reach to 1Mh/s.
But I don't know how to modify your code (in cl script), can you guide me or update your code, if my idea is ok?

PR #3750, opened 2023-05-29
Fix a typo in the help text

ghost:


jsteube:
I think this is not a typo but intentional, because it's not required to use the full parameter name.

ghost:
Fair enough. These are shortened as well:

```
--induction
--keyb
--truecrypt-keyf
--veracrypt-keyf
```

Issue #3400, opened 2022-07-29
/usr/share/hashcat/OpenCL/m-0004_a0-pure.cl: No such file or directory


Tobey123:
run hashcat in kali and mac
but all show the information below
/usr/share/hashcat/OpenCL/m-0004_a0-pure.cl: No such file or directory

But I can't find the file "m-0004_a0-pure.cl"

Thanks.

Chick3nman:
You've got a corrupt module or custom module that is broken. Download a fresh copy of hashcat and try again.

philsmd:
What is the full command that you use (all switches / arguments / parameters) ?

I agree with @Chick3nman that it seems to be either a problem of your installation/setup or how you use it. I can't reproduce this (new ?) problem ? Does other/older hashcat versions from https://hashcat.net/hashcat work for you ?

Tobey123:
hashcat -a 0 -m 16500 a.hash 1.dict
I‘d try downloading a fresh copy of hashcat and try again.
But still the same.

philsmd:
if you have a local copy, you need to run it with a prefix of "./" e.g. something like `./hashcat --help`

maybe try to clean up a little bit older versions or installed versions (that's probably the reason for the `/usr/share/hashcat/` path).

Tobey123:
Yes, I have a local copy, I run it with a prefix of "./" 
I try to install the new clean os and fresh copy of hashcat and try again.
But still same........Orz.........

philsmd:
Does it still say something about `/usr/share/hashcat/` ?

do you use your operating system package manager to install it ?

if you run it locally, it should not say anything about `/usr/share/hashcat`.

What distro do you use ? Please give more details. Thx

NAs3c:
Hello,

Having the same problem, trying to launch hashcat on Kali VM

```
┌──(kali㉿kali)-[~/redacted]
└─$ ~/Tools/hashcat-6.2.6/hashcat.bin -a 0 -m 16500 jwt.hash trivial.lst -r /usr/share/hashcat/rules/* 
hashcat (v6.2.6) starting

OpenCL API (OpenCL 3.0 PoCL 3.1+debian  Linux, None+Asserts, RELOC, SPIR, LLVM 15.0.6, SLEEF, DISTRO, POCL_DEBUG) - Platform #1 [The pocl project]
==================================================================================================================================================
[...]

PR #3760, opened 2023-06-07
do not allow negative kern_type

matrix:
Hi,

this is the fix for #3400 

In case of negative kern_type, for example from a call to module_kern_type_dynamic(), hashcat will show an error and exit.

```
bash-3.2$ ./hashcat -m 16500 y --force -d1
hashcat (v6.2.6-574-geb276f12f) starting

METAL API (Metal 263.9)
=======================
* Device #1: Apple M1, 5408/10922 MB, 8MCU

OpenCL API (OpenCL 1.2 (Aug  8 2022 21:29:55)) - Platform #1 [Apple]
* Device #2: Apple M1, skipped

Rules: 1

Optimizers applied:
* Zero-Byte
* Not-Iterated
* Single-Hash
* Single-Salt

Watchdog: Temperature abort trigger set to 100c

Invalid hash-mode selected: -1

Started: Wed Jun  7 21:11:34 2023
Stopped: Wed Jun  7 21:11:34 2023
bash-3.2$
```

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -9893,6 +9893,13 @@ int backend_session_begin (hashcat_ctx_t *hashcat_ctx)
       }
     }
 
+    if ((int) kern_type == -1)
+    {
+      event_log_error (hashcat_ctx, "Invalid hash-mode selected: -1");
+
+      return -1;
+    }
+
     // built options
 
     const size_t build_options_sz = 4096;


matrix:
another way could be perform the check only if module_kern_type_dynamic is set, so here:

```
bash-3.2$ git diff cf6f71eb8a32051de8e0fa89050714f23f02b12c
diff --git a/src/backend.c b/src/backend.c
index ea6fa7acd..e22fe1048 100644
--- a/src/backend.c
+++ b/src/backend.c
@@ -9890,6 +9890,13 @@ int backend_session_begin (hashcat_ctx_t *hashcat_ctx)
         if (hash_info) hash_info_ptr = hash_info[0];
 
         kern_type = (u32) module_ctx->module_kern_type_dynamic (hashconfig, digests_buf, salts_buf, esalts_buf, hook_salts_buf, hash_info_ptr);
+
+        if ((int) kern_type == -1)
+        {
+          event_log_error (hashcat_ctx, "Invalid dynamic hash-mode selected: -1");
+
+          return -1;
+        }
       }
     }
 

```

PR #3776, opened 2023-06-26
Add GPG (CAST5 (SHA-1($pass)))

thatux:
Fully functional PR: cracks GPG passphrase protected keys that are protected with CAST5 and SHA-1.


matrix:
Hi,
[3776.patch](https://github.com/hashcat/hashcat/files/11899076/3776.patch)
some updates to make it work with Apple Metal

thatux:
Thanks, looks fine, could you explain / refer to something that explains what PRIVATE_AS does?

Will test and commit later (not behind pc now).

matrix:
PRIVATE_AS is defined in this block, is about the address space

https://github.com/hashcat/hashcat/blob/f341bfc82ebaf04f2602b06bc2b4a20d9aa22275/OpenCL/inc_vendor.h#L27C1-L67C7

from Apple Metal Shading Language Specification, page 79: "Any variable that is a pointer or reference needs to be declared with one of the address space attributes discussed in this section. If an address space attribute is missing on a pointer or reference type declaration, a compilation error occurs."

https://developer.apple.com/metal/Metal-Shading-Language-Specification.pdf

thatux:
Ok thanks, I see it being defined on this line
https://github.com/hashcat/hashcat/blob/f341bfc82ebaf04f2602b06bc2b4a20d9aa22275/OpenCL/inc_vendor.h#L57 
as #define PRIVATE_AS  thread

jsteube:
Many thanks!

PR #3794, opened 2023-07-14
Remove *75

0xVavaldi:
*75 does not contribute to the password.

'5 will cut it off at length 5, so swapping the 6th and 8th character has no purpose.


jsteube:
Works. Maybe it's time to finally fix base64.rule, that is by removing that "unknown" section which shouldn't be there. We left it there for a while because base64.rule was used by many people as a reference.

PR #3798, opened 2023-07-15
add meson 

neheb:
Provides simpler and faster build recipes than the custom ones currently
in use.

Uses wraps(subprojects) from WrapDB instead of custom dependencies when
the host does not provide them.

Added a CI to test most platforms.

Added wraps in case host system is missing dependencies.

Supports cross compilation. Sample Windows cross build definition files
added.

Setting as draft as this is not run tested. Some features of the stock build recipes are missing, production and whatnot.

edit: maybe the MinGW backends should not be tested. Hashcat does not support msvcrt properly, especially when it comes to Unicode paths. UCRT is the way forward IMO.

edit2: looks like Ubuntu doesn't support UCRT. Looks like they need to fix that.


neheb:
Note that the current Makefile does not build against CLANG32/64 on MSYS2.

neheb:
setting this to ready. Everything has been implemented AFAIK.

dropdeadfu:
why did you chose to rewrite the build logic in meson instead of trying to make it work in the makefile?

neheb:
Because meson is simpler and more flexible. It’s fairly hard to mimic it with Makefile logic.

jsteube:
A second build system costs every time the build changes, not once. Each change has to land in both, or they drift and the second stops being built.

In the meanwhile "src/Makefile" has taken 99 commits and grown by about a thousand lines, "src/feeds/" has arrived as a subsystem "src/meson.build" has never seen, and the wraps here for zlib, xxhash and minizip-ng no longer describe what "deps/" carries. So this is a rewrite rather than a rebase, and what waits at the end of it is the obligation rather than the work.

That is what I am closing it on, rather than anything about meson itself.

The MSYS2 CLANG32/64 build you opened with is worth having on its own, and it is a smaller thing than this. If you still hit it, a report or a change to "src/Makefile" for that alone gets read.

Issue #3689, opened 2023-04-16
feature: specify expected stdin record count for time estimation


roycewilliams:
If piping into hashcat, _hashcat_ can't automatically know the input size (number of records), so time estimation doesn't work. But very often the _user_ already knows (or can easily calculate) this size. This is very common in combinator and similar attacks, or if files need to be unzipped dynamically to stdout, etc.

it would be *very* useful to be able to optionally tell hashcat the expected size, and then use this information to provide a _rough_ time estimate in the status screen.

IMO this would _dramatically_ improve user experience for power users. Hashcat is already doing all of the other work necessary to provide and display the time estimate in the status output. Only this information is missing. Today, the user has to calculate the expected runtime manually - at start time, but then if anything else is externally impacting speed (like a competing job or processing running, etc.).

With a command-line option to supply expected size, the size could be automatically calculated by external scripts that create piping attacks:

```
$ cat test.sh
#!/bin/bash

count1=$(wc -l found.all | awk '{print $1}')
count2=$(wc -l found.all | awk '{print $1}')
stdincount=$((count1 * count2))

echo "combinator found.all found.all | hashcat --stdincount=$stdincount [...]"
```

```
$ ./test.sh
combinator found.all found.all | hashcat --stdincount=122102875700289 [...]
```

davidbolvansky:
+1, that would be useful feature

davidbolvansky:
What about _--limit_ ?

Sadly it is not respected in internal computation.  Still zero.

"progress": [5760, 0],

davidbolvansky:
I am also wondering if this extra info could help increase performance… atleast allocations could be more precise..

EDIT: I implemented this new option..

mohemiv:
What's about if we pass the number of words as the first line in STDIN?

Something like:
```
$HASHCAT[5]
1
12
123
1234
12345
```

I don't say that it's the only correct solution, but I currently develop some password generator and I don't see any other way to pass a value from my generator to hashcat.

davidbolvansky:
How this solution avoids all problems you raised here? A new flag sounds much better than magic word as the first word.

mohemiv:
I cannot pass a value from the generator to the argument automatically. A generator can pass data only to the pipe.

davidbolvansky:
Aha, so you cannot control hashcat invocation?

then environmental variable could be a workaround..

mohemiv:
[...]

PR #3800, opened 2023-07-16
New option to set number of words in the wordlist

davidbolvansky:
Implements https://github.com/hashcat/hashcat/issues/3689

I will have usecases for this new flag also in Fitcrack.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -717,6 +717,7 @@ typedef enum user_options_defaults
   VERACRYPT_PIM_START      = 485,
   VERACRYPT_PIM_STOP       = 485,
   WORDLIST_AUTOHEX_DISABLE = false,
+  WORDLIST_COUNT           = 0,
   WORKLOAD_PROFILE         = 2,
 
 } user_options_defaults_t;
@@ -841,6 +842,7 @@ typedef enum user_options_map
   IDX_VERSION_LOWER             = 'v',
   IDX_VERSION                   = 'V',
   IDX_WORDLIST_AUTOHEX_DISABLE  = 0xff53,
+  IDX_WORDLIST_COUNT            = 0xff54,
   IDX_WORKLOAD_PROFILE          = 'w',
 
 } user_options_map_t;
@@ -2354,6 +2356,7 @@ typedef struct user_options
   bool         remove;
   bool         restore;
   bool         restore_disable;
+  bool         rule_buf_l_chgd;
   bool         self_test_disable;
   bool         show;
   bool         slow_candidates;
@@ -2365,6 +2368,7 @@ typedef struct user_options
   bool         username;
   bool         veracrypt_pim_start_chgd;
   bool         veracrypt_pim_stop_chgd;
+  bool         wordlist_count_chgd;
   bool         version;
   bool         wordlist_autohex_disable;
   #ifdef WITH_BRAIN
@@ -2441,6 +2445,7 @@ typedef struct user_options
   u32          usage;
   u32          veracrypt_pim_start;
   u32          veracrypt_pim_stop;
+  u64          wordlist_count;
   u32          workload_profile;
   u64          limit;
   u64          skip;
--- a/src/status.c
+++ b/src/status.c
@@ -1069,7 +1069,9 @@ time_t status_get_sec_etc (const hashcat_ctx_t *hashcat_ctx)
 
   time_t sec_etc = 0;
 
-  if ((user_options_extra->wordlist_mode == WL_MODE_FILE) || (user_options_extra->wordlist_mode == WL_MODE_MASK))
+  if ((user_options_extra->wordlist_mode == WL_MODE_FILE) ||
+      (user_options_extra->wordlist_mode == WL_MODE_STDIN) ||
+      (user_options_extra->wordlist_mode == WL_MODE_MASK))
   {
     if (status_ctx->devices_status != STATUS_CRACKED)
     {
--- a/src/straight.c
+++ b/src/straight.c
@@ -13,6 +13,7 @@
 #include "rp.h"
 #include "wordlist.h"
 #include "straight.h"
+#include "user_options.h"
 
 static int straight_ctx_add_wl (hashcat_ctx_t *hashcat_ctx, const char *dict)
 {
@@ -80,7 +81,15 @@ int straight_ctx_update_loop (hashcat_ctx_t *hashcat_ctx)
         return -1;
       }
 
-      const int rc = count_words (hashcat_ctx, &fp, straight_ctx->dict, &status_ctx->words_cnt);
+      int rc = 0;
+      if (user_options->wordlist_count > 0)
+      {
+        status_ctx->words_cnt = user_options->wordlist_count * straight_ctx->kernel_rules_cnt;
+      }
+      else
+      {
+        rc = count_words (hashcat_ctx, &fp, straight_ctx->dict, &status_ctx->words_cnt);
+      }
 
       hc_fclose (&fp);
 
@@ -97,6 +106,10 @@ int straight_ctx_update_loop (hashcat_ctx_t *hashcat_ctx)
 
         return 0;
       }
+    }
+    else if (user_options_extra->wordlist_mode == WL_MODE_STDIN)
+    {
+      status_ctx->words_cnt = user_options->wordlist_count * straight_ctx->kernel_rules_cnt;
     }
   }
   else if (user_options->attack_mode == ATTACK_MODE_COMBI)
--- a/src/usage.c
+++ b/src/usage.c
@@ -60,6 +60,7 @@ static const char *const USAGE_BIG_PRE_HASHMODES[] =
   "     --outfile-autohex-disable  |      | Disable the use of $HEX[] in output plains           |",
   "     --outfile-check-timer      | Num  | Sets seconds between outfile checks to X             | --outfile-check-timer=30",
   "     --wordlist-autohex-disable |      | Disable the conversion of $HEX[] from the wordlist   |",
+  "     --wordlist-count           | Num  | Sets number of words in the wordlist                 | --wordlist-count=14341564",
   " -p, --separator                | Char | Separator char for hashlists and outfile             | -p :",
   "     --stdout                   |      | Do not crack a hash, instead print candidates only   |",
   "     --show                     |      | Compare hashlist with potfile; show cracked hashes   |",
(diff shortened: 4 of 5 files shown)


0xVavaldi:
Perhaps an extra conditional check to make wordlist count only function in combination with STDIN. As I believe there is no reason for users to (accidentally) specify this outside of an stdin attack (and might cause issues/confusions otherwise)

davidbolvansky:
> Perhaps an extra conditional check to make wordlist count only function in combination with STDIN. As I believe there is no reason for users to (accidentally) specify this outside of an stdin attack (and might cause issues/confusions otherwise)

Well I see usecases...
If I know words count, why not to inform hashcat about that and skip useless counting words. If you pass wrong number, that is your resposibility... as with every other flag.

jsteube:
This is a very dangerous operation. If the numbers are not correct, it can cause hashcat to do out-of-boundary reads, which can crash hashcat with all sorts of weird errors. I'm not strictly against this, but it requires a good number of tests. Did you check this also with more rare parameters like -S, -j and -k?

davidbolvansky:
Yes, you are right, there could be surprises with other attack modes. So new patch pushed, flag restricted to attack mode 0 as this is my real motivation.

davidbolvansky:
@jsteube please review / merge

mohemiv:
Regarding `-j`, currently it's a bit annoying that hashcat count words every time when `-j` is used, even when there isn't any reject rule specified.

I have an attack (-a 0) that runs for 6 hours (sha512crypt), so there is no way I can use any rule file with a big wordlist. So, instead of using a rule file, I constructed a single rule that I have specified in `-j` parameter. Now every start of hashcat takes appox. 20-30 min since hashcat counts words in my wordlist. But it doesn't make any sense since the amount of words is always the same. 

@davidbolvansky I haven't properly checked, but it seems like the case with reject rules in `-j` parameter is not covered in the current PR code.

davidbolvansky:
-j is for hybrid attacks only no?

I dropped the changes for other attacks than straight attack (could be added as followup patch)

mohemiv:
The `-j` parameter can be used in -a 0. Especially in my case, when I have only 1 rule and I don't want to put it in a rule file.

davidbolvansky:
Aha, interesting. But in this case I would assume that kernel_rules_cnt is 1.

There are so many places in codebase where multiplication like this happens.

mohemiv:
I think the correct solution is to check whether `-j` parameter contains a reject rule and depending on this use cache (or --wordlist-count flag) or recalculate the cache (and throw an error if --wordlist-count is used). But it would take some time to implement.

I'm not a member of a hashcat team, but I like the idea of using --wordlist-count with -a 0 to skip an unnecessary counting.

mohemiv:
Just to clarify this a bit more. There are two types of rules: ordinary rules and reject rules. The formula `count(words) * count(rules)` is correct for ordinary rules. This formula is NOT correct for reject rules.

davidbolvansky:
so existing code is broken?

https://github.com/hashcat/hashcat/blob/f936f371ab36370f9d007d177ff45c9ac58b3e50/src/wordlist.c#L517

mohemiv:
It's not, because this code is executed when the previous condition is true.
```
if (run_rule_engine (user_options_extra->rule_len_l, user_options->rule_buf_l) == 0) {
...
keyspace *= straight_ctx->kernel_rules_cnt; 
```

davidbolvansky:
I see, hmm. So probably just copy that condition here as well.

What do you think, what we could do? -j with -a 0 looks kinda niche to me.

@jsteube ?

mohemiv:
> -j with -a 0 looks kinda niche to me.

I don't think that @jsteube would agree that support of this should be dropped.

davidbolvansky:
I didnt say that we should drop support for this combination of flags. Just this flag would be unsupported (not effective) with that combination of flags.

mohemiv:
What's interesting and maybe it's a bug, but the combination of parameters `-a 0 -j` has a huge application.

I have sha512crypt hashes and I use the optimized kernel. Here is a setup similar to mine:

```
$ cat /tmp/w.txt 
123456789012345

$ cat /tmp/r.txt 
:
]
] ] 
] ] ] 
] ] ] ]
] ] ] ] ]
] ] ] ] ] ]
] ] ] ] ] ] ]
] ] ] ] ] ] ] ]
] ] ] ] ] ] ] ] ]
] ] ] ] ] ] ] ] ] ]

$ cat /tmp/test.txt
$6$TEST$FuGD6BoQZY6QYL2Ndq9XtLz9qJ5e3oX6CtpU4tZbnsx.1J48nHz9sg.CIgMmfTwsIwZTLKKcANST02BYw/UjH/
```

The password is 1234567890. I can discover it by the following commands:
```
hashcat -D 1 -m 1800 /tmp/test.txt --potfile-disable /tmp/w.txt -j '] ] ] ] ]' -O
hashcat -D 1 -m 1800 /tmp/test.txt --potfile-disable /tmp/w.txt -r /tmp/r.txt
```
But I cannot discover it by this command:
```
hashcat -D 1 -m 1800 /tmp/test.txt --potfile-disable /tmp/w.txt -r /tmp/r.txt -O
```
Hashcat just doesn't send the password to the kernel and it's never "destrongified". 

So the only way to make `] ] ] ] ]` rule work is to put it to `-j` parameter. And I didn't invent the `] ] ] ] ]` rule, there are similar rules in the built-in best66.rule list. The `-S` flag also works, but I haven't read anywhere that this is the way to go.

davidbolvansky:
Yes, weird. Please create new issue for your findings.

jsteube:
As I said, this is very complicated. Especially for the maintenance in future. That combination of maintenance and complicated logic is the worst situation you can end up. We can't reduce the complicated part, so we need to reduce the maintenance. 

I can foresee that soon after we merge this users want to have same functionality but for the other attack modes. So we need to implement this functionality now that the thoughts are fresh. That means please provide support for all attack modes.

In addition I'd like some intensive tests done before merge. That is especially with rare combination of parameters like -S and -j, but also the other attack-modes, also fast hashes and slow hashes, as discussed above. Please provide a list, ideally with command so we can reproduce these tests locally.

I'm not trying to make this more complicated than it needs to be. These things just need to be done.

davidbolvansky:
I am going to commit to implement everything for every attack, sorry.

davidbolvansky:
I will wait for @jsteube’s opinion as he is the one who maintains hashcat.

As you said, you want small maintenence so scope of this patch is small and restricted.

jsteube:
Before we move forward with the PR, I want to highlight again the complexity of what this PR is trying to do.

I also don't enjoy waiting for the wordlist to load before hashcat starts. It's annoying. But there's a good reason I never added this flag myself. Again, I'm not strictly against it, but we need to do some brainstorming first to make sure we've thought everything through.

Maybe that's what's missing in the PR: a list of reasons why adding the parameter might be a bad idea, or at least a list showing how the PR handles certain problematic constructs in attack scenarios that would be affected by it. Without that, it feels like an attempt to solve a problem without considering the consequences.

So please support me in finding the traps before we fall into them. And if everything still looks good afterwards, then let's go ahead and add the parameter.

The main problem is: the number provided must be absolutely perfect. There's zero room for inaccuracy. Why?

- If the number is too large, hashcat will read data from unallocated memory, which could lead to out-of-bounds reads or null pointer dereferences and then hashcat will crash.
- If the number is too small, users will get false negatives, meaning a correct password might be skipped, even if it's in the wordlist.

In both cases, users would blame hashcat for behaving incorrectly, even though it's actually user error.

It doesn't help to print warnings or similar messages. After reading the same message ten times, most people will just ignore it and hope it still works. And the real problem: we have no way to verify that the input number is correct. That's the whole point of this PR. So we can't even warn them if it's wrong.

1. Should we add this to all attack modes, or start small?

On the positive: Adding it to all attack modes would be ideal, because then the parameter becomes more useful and people are more likely to use it. If not, it might only be used in specific overlays like Fitcrack. But it's a feature I've wanted for a long time, so shouldn't we take the time and do it right?

On the negative: It might help to go back and ask: why are we doing this in the first place? It's about startup time. Loading the wordlist takes time, and we want to reduce that delay. Attack mode 3 doesn't load wordlists at all, so it wouldn't really benefit.

Maybe limiting it to -a 0 is good enough?

2. Should we add checks in hashcat to detect invalid baseword counts?

I think we could do some. For example, if hashcat hits EOF in the wordlist before reaching the expected count, that could trigger a warning. But I don't see this implemented in the PR.

3. Some things from the PR:

I understand the change was made only to cover your particular use case. But it feels incomplete, because we have to live with it going forward. In the future, when we come across this code again, we or another developer who is not aware of this PR might ask: why does it handle only the -j option, and not the -k option?

```
bool         rule_buf_l_chgd;
```

---

This attribute is not supposed to be set manually. We use count_words() or sp_get_sum() to handle that.


```
status_ctx->words_cnt = user_options->wordlist_count * straight_ctx->kernel_rules_cnt;
```

What I mean is that this code change should go into count_words() instead.

---

This part also feels incomplete. With the information available, we could take it further. For example, we could use it to calculate the ETA again in the status display.

```
else if (user_options_extra->wordlist_mode == WL_MODE_STDIN)
{
  status_ctx->words_cnt = user_options->wordlist_count * straight_ctx->kernel_rules_cnt;
}
```

davidbolvansky:
Unfortunely I am no longer working on Fitcrack/hashcat.

I agree with your concerns but unable to address them anymore. This one should be closed.

0xVavaldi:
I feel like there's some room for discussion on this still. I think the addition of the flag provides room to improve the stdin at the very least.

Stdin lacks proper tuning because of stdin's limits, but also suffers from the issue that it's not known how many candidates exist (something that HTP preprocessor flags --skip --limit, and --keyspace cover). As a small trial it could be worth to add there, but the thought of fully swapping to the keyspace provided does add the issues Atom specified. Mainly the thought of adding out of bound memory access worries me.

Maybe a first step is implementing it purely for auto tuning and runtime estimation purposes, not for the actual accesses. If we keep the discussion going and try to map pitfalls it might be implemented more fully later.

How does hashcat behave if the size of a wordlist changes mid-run? Perhaps some safe exit mechanism could be worth investigating? (I lack the technical depth on C & memory mapping to properly map the possibilities there)

jsteube:
> but also suffers from the issue that it's not known how many candidates exist 

I'm not sure to which problem you are referring here. The main issue is that we can't predict the ETA.

> Maybe a first step is implementing it purely for auto tuning and runtime estimation purposes

From implementation perspective, there should be no two code paths when it comes to auto-tuning whatever it is stdin or not based attack. Do you have a command line that demonstrates the issue which you mean here? That could be interesting.

> How does hashcat behave if the size of a wordlist changes mid-run? 

Bad, that can't happen. I think there will be a race condition in which sometimes EOF will just prevent a crash, and in other situations I could also see smaller issue like finished < 100% in status display, but then also crashes of all sorts.

PR #3825, opened 2023-08-10
Fix a wrong expression in rp.c (static analysis)

mohemiv:
During analysis of hashcat with PVS-Studio a bug was found:
```
rp.c (773) | V1051 | Consider checking for misprints. It's possible that the 'rule_len' should be checked here.
```

The manual reviewing has confirmed that a misprint was indeed present.


mohemiv:
I've reviewed the previous commits and it looks like there isn't actually any bugs. But it might be better to change line order to avoid confusions.

PR #3827, opened 2023-08-11
clang-tidy: align and pack structs

neheb:
Found with: altera-struct-pack-align

### DIFF
--- a/OpenCL/inc_ecc_secp256k1.h
+++ b/OpenCL/inc_ecc_secp256k1.h
@@ -216,7 +216,7 @@ typedef struct secp256k1
 {
   u32 xy[SECP256K1_PRE_COMPUTED_XY_SIZE]; // pre-computed points: (x1,y1,-y1),(x3,y3,-y3),(x5,y5,-y5),(x7,y7,-y7)
 
-} secp256k1_t;
+} __attribute__((aligned(128))) secp256k1_t;
 
 
 DECLSPEC u32  transform_public (PRIVATE_AS secp256k1_t *r, PRIVATE_AS const u32 *x, const u32 first_byte);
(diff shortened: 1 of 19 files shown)


jsteube:
This is interesting, but did you try it also on all the different compute runtimes like cuda, nvidia opencl, amd opencl, rocm, pocl, intel cpu opencl, intel gpu opencl, etc?

neheb:
haha no. I don't have the hardware necessary to do so. I mostly posted this PR to point out that clang-tidy has OpenCL related checks. This one supports automated fixes.

jsteube:
This looks a bit dangerous:

```
OpenCL/m16800-pure.cl: In function ‘m16800_init’:
OpenCL/m16800-pure.cl:110:81: warning: taking address of packed member of ‘struct wpa_pmkid’ may result in an unaligned pointer value [-Waddress-of-packed-member]
  110 |   sha1_hmac_update_global_swap (&sha1_hmac_ctx0, esalt_bufs[DIGESTS_OFFSET_HOST].essid_buf, esalt_bufs[DIGESTS_OFFSET_HOST].essid_len);
      |                                                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~
```

And it's the only one that pops up because it's used in emulation mode from the host. Very like this affects a lot more modules.

Can you share benchmarks or specific reasons why this alignment is beneficial for these structs? Were there performance issues or alignment errors being addressed?

neheb:
I'll do some testing with the latest clang-tidy. It's possible the check is buggy.

neheb:
I think I'll just close this. This might only be relevant for FPGAs. For brevity, the explanation is here:

https://releases.llvm.org/20.1.0/tools/clang/tools/extra/docs/clang-tidy/checks/altera/struct-pack-align.html

PR #3828, opened 2023-08-11
clang-tidy: use const pointer

neheb:
Found with readability-non-const-parameter


jsteube:
This should be safe to merge (hopefully :))

neheb:
I mean this would be a compile time error if so.

PR #3829, opened 2023-08-11
Update gzlog.c

gogolovefish:
fix memory leak


jsteube:
Going to merge, but note that this code comes from the original gz sources. With the next update it can get overwritten.

PR #3838, opened 2023-08-17
Prevent file-unsafe characters in --session

PenguinKeeper7:
Prevent file-unsafe characters in `--session`, previous behaviour:
```
> ./hashcat -m 99999 -a 3 abcdef abcdef2 --session "Hello:Hello" 
...
./Hello:Hello.outfiles: Invalid argument

./Hello:Hello.outfiles: Invalid argument
```
When only the "Hello" file would be created in a Windows environment

### DIFF
--- a/src/user_options.c
+++ b/src/user_options.c
@@ -1384,6 +1384,22 @@ int user_options_sanity (hashcat_ctx_t *hashcat_ctx)
     }
   }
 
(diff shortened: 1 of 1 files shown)


roycewilliams:
Also useful when Linux mounts NTFS volumes in MS-compability mode - thanks!

PR #3839, opened 2023-08-18
fix some const-qual warnings

neheb:


jsteube:
I can see some of these changes are also part of PR https://github.com/hashcat/hashcat/pull/3798 I hope merging this will not break the other PR...

PR #3857, opened 2023-08-27
Mode 17230: Remove lower bound restriction for hash count

davidbolvansky:
I was able to create zip archive with one file, extracted hash (-c, checksum only) and cracked it (dict attack) with hashcat + this patch.


davidbolvansky:
From hashcat discord:

"the reason for having a minimum amount of files is because you will get lots of false positives
when having the checksums only you either have 8 or 16 bits (depending on the checksum size) per file.. so 3 already is fairly low to be sure."

PR #3871, opened 2023-09-10
Add Dogechain wallet extraction script

### LINKED ISSUE #3869: dogechain.info extraction issues
**Describe the issue**
The current dogechain.info module added in https://github.com/hashcat/hashcat/pull/3746 doesn't have an extraction script, meaning people have to format the hash manually as `$dogechain$0*iterations*payload*salt`. As well as this, the current module [demands exactly 320 character payloads](https://github.com/hashcat/hashcat/blob/5b06ffe63f407bca5fe8960e0bdded6b5cd8d9d5/src/modules/module_32500.c#L106C1-L109C26), while real-world payloads can be any length so either handling this through the extraction script or updating the module would be highly UX-friendly. Its currently unknown if it's possible for payloads to be less than 320 characters.

Important note: There are 2 known formats, the newest being: https://github.com/hashcat/hashcat/issues/3868 and the older format: https://github.com/hashcat/hashcat/pull/3746, -m 32500 is the older format. At the time of writing, there is no new format module,

ventaquil:
Fixed #3869 

Please, check it on your own. I don't have old wallet to check it (newly created has `cipher: AES-GCM` field).

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -134,6 +134,7 @@
 - Help: show supported hash-modes only with -hh
 - Makefile: prevent make failure with Apple Silicon in case of partial rebuild
 - Rules: Rename best64.rule to best66.rule and remove the unknown section from it
+- Tools: Added new dogechain2hashcat.py extraction script.
 
 * changes v6.2.5 -> v6.2.6
 
--- a/tools/dogechain2hashcat.py
+++ b/tools/dogechain2hashcat.py
@@ -1,0 +1,45 @@
+#!/usr/bin/env python3
+
+#
+# Author......: See docs/credits.txt
+# License.....: MIT
+#
+
+import json
+from argparse import ArgumentParser
+from base64 import b64decode, b64encode
+
+
+if __name__ == "__main__":
+    # prepare parser and parse args
+    parser = ArgumentParser(description="dogechain2hashcat extraction tool")
+    parser.add_argument("path", type=str, help="path to dogechain wallet")
+    args = parser.parse_args()
+
+    try:
+        # open wallet file
+        with open(args.path) as file:
+            wallet = json.load(file)
+        # verify basic fields
+        if ("pbkdf2_iterations" not in wallet) or ("payload" not in wallet) or ("salt" not in wallet):
+            parser.error("Unsupported wallet format")
+        # variant
+        if ("cipher" not in wallet) or (wallet["cipher"].upper() == "AES-CBC"):
+            variant = 0
+        elif wallet["cipher"].upper() == "AES-GCM":
+            variant = 1
+        else:
+            parser.error("Unsupported wallet cipher variant")
+        # iterations
+        iterations = wallet["pbkdf2_iterations"]
+        # payload
+        payload = b64decode(wallet["payload"])
+        payload = payload[:240]  # payload length must be equal to 320
+        payload = b64encode(payload)
+        payload = payload.decode()
+        # salt
+        salt = wallet["salt"]
+        # print
+        print("$dogechain$" + "*".join([str(variant), str(iterations), payload, salt]))
+    except IOError as e:
+        parser.error(e.strerror.lower())


dalmatele:
I think your function is wrong. It's not simple like truncating the characters in the payload

ventaquil:
Hey @dalmatele 

I was quite long time ago when I worked with this script...

When you take a look at the module's code you will see the hardcoded payload length equal to 240 bytes (not more nor less):

https://github.com/hashcat/hashcat/blob/6716447dfce969ddde42a9abe0681500bee0df48/src/modules/module_32500.c#L106-L111

My plan was to do not reinventing the wheel but if you have any suggestions I am open to implement better approach.

dalmatele:
@ventaquil : If you can download the javascript from dogechain.info, you can have below script:
https://dogechain.info/js/secure/frontend.js
Let's see the `loadPayloadKeeper` function.
They's applying some tricks to decode the payload.

dalmatele:
> Hey @dalmatele
> 
> I was quite long time ago when I worked with this script...
> 
> When you take a look at the module's code you will see the hardcoded payload length equal to 240 bytes (not more nor less):
> 
> https://github.com/hashcat/hashcat/blob/6716447dfce969ddde42a9abe0681500bee0df48/src/modules/module_32500.c#L106-L111
> 
> My plan was to do not reinventing the wheel but if you have any suggestions I am open to implement better approach.

This code is applied for the decoded payload. Current payload you get from the Dogechain is not decoded payload we can use for this function

ventaquil:
Do you have any wallet which I can use for tests? It would be very helpful to give valid password and wallet pair. I cannot find any backup wallet which I used for tests.

ventaquil:
Backup gist with `frontend.js` code - https://gist.github.com/ventaquil/897d73cdd9c9ace5e54ecf3a2589da52.

dalmatele:
@ventaquil 
```
{
  "guid": "52500558-b3fa-4318-b6a3-3c55835c6575",
  "payload": "Dw5gN5A0eMzK0aFzkn+sIymgcy5ncpGt7wSyJLFbCJvTJnbA0rNRPWz6LQRZeaoqOFugYjXqlj/HMjOGTlAmUWjZSbOCkngzQBwZe1jKY0iVWIKI7SCHE7yI6YQN0bV9HoAo4zl+nDxYC/EN+ue11OqWOYZNhDxwyn/poyRd7bcigeuvcjam292+pglc2koAmOJeuOg0m+NvdvS7Mw07mAGCyLVrxLmj5efDtIHWrdrGWv5SgmTUUr1pUEOJG5Ps;LzZwl72onfBNml5KA22HBJMEJwyfSUWxOyXlA36kWJA6FAdVYadYX7Ln3+Xe9BG2662d6lLI+b3SAK9njFtGe0O0DOnS+HIyVCBvPAkMsj874dNB2J9bJfpm5/kuh58EFkZoJXpXywscBlPHu7TlbTieb8B1g/SPpkLrXvQL22XM6NVI927wB1s081KFjcT8IBdMmMZIp/qVtlkyoqKfQBwqtmGb1sOYsa0r2n8aBcI+RFP6O7Md6LCMfZeRYIlU;RHpIRAm+DoUA1oribmXhxJ44LiXL+BpOdanK/P3jFfs+FVPagu//ItCgfUMlmHgt5x7Qhlu7JZMWJU3WdQOeixdxbpI1EEseAdOtRG/sIwv2Vf4vo16effuZuQm/MO8xBngeKE56splt6A39FHQUG4v3YtlDb1w6yVphyDBrhIFxUsdJyCxjzWappyE00FDdG+OOnkRKtBGRtN/yF1UIaUvs699ssM88FnUNK4xGc0WmPpJpL8ToLKIKTmD9c6VJ;U63oZYYqy9EvT85dOAuIR0TaQmRACJvzETyH3laoUXLGv9ktjYd9WdXJX/7G9gQnmzwCd1J65/GsitvbI9Ch0OHeCW3iIZqjVGGg4sU91RWR3BVgl4d5XLNxmHrUVHkOrEXMmJRgrF9oWXeYSsmNMQiCk5StgJppBE9WMc6x//tDPMVWMOWICmM6E1Qn+zlIrjtizPIWjGdckTgaKw7a415Yi1yFyh211ueMczycIvshhEKFGJO6f0UuwLej11Rv;su0DZr+7yKjxsF0guCzmBUERevBP18gRrJQclCExf86CuM5KrBXZAijCy8J53g8eITNCiCFQORMD5gDRHcfbypJ/hGSh0VBdCcxJ8CQCc3K8Ok+BJ0le7XsxL23oC1XoU/ORNZjN7yg7O21Ez+RArDGItEfdHHjMpox9rNyRHyjOExDFyz55Zb3rAF8PrTUL2dnOmo0BF8uvW45sZzxJdFztFwg/FyeycFRbwhQ0i/TfzW679dVvQkdOpiG31c4X",
  "salt": "Gc55/tevRG/ODxUN0ENiRg==",
  "pbkdf2_iterations": 5000,
  "cipher": "AES-CBC"
}
```
Here you go. Password of this wallet is 'btcr-test-password'

dalmatele:
And maybe you need their public key to decrypt:
```
LS0tLS1CRUdJTiBQVUJMSUMgS0VZLS0tLS0KTUlJQ0lqQU5CZ2txaGtpRzl3MEJBUUVGQUFPQ0FnOEFNSUlDQ2dLQ0FnRUE5dUtucklvbXdLdFJQRE9nOU00RgpUS05teDIxYmdrbDMvOGhnS0pZaFQvR2g0d1NDUlJwT3NJZndSNTRNY2F2VWE2Q3NQSXdHaXRldHBzaDhmS2xCClVBK296RDZ3R0NsN2dJL05neExBTkczQ3lNZDhtWXUvSERsMkVCMXBGV3NINUJqTnk2YVpqbFRxV0hJWEU3a1gKMFR2Y0VxQ3JkcUE3VW4vV01IYkFwRVlDV1pZVXJ1MUVVOU9reEQreXk1djQ0Zm54WG5VMXB6a3RPbjdzZmVQSQozdm9zejNzcnd3dGY0Y3lFSkczUDdZdXFnQ3I3b3c2SFI5WEdrR3NCZGJtNlNNTU83NVEwUmxTNTJqNHcwZXFrCkRrNE9hTy9qZ3hPNlByZTBDb2pIbGZTMzBqTE1hN1h4VW9NT01PN3NydURFRW1JUVN0c3ZUaGJrVHlRQWg1NXkKbzlnOXhBbUhyOGhBTzFhWnRTTWhUbXdXSWxjRERXSStCOWZHN2cvcis2b2N6d3VHcE44Sk0vUS9Xa01HRW16dQpvYThTMFZIdmdaR0pXbUpNMGpGWDJvTjdEb1VJTGRoUGxKVUl4T25UQXdpUFRXYmV5VmpLcy8vWjZxSVdVRFM2CldPWm1sOE1ZQWJQMHVteVZlQytzNTczaG0wVlJIK0tBMTZqQXcwdGt2Z1ZudVdxcVVNRXByaWVXVjhCMmYwR2EKZ0FJWERGS2llTzBPMzNVcXI2OHRQMlp2UmFWZWhQclNGV3ZOcEJGRHJnZm51TmhNdWNKRHJnbjBmR0RGOWpydAp1bDRRYnplamZUU0liM2NQc1N3VnlGZlpOMVFzR3Q4V2pqRWhCVDYrR3BOME05MlZNeldxTDVEdFB5a2pvNFpLCk5QZnE1cEZEQk1YUWplMFFwdXNLNEhNQ0F3RUFBUT09Ci0tLS0tRU5EIFBVQkxJQyBLRVktLS0tLQo=
```

ventaquil:
I will take a look into that :+1:

jsteube:
And update on this?

jmvermeulen:
Hey, I took a look at this while working on another hashcat module and I think I figured out what's going on with the test wallet.

The issue isn't really the truncation - that part is actually fine for older dogechain wallets. It's the same approach blockchain2hashcat.py uses (dogechain is a fork of blockchain.info after all). Module 32500 only needs a few AES blocks to validate the key, so 240 bytes is plenty.

The problem is that dogechain added RSA transport encryption at some point. When you download a wallet now, the payload comes back wrapped in an ephemeral RSA session key:

1. Browser generates a 1536-bit RSA keypair in memory on login
2. Sends the public key to the server
3. Server encrypts the AES payload with that RSA key
4. Payload arrives as semicolon-separated RSA-encrypted segments
5. Browser decrypts with `loadPayloadKeeper()` using the private key

That private key only ever existed in the browser's memory during the session - it's never stored anywhere. So if someone saved the wallet JSON from an API response (which is what dalmatele's test wallet looks like - 5 segments of 192 bytes, matching 1536-bit RSA), the payload can't be unwrapped offline. There's no way around that.

The script should probably just detect this and give users a clear error instead of producing a bad hash. Something like checking for semicolons in the payload field - old-format wallets won't have them.

I'd suggest:
- Add semicolon detection with an error message explaining the RSA limitation
- Add AES-GCM detection (variant 1) with a note that it needs a different module
- Document that only old-format (pre-RSA) wallets are supported
- Find an old-format test wallet for validation (dalmatele's won't work for the reason above)

Here's a minimal patch that adds just those two checks without changing the rest of the script:

```diff
--- a/tools/dogechain2hashcat.py
+++ b/tools/dogechain2hashcat.py
@@ -4,6 +4,7 @@
 # License.....: MIT
 #

+import sys
 import json
 from argparse import ArgumentParser
 from base64 import b64decode, b64encode
@@ -24,13 +25,21 @@ if __name__ == "__main__":
         if ("pbkdf2_iterations" not in wallet) or ("payload" not in wallet) or ("salt" not in wallet):
             parser.error("Unsupported wallet format")
         # variant
-        if ("cipher" not in wallet) or (wallet["cipher"].upper() == "AES-CBC"):
+        cipher = wallet.get("cipher", "AES-CBC").upper()
+        if cipher == "AES-CBC":
             variant = 0
-        elif wallet["cipher"].upper() == "AES-GCM":
-            variant = 1
+        elif cipher == "AES-GCM":
+            print("Error: AES-GCM wallets require a different module (not yet available).", file=sys.stderr)
+            sys.exit(1)
         else:
             parser.error("Unsupported wallet cipher variant")
+        # detect RSA-wrapped payload (newer dogechain wallets)
+        if ";" in wallet["payload"]:
+            print("Error: This wallet uses RSA transport encryption.", file=sys.stderr)
+            print("The payload was encrypted with an ephemeral browser session key", file=sys.stderr)
+            print("that is not stored in the wallet file. Cannot extract offline.", file=sys.stderr)
+            sys.exit(1)
         # iterations
         iterations = wallet["pbkdf2_iterations"]
         # payload
```

With those changes this is mergeable imo. The extraction script would cover the same scope as blockchain2hashcat.py does for blockchain.info - just the older wallet format.

jsteube:
The wallet posted in this thread cannot be unwrapped offline, which is worth settling first because it is the reason the script looked wrong. Its payload splits on semicolons into 5 segments of exactly 192 bytes:

```
  seg 0: 256 b64 chars -> 192 bytes = 1536 bits
  seg 1: 256 b64 chars -> 192 bytes = 1536 bits
  ... 5 of them
```

One 1536 bit RSA block per segment, which matches what @dev-goalz traced in April: the keypair is generated in the browser at login and the private half never leaves that page, so a payload captured from an API response has no key left anywhere to decrypt it. The truncation to 240 bytes was never the problem.

Could you add the two checks from his patch, the semicolon test and the AES-GCM refusal?

The second has a harder reason than the patch gives. The variant digit is inside the signature rather than a field of its own:

```
static const char *SIGNATURE_DOGECHAIN = "$dogechain$0";
```

so a line written as $dogechain$1 fails the signature check and hashcat refuses to load it. The variant = 1 branch has never produced a hash that could be used.

That also settles whether removing it gives anything up later. AES-GCM needs a kernel of its own and therefore a mode of its own, and a new mode brings its own hash format with it, so the variant field here is not holding a place for one.

Issue #3878, opened 2023-09-18
invalid base58 check


benjiqq:
**Describe the bug**

I believe the base58 check here is invalid, reporting false solutions, should be sth like this

https://github.com/hashcat/hashcat/pull/1805/files

```
DECLSPEC int is_valid_base58_8 (const u8 v)
{
  if (v > (u8) 'z') return 0;
  if (v < (u8) '1') return 0;
  if ((v > (u8) '9') && (v < (u8) 'A')) return 0;
  if ((v > (u8) 'Z') && (v < (u8) 'a')) return 0;

  return 1;
}

```

should be IMO

```
DECLSPEC int is_valid_base58_8 (const uint8_t v) 
{
  const char *base58_chars = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";

  for (int i = 0; i < 58; i++) 
  {
    if (v == base58_chars[i]) 
    {
      return 1;
    }
  }

  return 0;
}
```

PenguinKeeper7:
For clarity: it checks for base58 characters with ASCII ranges between 49-57 (1-9), 65-90 (A-Z), 97-122 (a-z) but does not exclude capital `O`, capital `I` or lowercase `l` as base58 does. Fixed in PR https://github.com/hashcat/hashcat/pull/3879

PenguinKeeper7:
Issue can be closed - Fixed in https://github.com/hashcat/hashcat/pull/3879

PR #3879, opened 2023-09-18
Fix base58 charset check

PenguinKeeper7:
See https://github.com/hashcat/hashcat/issues/3878

### DIFF
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -2815,6 +2815,11 @@ DECLSPEC int is_valid_base58_8 (const u8 v)
   if ((v > (u8) '9') && (v < (u8) 'A')) return 0;
   if ((v > (u8) 'Z') && (v < (u8) 'a')) return 0;
 
+  // https://github.com/hashcat/hashcat/issues/3878
+  if (v == 'O') return 0;
+  if (v == 'I') return 0;
+  if (v == 'l') return 0;
+
   return 1;
 }
 


magnumripper (commented):
Assuming this function is performance critical, you might want to replace all those lockstep conditionals (the old and the new) with a single lookup in a static 128-byte table.

PenguinKeeper7:
> Assuming this function is performance critical, you might want to replace all those lockstep conditionals (the old and the new) with a single lookup in a static 128-byte table.

I tested performance for this commit, it seemed identical to pre-commit. The tiny single u8 comparison is nothing in comparison to the computation of AES/SHA256 that the few kernels that use this check also have to perform. Seems like overcomplication but I can test if you feel it'd affect performance above margin of error?

GuzziRaz:
> I tested performance for this commit, it seemed identical to pre-commit. The tiny single u8 comparison is nothing in comparison to the computation of AES/SHA256 that the few kernels that use this check also have to perform. Seems like overcomplication but I can test if you feel it'd affect performance above margin of error?

Nah probably not worth it then. Just saying it'd be easy and perhaps the function is used in other places?

Anyway good find on the initial bug - kinda hard to spot if you're not looking for it.

magnumripper:
> I tested performance for this commit, it seemed identical to pre-commit. The tiny single u8 comparison is nothing in comparison to the computation of AES/SHA256 that the few kernels that use this check also have to perform. Seems like overcomplication but I can test if you feel it'd affect performance above margin of error?

Nah probably not worth it then. Just saying it'd be easy and perhaps the function is used in other places?

Anyway good find on the initial bug - kinda hard to spot if you're not looking for it.

PR #3895, opened 2023-10-18
Add 4060/4060Ti to autotune alias

PenguinKeeper7:


jsteube:
Many thanks! If you have this GPU, optimized values for SCRYPT based hash-modes would be nice to have, too.

PenguinKeeper7:
Unfortunately not, nor do I know anyone with one but I'll keep an eye open if I see anyone with one!

PR #3901, opened 2023-11-08
add bip39 bitcoin wallet module

seed-cat:
TLDR: I have implemented the world’s fastest bitcoin seed word and passphrase recovery as a new hashcat module.

**Purpose**
While hashcat already provides support for bitcoin “brain wallets” and electrum style wallets, these types of wallets have been falling into disuse because of how easily funds may be [stolen if one only relies on user-generated entropy](https://fc16.ifca.ai/preproceedings/36_Vasek.pdf).

Therefore since 2013, modern wallets began to switch to the [BIP-39 mnemonic code standard](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) consisting of “seed words” and an optional passphrase, which offers greater cracking difficulty due to the 128-bits of machine-generated entropy and a 2048-round PBKDF2-HMAC-SHA512 key-stretching algorithm which makes hashing much slower.  As the majority of bitcoin users switch to these wallets, need for fast recovery tools will continue to increase.

For instance imagine if the seed words are known, but the user-generated passphrase has been lost.  Case in point, I originally wrote this module to successfully recover bitcoin I had forgotten the passphrase to.  After failing with my CPU-based implementation, I chose to leverage hashcat which already implements most of the necessary GPU-optimized algorithms and has excellent password generation support.

Another possible use-case occurs if information about the seed words is lost.  Misspellings or misorderings of seed words may easily result in lost funds if the user has not backed-up the seed words 100% correctly.  For instance, the seed words “wood” and “wool” might easily be mistaken for each other or be written down in the wrong order.  Words may also become partially lost over time from events such as [fire damage](https://blog.lopp.net/metal-bitcoin-seed-storage-stress-tests-round-vi/).

**Implementation**
The 2048-rounds of PBKDF2-HMAC-SHA512 proved to be the bottle-neck of the hashing algorithm.  Therefore I opted to implement a “pure slow” kernel attack.  The major divergence from the existing PBKDF2-HMAC-SHA512 module are the special salts and ECC key derivations required by the BIP-39 and BIP-32 specifications.

One aspect of the BIP-39 standard that proved difficult to implement is the variable-length SHA256 checksum that is stored in the last seed word.  At first, I attempted to try all possible seed words and short-circuit the kernel functions if the checksum was invalid.  However, this actually decreased performance due to the fact that GPUs do not handle divergent branches well.

So instead I opted to require users to pass in seed word entropy via the password itself by encoding each unknown 11-bit seed word entropy as a binary charset (the first character being 5-bits and the second character being 6-bits).  The last bits of entropy can be passed as a single character, but varies depending on the total number of seed words.  For instance with 12-words the last word contains 7-bits of entropy but with 24-words the entropy is only 3-bits.

By calculating the checksum, the number of possible passwords can be reduced by orders of magnitude.  For instance, if 3 words of a 24-word seed phrase are unknown, a naive implementation would need to attempt 2^33 hashes.  However, we can actually filter the input to 2^25 hashes (two 11-bit words and 3-bits of additional entropy) reducing the password space by 256x.

Additional complexity arises from the fact that different bitcoin wallets may use arbitrary derivation  paths and several common address types.  Since the kernel remains identical in these cases, I opted to allow users to specify these via their hash salt, rather than attempt to split the types into different modules.

**Benchmarks**
Hashcat’s goal is to be the world’s fastest password recovery utility.  I found only two other GPU-based implementations of the BIP-39 algorithm.  [BtcRecover](https://github.com/gurnec/btcrecover) partially leverages GPU-acceleration based on John the Ripper, but only for some operations and is CPU-bound when guessing passphrases.  The other [implementation by Jon Cantrell](https://medium.com/@johncantrell97/how-i-checked-over-1-trillion-mnemonics-in-30-hours-to-win-a-bitcoin-635fe051a752) is pure GPU, but was written specifically to win a contest by guessing seed words, not create a general cracking utility.

To get the most apples-to-apples comparison, I ran on the same 2080 TI hardware used by the Cantrell implementation since I was unable to get his original source code to run.  BtcRecover and my hashcat module were similarly limited to searching the same derivation path with missing seed words.  Note that because BtcRecover naively handles checksums, it would check 256x more passwords than the other implementations in this specific case, in addition to having a slower hashrate.

|Implementation|Performance|
|-------------------|----------------|
|btcrecover|49.0 kH/s|
|johncantrell97|142.9 kH/s|
|hashcat (address)|300.5 kH/s|
|hashcat (xpub)|507.9 kH/s|

In my tests, the hashcat implementation is at least twice as fast as the other pure GPU implementation, likely due to the optimizations in the hashcat kernel compared to the naive CPU to GPU port done by Cantrell.

To give users the option of avoiding the ECC performance hit I added support for master xpubs as a hash target which provide 128-bits of entropy and are exportable from most bitcoin wallet software.  The hashcat xpub implementation is almost identical in speed to the existing PBKDF2-HMAC-SHA512 module set at 2048 iterations, proving that my implementation does not add much overhead.

**Future Improvements**
The ECC operations required to go from bitcoin private to public key reduced the hashrate by around 40%.  The performance hit of ECC could probably be eliminated if we [cached ~67MB of secp256k1 calculations which could offer a 20-30x performance](https://github.com/XopMC/CudaBrainSecp) boost at the cost of additional code complexity and memory overhead.  This could also greatly improve the hashrate of other wallet modules.

Working on the kernel made me realize there is no easy way within hashcat to filter out bad password candidates in the kernel (seed words with invalid checksums) without taking a significant performance hit due to divergent branching.  Since this is a slow hash, we could filter password candidates via a valid seed word generator utility.  However, the ideal solution would be add support for module password filtering on the kernel-level since it overcomes the stdin bottle-neck and allows for hashcat brain support.

Since BIP-39 is quite a technical spec, building more utilities for bitcoin wallet recovery would be good step for improving usability.  For instance, users may not know what likely derivation paths or address types their wallet software has chosen.  If the seed words are corrupt, users may wish to try all possible combinations from a list of possible candidates, filtering out those with bad checksums.


braindamagedjustnow:
Why did you closed this?

PR #3907, opened 2023-11-20
Mode 33400: mega.nz password-protected link (PBKDF2-HMAC-SHA512)

DCNick3:
mega.nz is a centralized file-sharing service with some nice security features, like client-side file decryption. Apparently it is [not widely known](https://hashcat.net/forum/thread-10499.html), but mega password-protected links are implemented fully on the client side using `PBKDF2-HMAC-SHA512` to derive a key and an `HMAC-SHA256` to verify the derived key.

This PR adds support for cracking passwords for such links.

### Relevant documentation
- [mega.nz whitepaper](https://mega.nz/SecurityWhitepaper.pdf)
- [format description in source code](https://github.com/meganz/webclient/blob/4790e52b1eb1856a47e60b52615ffc58242c2d1c/js/ui/export.js#L1)
- [password decryption function](https://github.com/meganz/webclient/blob/4790e52b1eb1856a47e60b52615ffc58242c2d1c/js/mobile/mobile.password.decryption.js#L177).

### Example hash

`P!AgD________U2XVjJi1vxkJgMPf5rkQYUn1H_6WI_sKtiic69mqBKP_____________________O_PDG0Om7BSapL1QoRAgUrz9vzaZmrYnU8t-Au6hteg`

It corresponds to the following link:

https://mega.nz/#P!AgD________U2XVjJi1vxkJgMPf5rkQYUn1H_6WI_sKtiic69mqBKP_____________________O_PDG0Om7BSapL1QoRAgUrz9vzaZmrYnU8t-Au6hteg

This hash follows the algorithm, but does not correspond to a real file.

### Notes

I did not test this on file links (they have `01` at the second byte and a different key length), but I did try to support it.

I am unsure of some of the choices I made in the PR and would like some input from the maintainers:

1. What should the category for such a hash should be? I've used `HASH_CATEGORY_APPLICATION_DATABASE`, but it's not really an application and more like a centralized service..
2. I am not sure if the use of `COMPARE_M` in the kernel makes sense. Intuitively, we already know the digest we want to compare with so, probably `COMPARE_S` should be used instead? I was not able to figure out how to use it here though


WhiteyCookie:
Why isn't this already implemented? Can't wait. Plz hashcat people, do it before Christmas :)

src/modules/module_33400.c
```diff
@@ -0,0 +1,295 @@
...
+  u32 *digest = (u32 *) digest_buf;
+
+  mega_t *mega = (mega_t *) esalt_buf;
+
+  if (line_buf[0] != 'P' || line_buf[1] != '!') return (PARSER_SIGNATURE_UNMATCHED);
+
+  // the base64-decoded data after the P! is either 88 or 104 bytes, depending on whether it is a folder or a file link
+  u8 decoded_buf[0x100];
+
+  if (line_len > 140) return (PARSER_GLOBAL_LENGTH);
```
ricnava00: This should be 142
The base64 is ceil(104/3)*4=140 chars long, plus 2 for P!

DCNick3:
Rebased & fixed the length check

jsteube:
Sorry for delay, and top PR, thanks!

PR #3909, opened 2023-11-23
Add faq link to potfile advice

PenguinKeeper7:
Add the FAQ link to the potfile advice as this seems to trip a lot of beginners up, "Yeah, but what actually is a potfile? And why are my hashes in it?"

(Thanks to @roycewilliams for the shortened link!)


jsteube:
Nice idea, thanks

PR #3938, opened 2024-01-25
Increase virtual backend limit

fse-a:
This change increases the maximum device limit to 256, which is particularly useful when using the new 'virtual backend device' functionality of upcoming Hashcat 7.

The backend_devices_filter is changed to a boolean array where each index refers to a specific virtual backend device. The last place in the array ( DEVICES_MAX + 1) is reserved to determine if filtering is needed. If this value is true, then no filter was specified and all (virtual) devices are enabled.

Let us know what you think and if any changes are needed. 

Kind regards

(Pelle & Ewald)

### DIFF
--- a/include/common.h
+++ b/include/common.h
@@ -131,7 +131,7 @@ but this is needed for VS compiler which doesn't have inline keyword but has __i
 
 #define CPT_CACHE           0x20000
 #define PARAMCNT            64
-#define DEVICES_MAX         128
+#define DEVICES_MAX         256
 #define EXEC_CACHE          128
 #define SPEED_CACHE         4096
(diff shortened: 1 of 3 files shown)


PenguinKeeper7:
Not blocking the PR or disputing code, just curious. What benefit do you see in having greater than 128 virtual devices? What situation would be limited by the previous upper-bound?

jsteube:
@PenguinKeeper7 Hashcat 7 will feature devices other than just GPU and CPU. Some of these devices must be virtually divided into multiple devices (even if they are only one physical piece). This is something determined by the API of the device (depending on the vendor) and we lack control over it

PR #3939, opened 2024-01-29
Flush debug-file more often

PenguinKeeper7:
Instead of only flushing/unlocking the debug file at the end of Hashcat, flush after each write, aligning it with the outfile and potfile's behaviour

### DIFF
--- a/src/debugfile.c
+++ b/src/debugfile.c
@@ -76,6 +76,11 @@ void debugfile_write_append (hashcat_ctx_t *hashcat_ctx, const u8 *rule_buf, con
     if ((debug_mode == 3) || (debug_mode == 4) || (debug_mode == 5)) hc_fputc (':', &debugfile_ctx->fp);
   }
 
+  if (hc_lockfile (&debugfile_ctx->fp) == -1)
+  {
+    event_log_error (hashcat_ctx, "%s: Failed to lock file.", debugfile_ctx->filename);
+  }
+
   hc_fwrite (rule_buf, rule_len, 1, &debugfile_ctx->fp);
 
   if ((debug_mode == 4) || (debug_mode == 5))
@@ -104,6 +109,13 @@ void debugfile_write_append (hashcat_ctx_t *hashcat_ctx, const u8 *rule_buf, con
   }
 
   hc_fwrite (EOL, strlen (EOL), 1, &debugfile_ctx->fp);
+
+  hc_fflush (&debugfile_ctx->fp);
+
+  if (hc_unlockfile (&debugfile_ctx->fp))
+  {
+    event_log_error (hashcat_ctx, "%s: Failed to unlock file.", debugfile_ctx->filename);
+  }
 }
 
 int debugfile_init (hashcat_ctx_t *hashcat_ctx)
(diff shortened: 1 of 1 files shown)


jsteube:
After patch, there would still be this left in debugfile_destroy():

```
  if (debugfile_ctx->filename)
  {
    hc_unlockfile (&debugfile_ctx->fp);
...
```

So basically double unlock. I think this needs to be removed, but didn't check.

PenguinKeeper7:
There's a chance it might fail the first time and throw the `Failed to unlock file` error and I don't see any reason why double-freeing it would cause any problems, especially when it's only potentially duplicated once per run other than potentially just being a little overly protective. I'm happy to remove it, I just don't see any downsides

jsteube:
Please remove it, no need to guarantee running into double unlock
If you want to fix it correctly, catch the exceptions and deal with them (whatever that means you must decide).
Do you really need to lock/unlock just to flush?

PenguinKeeper7:
>Do you really need to lock/unlock just to flush?

Probably not with just a single session but with multiple running at once, it may cause problems. I only did that because it's the precedent in Hashcat to lock -> flush -> unlock as seen below with the only real difference being that I check return values. Many of these also have the same double-unlock "problem" in the destroy functions

Seemingly the only flushes that Hashcat does that aren't locked/unlocked is the pidfile and restore file but both have external protections so they don't need it

Kernel file write:
https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/backend.c#L797-L817

Logfile:
https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/logfile.c#L55-L69

Potfiles:
https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/potfile.c#L304-L313

Loopback:
https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/loopback.c#L157-L163

Hashes:
https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/hashes.c#L266-L268

jsteube:
The problem is that there's a unlock left in your destroy function, but only in your code. All the other code you listed is not affected. 

https://github.com/hashcat/hashcat/blob/7965e85dd6a9e1acb1543906e21ca421686d39fb/src/debugfile.c#L171-L175

You can try to see it symmetric. On the same level where's the lock, there should be the unlock. You removed the lock, but not the unlock.

If exceptions in between are catched and break/return, these exception must also do the unlock. If the exception are not caught, that's also fine.

PenguinKeeper7:
> The problem is that there's a unlock left in your destroy function, but only in your code. All the other code you listed is not affected.

Yep, I was specifically replying to your quote "Do you really need to lock/unlock just to flush?", I wasn't using those links to talk about the destroy/double-unlock thing. 

See https://github.com/hashcat/hashcat/pull/3939/commits/b8a1d872ac82e0ddd074793c9480f218f9bc58ff

jsteube:
Very nice, thanks!

PR #3952, opened 2024-02-21
Add dynamic iteration count support for hashmodes 26600 and 26610 (Metamask)

its5Q:
Newer metamask versions have an iteration count of 600000 and the iteration count is stored in the vault's JSON. This PR adds support for vaults created with the new version by introducing hashmodes that support an arbitrary iteration count. It also updates the metamask2hashcat tool to parse iteration count if it's present.


tools/metamask2hashcat.py
```diff
@@ -51,18 +51,30 @@ def metamask_parser(file, shortdata):
...
         exit(1)
 
+
     if isMobile is False:
 
+      if 'keyMetadata' in j and 'params' in j['keyMetadata'] and 'iterations' in j['keyMetadata']['params']:
+        iter_count = j['keyMetadata']['params']['iterations']
+      else:
+        iter_count = 10000
+
```
ventaquil:
```suggestion
      try:
        iter_count = j['keyMetadata']['params']['iterations']
      except KeyError:
        iter_count = 10_000
```

```pycon
>>> params = {}
>>> try:
...     value = params["my"]["unique"]["key"]
... except KeyError:
...     value = 1_000
...
>>> value
1000
```

src/modules/module_26620.c
```diff
@@ -0,0 +1,415 @@
...
+
+  token.len[0]     = 10;
+  token.attr[0]    = TOKEN_ATTR_FIXED_LENGTH
+                   | TOKEN_ATTR_VERIFY_SIGNATURE;
+
+  token.sep[1]     = '$';
+  token.len_min[1] = 1;
+  token.len_max[1] = 6;
+  token.attr[1]    = TOKEN_ATTR_VERIFY_LENGTH
+                   | TOKEN_ATTR_VERIFY_DIGIT;
```
ventaquil: Cannot `TOKEN_ATTR_OPTIONAL_ROUNDS` be used in old module?

its5Q: Let me take a look at that

its5Q: I've added support for dynamic iteration count to the old modules with TOKEN_ATTR_OPTIONAL_ROUNDS. One thing though, this attribute works kinda unintuitive and weird, as it actually acts both as a rounds count token, and a salt token at the same time, if that makes sense. Is that intended?

its5Q:
I've updated the metamask2hashcat tool to use the new format and added a test vault JSON that you could use it on to try out the new dynamic iteration count support. If everything else looks good, I think it can be merged.

src/modules/module_26620.c
ventaquil: It can be a little bit unintuitive. I have seen another option with two different tokenizers.

Take a look.

https://github.com/hashcat/hashcat/blob/fafb277e0736a45775fbcadc1ca5caf0db07a308/src/modules/module_07100.c#L118-L308

An opinion from maintainers would be very helpful.

learnerLj:
thanks for your PR. Maybe the original `ST_HASH` should change.

jsteube:
We've got a bit of a luxury problem: there are now three PRs (and even more on some external sites) that all do all more or less the same thing:

https://github.com/hashcat/hashcat/pull/3952
https://github.com/hashcat/hashcat/pull/4120
https://github.com/hashcat/hashcat/pull/4142

That's really cool, and I appreciate everyone's effort!

That said, I can only merge one. Since I'm not too familiar with MetaMask and the details, is there anything technically important that's not already covered in https://github.com/hashcat/hashcat/pull/3952?

I don't want to downplay anyone's work, which is why I want to discuss this first. If there's no strong technical reason to pick a different one, I'd lean toward #3952 since it was first to appear and is nicely integrated into the existing module, which helps with lower maintenance.

learnerLj:
> I'd lean toward https://github.com/hashcat/hashcat/pull/3952 since it was first to appear and is nicely integrated into the existing module, which helps with lower maintenance.

Yes, I agree with you. #3952 maintains backward compatibility. But I'm not sure whether we should update the Perl script in the tests and the help info.

sieger82:
I only created #4120 for one project I was working on and was not aware that #3952 already existed. My code was not thoroughly tested, so it's probably better to merge #3952

jsteube:
> Yes, I agree with you. #3952 maintains backward compatibility. But I'm not sure whether we should update the Perl script in the tests and the help info.

Good point. But none of the PR's have updated the perl test script or did I miss something?

its5Q:
I think I can work on updating the perl test script

cyclone-github:
Just FYI in case it's useful, the widely used 26620 module below builds off the official 26600 module by adding support for all dynamic iteration counts.
https://github.com/cyclone-github/hashcat_26620_kernel

The metamask_pwn toolkit below contains an extractor which works on all versions of metamask wallets (as of this writing). 
https://github.com/cyclone-github/metamask_pwn

jsteube:
> Just FYI in case it's useful, the widely used 26620 module below builds off the official 26600 module by adding support for all dynamic iteration counts. https://github.com/cyclone-github/hashcat_26620_kernel
> 
> The metamask_pwn toolkit below contains an extractor which works on all versions of metamask wallets (as of this writing). https://github.com/cyclone-github/metamask_pwn

I'm not sure what's missing with the #3952 PR, but if there's anything missing you have in your version, it would be really nice if you could add that to our version, too.

Issue #3959, opened 2024-02-29
Feature req [masks]: add increment backwards flag


ysrn:
Currently when using the increment feature it goes from the start of the pattern to its end. I think it would be super useful to have a flag to reverse that direction. I'm hitting up against this use-case a lot, and thought it might be useful to others as well.

Example:
If I'd like to run through all masks starting with a single digit and ending with 2-4 lowercase, that's easy: -i ?d?l?l?l?l

Now, if I'd like to run through all masks **ending** with a single digit, but starting with 2-4 lowercase. I'm forced to verbosely specify each mask in order: -i ?l?l?d then -i ?l?l?l?d then -i ?l?l?l?l?d.

If I had --increment-backwards, then I could use a simple  -i --increment-backwards ?l?l?l?l?d.

Chick3nman:
While this isn't a rebuttal you could work around this currently by specifying the masks in a mask file to avoid having to run multiple attacks. An "inverse increment" or similar flag could still prove useful.

dmgarry:
Just thinking about this... May or May not be possible as I can't try it at the minute but if you stdout... put your mask in reverse so last character first and so on... increment it the way you want... pipe it back into hashcat applying the reverse rule?

Chick3nman:
You could do that @dmgarry but you lose the speed/efficiency of running the mask attack from within the kernel. By using mask files you achieve the same thing but without the potential speed loss.

PenguinKeeper7:
Please test https://github.com/hashcat/hashcat/pull/3960 and see if it meets what you want. Here's an example command, using your listed situation of an unknown amount of letters, then one number at the end
```
> ./hashcat -m 99999 abcd1 -a 3 ?l?l?l?l?l?l?l?d -i --increment-inverse --quiet

abcd1:abcd1
```

ysrn:
Woah that was fast! I tested it and played around a bit. It does exactly what I described/needed and will be incredibly useful, thank you!

ysrn:
Found a small bug by looking at the code. Using an escaped question mark ?? in the pattern will lead to it being misinterpreted as a pattern token instead of a literal. So the pattern ?l??1?d will error twice with "Custom charset 1 is undefined" in inverse mode.

PenguinKeeper7:
[...]

PR #3960, opened 2024-02-29
Add --increment-inverse

PenguinKeeper7:
See: https://github.com/hashcat/hashcat/issues/3959

~~(Submitted as a draft as it could do with a small clean-up and will likely need to be changed with future Atom's large commit anyway)~~


virodoran:
This is awesome, I have been wanting this as a hashcat feature for ages.

One thing I'd suggest though is to not have it require the `-i` flag. In fact, I'd actually rather see it be mutually exclusive with `-i` so when running attacks you would choose between `--increment` and `--increment-inverse`.

I'd love a `-ii` shorthand (for increment inverse) or something too, but I guess hashcat doesn't have any other flags like that so maybe it would be weird... and `-I` is already taken unfortunately.

roycewilliams:
> I'd love a `-ii` shorthand (for increment inverse) or something too, but I guess hashcat doesn't have any other flags like that so maybe it would be weird... and `-I` is already taken unfortunately.

There are `-II` and `-hh`, at least - but so far, they seem to mean "do it more".

PenguinKeeper7:
> 
> There are `-II` and `-hh`, at least - but so far, they seem to mean "do it more".

What's your opinion on `-ii`? Do you think it'd be out of place? I'm happy to implement it and I think it might be cool

Shooter3k:
> > There are `-II` and `-hh`, at least - but so far, they seem to mean "do it more".
> 
> What's your opinion on `-ii`? Do you think it'd be out of place? I'm happy to implement it and I think it might be cool

personally, I love the -ii idea

PenguinKeeper7:
@virodoran, `--increment` and `--increment-inverse` are no longer required together (https://github.com/hashcat/hashcat/pull/3960/commits/0fb6e68c7604d4f9c78b6c20251301a7af61f7d2)
@virodoran / @Shooter3k, `-ii` is now an alias for `--increment-inverse` (https://github.com/hashcat/hashcat/pull/3960/commits/18a35bdffdffcc8a7f06ebdfe2c2ddca18f941a4)

include/types.h
```diff
@@ -2349,6 +2351,7 @@ typedef struct user_options
   bool         hex_salt;
   bool         hex_wordlist;
   bool         increment;
+  bool         increment_inverse;
```
ventaquil: Isn't it better to use an `enum` type for it?

```suggestion
  increment_t increment;
```

It could be just like the following:

```c
typedef enum increment {
    INCREMENT_NONE,
    INCREMENT_NORMAL,
    INCREMENT_INVERSED,
} increment_t;
```

PenguinKeeper7: Feels unnecessary imo but resolved:
https://github.com/hashcat/hashcat/pull/3960/commits/3db7f01dfa27cf7d1ea03c311be343a6bd4d9dd6

jsteube:
The implementation of -ii in user options could use some improvement. It would be great if you could simplify it perhaps by using the same ++ code style that -hh and -II is using, or removing -ii and stick with only long options instead. Also I have problems merging this locally and can't test it.

PenguinKeeper7:
> Also I have problems merging this locally and can't test it.

Yeah, because this PR is so old and master continued without it, so it's now broken due to backwards-incompatible changes. It was functional originally but there's been 544 commits, 310k additions and 104k removed lines since this was created, so a lot has changed - in this case, types.h conflicts so it should now be safe to merge.

Code has also been cleaned up, removing the admittedly janky parsing function. https://github.com/hashcat/hashcat/pull/3960/commits/088f4a804b7217e6dce73f10c6d65139c76c97f0

jsteube:
Of course, and sorry for the mess with the merges. I didn't mean to suggest it was your fault that this PR stopped working. I'd really like to include it in the next release, so I thought it would be great if it could be fixed before the release. 

I tried the updated version and saw that it starts with a compiler warning:

```
$ make -s
src/user_options.c: In function ‘user_options_init’:
src/user_options.c:251:43: warning: implicit conversion from ‘enum user_options_defaults’ to ‘increment_t’ {aka ‘enum increment’} [-Wenum-conversion]
  251 |   user_options->increment                 = INCREMENT;
      |                                           ^
```

But it seems to work fine:

```
$ ./hashcat --stdout -a 3 ?lest123 -ii
3
23
123
t123
st123
est123
sest123
mest123
cest123
best123
aest123
pest123
lest123
jest123
dest123
test123
rest123
kest123
hest123
gest123
fest123
nest123
iest123
eest123
west123
vest123
oest123
yest123
zest123
qest123
uest123
xest123
```

Could you please fix the compiler warning?

PenguinKeeper7:
Fascinating! I was compiling with an older version of gcc (9.4.0), which didn't trigger this warning but using a newer environment does trigger it (11.3.0). I wonder if the verbosity of what a compiler considers a warning changed at some point, newer gcc just doesn't seem to like implicit enum conversion anymore, so just explicit casting is completely safe.
Either way, should now be fixed https://github.com/hashcat/hashcat/pull/3960/commits/822cfd159de875209c7a25fc87e6200748ee520b, feel free to re-test. Thanks for the report!

PR #3961, opened 2024-02-29
Make skipped device warning respect --quiet

PenguinKeeper7:
The other usual device information does not show when using --quiet, only the skipped warning does, making it a little out of place.
Old:
```
> ./hashcat -m 99999 abcdef -a 3 abcdef --quiet
The device #2 specifically listed was skipped because it is an alias of device #1

abcdef:abcdef
```
New:
```
> ./hashcat -m 99999 abcdef -a 3 abcdef --quiet
abcdef:abcdef
```


mohemiv:
I had to patch hashcat so that on my system it wouldn’t output 12 extra warnings every run. I don´t know why the warning was added in the first place.

Maybe we should delete the message completely?

PR #3980, opened 2024-04-02
Raise Bitwarden Iteration Limit

PenguinKeeper7:
Replace old 999,999 iteration cap with a larger cap to satisfy: https://hashcat.net/forum/thread-11885.html. Unknown if there is any Bitwarden-native cap on iterations.

### DIFF
--- a/src/modules/module_23400.c
+++ b/src/modules/module_23400.c
@@ -167,14 +167,14 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
   const u32 iter1 = hc_strtoul ((const char *) iter1_pos, NULL, 10);
 
   if (iter1 <     1) return (PARSER_SALT_ITERATION);
-  if (iter1 > 10^10) return (PARSER_SALT_ITERATION);
+  if (iter1 > 4294967294) return (PARSER_SALT_ITERATION);
 
   salt->salt_iter = iter1 - 1;
 
   const u32 iter2 = hc_strtoul ((const char *) iter2_pos, NULL, 10);
 
   if (iter2 <     1) return (PARSER_SALT_ITERATION);
-  if (iter2 > 10^10) return (PARSER_SALT_ITERATION);
+  if (iter2 > 4294967294) return (PARSER_SALT_ITERATION);
 
   salt->salt_iter2 = iter2 - 1;
 
--- a/src/modules/module_23400.c
+++ b/src/modules/module_23400.c
@@ -9,6 +9,7 @@
 #include "bitops.h"
 #include "convert.h"
 #include "shared.h"
+#include "limits.h"
 
 static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
 static const u32   DGST_POS0      = 0;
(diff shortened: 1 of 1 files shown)


BaelfireNightshd:
This doesn't appear to be able to load the hash (though it can from the benchmark). I ran the following command
`./hashcat -a 0 -m 23400 '$bitwarden$2*100000*2*bm9yZXBseUBoYXNoY2F0Lm5ldA==*+v5rHxYydSRUDlan+4pSoiYQwAgEhdmivlb+exQX+fg=' /usr/share/SecLists/Passwords/Leaked-Databases/rockyou.txt`
which uses the [same hash from the benchmark](https://github.com/PenguinKeeper7/hashcat/blob/29b39c5b832d7dab174d2589d4c06c3974eecc28/src/modules/module_23400.c#L30) and got the following output. 
```
 hashcat (v6.2.6) starting

* Device #1: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
* Device #2: WARNING! Kernel exec timeout is not disabled.
             This may cause "CL_OUT_OF_RESOURCES" or related errors.
             To disable the timeout, see: https://hashcat.net/q/timeoutpatch
The device #2 specifically listed was skipped because it is an alias of device #1

CUDA API (CUDA 12.4)
====================
* Device #1: NVIDIA GeForce RTX 3080, 8973/9999 MB, 68MCU

OpenCL API (OpenCL 3.0 CUDA 12.4.89) - Platform #1 [NVIDIA Corporation]
* Device #2: NVIDIA GeForce RTX 3080, skipped

Hash '$bitwarden$2*100000*2*bm9yZXBseUBoYXNoY2F0Lm5ldA==*+v5rHxYydSRUDlan+4pSoiYQwAgEhdmivlb+exQX+fg=': Salt-iteration count exception
No hashes loaded.

Started: Sun Apr 14 19:46:10 2024
Stopped: Sun Apr 14 19:46:11 2024
```

src/modules/module_23400.c
```diff
@@ -166,15 +166,15 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   const u32 iter1 = hc_strtoul ((const char *) iter1_pos, NULL, 10);
 
-  if (iter1 <      1) return (PARSER_SALT_ITERATION);
-  if (iter1 > 999999) return (PARSER_SALT_ITERATION);
+  if (iter1 <     1) return (PARSER_SALT_ITERATION);
+  if (iter1 > 10^10) return (PARSER_SALT_ITERATION);
```
ventaquil: `^` is xor operator - take a look here https://en.cppreference.com/w/cpp/keyword/xor.

`iter1 > 10^10` will be compiled into `iter1 > 0` which almost always will be true.

ventaquil (changes requested):

PenguinKeeper7:
Not sure how I didn't see this, thanks. Must've been something I added after testing to simplify the code a bit

ventaquil:
What do you think about creating some consts like `ITER_DIGITS_MIN`/`ITER_DIGITS_MAX`?

```c
const static unsigned ITER_DIGITS_MIN = 1;
const static unsigned ITER_DIGITS_MAX = 10;

// validate with those values
const unsigned iter_min = pow(10, ITER_DIGITS_MIN - 1);
const unsigned iter_max = pow(10, ITER_DIGITS_MAX - 1);
if (iter1 < iter_min) return (PARSER_SALT_ITERATION);
if (iter1 > iter_max) return (PARSER_SALT_ITERATION);
```

PenguinKeeper7:
Well, it's just set to int-max for now. The kernel itself would have to be re-designed if you wanted to go higher and there could potentially be accidental accuracy/overflow issues if you do go higher than u32 and it's not caught. Defining consts would only be purely a readability thing, it wouldn't enable easy future expansion

ventaquil:
Hmm... Well, isn't it simpler to use `const static unsigned INTMAX = ~0`? Maybe with a tiny comment why this value was chosen.

It's more about readability rather than option to quick and easy increase iter size.

src/modules/module_23400.c
```diff
@@ -166,15 +167,15 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   const u32 iter1 = hc_strtoul ((const char *) iter1_pos, NULL, 10);
 
-  if (iter1 <      1) return (PARSER_SALT_ITERATION);
-  if (iter1 > 999999) return (PARSER_SALT_ITERATION);
+  if (iter1 <     1) return (PARSER_SALT_ITERATION);
+  if (iter1 > UINT_MAX) return (PARSER_SALT_ITERATION);
```
ventaquil: `iter1` is `u32` so condition `1 <= iter1 <= UINT_MAX` isn't just `iter != 0`?

Is it possible to store more than `UINT_MAX` in `iter1`?

```suggestion
  if (iter1 == 0) return (PARSER_SALT_ITERATION);
```

PR #3984, opened 2024-04-05
 Added hash-mode: Citrix NetScaler (PBKDF2-HMAC-SHA256) 

stumblebot:
Here's code for a new hash-mode. It handles Citrix NetScaler's most recent (since 13.1) local system user password hashing scheme, PBKDF2-HMAC-SHA256 with 2500 rounds.

This is an extension of work done by my colleague @dru1d-foofus and I. You can find some additional background and documentation : [here](https://github.com/CDW-OffSec/NetScaler-Password-Hash-Type-5)

As a quick summary, this is basically a re-worked implementation of the existing hash mode 10900. The kernel is essentially unchanged and the module has been modified to reflect the the different token parsing scheme and input type (hex instead of base64). I've also made some changes to buffer sizes to reflect the static length of the salt.

I've tested this on a handful of manually generated hashes pulled from a Citrix NetScaler appliance and it works as expected on an OpenCL cpu-based platform, as well as a CUDA-driven gpu-based platform. 

Thanks!


matrix:
Hi @stumblebot, could you rename to 10910? Thanks

Issue #3988, opened 2024-04-16
Profiling of --show revealed a secret


mohemiv:
Hello there,

I've done the profiling of the --show options and I have found out how --show can increase performance by two times or more, and sometimes by a hundred times.

Let's take a look into the `potfile_remove_parse` function in `src/potfile.c`:
```C
int potfile_remove_parse (hashcat_ctx_t *hashcat_ctx)
{

  [REDACTED]

  while (!hc_feof (&potfile_ctx->fp))
  {
    size_t line_len = fgetl (&potfile_ctx->fp, line_buf, HCBUFSIZ_LARGE);

    if (line_len == 0) continue;

    char *last_separator = strrchr (line_buf, hashconfig->separator);

    if (last_separator == NULL) continue; // ??

    char *line_pw_buf = last_separator + 1;

    size_t line_pw_len = line_buf + line_len - line_pw_buf;

    char *line_hash_buf = line_buf;

    int line_hash_len = last_separator - line_buf;

    line_hash_buf[line_hash_len] = 0;

    if (line_hash_len == 0) continue;

    if (hash_buf.salt)
    {
      memset (hash_buf.salt, 0, sizeof (salt_t));
    }

    if (hash_buf.esalt)
    {
      memset (hash_buf.esalt, 0, hashconfig->esalt_size);
    }

    if (hash_buf.hook_salt)
    {
      memset (hash_buf.hook_salt, 0, hashconfig->hook_salt_size);
    }

    [REDACTED]

        {
          const int parser_status = module_ctx->module_hash_decode (hashconfig, hash_buf.digest, hash_buf.salt, hash_buf.esalt, hash_buf.hook_salt, hash_buf.hash_info, line_hash_buf, line_hash_len);

          if (parser_status != PARSER_OK) continue;

          if (hashconfig->potfile_keep_all_hashes == true)
          {
            potfile_update_hashes (hashcat_ctx, &hash_buf, line_pw_buf, (u32) line_pw_len, all_hashes_tree);

            continue;
          }

          hash_t *found = (hash_t *) hc_bsearch_r (&hash_buf, hashes_buf, hashes_cnt, sizeof (hash_t), sort_by_hash, (void *) hashconfig);

          potfile_update_hash (hashcat_ctx, found, line_pw_buf, (u32) line_pw_len);
        }

    [REDACTED]
```

What is going on here is as follows:
1. Hashcat reads a line from the potfile.
2. Hashcat memsets buffers.
[...]

PR #3989, opened 2024-04-17
Improving performance of --show #1

mohemiv:
Related: https://github.com/hashcat/hashcat/issues/3988

Let's take a look into `src/hashcat.c` file:
```C
    while (!hc_feof (&fp))
    {
      const size_t line_len = fgetl (&fp, line_buf, HCBUFSIZ_LARGE);

      if (line_len == 0) continue;

      char *hash_buf = NULL;
      int   hash_len = 0;

      hlfmt_hash (hashcat_ctx, hashlist_format, line_buf, line_len, &hash_buf, &hash_len);

      bool hash_fmt_error = false;

      if (hash_len < 1)     hash_fmt_error = true;
      if (hash_buf == NULL) hash_fmt_error = true;

      if (hash_fmt_error) continue;

      int parser_status = module_ctx->module_hash_decode (hashconfig, digest, salt, esalt, hook_salt, hash_info, hash_buf, hash_len);
```

This code uses module_hash_decode in a loop and doesn't memset `salt`, `esalt`, or `hook_salt`. Therefore, we can conclude that it's safe to use module_hash_decode without memsetting the buffers.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -67,6 +67,7 @@
 ##
 
 - DEScrypt Kernel (1500): Improved performance from 950MH/s to 2200MH/s (RX6900XT) on HIP backend by workaround invalid compile time optimizer
+- Performance of --show feature has been improved by removing unnecessary memset calls in potfile_remove_parse function.
 
 ##
 ## Bugs
--- a/src/potfile.c
+++ b/src/potfile.c
@@ -535,21 +535,6 @@ int potfile_remove_parse (hashcat_ctx_t *hashcat_ctx)
 
     if (line_hash_len == 0) continue;
 
-    if (hash_buf.salt)
-    {
-      memset (hash_buf.salt, 0, sizeof (salt_t));
-    }
-
-    if (hash_buf.esalt)
-    {
-      memset (hash_buf.esalt, 0, hashconfig->esalt_size);
-    }
-
-    if (hash_buf.hook_salt)
-    {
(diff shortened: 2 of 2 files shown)


jsteube:
It's the opposite. I think we need to add a memset() in the section you highlighted, since it currently lacks one. We can't guarantee how module plugins are written or whether plugin authors memset() their decoded data properly.

jsteube:
In "src/potfile.c", the read loop in potfile_remove_parse() reuses a single hash_t for every line of the potfile, which is what the three removed calls are there for:

```
    if (hash_buf.salt)
    {
      memset (hash_buf.salt, 0, sizeof (salt_t));
    }
```

A plugin's module_hash_decode() is not required to fill every byte of the structures it receives, so this clears what the previous line left behind rather than repeating work the decoder already did.

The direction I would take here is the opposite one, because a fourth memset is missing. The buffer hash_buf.digest is allocated once before the loop and never cleared per line, so only the first line starts on zeroes. A module that fills fewer bytes than dgst_size leaves a stale tail for sort_by_hash() to compare, and I will add that memset separately.

For the --show cost, the larger item sits in the decoder itself. Every call to module_hash_decode() opens with memset(&token, 0, sizeof(hc_token_t)), roughly 3.7 KB per potfile line, while most modules declare between 2 and 10 of the 128 tokens that structure holds. Closing this one.

PR #4009, opened 2024-05-26
Create SECURITY.md

Ramdabbe:


Chick3nman:
This appears to be an empty template for a SECURITY.md file, and given the completely empty issue I'm inclined to believe that this PR is spam and not of any legitimate value. I will be closing this for now, but if this is indeed NOT spam, feel free to reopen with justification.

PR #4039, opened 2024-07-04
More correct calculation size for copying full memory buffers

GermanAizek:
@jsteube, carefully review code, I could have made inconspicuous mistake.

### DIFF
--- a/src/modules/module_06211.c
+++ b/src/modules/module_06211.c
@@ -236,7 +236,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   // fake digest
 
-  memcpy (digest, tc->data_buf, TC_DATA_LEN / 4);
+  memcpy (digest, tc->data_buf, sizeof(tc->data_buf));
 
   return (PARSER_OK);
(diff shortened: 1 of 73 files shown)


ventaquil:
How about it?

https://github.com/hashcat/hashcat/pull/4039/files#diff-38a9c12141e4dd7627bb7684fca8f8e319c45a300f9babd390b6548c77a6133aL231

src/modules/module_06211.c
```diff
@@ -236,7 +236,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   // fake digest
 
-  memcpy (digest, tc->data_buf, TC_DATA_LEN / 4);
+  memcpy (digest, tc->data_buf, sizeof(tc->data_buf));
```
ventaquil, line 239: Due to convention there should be space between the function name and the bracket.

```suggestion
  memcpy (digest, tc->data_buf, sizeof (tc->data_buf));
```

Lilblok (commented):
https://www.facebook.com/profile.php?id=100007286296213

PR #4074, opened 2024-09-11
Added m10510 for PDF 1.3 - 1.6 using RC4-40

### LINKED ISSUE #4047: PDF decryption problem
When the hash value starts with `$pdf$1*3*40*,` hashcat cannot decrypt it.

magnumripper:
Fixes #4047 

<s>Note: Self-test missing because I couldn't find nor produce a PDF for making one.</s>The posted test vector is a mockup but the format is tested on a real '`$pdf$1*3*40*`' document and cracked it just fine. 

This is based on 10500 (diffing against its files shows the changes are minimal).

### DIFF
--- a/OpenCL/m10510-pure.cl
+++ b/OpenCL/m10510-pure.cl
@@ -1,0 +1,360 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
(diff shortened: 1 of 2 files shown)


jsteube:
Many thanks!

Issue #4079, opened 2024-09-18
Add support to radmin3_to_hashcat.pl for dumping multiple users


lakiw:
**Describe the feature**
The current version of radmin3_to_hashcat.pl only supports dumping one user from a registry dump. This can cause problems if there are multiple users present. This behavior was seen in the CMIYC2024 competition.

Here is a write-up describing that issue in more detail: [CMIYC 2024: RAdmin3 Challenge
](https://reusablesec.blogspot.com/2024/08/cmiyc-2024-radmin3-challenge.html)

I have a fix for this that I'm still testing, but I wanted to open up an issue so I could tie my Pull Request to it.

**Current behavior**
- The current version of radmin3_to_hashcat.pl struggles if there are additional header information from the registry dumping file and will not identify the registry keys containing the hashes if that header content is present.
- The current version of radmin3_to_hashcat.pl will only dump the first hash it encounters in the registry file, ignoring any other users that appear after it.
- The current version of radmin3_to_hashcat.pl can also struggle with various file encodings that registry files might be dumped as.  

**Expected behavior**
- The script should dump all user hashes found in the registry file
- The script should ignore any header data found in the registry file
- The script should support registry files encoded as ASCII, UTF-8, and UTF-16LE

PR #4080, opened 2024-09-19
Updated radmin3_to_hashcat.pl to support multiple users

lakiw:
This is a proposed fix for Issue #4079

I tried to keep much of the core code that processes the hashes the same. The big difference is I set it to run inside a loop. That way it can process extract the hashes for multiple different users in the same registry dump.

The major change is in how the script reads in the registry file.
- It now starts off by detecting the file encoding to deal with ASCII vs. UTF16-LE.
- It then strips out any header info in the registry dump until it reaches the first radmin3 user registry key
- Next it loops through the the registry and stores each radmin user info in an array.
- Finally the array can be looped through to run the original code on a single radmin user at a time.

I tested this using the radmin.reg files from the 2024 CMIYC Street team challenge files. I also tried it against an ASCII file using the test registry file present in the radmin3_to_hashcat.pl comments.

While I would like to test this on a real radmin3 registry dump, I don't have access to that tool unfortunately.

### DIFF
--- a/tools/radmin3_to_hashcat.pl
+++ b/tools/radmin3_to_hashcat.pl
@@ -48,6 +48,8 @@
 use strict;
 use warnings;
 use utf8;
+use Encode;
+use Encode::Guess;
 
 
 #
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks @lakiw for contribution! I have no database to test with, but I have no doubt your code is clean.

PR #4088, opened 2024-10-06
Added Node Express cookie-session middleware module

d0ge:
Hello!

This PR is for the Node js Express cookie-session middleware (keygrip). Functionally is same as JWT with extra support of the SHA-1 and SHA-224 hmac functions. Since the hash is stored separately from the message, I used the $ char as a separator, although you can use any symbol not included in base64.

 I've never written in Perl before but if its something thats required on my end just let me know and I'll see what I can do!

Cheers

### DIFF
--- a/src/modules/module_90909.c
+++ b/src/modules/module_90909.c
@@ -1,0 +1,471 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#include "common.h"
+#include "types.h"
+#include "modules.h"
+#include "bitops.h"
+#include "convert.h"
+#include "shared.h"
+#include "emu_inc_hash_md5.h"
+#include "memory.h"
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_4_16;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_NETWORK_PROTOCOL;
+static const char *HASH_NAME      = "ExpressJS";
+static const u64   KERN_TYPE      = 90901;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_NOT_ITERATED;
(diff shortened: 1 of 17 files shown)


src/modules/module_90909.c
```diff
@@ -0,0 +1,471 @@
...
+  {
+    kern_type = KERN_TYPE_EXPRESS_HS512;
+  }
+  else
+  {
+    return (PARSER_HASH_LENGTH);
+  }
+
+  return kern_type;
+}
```
ventaquil, line 126: I am not a big fan of dynamic kernel resolution. When it's possible you should create as many modules as many kernels you have IMHO. The biggest advantage is that you will have more accurate benchmark result for each kernel.

matrix:
> Hello!
> 
> This PR is for the Node js Express cookie-session middleware (keygrip). Functionally is same as JWT with extra support of the SHA-1 and SHA-224 hmac functions. Since the hash is stored separately from the message, I used the $ char as a separator, although you can use any symbol not included in base64.
> 
> I've never written in Perl before but if its something thats required on my end just let me know and I'll see what I can do!
> 
> Cheers

Hi @d0ge, since there are so many kernels, it would be good to have the test units in perl. Would you be able to do that?
Thanks

d0ge:
Hi @matrix,
Unfortunately, I don’t have any experience with Perl, so I’d need some help to write the tests. Let me know if someone can assist or if there's another way I can contribute!

matrix:
Hi @d0ge, is also fine in python if that is more convenient for you.

The goal is to see if actually what you have written works.
You can either take a hint from what you see inside hashcat or make up a script from scratch that will do it.

Thanks :)

d0ge:
Hi @matrix, 
May I ask you to take a look at following Perl test. Does it make sense to you?
Cheers

d0ge:
I tested the module using the provided Perl test file `perl test.pl single 90909 | bash`, and everything is working as expected. All tests pass successfully.

matrix:
@d0ge it's not the way, but starting from another point of view: this look similar to hash-mode 16500, have you seen it yet?

d0ge:
@matrix Yes, as I mentioned earlier, the Express cookie-session module is indeed based on hash-mode 16500, so the test follows the same structure.

matrix:
> @matrix Yes, as I mentioned earlier, the Express cookie-session module is indeed based on hash-mode 16500, so the test follows the same structure.

It's duplicated, not based ... You need reuse almost all kernels, except of sha1 and sha224 (that don't even work). All the others kernel must be deleted.

Please provide also test vectors from real world for each algo supported.

Thanks

PR #4093, opened 2024-10-19
Fixed MIT License file

FoxVukOff:


README.md
```diff
@@ -4,7 +4,7 @@
 
 ### License ###
 
-**hashcat** is licensed under the MIT license. Refer to [docs/license.txt](docs/license.txt) for more information.
+**hashcat** is licensed under the MIT license. Refer to License tab for more information.
```
ventaquil, line 7: License tab is GitHub feature, when someone download source code and open the `README.md` there are no tabs.

Chick3nman:
As ventaquil said, this would only apply to those viewing the project from within Github. This change makes no sense if the source is downloaded and viewed locally.

PR #4107, opened 2024-10-31
Add support to character class rules

### LINKED ISSUE #4030: Rule to remove all special characters or numbers
Rules like `@` allow us to remove numbers such as: `@0@1@2@3@4@5@6@7@8@9`, but take up a lot of space in the queue when trying to stack rules, especially when combining it with special characters.

I would like to see a rule added that allows us to purge all characters part of a specific mask.

For example:
`Pd` Purges all ?d
`Ps` Purges all ?s

Ideally we'd even support custom charsets with -1 -2 -3. But I understand this might be a technical limitation.
`P1` Purges all characters in custom charset 1

matrix:
Fix #4030

As per the change made on d3ad0ne.rule, to remove the '?' character we will need to use the new syntax '@??' instead of the old '@?'

It will be possible to use the following classes: '?l' '?u' '?d' '?h' '?H' '?s'
These classes mimic the operation of “Built-in Charsets”, shown in the help.

poc

```
$ cat purge.rule 
@?d
@a
@?s
@c
$ for word in $(echo -n "h2ash?1c4at" | ./hashcat --stdout -r purge.rule | grep ^h); do echo -n $word | md5sum | awk '{print $1}'; done > 0.hashes
$ cat 0.hashes
2e1d68c756cb81ce137d8193d357ff2d
8ae1486f8e5ce87c1980152667399b89
2e5256f63dfce0b7315e2c17bd82ccfb
ce20562042a05aa49da5bc93b945e3da
$ echo -n "h2ash?1c4at" | ./hashcat -m 0 0.hashes -r purge.rule --debug-mode=4 --debug-file=purge.pure.debug -D1 --potfile-disable --quiet
2e1d68c756cb81ce137d8193d357ff2d:hash?cat
8ae1486f8e5ce87c1980152667399b89:h2sh?1c4t
2e5256f63dfce0b7315e2c17bd82ccfb:h2ash1c4at
ce20562042a05aa49da5bc93b945e3da:h2ash?14at
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 0 (MD5)
Hash.Target......: 0.hashes
Time.Started.....: Thu Oct 31 15:36:32 2024 (0 secs)
Time.Estimated...: Thu Oct 31 15:36:32 2024 (0 secs)
Guess.Base.......: Pipe
Guess.Mod........: Rules (purge.rule)
Speed.#1.........:    24845 H/s (0.01ms) @ Accel:512 Loops:4 Thr:1 Vec:4
Recovered........: 4/4 (100.00%) Digests (total), 4/4 (100.00%) Digests (new)
Progress.........: 4
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#1...: Salt:0 Amplifier:0-4 Iteration:0-4
Candidates.#1....: hash?cat -> h2ash?14at
Hardware.Mon.SMC.: Fan0: 20%
Hardware.Mon.#1..: Temp: 43c
$ cat purge.pure.debug
h2ash?1c4at:@?d:hash?cat
h2ash?1c4at:@a:h2sh?1c4t
h2ash?1c4at:@?s:h2ash1c4at
h2ash?1c4at:@c:h2ash?14at
$ echo -n "h2ash?1c4at" | ./hashcat -m 0 0.hashes -r purge.rule --debug-mode=4 --debug-file=purge.opt.debug -D1 --potfile-disable -O --quiet
2e1d68c756cb81ce137d8193d357ff2d:hash?cat
8ae1486f8e5ce87c1980152667399b89:h2sh?1c4t
2e5256f63dfce0b7315e2c17bd82ccfb:h2ash1c4at
ce20562042a05aa49da5bc93b945e3da:h2ash?14at
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 0 (MD5)
Hash.Target......: 0.hashes
Time.Started.....: Thu Oct 31 15:38:21 2024 (0 secs)
Time.Estimated...: Thu Oct 31 15:38:21 2024 (0 secs)
Kernel.Feature...: Optimized Kernel
Guess.Base.......: Pipe
Guess.Mod........: Rules (purge.rule)
Speed.#1.........:    27778 H/s (0.00ms) @ Accel:512 Loops:4 Thr:1 Vec:4
Recovered........: 4/4 (100.00%) Digests (total), 4/4 (100.00%) Digests (new)
Progress.........: 4
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#1...: Salt:0 Amplifier:0-4 Iteration:0-4
Candidates.#1....: hash?cat -> h2ash?14at
Hardware.Mon.SMC.: Fan0: 20%
Hardware.Mon.#1..: Temp: 43c
$ echo -n "h2ash?1c4at" | ./hashcat -m 0 0.hashes -j '@??' -D1 --potfile-disable --quiet
2e5256f63dfce0b7315e2c17bd82ccfb:h2ash1c4at
Session..........: hashcat
Status...........: Exhausted
Hash.Mode........: 0 (MD5)
Hash.Target......: 0.hashes
Time.Started.....: Thu Oct 31 15:39:10 2024 (0 secs)
Time.Estimated...: Thu Oct 31 15:39:10 2024 (0 secs)
Guess.Base.......: Pipe
Speed.#1.........:     4854 H/s (0.01ms) @ Accel:512 Loops:1 Thr:1 Vec:4
Recovered........: 1/4 (25.00%) Digests (total), 1/4 (25.00%) Digests (new)
Progress.........: 1
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#1...: Salt:0 Amplifier:0-1 Iteration:0-1
Candidates.#1....: h2ash1c4at -> h2ash1c4at
Hardware.Mon.SMC.: Fan0: 20%
Hardware.Mon.#1..: Temp: 43c
```

At the moment, I have not yet tried to generate random rules. It will probably be necessary to add a '?' to a possible '@?' generated rule.

Thanks :)


PenguinKeeper7:
Cool commit although in the interest of attempting to maintain some backwards compatibility with existing rulesets, may it be a good idea to interpret `@? T0` as if it were `@?? T0`, where "space" is effectively acting as an alias for `?`?. This case block would just have to be duplicated, but as `case ' '`:
https://github.com/hashcat/hashcat/blob/33092b13c314f60c91335660f290a54fe479352f/src/rp.c#L396-L400
This would of course still break backwards compatibility where there isn't a space, such as `@?T0` but hopefully it may minimise the impact of this commit such as the conflict in the d3ad0ne ruleset, whilst also (theoretically?) not negatively impacting any current rule parsing

matrix:
> Cool commit although in the interest of attempting to maintain some backwards compatibility with existing rulesets, may it be a good idea to interpret `@? T0` as if it were `@?? T0`, where "space" is effectively acting as an alias for `?`?. This case block would just have to be duplicated, but as `case ' '`:
> 
> https://github.com/hashcat/hashcat/blob/33092b13c314f60c91335660f290a54fe479352f/src/rp.c#L396-L400
> 
> 
> This would of course still break backwards compatibility where there isn't a space, such as `@?T0` but hopefully it may minimise the impact of this commit such as the conflict in the d3ad0ne ruleset, whilst also (theoretically?) not negatively impacting any current rule parsing

Hi, isn't it easier to update any rules? this way you avoid misalignments with other tools like john/mdxfind

PenguinKeeper7:
Sure, just thinking that quite a large number of existing rules will have to be updated, common ones including [ORTRTA](https://github.com/NotSoSecure/password_cracking_rules/blob/master/OneRuleToRuleThemAll.rule#L44090)/[ORTRTS](https://github.com/stealthsploit/OneRuleToRuleThemStill/blob/main/OneRuleToRuleThemStill.rule#L36617)/[Dive](https://github.com/hashcat/hashcat/blob/master/rules/dive.rule#L55735) and this behaviour wouldn't change anything, JtR rules would still work with Hashcat, it'd only increase backwards compatibility for rules written for Hashcat.

This is obviously not scientific, but just from my own personal rules folder of 257 rulesets:

Rulesets requiring updates:
Current PR - 49
My idea - 27 (18 if including the end of the line as another alias for `??`)

Rule functions that would need to be changed:
Current PR -  336
My idea - 147 (98 if including the end of the line as another alias for `??`)

I can see both ways though, I'm not disagreeing with you, just wanted to put the thought in your head in case you liked it. Either way, it causes many problems, both for forward and backward compatibility, as well as messaging about the new format of @ being more subtle for certain rules, due to this exception. There's no correct answer, so feel free to reject this opinion of mine :) I'm just throwing my hat in the ring, not trying to complain or block it

(P.S [Dive](https://github.com/hashcat/hashcat/blob/master/rules/dive.rule#L55735) as well as d3ad0ne would also have to be updated as part of this PR)

matrix:
But the dive rule is "Replace" not "Purge", or am I wrong?
What if after "@?" there is no space?
I could try to make it work '@?' but I need to see if it can cause any problems

ghost:
I hate it when people break backward compatibility. There's no need to use `@`. There are other unused characters, `~`, for example.

matrix:
@PenguinKeeper7 check it

PenguinKeeper7:
Haven't tested it yet but looks good

solardiz:
> common ones including [ORTRTA](https://github.com/NotSoSecure/password_cracking_rules/blob/master/OneRuleToRuleThemAll.rule#L44090)/[ORTRTS](https://github.com/stealthsploit/OneRuleToRuleThemStill/blob/main/OneRuleToRuleThemStill.rule#L36617)

For compatibility with these two, can also add treating of `'` the same as space. And we'll know not to use these two characters to identify any class in the future (not that we were going to anyway).

matrix:
> > common ones including [ORTRTA](https://github.com/NotSoSecure/password_cracking_rules/blob/master/OneRuleToRuleThemAll.rule#L44090)/[ORTRTS](https://github.com/stealthsploit/OneRuleToRuleThemStill/blob/main/OneRuleToRuleThemStill.rule#L36617)
> 
> For compatibility with these two, can also add treating of `'` the same as space. And we'll know not to use these two characters to identify any class in the future (not that we were going to anyway).

Hashcat use `'` for "Truncate word at position N". I don't think it's appropriate to treat it as for space.
if atom approves the code I wrote then I will do a PR where needed to directly change the syntax.

(edit: and John for "truncate the word at length N")

solardiz:
> Hashcat use `'` for "Truncate word at position N". I don't think it's appropriate to treat it as for space.

Yes (and JtR treats it that way too), so I didn't mean treating it as space for all purposes - but only for purposes of the `@` command. It would then be processed as a command on its own. So one character in a rule line would play two roles. I do admit this is more of a hack than with space, which is a no-op, and it may also be harder/dirtier to implement (need lookahead or `ungetc`-alike). And maybe this sort of hack would be more worth it if extended even further, to treat all non-alphanumeric characters like that (limiting us to usage of only letters for future classes, as digits are for user-defined classes in JtR). I'm really not sure. And then it's unclear yet to what extent would this drastic move address the backwards compatibility problem e.g. for @PenguinKeeper7's many rulesets.

matrix:
Doing the math with the missing rules to support also classes usage (?C) as crack/john does, if we add one more character to identify this (maybe) new rule then we need 7 more. As soon as we agree internally then we will figure out which of the two paths to choose.

@PenguinKeeper7 do you have any other use cases?

solardiz:
> Doing the math with the missing rules to support also classes usage (?C) as crack/john does, if we add one more character to identify this (maybe) new rule then we need 7 more.

Per my math, there are 9 class-supporting commands in John, so I guess it should be 8 more? Anyway, a possible syntax could be to prefix those rules with `?` (I think not taken as a rule command character anywhere yet?), so e.g. `?@a` to purge all letters. Then you need just one new command character `?` for all 9 commands. It won't be compatible with John's current rules nor conflicting with them, so we'll have to separately implement it in John for future hashcat rules compatibility.

matrix:
> Per my math, there are 9 class-supporting commands in John, so I guess it should be 8 more?

I count 8, what would be 9?

Screen Shot 2024-11-02 at 15 30 29

&nbsp;
> Anyway, a possible syntax could be to prefix those rules with ?

I would like to avoid using ?, but as previously suggested I could use ~
I don't see it used on john either.
I'll work on it a bit, see how it goes :)

solardiz:
> I count 8, what would be 9?

You seem to be looking at the webpage, which shows documentation from "core". We have extra commands in "jumbo":

```
$ fgrep '?C' doc/RULES
s?CY	replace all characters of class C in the word with Y
@?C	purge all characters of class C from the word
!?C	reject the word if it contains a character in class C
/?C	reject the word unless it contains a character in class C
=N?C	reject the word unless character in position N is in class C
(?C	reject the word unless its first character is in class C
)?C	reject the word unless its last character is in class C
%N?C	reject the word unless it contains at least N characters of class C
e?C	title case, with chars from class C as separation characters.
```

> I would like to avoid using ?, but as previously suggested I could use ~

Any specific reason not to use `?`? Either works.

matrix:
> Any specific reason not to use `?`? Either works.

Nothing special, it obviously works with `?` as well as with any other unused character. 
Since it is generically used often, even inside the syntax of these new rules, I would like to avoid it.

solardiz:
Another idea is to enable the character class support with a line like:

```
!! hashcat logic 2
```

which I guess would have effect until the end of rules file, or maybe until overridden with:

```
!! hashcat logic 1
```

Then the same Crack-derived syntax that John uses could also be used by hashcat in that mode, without compatibility issues for old hashcat rulesets.

This is similar to what we do in John, where we currently have these directives:

```
[List.Rules:d3ad0ne]
!! hashcat logic ON
.include <rules/d3ad0ne.rule>
!! hashcat logic OFF
```

so we'd add support for this/these additional hashcat logic version overrides.

A drawback of this approach is that it'd be difficult to arbitrarily mix lines from different-version rulesets. We already have this drawback when using hashcat rulesets in John along with our John-native rulesets - we cannot _easily_ optimize line order across such mix of rulesets to produce a single optimized ruleset (we can and already did optimize the order of different included rulesets, including mixing hashcat logic with John-native ones). With some more advanced processing we could, but the result would be cluttered with many of those logic switch lines.

Yet another idea is to have similar version overrides via a rule command affecting only the rest of line, e.g. `~2` for new hashcat logic, `~1` for old/current hashcat logic, and `~0` for John logic (which I doubt you'd implement, but in John we probably would).

Just thinking out loud, not advocating for any of these in particular.

matrix:
> ```
> e?C	title case, with chars from class C as separation characters.
> ```

This rule is missing. I didn't understand how it works (I haven't tried it yet) and if it is compatible with 'e' of hashcat.
(edit: ok, tried it and it seems compatible with hashcat 'e')

matrix:
@solardiz we wrote at the same time :D
the new version I have committed now does not create backwards compatibility issues

solardiz:
> I didn't understand how it works (I haven't tried it yet) and if it is compatible with 'e' of hashcat

It is supposed to be mostly compatible, except that it supports character classes. As I recall, we added it precisely for hashcat compatibility.

solardiz:
> the new version I have committed now does not create backwards compatibility issues

Skimming the code, the syntax appears to be `~` as prefix to Crack/John-style commands, so e.g. `~@?a`. The `?` is then redundant (and I had thought you'd omit it), but is OK to have/require as it keeps those commands more similar to what's in John (it'd even ease our implementation of this hashcat compatibility feature, where `~` would simply set a flag, so I appreciate your doing it in this way).

matrix:
> > the new version I have committed now does not create backwards compatibility issues
> 
> Skimming the code, the syntax appears to be `~` as prefix to Crack/John-style commands, so e.g. `~@?a`. The `?` is then redundant (and I had thought you'd omit it), but is OK to have/require as it keeps those commands more similar to what's in John (it'd even ease our implementation of this hashcat compatibility feature, where `~` would simply set a flag, so I appreciate your doing it in this way).

Yes, I used `~` as a prefix for all rules that use character classes, and then used the same syntax to express character classes (it seemed more intuitive to do it this way). I will soon add the 'e' rule that uses the character classes.

matrix:
Added but something doesn't work well with actual OpenCL code.

Example:

```
$ echo -n "has?hcat" | ./hashcat --stdout -j '~e?h'
HaS?hcaT
$ cat tsep.rule 
~e?h
$ echo -n "has?hcat" | ./hashcat --stdout -r tsep.rule 
HaS?hcAT

```

jsteube:
Thanks for the PR, it's really nice.

I see there has been some progress made between the initial PR and the latest version.
Could you please update the initial PR text so that users understand how to use the changes? 

Also, please fix the merge conflict, is just a single line.

I would prefer not to change all existing kernels, since that could break private kernel repositories (kernels that are not submitted to GitHub). You can stick with inc_rp_common.cl and just include it from the specific inc_rp* files that need it.

Did you test if it also works with the host rule engine when using --stdout?

matrix:
> Thanks for the PR, it's really nice.
> 
> I see there has been some progress made between the initial PR and the latest version. Could you please update the initial PR text so that users understand how to use the changes?

I'll do it

> Also, please fix the merge conflict, is just a single line.

done

> I would prefer not to change all existing kernels, since that could break private kernel repositories (kernels that are not submitted to GitHub). You can stick with inc_rp_common.cl and just include it from the specific inc_rp* files that need it.

I try to evaluate an alternative, although I like it the way it is already. We know that with time things change, a minimum of adaptation is necessary ;)

> Did you test if it also works with the host rule engine when using --stdout?

the last comment I had left is really about stdout. We also discussed this in private, and you had told me that the behavior I described in the comment is normal. Do you confirm that?

jsteube:
> I try to evaluate an alternative, although I like it the way it is already. We know that with time things change, a minimum of adaptation is necessary ;)

Whatever alternative is ok, just make sure to not touch existing OpenCL/m* kernel because that would also impact private kernel repositories.

> the last comment I had left is really about stdout. We also discussed this in private, and you had told me that the behavior I described in the comment is normal. Do you confirm that?

Confirmed, but not correct. It was not related to ~ character, but to title rule 'e'.

What I actually meant is if you had tested the change in stdout mode, and if the output is in sync with what is happening on the device? Can you provide some commands to reproduce locally? Maybe some cracked hash using a character class rule, and then another command in stdout, resulting in the same plaintext.

matrix:
> Whatever alternative is ok, just make sure to not touch existing OpenCL/m* kernel because that would also impact private kernel repositories.
> 
> >

this is done

matrix:
> What I actually meant is if you had tested the change in stdout mode, and if the output is in sync with what is happening on the device? Can you provide some commands to reproduce locally? Maybe some cracked hash using a character class rule, and then another command in stdout, resulting in the same plaintext.

This is the output of the bash script created for testing characters rules

```
$ ./chrules.sh 
> Password for Pure tests: !"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\]^_`abcdefghijklmnopqrstuvwxyz{|}~
> Password for Optimized tests: 756nZ-Y.Z)%N>4j[
>> Testing if hashcat works without rules ...
> Pure Initial test success!
> Optimized Initial test success!

[0] >> Testing character class: ?l
[0] > Creating rule file ...
[0] > Rule file content: 
~@?l
~@??
~s?lX
~e?l

[0] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[0] > Cracking md5 hash using current character class rule ...
[0] > Pure Test success!
[0] > Optimized Test success!

[1] >> Testing character class: ?u
[1] > Creating rule file ...
[1] > Rule file content: 
~@?u
~@??
~s?uX
~e?u

[1] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[1] > Cracking md5 hash using current character class rule ...
[1] > Pure Test success!
[1] > Optimized Test success!

[2] >> Testing character class: ?d
[2] > Creating rule file ...
[2] > Rule file content: 
~@?d
~@??
~s?dX
~e?d

[2] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[2] > Cracking md5 hash using current character class rule ...
[2] > Pure Test success!
[2] > Optimized Test success!

[3] >> Testing character class: ?h
[3] > Creating rule file ...
[3] > Rule file content: 
~@?h
~@??
~s?hX
~e?h

[3] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[3] > Cracking md5 hash using current character class rule ...
[3] > Pure Test success!
[3] > Optimized Test success!

[4] >> Testing character class: ?H
[4] > Creating rule file ...
[4] > Rule file content: 
~@?H
~@??
~s?HX
~e?H

[4] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[4] > Cracking md5 hash using current character class rule ...
[4] > Pure Test success!
[4] > Optimized Test success!

[5] >> Testing character class: ?s
[5] > Creating rule file ...
[5] > Rule file content: 
~@?s
~@??
~s?sX
~e?s

[5] > Generating md5 hash with stdout mode for Pure and Optimized tests ...
[5] > Cracking md5 hash using current character class rule ...
[5] > Pure Test success!
[5] > Optimized Test success!

>> Done with 6 tests, errors detected: 0
```

The script is the following:

```
#!/bin/bash

HC_OPTS_START="--quiet --potfile-disable -d 1"
HC_OPTS="--quiet --potfile-disable -d 1 --debug-mode=4 --debug-file=purge.debug"
CH_CLASSES="?l ?u ?d ?h ?H ?s"

generate_strong_printable_string() {
  local lower upper digit symbol rest all final

  # Forzare locale per tr
  lower=$(LC_CTYPE=C tr -dc 'a-z' < /dev/urandom | head -c4)
  upper=$(LC_CTYPE=C tr -dc 'A-Z' < /dev/urandom | head -c4)
  digit=$(LC_CTYPE=C tr -dc '0-9' < /dev/urandom | head -c4)
  symbol=$(LC_CTYPE=C tr -dc '!"#$%&'\''()*+,-./:;<=>?@[\]^_`{|}~' < /dev/urandom | head -c4)
  rest=$(LC_CTYPE=C tr -dc '[:alnum:]!"#$%&'\''()*+,-./:;<=>?@[\]^_`{|}~' < /dev/urandom | head -c15)

  all="$lower$upper$digit$symbol$rest"

  # Mischia i caratteri senza shuf
  final=$(echo "$all" | fold -w1 | awk 'BEGIN {srand()} {a[NR]=$0} END {for (i=1;i<=NR;i++) {j=int(rand()*NR)+1; print a[j]; a[j]=a[NR]; NR--}}' | tr -d '\n')

  echo "$final"
}

pw_pure=$(printf "%b" "$(printf '\\x%x' {33..126})")
pw_opti=$(generate_strong_printable_string)

echo -n "> Password for Pure tests: "
echo $pw_pure

echo -n "> Password for Optimized tests: "
echo $pw_opti

echo ">> Testing if hashcat works without rules ..."
echo -n ${pw_pure} | md5sum | awk '{print $1}' > tmp_md5_hash_pure
echo -n ${pw_pure} | ./hashcat ${HC_OPTS_START} -m 0 tmp_md5_hash_pure &>/dev/null
if [ $? -ne 0 ]; then
  echo "! Pure Initial test failed ..."
else
  echo "> Pure Initial test success!"
fi

echo -n ${pw_opti} | md5sum | awk '{print $1}' > tmp_md5_hash_opti
echo -n ${pw_opti} | ./hashcat ${HC_OPTS_START} -m 0 -O tmp_md5_hash_opti &>/dev/null
if [ $? -ne 0 ]; then
  echo "! Optimized Initial test failed ..."
else
  echo "> Optimized Initial test success!"
fi

echo

err=0

idx=0
for class in ${CH_CLASSES}; do
  echo "[$idx] >> Testing character class: ${class}"

  echo "[$idx] > Creating rule file ..."
  echo "~@${class}" > tmp_rule
  echo "~@??" >> tmp_rule
  echo "~s${class}X" >> tmp_rule
  #echo "~=N${class}" >> tmp_rule
  #echo "~(${class}" >> tmp_rule
  #echo "~)${class}" >> tmp_rule
  #echo "~%N${class}" >> tmp_rule
  echo "~e${class}" >> tmp_rule
  echo "[$idx] > Rule file content: "
  cat tmp_rule
  echo

  echo "[$idx] > Generating md5 hash with stdout mode for Pure and Optimized tests ..."
  for word in $(echo -n ${pw_pure} | ./hashcat ${HC_OPTS} --stdout -r tmp_rule); do echo -n $word | md5sum | awk '{print $1}'; done > tmp_md5_hash_pure
  for word in $(echo -n ${pw_opti} | ./hashcat ${HC_OPTS} --stdout -r tmp_rule); do echo -n $word | md5sum | awk '{print $1}'; done > tmp_md5_hash_opti
  echo "[$idx] > Cracking md5 hash using current character class rule ..."
  echo -n ${pw_pure} | ./hashcat ${HC_OPTS} -m 0 tmp_md5_hash_pure -r tmp_rule &>/dev/null
  if [ $? -ne 0 ]; then
    echo "[$idx] ! Pure Test failed ..."
    (( err++ ))
  else
    echo "[$idx] > Pure Test success!"
  fi

  echo -n ${pw_opti} | ./hashcat ${HC_OPTS} -m 0 tmp_md5_hash_opti -r tmp_rule -O &>/dev/null
  if [ $? -ne 0 ]; then
    echo "[$idx] ! Optimized Test failed ..."
    (( err++ ))
  else
    echo "[$idx] > Optimized Test success!"
  fi

  (( idx++ ))
  echo
done

echo ">> Done with $idx tests, errors detected: $err"
```

Let me know!

jsteube:
Looks really good. I didn't check, but did you add support to random rule generator -g ? Is not so important, but could be nice.

matrix:
> Looks really good. I didn't check, but did you add support to random rule generator -g ? Is not so important, but could be nice.

I think no, I remember it wasn't easy/immediate

Issue #4125, opened 2024-11-29
RAM Usage 100% - Apple Silicon Unified Memory


codecuriously:
**GitHub is for bugs and features - not for support**
For support, please use the hashcat forums https://hashcat.net/forum/

**Check the FAQ**
Some items that might appear to be issues are not issues. Please review the hashcat FAQ https://hashcat.net/wiki/doku.php?id=frequently_asked_questions before submitting a bug report.

**Describe the bug**
When running "hashcat -II" RAM usage spikes to 100% and the entire OS freezes and/or crashes. Issue appears to be similar to that fixed for other systems under https://github.com/hashcat/hashcat/issues/3426. Appears to be an issue with "unified memory" systems. Using Asahi Linux on Apple Silicon M2 MAX. Would really appreciate any help getting this fixed.

**To Reproduce**
Under Asahi Linux on an Apple Silicon processor run "hashcat -II"

**Expected behavior**
RAM not at 100% and rather handled the same as other "unified memory" systems.

**Hardware/Compute device (please complete the following information):**
- Compute device name: Apple M2 Max (G14C B1)
- OpenCL/CUDA driver name: Unsure sorry. Posted "hashcat -II" below.
- OpenCL/CUDA driver version: 24.3.0-asahi20241111 
- OpenCL/CUDA driver source: Unsure sorry. Posted "hashcat -II" below.

**Hashcat version (please complete the following information):**
 - OS: Linux
 - Distribution: Fedora Linux Asahi Remix 40 (KDE Plasma)
 - Version: 6.2.6 
 
**Diagnostic output compute devices:**

```
For NV: Post nvidia-smi output. This tool also exist on Windows
For AMD ROCm: Post rocm-smi and rocminfo output
Not sure what to provide for Apple Silicon. Happy to provide further detail.
```

```
Post clinfo output
Number of platforms                               3
  Platform Name                                   rusticl
  Platform Vendor                                 Mesa/X.org
  Platform Version                                OpenCL 3.0 
  Platform Profile                                FULL_PROFILE
  Platform Extensions                             cl_khr_byte_addressable_store cl_khr_create_command_queue cl_khr_expect_assume cl_khr_extended_versioning cl_khr_icd cl_khr_il_program cl_khr_spirv_no_integer_wrap_decoration cl_khr_suggested_local_work_size
  Platform Extensions with Version                cl_khr_byte_addressable_store                                    0x400000 (1.0.0)
[...]

PR #4126, opened 2024-12-01
Fix RAM usage bug for Linux Apple Silicon #4125

codecuriously:
Fixes bug #4125 where RAM usage spiked to 100% causing Apple Silicon systems under Linux to crash. Same fix as that merged under #3910.


jsteube:
We need to fix using a different strategy. Using brute force to find free memory is causing too many crashes on too many OpenCL platforms. I'll do this in a separate commit.

PR #4131, opened 2024-12-15
Devin/1734109436 add enc8 hash support

Explorer1092:


.vscode/settings.json
matrix, line 1: please remove files that are not needed by hashcat

OpenCL/m33000_a3-pure.cl
```diff
@@ -0,0 +1,216 @@
...
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_md5.cl)
+#endif
+
+/**
```
matrix, line 17: if you want to leave comments, please write them in English. Otherwise remove them

OpenCL/m33000_a3-pure.cl
```diff
@@ -0,0 +1,216 @@
...
+
+  u32x s[64] = { 0 };
+
+  for (u32 i = 0, idx = 0; i < salt_len; i += 4, idx += 1)
+  {
+    s[idx] = salt_bufs[SALT_POS_HOST].salt_buf[idx];
+  }
+
+  // 打印第一个密码尝试的信息
+  if (gid == 0 && lid == 0)
```
matrix, line 65: please remove debug sections

src/test_enc8.txt
matrix, line 1: please remove files that are not needed by hashcat

OpenCL/m33000_a3-pure.cl
matrix, line 1: a0 and a1 kernels are missing, could you add them?

matrix (changes requested):
Would you be able to add also the test unit in perl?

jsteube:
The number this takes is spoken for. Mode 33000 on master is md5($salt1.$pass.$salt2), so this cannot build as it stands.

The rest would need rebuilding around that anyway. One a3 kernel ships where a mode running inside the kernel needs a0, a1, a3 and a4, there is no test module, and the comment beside ST_HASH marks it as an example rather than a value the mode produces. ".vscode/settings.json" and "src/test_enc8.txt" are in the diff as well.

So I am closing this one. If enc8 is worth having, a fresh PR with a free number, the full kernel set and a test module is the way to it.

Issue #3515, opened 2022-11-27
Add color for status items


pbies:
Add color for status items in console like:
- Time Estimated (the days/hours/minutes/seconds) - cyan
- Speed - yellow
- Recovered - green
- Progress percent - red

**Current behavior**
No color used in console.

**Expected behavior**
Option for command line version - both Linux and Windows - to color some of the status items.

**Hashcat version (please complete the following information):**
 - OS: Linux+Windows
 - Version: 6.2.6

**Additional context**
This should be optional, but when used - very helpful to quickly put eyes on the interesting info.

RLStrug:
This would be nice
Maybe this feature could come up with a command line option like grep, ls, etc.  (`--color=(yes|no|auto)`)

Xanadrel:
Partial duplicate of #3158

Issue #3158, opened 2022-02-07
add color found hashes


mobola1edu:
example:
![hashcat](https://user-images.githubusercontent.com/80366344/152785410-ecee408a-0f17-44fa-be5d-c76000b14d97.png)

kost123:
Can we also get some fireworks and some confetti aswell please.

Chick3nman:
This is unlikely to be added due to how coloring works currently using Log Levels and how it differs platform to platform. Was there a specific reason you wanted this?

mobola1edu:
Yes. It will take me less time to realize that I have found hashes than I would by looking at Recoreved or opening a potfile

roycewilliams:
I see the appeal somewhat, but the additional complexity may not be worth the benefit. The complexity of adding this on a per-hashtype basis is higher than it might appear, I'm told.

banderlog:
fastest way is to modify `main_cracker_hash_cracked()` [in main.c](https://github.com/hashcat/hashcat/blob/6716447dfce969ddde42a9abe0681500bee0df48/src/main.c#L341) and recompile hc from source

For linux/bash:
```diff
static void main_cracker_hash_cracked (MAYBE_UNUSED hashcat_ctx_t *hashcat_ctx, MAYBE_UNUSED const void *buf, MAYBE_UNUSED const size_t len)
{
  outfile_ctx_t         *outfile_ctx        = hashcat_ctx->outfile_ctx;
  status_ctx_t          *status_ctx         = hashcat_ctx->status_ctx;
  user_options_t        *user_options       = hashcat_ctx->user_options;
  user_options_extra_t  *user_options_extra = hashcat_ctx->user_options_extra;

  if (outfile_ctx->fp.pfp != NULL) return; // cracked hash was not written to an outfile

  if ((user_options_extra->wordlist_mode == WL_MODE_FILE) || (user_options_extra->wordlist_mode == WL_MODE_MASK))
  {
    if (outfile_ctx->filename == NULL) if (user_options->quiet == false) clear_prompt (hashcat_ctx);
  }

+ // make cracked hash output in green
+ fputs("\033[32m", stdout);
  fwrite (buf, len,          1, stdout);
+ fwrite("\033[0m", 4, 1, stdout);
  fwrite (EOL, strlen (EOL), 1, stdout);

  if ((user_options_extra->wordlist_mode == WL_MODE_FILE) || (user_options_extra->wordlist_mode == WL_MODE_MASK))
  {
[...]

PR #4136, opened 2024-12-24
Colorize cracked hashes

banderlog:
Make cracked hashes cyan in stdout, if program compiled on non _WIN os

partially solves:
- https://github.com/hashcat/hashcat/issues/3158 
- https://github.com/hashcat/hashcat/issues/3515


ventaquil:
Shouldn't support `--color always|auto|never`?

banderlog:
idnk, should it? it only colors cracked hashes because they are simply buffer output, without any logging stuff.

ventaquil:
From my point of view it should support option to decide whether I want or not to colorize output.

I use Hashcat as a library in my program and extra characters could break the whole application, needs manually correction and trimming color escape sequences. A better solution is to provide the `color` option in the `user_options_t` struct with `always|auto|never` (where `never` is `0` so it will be set by default). It won't damage any external application by far.

banderlog:
I added `--color` option, which is boolean and currently available only on non WIN machines:

```bash
# will provide cracked hashhes in cyan color
./hashcat -a3 -w3 -m0  698d51a19d8a121ce581499d7b701668 ?d?d?d --potfile-disable --color

# usual behavior
./hashcat -a3 -w3 -m0  698d51a19d8a121ce581499d7b701668 ?d?d?d --potfile-disable
```

ventaquil:
Much better imho.

magnumripper:
FWIW: In John the Ripper we do it like this:
```
    if (user_options->color == true && isatty(fileno(stdout))) fputs("\033[0;36m", stdout);
```
This means color codes will never be printed to files, only to terminal. A similar test can obviously be made for `stderr` if needed.

Also, we use the exact same code and ANSI sequences for Windows, not sure why you disabled that?

banderlog:
> This means color codes will never be printed to files, only to terminal.

O, thank you, I'll add this condition

> Also, we use the exact same code and ANSI sequences for Windows, not sure why you disabled that?

Mainly, because I cannot test it -- no working windows pc/laptop with opencl device under my reach

banderlog:
At last, I've checked it with WSL build - everything works.
With @magnumripper snippet it colorizes only for stdout, no garbage in redirected output.
Colorization still should be enabled explicitly via `--color` option

magnumripper:
> A better solution is to provide the `color` option in the `user_options_t` struct with `always|auto|never` (where `never` is `0` so it will be set by default). It won't damage any external application by far.

Some warnings are already printed in color. Maybe this option should affect that too? If not (and not planned), the new option should probably be called `--color-cracked`.

banderlog:
OK, sounds legit. Now `--color` changed to `--color-cracked`
@magnumripper @ventaquil

magnumripper:
Meanwhile someone else took 0xff56
```
src/user_options.c:561:12: error: duplicate case value: 'IDX_BENCHMARK_MAX' and 'IDX_COLOR' both equal '65366'
  561 |       case IDX_COLOR:                     user_options->color_cracked             = true;                            break;
      |            ^
src/user_options.c:437:12: note: previous case defined here
  437 |       case IDX_BENCHMARK_MAX:             user_options->benchmark_max             = hc_strtoul (optarg, NULL, 10);   break;
      |            ^
1 error generated.
```
```
$ git grep "0xff56\b"
include/types.h:  IDX_BENCHMARK_MAX             = 0xff56,
include/types.h:  IDX_COLOR                     = 0xff56,
```
Easy fix for you though. Not sure how we're suppose to pick them but right now 0xff58 is available.

banderlog:
> Meanwhile someone else took 0xff56

I shoul'd have anticipate that, considering that the first pull request was almost 5 months ago.
Fixed, thanks for id value.

> Not sure how we're suppose to pick them

Last time I just took the next available one

magnumripper:
> At last, I've checked it with WSL build - everything works.
> With @magnumripper snippet it colorizes only for stdout, no garbage in redirected output.

Turns out hc have a function in terminal.c that maybe should be used instead:
```
bool is_stdout_terminal (void)
{
#if defined(_WIN)
  return _isatty(_fileno (stdout));
#else
  return isatty (fileno (stdout));
#endif
}
```
You can obviously use it after including terminal.h. Not sure why it's needed but you may want to use that instead.

Sorry for suggesting so many changes 🤣 

EDIT: Checked for _isatty in JtR and it's indeed used, but not in all places. Maybe I should fix that. Perhaps it's a compatibility thing for older Windows versions?

magnumripper:
https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-isatty%3Fview%3Dmsvc-170%23:~:text%3DThe%2520Microsoft%252Dimplemented%2520POSIX%2520function,the%2520function%2520is%2520still%2520supported.&ved=2ahUKEwi7j5K1kcONAxUqBxAIHZHxHO0QFnoECCIQAw&usg=AOvVaw11seaeMpI_0wfc2294KNNH

"_The Microsoft-implemented POSIX function name isatty is a deprecated alias for the_ __isatty_ _function. By default, it generates Compiler warning (level 3) C4996. The name is deprecated because it doesn't follow the Standard C rules for implementation-specific names. However, the function is still supported._"

Why the heck are they deprecating a POSIX function? Maybe because Windows. You can't live _with_ it, and you can't _live_ with it.

banderlog:
> Why the heck are they deprecating a POSIX function? Maybe because Windows.

Strange shit happens in the Windows development all the time `¯\_(ツ)_/¯`

> Not sure why it's needed but you may want to use that instead

I am also do not know why they did it, but for the sake of unification, let's use it

banderlog (commented):
thx

jsteube:
Could you please adapt the naming convention? In hashcat's it's typical the same as the parameter. So "COLOR" would be "COLOR_CRACKED", and "IDX_COLOR" is "IDX_COLOR_CRACKED".

banderlog:
@jsteube done

jsteube:
Very cool, thanks for the contribution!

PR #4139, opened 2025-01-06
Adding keybag2hashcat.py script

zyronix:
For iPhone's where you are able to decrypt the keybag file using for example:
https://github.com/nt0xa/systembag.kb/blob/master/decrypt_systembag.py

You can use this tool to convert the decrypted keybag (tested with keybag version 3 and 4) into a 26500 hashcat hash. The script can be tested with https://github.com/nt0xa/systembag.kb/blob/master/example/keybag

### DIFF
--- a/tools/keybag2hashcat.py
+++ b/tools/keybag2hashcat.py
@@ -1,0 +1,171 @@
+import argparse
+import logging
+import sys
+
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks :)

PR #4142, opened 2025-01-07
fix metamask module 26600

learnerLj:
Description of Changes

This PR is inspired by [#3952](https://github.com/hashcat/hashcat/pull/3952), but I found it could not run correctly in my specific case. The changes in this PR have passed local testing.

Changes Made:
1. Updated the ST_HASH and the conversion tool.
2. Updated the test data.

Remaining Tasks:
1. Update the metamask type in module 26600 and the corresponding Perl test script.
2. Address the self-test issue to ensure the remaining parts pass.

Note:

The current fix still requires disabling the self-test to pass the remaining issues.


.gitignore
```diff
@@ -1,6 +1,8 @@
 *.exe
 *.bin
 *.app
+.idea
+.DS_Store
```
ventaquil: Those lines shouldn't be pushed.

Editor related lines should be configured on your local machine in global git ignore. Check this [Stack Overflow answer](https://stackoverflow.com/a/7335487).

ventaquil (changes requested):

.gitignore
learnerLj: Thank you for pointing this out. I adapt your advice.

its5Q:
Hi, what exactly didn't work in #3952 for you?
I've tested it with hashes with and without iteration count supplied and it worked for me, I've used the updated conversion script and a new test vault in my PR.

learnerLj:
> Hi, what exactly didn't work in #3952 for you? I've tested it with hashes with and without iteration count supplied and it worked for me, I've used the updated conversion script and a new test vault in my PR.

Oh, I miss up the position between the iterations and salt.  Your PR has higher quality.  I have another question, should we update the perl script in tests and the help info?

jsteube:
@learnerLj Could you please join this discussion? https://github.com/hashcat/hashcat/pull/3952#issuecomment-3058357847

jsteube:
Closing as discussed in https://github.com/hashcat/hashcat/pull/3952

Thanks for your contribution!

PR #4157, opened 2025-02-19
mod 8300 (NSEC3): support salt length up to what the standard allows

Dook97:
Currently there is an arbitrary limit of 32 chars on NSEC3 salt length, while the standard theoretically allows up to 510 chars (255 octets represented in hex). This PR fixes that.

### DIFF
--- a/src/modules/module_08300.c
+++ b/src/modules/module_08300.c
@@ -64,7 +64,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[2]     = ':';
   token.len_min[2] = 0;
-  token.len_max[2] = 32;
+  token.len_max[2] = 510; // max salt length is 255B, human readable repr is hex so 2 chars per byte
   token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH;
 
   token.sep[3]     = ':';
--- a/src/modules/module_08300.c
+++ b/src/modules/module_08300.c
@@ -19,7 +19,8 @@ static const u32   DGST_SIZE      = DGST_SIZE_4_5;
 static const u32   HASH_CATEGORY  = HASH_CATEGORY_NETWORK_SERVER;
 static const char *HASH_NAME      = "DNSSEC (NSEC3)";
 static const u64   KERN_TYPE      = 8300;
-static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE;
+static const u32   OPTI_TYPE      = OPTI_TYPE_ZERO_BYTE
+                                  | OPTI_TYPE_REGISTER_LIMIT;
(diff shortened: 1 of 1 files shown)


matrix:
please update also test unit and check if your changes works

jsteube:
token[2] decode into salt_buf[], and in optimized kernels we have a 32 byte limit:

```
  u32 salt_buf0[4];
  u32 salt_buf1[4];

  salt_buf0[0] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 0]);
  salt_buf0[1] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 1]);
  salt_buf0[2] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 2]);
  salt_buf0[3] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 3]);
  salt_buf1[0] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 4]);
  salt_buf1[1] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 5]);
  salt_buf1[2] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 6]);
  salt_buf1[3] = hc_swap32_S (salt_bufs[SALT_POS_HOST].salt_buf[ 7]);

  const u32 salt_len = salt_bufs[SALT_POS_HOST].salt_len;
```

From theoretical perspective the change will produce false negatives if -O is used.  But in pure kernel mode it should work.

Also I just realized that 510 is not registered as hex, so the decoder doesn't know to decode it from hex and so it will try to copy 510 byte into a 256 byte buffer and overflow.

This PR needs updates. Easiest is probably to add a check if length >= 32 in optimized mode, then reject, and if set the max to instead of 510, or add coding using hex_decode() to decode first.

Dook97:
Thanks for the insight. This was just a quick hack I did because a dataset I was cracking came with 20B salts and this seemed to work in my case.

I'm not at all familiar with hashcat's codebase or inner datastructures, so it might take a long time for me to get around to fixing this. The easiest improvement right now might be to store the hash in binary instead of human readable hex representation - that's doubling max salt len for free.

jsteube:
It will not change anything, the maximum salt length is 32 byte, no matter how the input data is encoded.

But there's actually no need to change the kernel. It's like I said, the pure kernel does support 256 byte. The only thing that needs to be decoder. When it detect hashcat runs the hash-mode in optimized mode, then it needs to reject the hash.

jsteube:
It's kind of complex algorithm that doesn't really fit into optimized mode with only 54 bytes of buffer space for a single transform. For serious cracking of NSEC, make sure to stick to pure mode, and in that case domains and salt both can have length up to 256. I'll also push an updated unit test later on.

Thanks for the contribution!

PR #4158, opened 2025-02-21
Add apfs2hashcat.py

holly-o:
Created apfs2hashcat.py for extracting hashes from APFS volumes. Existing open-source projects are insufficient. This script will extract hashes for multiple users and warn if the hash is for the iCloud recovery key rather than a local user.


jsteube:
Thanks Holly, Top!

PR #4164, opened 2025-02-27
Add module 30601 for bcrypt-sha256 v2

twier:
Module 30601 implements hashing according to version 2 of passlib's bcrypt-sha256. Code is based mostly on module 30600 and also on modules 3200 and 28300.

The existing module 30600 implements a different variant of bcrypt-sha256 which seems to be incompatible with passlib's hashes, as it foregoes the base64 encoding of the intermediate hash before handing it to bcrypt.

The implemented version 2 of the algorithm works as follows:

1. The password is hashed using HMAC-SHA2-256 with the base64-encoded salt as key (bcrypt base64 alphabet).
2. The 32 byte digest is base64-encoded (RFC4648 base64 alphabet) and `=`-padded to 44 bytes.
3. The result is encrypted via bcrypt with the original binary salt.
The resulting final digest is again base64-encoded for the hashline (bcrypt base64 alphabet).

This module expects the hash in the format defined by passlib. Example hash for "password":
`$bcrypt-sha256$v=2,t=2b,r=12$n79VH.0Q2TMWmt3Oqt9uku$Kq4Noyk3094Y2QlB8NdRT8SvGiI4ft2`

This format has been specifically requested in the past:
https://hashcat.net/forum/thread-8377.html

Possible future improvements:
- Version 1 of the hash could also be added to this, but the format of the generated hashes differs significantly in length, because of the increased verbosity of the new format.
- The provided test module works on my local hardware, but test success depends on commandline switches. If I specify a device, the single mode tests succeed and without that switch the multi mode tests do. I am not familiar enough with the test environment to debug this behavior at the moment.

Implementation reference:
https://passlib.readthedocs.io/en/stable/lib/passlib.hash.bcrypt_sha256.html

### DIFF
--- a/OpenCL/m30601-pure.cl
+++ b/OpenCL/m30601-pure.cl
@@ -1,0 +1,1119 @@
+
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ * 
+ * Based on module m30600
+ */
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha256.cl)
+#endif
+
+#if   VECT_SIZE == 1
+#define uint_to_hex_lower8(i) make_u32x (u16_bin_to_u32_hex ((i)))
+#elif VECT_SIZE == 2
+#define uint_to_hex_lower8(i) make_u32x (u16_bin_to_u32_hex ((i).s0), u16_bin_to_u32_hex ((i).s1))
+#elif VECT_SIZE == 4
+#define uint_to_hex_lower8(i) make_u32x (u16_bin_to_u32_hex ((i).s0), u16_bin_to_u32_hex ((i).s1), u16_bin_to_u32_hex ((i).s2), u16_bin_to_u32_hex ((i).s3))
+#elif VECT_SIZE == 8
+#define uint_to_hex_lower8(i) make_u32x (u16_bin_to_u32_hex ((i).s0), u16_bin_to_u32_hex ((i).s1), u16_bin_to_u32_hex ((i).s2), u16_bin_to_u32_hex ((i).s3), u16_bin_to_u32_hex ((i).s4), u16_bin_to_u32_hex ((i).s5), u16_bin_to_u32_hex ((i).s6), u16_bin_to_u32_hex ((i).s7))
+#elif VECT_SIZE == 16
+#define uint_to_hex_lower8(i) make_u32x (u16_bin_to_u32_hex ((i).s0), u16_bin_to_u32_hex ((i).s1), u16_bin_to_u32_hex ((i).s2), u16_bin_to_u32_hex ((i).s3), u16_bin_to_u32_hex ((i).s4), u16_bin_to_u32_hex ((i).s5), u16_bin_to_u32_hex ((i).s6), u16_bin_to_u32_hex ((i).s7), u16_bin_to_u32_hex ((i).s8), u16_bin_to_u32_hex ((i).s9), u16_bin_to_u32_hex ((i).sa), u16_bin_to_u32_hex ((i).sb), u16_bin_to_u32_hex ((i).sc), u16_bin_to_u32_hex ((i).sd), u16_bin_to_u32_hex ((i).se), u16_bin_to_u32_hex ((i).sf))
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+// // Unix crypt alphabet
+// CONSTANT_VK u32 bin2base64[0x40] =
+// {
+//   0x2e, 0x2f, 0x41, 0x42, 0x43, 0x44, 0x45, 0x46, 0x47, 0x48, 0x49, 0x4a, 0x4b, 0x4c, 0x4d, 0x4e, 0x4f, 0x50,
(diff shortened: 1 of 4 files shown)


matrix:
Hi, would you be able to test on CPU as well? I see a few little issues
Thanks

matrix:
You need also modify tunings/Modules_bcrypt.hctune by replicate the "30600" settings for "30601".

Following my local version:

```
$ cat tunings/Modules_bcrypt.hctune 

#Device                                         Attack  Hash    Vector  Kernel  Kernel
#Name                                           Mode    Type    Width   Accel   Loops

DEVICE_TYPE_CPU                                 *       3200    1       N       A
DEVICE_TYPE_CPU                                 *       9000    1       N       A
DEVICE_TYPE_CPU                                 *       18600   1       N       A
DEVICE_TYPE_CPU                                 *       25600   1       N       A
DEVICE_TYPE_CPU                                 *       25800   1       N       A
DEVICE_TYPE_CPU                                 *       28400   1       N       A
DEVICE_TYPE_CPU                                 *       30600   1       N       A
DEVICE_TYPE_CPU                                 *       30601   1       N       A
```

jsteube:
Please update tunings/Modules_bcrypt.hctune as @matrix described

twier:
> Hi, would you be able to test on CPU as well? I see a few little issues Thanks

```
$ .\hashcat.exe -D 1 -b -m 30601
hashcat (v6.2.6) starting in benchmark mode

hiprtcCompileProgram is missing from HIPRTC shared library.

Support for HIPRTC was dropped by AMD Adrenalin Edition 22.7.1 and later.
This is not a hashcat problem.

Please install the AMD HIP SDK

CUDA API (CUDA 12.9)
====================
* Device #1: NVIDIA GeForce RTX 2060 SUPER, skipped

OpenCL API (OpenCL 3.0 CUDA 12.9.40) - Platform #1 [NVIDIA Corporation]
* Device #2: NVIDIA GeForce RTX 2060 SUPER, skipped

OpenCL API (OpenCL 3.0 WINDOWS) - Platform #2 [Intel(R) Corporation]
* Device #3: AMD Ryzen 7 7800X3D 8-Core Processor, 16189/32442 MB (8110 MB allocatable), 16MCU

Benchmark relevant options:
===========================
* --backend-devices-virtual=1
* --opencl-device-types=1
* --optimized-kernel-enable

----------------------------------------------------------------------------------
* Hash-Mode 30601 (bcrypt-sha256 v2 bcrypt(HMAC-SHA256($pass))) [Iterations: 4096]
----------------------------------------------------------------------------------

Speed.#3.........:       99 H/s (81.14ms) @ Accel:16 Loops:128 Thr:1 Vec:1

Started: Sun Jul 13 19:59:18 2025
Stopped: Sun Jul 13 19:59:33 2025
```
With the new fixes, CPU seems to be working as well.

twier:
> You need also modify tunings/Modules_bcrypt.hctune by replicate the "30600" settings for "30601".
> 
> Following my local version:
> 
> ```
> $ cat tunings/Modules_bcrypt.hctune 
> 
> #Device                                         Attack  Hash    Vector  Kernel  Kernel
> #Name                                           Mode    Type    Width   Accel   Loops
> 
> DEVICE_TYPE_CPU                                 *       3200    1       N       A
> DEVICE_TYPE_CPU                                 *       9000    1       N       A
> DEVICE_TYPE_CPU                                 *       18600   1       N       A
> DEVICE_TYPE_CPU                                 *       25600   1       N       A
> DEVICE_TYPE_CPU                                 *       25800   1       N       A
> DEVICE_TYPE_CPU                                 *       28400   1       N       A
> DEVICE_TYPE_CPU                                 *       30600   1       N       A
> DEVICE_TYPE_CPU                                 *       30601   1       N       A
> ```

I have changed the file accordingly, thanks for the hint complete with the appropriate code!

jsteube:
Top. Many thanks for the contribution!

jsteube:
Hey there! Thanks again for contributing this hash-mode 30601.

I just ran into some self-test issues on an Intel Arc GPU. At first I thought it might be a runtime issue, but then I noticed that hash-mode 30600 is working fine. So I had a closer look into your code (which bases on hash-mode 30600) and found something that might be the cause.

There is a subtle (and honestly pretty tricky) issue here, and to be fair, hashcat does not document this part very well. It is a bit unusual, but it is all for performance reasons. Check out this section from the plugin development guide: https://github.com/hashcat/hashcat/blob/master/docs/hashcat-plugin-development-guide.md

It says:

> The buffer you provide must be padded to a size of a multiple of the block size of whatever crypto primitive you are using. For example 64 byte for MD5, SHA256, etc. and 128 byte for SHA512 and others. If you do not know the block sizes, check the algorithm specs. For instance, if you want to statically append a 5 byte string to a password from inside the kernel, you could use sha1_update (&ctx, buf, 5);, the important thing here is that buf[] must be declared as u32 buf[16]. That is because the block size of SHA1 is 64 byte.


In your code, you have:

```
u32 s[8] = { 0 };
...
sha256_hmac_init (&ctx0, s, string_salt_len);
```

This leads to out-of-bounds reads in the crypto primitive. Whether or not it crashes or misbehaves depends on the compute runtime, so it might have worked fine on your system during development, but causes issues on mine (Intel NEO).

No worries though. This is a really easy mistake to make, and not obvious at all if you are new to hashcat internals. You may have even run into this yourself and tried to work around it by copying data into registers manually, which can appear to fix the issue, even though the underlying cause remains hidden.

With that in mind, and after a small change to the salt structure, we can actually switch back to the more natural (and preferred) function calls like this:

```
sha256_hmac_ctx_t ctx0;

sha256_hmac_init_global_swap (&ctx0, esalt_bufs[DIGESTS_OFFSET_HOST].string_salt_buf, esalt_bufs[DIGESTS_OFFSET_HOST].string_salt_len);

sha256_hmac_update_global_swap (&ctx0, pws[gid].i, pws[gid].pw_len);

sha256_hmac_final (&ctx0);
```

This also lets us remove the large pw[] buffer, which was using 64 registers per thread. It probably will not have a big impact on performance, since this is only in the init() function.

Thanks again for the contribution. This stuff is not easy to get right on the first try, and it is great to see new developers diving in!

See this commit for details: https://github.com/hashcat/hashcat/commit/d536d6e8329bae9731c165337a7148d2ad4f7924

PR #4180, opened 2025-04-09
Allow flexible current directories on Windows

### LINKED ISSUE #1539: Hashcat 4.1.0 Windows PATH Bug
I have the hashcat-4.1.0 directory extracted into the root of the C: drive on a PC (C:\hashcat-4.1.0) and have the Windows PATH set to include this folder.

The machine is running 64bit Windows 10 Pro.

I tried calling the hashcat64 binary from another folder through command prompt or powershell e.g. ```C:\example>hashcat64 -a 3 -m 10.....```

However, I get this error:

```./hashcat.hctune: No such file or directory```

I have confirmed that this file is in fact present within the hashcat-4.1.0 directory. This issue does not exist with hashcat-4.0.1

Hope that helps!

### PR DESCRIPTION
**Pull request by PenguinKeeper7**
This PR allows Hashcat to be run from any directory, not requiring Hashcat to be the current directory. This issue only affects Windows, Linux already appears to work by default so this just restores a little feature parity.

Considerations:
I have tested this personally a fair amount but because Hashcat uses thousands of paths, I can't test every them all and I encourage people to try many different varied attacks, on many compilers to ensure this is a safe change.

Fixes https://github.com/hashcat/hashcat/issues/1539 and others

### DIFF
--- a/include/shared.h
+++ b/include/shared.h
@@ -41,6 +41,8 @@ u64 mydivc64 (const u64 dividend, const u64 divisor);
 
 char *filename_from_filepath (char *filepath);
 
+int utf8_to_widechar (const char *utf8, wchar_t **wide_out);
+
 void naive_replace (char *s, const char key_char, const char replace_char);
 void naive_escape (char *s, size_t s_max, const char key_char, const char escape_char);
 
@@ -57,6 +59,7 @@ bool hc_path_is_file (const char *path);
 bool hc_path_is_directory (const char *path);
 bool hc_path_is_fifo (const char *path);
 bool hc_path_is_empty (const char *path);
+bool hc_access (const char *path, const int mode);
 bool hc_path_exist (const char *path);
 bool hc_path_read (const char *path);
 bool hc_path_write (const char *path);
@@ -72,6 +75,8 @@ void hc_string_trim_leading (char *s);
 int hc_get_processor_count (void);
 
 bool hc_same_files (char *file1, char *file2);
+
+int hc_stat (const char* path, struct stat *buf);
 
 u32 hc_strtoul  (const char *nptr, char **endptr, int base);
 u64 hc_strtoull (const char *nptr, char **endptr, int base);
--- a/src/backend.c
+++ b/src/backend.c
@@ -8504,12 +8504,7 @@ static bool load_kernel (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_p
 
       hc_asprintf (&nvrtc_options[3], "compute_%d%d", device_param->sm_major, device_param->sm_minor);
 
-      // untested on windows, but it should work
-      #if defined (_WIN) || defined (__CYGWIN__) || defined (__MSYS__)
-      hc_asprintf (&nvrtc_options[4], "-D INCLUDE_PATH=%s", "OpenCL");
-      #else
       hc_asprintf (&nvrtc_options[4], "-D INCLUDE_PATH=%s", folder_config->cpath_real);
-      #endif
 
       hc_asprintf (&nvrtc_options[5], "-D XM2S(x)=#x");
       hc_asprintf (&nvrtc_options[6], "-D M2S(x)=XM2S(x)");
--- a/src/backend.c
+++ b/src/backend.c
@@ -539,7 +539,7 @@ bool read_kernel_binary (hashcat_ctx_t *hashcat_ctx, const char *kernel_file, si
   {
     struct stat st;
 
-    if (stat (kernel_file, &st))
+    if (hc_stat (kernel_file, &st))
     {
       hc_fclose (&fp);
 
--- a/src/brain.c
+++ b/src/brain.c
@@ -1631,7 +1631,7 @@ bool brain_server_read_hash_dump (brain_server_db_hash_t *brain_server_db_hash, 
 
   memset (&sb, 0, sizeof (struct stat));
 
-  if (stat (file, &sb) == -1)
+  if (hc_stat (file, &sb) == -1)
   {
     brain_logging (stderr, 0, "%s: %s\n", file, strerror (errno));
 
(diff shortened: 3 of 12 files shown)

### CODE REVIEW

**Comment by jsteube** [comment]
Could you please test your patch with a username that contains UTF16 characters? But it's important to create this user first and not rename it. This way, its home folder will also contain UTF16 characters. Then install the hashcat patch in this folder and try to see if everything works "normally." I remember having problems with exactly that back then.

Try this username: 测试用户长名称

**Comment by PenguinKeeper7** [comment]
Hello @jsteube. You're not wrong that it breaks - thanks for the report - but wouldn't this be identical in Linux, due to it attempting to squash multibyte characters into simple single-byte chars and therefore need a larger upgrade of all path handling to use wider character types? (When calling from a separate file location, with the multibyte string in the file name, of course)

**Comment by jsteube** [comment]
On Linux it should be fine. Difference is UTF8 and UTF16. Strings in UTF16 will include zero bytes which cause strlen() and friends to fail.

**Comment by PenguinKeeper7** [comment]
Well that's annoying. I've managed to convert seemingly the entire codebase to be multibyte-compatible (UTF16-LE isn't a problem, the multibyte in some of the OS calls like `stat` breaks, [this is not unique to us](https://github.com/ocornut/imgui/issues/917)) but CUDA/OpenCL remains grumpy because of the strange pathing. I think the only solutions here are janky, like copying the entire contents of the OpenCL folder to a temp file that has a path of full ASCII each run, although this is obviously not ideal, so I think the non-janky, most compassing option is to probably allow people with single-byte paths to continue using Hashcat from any location, while multibyte people have to `cd`, which I think is probably a relatively low percentage of users anyway.

If anyone knows of a way to get nvrtc/CUDA to play happy with multibyte Windows paths, I'm all ears
```

shared_kernel(7): catastrophic error: cannot open source file "[redacted]/hashcat-1539-flexible-pathing/\u00e6\u00b5‹\u00e8¯•\u00e7”¨\u00e6ˆ·\u00e9•¿\u00e5\u00e7§°/OpenCL//inc_vendor.h"
  #include M2S(INCLUDE_PATH/inc_vendor.h)
                                         ^

1 catastrophic error detected in the compilation of "shared_kernel".
Compilation terminated.

* Device #1: Kernel [redacted]/hashcat-1539-flexible-pathing/测试用户长名称/OpenCL/shared.cl build failed.

* Device #1: Kernel [redacted]/hashcat-1539-flexible-pathing/测试用户长名称/OpenCL/shared.cl build failed.
```

**Comment by matrix** [comment]
have you tried converting the path to UTF-8 before setup INCLUDE_PATH?

**Comment by PenguinKeeper7** [comment]
It's already UTF-8. UTF16-LE was never a problem

**Comment by matrix** [comment]
ok. If you post the rest of the code you modified it will come easier. However there is the M2S macro that is used as a workaround on Apple to get the includes to work that is not helping you. You would have to try forcing the path on the includes without the macro to see if the problem is just that or is it also on nvrtc/CUDA/OpenCL/etc ...

**Comment by PenguinKeeper7** [comment]
Converted to draft as the branch is no longer fully tested, but you should be able to get down to the CUDA/OpenCL failure with a multibyte path. The fix in https://github.com/hashcat/hashcat/pull/4180/commits/6ad985fcd31feba31a263a29690834bc689e1c6f is just temporary, to fix a minor version control issue and will be prettified before release

**Comment by matrix** [comment]
> Converted to draft as the branch is no longer fully tested, but you should be able to get down to the CUDA/OpenCL failure with a multibyte path. The fix in [6ad985f](https://github.com/hashcat/hashcat/commit/6ad985fcd31feba31a263a29690834bc689e1c6f) is just temporary, to fix a minor version control issue and will be prettified before release

I don't have windows to test it, you should try modifying the code of one of the kernels, removing the M2S macro ... although I don't think it will come to anything useful

**Comment by PenguinKeeper7** [comment]
Tried and failed:
```
#include M2S(INCLUDE_PATH/inc_vendor.h)
#include INCLUDE_PATH/inc_vendor.h
#include "INCLUDE_PATH/inc_vendor.h"
#include "./inc_vendor.h"
#include "inc_vendor.h"
#include "[redacted]\hashcat-1539-flexible-pathing\测试用户长名称\OpenCL\inc_vendor.h"
#include [redacted]\hashcat-1539-flexible-pathing\测试用户长名称\OpenCL\inc_vendor.h"
#include M2S("[redacted]/hashcat-1539-flexible-pathing/测试用户长名称/inc_vendor.h")
#include M2S("[redacted]\hashcat-1539-flexible-pathing\测试用户长名称\inc_vendor.h")
```
Open to suggestions!

matrix:
> #include "[redacted]\hashcat-1539-flexible-pathing\测试用户长名称\OpenCL\inc_vendor.h"

if this didn't work ... the problem is not dependent on hashcat. No way for me

jsteube:
Just had an idea. Since this is a problem in an external program, hashcat can't really fix it, as @matrix already pointed out. 

Maybe we can workaround the problem. On Linux, a user can set a symbolic link with ln -s, but I don't know if that works on NTFS/Windows. I googled a bit and it seems on NTFS there's not a symlink, but CreateHardLink(). The idea would be to set a path containing only UTF8, or perhaps a pure ASCII random path, and symlink it with the UTF16 path. Then we could point the NVRTC/HIPRTC etc to the left UTF-8 side of the symlink.

I have never worked myself with CreateHardLink() so maybe this is the wrong API call, and I read somewhere that CreateHardLink() required admin rights, that would be a show-stopper, but I think is worth a try.

jsteube:
Both halves of this are handled now, so I'm closing this PR.

The working directory half landed as #4654. It reads the install directory as this branch does, and it also hands that directory to the HIP and OpenCL compilers, where this one leaves the relative name in place. It covers a UNC path as well: inside a -D value the leading // of //server/share starts a C comment, which leaves the include path empty.

The nvrtc half is #4885. In the meanwhile master gained an application manifest declaring the process code page as UTF-8, and I expected that to settle it, but nvrtc converts the include path with something else regardless. I reproduced your failure on the directory name from this thread with the manifest in effect, and the issue carries 2 ways out that do compile, so the wide character conversion this branch carries is not what is needed any more.

PR #4183, opened 2025-04-12
Fixing BUILD_WSL.md

whoot:
The win-iconv GitHub repo does not contain the Makefile (anymore?).
You need to download the release archiv instead.


ventaquil:
They are using CMake now so it looks like temporary solution - specially it binds to hardcoded version.

jsteube:
Fixed with https://github.com/hashcat/hashcat/pull/4193

PR #4188, opened 2025-04-22
Modules: Check UnpackSize to raise false positive with hc_decompress_rar

### LINKED ISSUE #4146: RAR3 -m 23800 emits false positives (and there's a trivial solution)
The -m23800 "RAR3-p (Compressed)" emits false positives due to CRC-32 collisions. The root problem is that the (external dependency) unrar code just returns on failure, without signalling that it encountered problems in the compressed data (the plain C implementation used in JtR returns true or false, so we never had this problem). So hashcat calculates a CRC on what it got and boing, we have a collision once in four billion candidates. That means 1-2 collisions per day on a decent GPU.

The solution is trivial but I currently don't have time to make a PR for you: You need to compare the expected unpacked size (which we thankfully know) with whatever actual unpacked size you get from unrar. If sizes don't match, reject. If they match AND the CRC-32 also matches, we have a crack! I have tested this scheme (not in hashcat) extensively and it works like a charm.

matrix:
Fix #4146

### DIFF
--- a/OpenCL/m23800-optimized.cl
+++ b/OpenCL/m23800-optimized.cl
@@ -44,6 +44,8 @@ typedef struct rar3_hook
 
   u32 first_block_decrypted[4];
 
+  u32 unpack_failed;
+
   u32 crc32;
 
 } rar3_hook_t;
@@ -494,6 +496,8 @@ KERNEL_FQ void m23800_comp (KERN_ATTR_TMPS_HOOKS_ESALT (rar3_tmp_t, rar3_hook_t,
   const u64 gid = get_global_id (0);
 
   if (gid >= GID_CNT) return;
+
+  if (hooks[gid].unpack_failed == 1) return;
 
   u32 crc32 = hooks[gid].crc32;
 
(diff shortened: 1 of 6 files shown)


magnumripper (commented):
There's theoretically a bug, very rarely triggering so also hard to test: What if the **expected** CRC happens to be 0x00000000? This will happen in one file out of four billion so very rare but we've seen things like that surface before.

To fix that, instead of `return 0` you could `return CurCRC32^0xcccccccc` or whatever can't _ever_ match the expected one.

Other than that, this change looks right and I'll test it soon.

matrix:
ok, I have a modification locally to avoid processing the crc32 if the unpack process fails since whatever value is set to crc32 might turn out to be valid. Meanwhile, let me know if that single check on UnpackSize is enough for you. Thanks!

magnumripper:
Tested now, false positive is gone and no regression found.

FunnyLarry:
great! But it seems to me that as a result of this fix there were real problems with speed, maybe I'm wrong.

https://github.com/hashcat/hashcat/issues/4190

magnumripper:
> great! But it seems to me that as a result of this fix there were real problems with speed, maybe I'm wrong.
> 
> #4190

I find that hard to believe. More likely something else changed - did you compare apples to apples? For example, the RAR KDF is severly impacted by password length - a four character password is way faster to test than a ten character one.

jsteube:
This is great. Thanks for patching and testing

PR #4225, opened 2025-05-29
add -std=c++14 for UNRAR only on MacOS

matrix:
@jsteube, so it doesn't give error. Even using clang/clang++ on linux works


jsteube:
Looking good, thanks.

PR #4247, opened 2025-06-09
Create cachedata2hashcat.py

Banaanhangwagen:
Tool to extract hash from CacheData.


jsteube:
This is great, thanks for the contribution!

PR #4252, opened 2025-06-17
Fix incorrect comparison result in sort_pot_orig_line()

visitorckw:
The original sort_pot_orig_line() function returned 0 even when a < b, violating the C standard requirements for qsort() comparison functions. Specifically, it broke antisymmetry and transitivity, which can result in undefined behavior.

In some versions of glibc, this leads not only to incorrect sorting but also potential memory corruption[1].

Fix the issue by returning -1 when a < b, restoring compliance with the standard.

Link: https://www.qualys.com/2024/01/30/qsort.txt [1]
Fixes: 6adc217ba ("Keep output of --show and --left in the original ordering of the input hash file")


jsteube:
Good catch, thanks!

PR #4254, opened 2025-06-17
Added remote-bridges, plugins and manual.

fse-a:
Remote Bridge Argon2id and Remote Bridge Scrypt

This is a proof-of-concept where you can use Hashcat to crack Argon2id and Scrypt passwords using a local GPU and remote CPU computer.

Here's how it works: 

There are two software components: Hashcat and hasher (https://github.com/fse-a/hasher)
•	Hashcat is launched on an Ubuntu machine with a GPU. This includes the IP addresses of the CPU computers to be used and the port number on which they listen for connections.
•	On the CPU computer(s), the Java application hasher-1.0.0 is launched, providing the same port number to listen for connections from Hashcat. For Argon2id, it is also important to reserve enough memory with the -Xmx option so that the Java process can handle high memory parameters.

Once Hashcat is started, the passwords are divided into blocks. Each block gets its own bridge unit and its own connection to a CPU computer. Once it starts a password block the first part is calculated on the GPU. The intermediate results are then send to the various CPU computers. They do the 'heavy' middle calculations and send the results back to their bridge unit. Once the last results are in, Hashcat calculates the final part on the GPU and checks to see if the password has been found. If the password is not found, Hashcat will continue on with the next password block.

Various concepts are tested with this: 
•	Sensitive information such as your private hash and dictionaries remain local and are not shared with third parties because only the middle part of the algorithm is run elsewhere. 
•	It creates the ability to use a significant amount of computing capacity in the short term without the need for additional hardware purchase and subsequent maintenance.

We hope this new feature inspires other developers :-) 

Kind regards,

fse-a


jsteube:
The is a fantastic use-case for the Assimilation Bridge. But before I can merge, the Git actions need to be fixed.

PR #4256, opened 2025-06-18
Fix Typo: change 'generatic' to 'generic' in both hashcat-assimilation-bridge-development.md and usage.c

DhruvTheDev1:
I believe the generetic is a typo and should be generic in the help/documentation strings for bridge parameters in the bridge_init() function.


PenguinKeeper7:
If generetic is a typo, this also needs to be updated, from the `--help` text:
https://github.com/hashcat/hashcat/blob/13245b5563d4f815523d3f4b1d94999bcec44722/src/usage.c#L97-L100

DhruvTheDev1:
Hi all,
This PR fixes a typo in both docs/hashcat-assimilation-bridge-development.md and src/usage.c,
The word 'generatic' isn't a standard term and looks like a typo for 'generic'

DhruvTheDev1:
> If generetic is a typo, this also needs to be updated, from the `--help` text:
> 
> https://github.com/hashcat/hashcat/blob/13245b5563d4f815523d3f4b1d94999bcec44722/src/usage.c#L97-L100

Hi, 
Thanks for this! I made the relevant changes and updated the PR to make the changes clearer and easier to track.

jsteube:
Spaces need to be fixed

PR #4264, opened 2025-06-23
Fix: correct typo and "Hashcat" to "hashcat"

DhruvTheDev1:
- Fixed typo: RUN_EXTENTED to  RUN_EXTENDED
- Capitalised Hashcat

### DIFF
--- a/docs/hashcat-assimilation-bridge-development.md
+++ b/docs/hashcat-assimilation-bridge-development.md
@@ -6,12 +6,12 @@ The following section is for plugin and bridge developers. It contains low-level
 
 ## Update existing plugins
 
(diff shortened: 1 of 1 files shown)


jsteube:
It’s part of our style, we intentionally write hashcat in lowercase. If you see it capitalized, that’s a typo. So the basic rule is: always use lowercase hashcat.

DhruvTheDev1:
> It’s part of our style, we intentionally write hashcat in lowercase. If you see it capitalized, that’s a typo. So the basic rule is: always use lowercase hashcat.

Hi, Thanks for the feedback!!
I changed all the "Hashcat" to "hashcat"
Also fixed the typo RUN_EXTENTED to RUN_EXTENDED

PR #4267, opened 2025-06-25
Fix socket leak in brain_server()

lhywk:
**Describe**

Hi,

Fixes a resource leak in the `brain_server()` function by ensuring the socket file descriptor (`server_fd`) is properly closed when an error occurs after `socket()` is successfully created but before the server is fully initialized.

**Expected Behavior**

When `setsockopt()`, `getaddrinfo()`, `bind()`, or `listen()` fails, the `server_fd` should be closed to avoid file descriptor leaks.

**Actual Behavior**

Previously, if any of those operations failed, the `server_fd` remained open, causing a resource leak.

This patch prevents a subtle file descriptor leak with minimal changes, improving the server’s stability in real-world use.

Thanks for reviewing.

### DIFF
--- a/src/brain.c
+++ b/src/brain.c
@@ -3052,6 +3052,8 @@ int brain_server (const char *listen_host, const int listen_port, const char *br
   {
     brain_logging (stderr, 0, "setsockopt: %s\n", strerror (errno));
 
+    close(server_fd);
+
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks :)

Issue #3673, opened 2023-04-08
feature: show max length in status


roycewilliams:
Idea .... take the value shown at:

`Maximum password length supported by kernel: 31`

and append it here:

`Kernel.Feature...: Optimized Kernel (max length: 31)`

(for both types of kernel)

Chick3nman:
There are times where it might be good to note that the lengths are not well respected, specifically in -a 9 and alternative attack modes. But any of the places where that may come up are places we should probably gate new users from anyway.

matrix:
Addressed here https://github.com/hashcat/hashcat/pull/3724

roycewilliams:
> Addressed here [#3724](https://github.com/hashcat/hashcat/pull/3724)

This looks like a hash-info-level change. My feature is a status-level change. Or maybe I'm missing your point? 😁

matrix:
> > Addressed here [#3724](https://github.com/hashcat/hashcat/pull/3724)
> 
> This looks like a hash-info-level change. My feature is a status-level change. Or maybe I'm missing your point? 😁

I think it makes more sense to show this information on hash-info instead of a status update

roycewilliams:
I'm actually in favor of both. It definitely makes sense to be in the hash info. But my goal in the original issue was to make it always available in the status, so you never have to look it up -- in the middle of an attack, you always know exactly what the maximum length is. I think this is something that will benefit users generally.

I consider it to be as fundamental to everyday use of hashcat as whether a pure or optimized kernel is running (which is why I am suggesting putting it right on the same line, since max length is a fundamental property of choice of kernel)

jsteube:
I think @roycewilliams has a point. There's a difference in the idea between https://github.com/hashcat/hashcat/pull/4270 and https://github.com/hashcat/hashcat/pull/3724 and both are useful, so we should merge both.

roycewilliams:
Merged with modifications -- thanks, closing!

PR #4270, opened 2025-06-27
show max length in Kernel.Feature status

roycewilliams:
Simple attempt to address my own #3673


matrix:
Hi, check this PR https://github.com/hashcat/hashcat/pull/3724
I think it already includes everything you need

roycewilliams:
See discussion in #3673

jsteube:
I'd also like to add more detail to the status screen. I think it's the place where users are most open to learning about hashcat internals and figuring out how to configure more focused attacks.

That said, was it intentional to use the term "max length"? It feels a bit ambiguous (salt length, hash length, ...). If you could explain your thinking, I’m sure it would make more sense.

If we're using this as an opportunity to teach users more about hashcat’s internals, we should try to work in terms like "general" and/or "kernel-feature specific" somewhere. I’m not quite sure how, though there’s a risk it could get too wordy, and we should keep the status view short.

roycewilliams:
@atom Hmm, yes, "max length" is ambiguous - someone else pointed that out on Discord. "Max pw length" or "max password length" would be more clear, as would making "bytes" explicit, so maybe:

max password length: 31 bytes

Or pw If some other information should be on the same line.

roycewilliams:
Or maybe

(salted)
`Kernel.Feature...: Optimized Kernel (passwords 8-63 bytes, salts 0-256 bytes)`

(unsalted)
`Kernel.Feature...: Pure Kernel (passwords 0-256 bytes)`
`Kernel.Feature...: Optimized Kernel (passwords 0-55 bytes)`

etc.
?

jsteube:
I salts is actually too much, most hashes nowadays are embedded salts and not generic constructs (there's no salt length option). So the second option makes more sense. Please update to "Password Length 0-55 bytes".

roycewilliams:
@jsteube Done (but I kept it as lower case, since it is supplemental information. Can capitalize if needed, but I recommend this way for idiomatic readability.)

Sample from local testing, -m 1000:

```
Kernel.Feature...: Optimized Kernel (password length 0-27 bytes)
[...]
```

jsteube:
Looks good to me, thanks!

PR #4272, opened 2025-06-28
workaround for HIP bug and avoiding a potential same bug on CUDA

matrix:
found a bug on HIP, documented here https://rocm.docs.amd.com/projects/HIP/en/docs-develop/doxygen/html/bug.html

```
HIP-Clang always returns 0 for regsPerBlock
```

Added a workaround and a safe check also on CUDA

POC on windows

Screenshot 2025-06-28 at 21 18 21


jsteube:
Top

PR #4273, opened 2025-06-29
Backend: Splitting backend_ctx_devices_init into smaller runtime-specific functions

matrix:
Refactoring backend_ctx_devices_init() by splitting runtime-specific functions into the following:
- backend_ctx_devices_init_cuda
- backend_ctx_devices_init_hip
- backend_ctx_devices_init_metal
- backend_ctx_devices_init_opencl

Tested on linux/windows/osx

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -127,6 +127,7 @@
 - Apple Driver: Automatically enable GPU support on Apple OpenCL instead of CPU support
 - Apple Driver: Updated requirements to use Apple OpenCL API to macOS 13.0 - use
 - Backend: Updated filename chksum format to prevent invalid cache on Apple Silicon when switching arch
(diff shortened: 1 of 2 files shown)


jsteube:
Nice idea, but over complicated. Instead of creating all these parameters, why not set directly?

```
  backend_ctx->need_adl           = need_adl;
  backend_ctx->need_nvml          = need_nvml;
  backend_ctx->need_nvapi         = need_nvapi;
  backend_ctx->need_sysfs_amdgpu  = need_sysfs_amdgpu;
  backend_ctx->need_sysfs_cpu     = need_sysfs_cpu;
  backend_ctx->need_iokit         = need_iokit;
```

matrix:
I hadn't checked the structure, since there are certainly some that can be optimized

matrix:
Do you think it's okay now or do you want to change something else?

PR #4293, opened 2025-07-06
Added workaround to get rid of internal runtimes memory leaks + Avoid deprecated HIP functions

matrix:
## Added workaround to get rid of internal runtimes memory leaks

As of now, especially in the benchmark mode, hashcat will not go to create and destroy context and command-queue for each enabled device each time it switches from one hash-mode to the next. Specifically using OpenCL with an NVIDIA device, it was not possible to complete the benchmark because clCreateContext has memory leaks that slowly consume all available GPU memory until hashcat can activate a new context and disable the device.

## Avoid deprecated HIP functions

All hipCtx* features have been declared deprecated, so we have replaced them with the new ones, also fixing a critical bug on handling multiple AMD devices in the same system.


jsteube:
Great finding! Thanks

PR #4294, opened 2025-07-08
Fix terminal TODO: Implement Windows system info display

Oblivionsage:
- Replace hardcoded 'N/A' values with actual Windows system information
- Add GetSystemInfo() for processor architecture detection
- Add GetVersionEx() for Windows version information
- Support both machine-readable and human-readable output formats
- Follow existing Linux uname() implementation pattern
- Maintain cross-platform compatibility

Resolves TODO comment in src/terminal.c line 1257

### DIFF
--- a/src/terminal.c
+++ b/src/terminal.c
@@ -1254,21 +1254,60 @@ void backend_info (hashcat_ctx_t *hashcat_ctx)
     }
 
     #if defined (_WIN) || defined (__CYGWIN__) || defined (__MSYS__)
-    // TODO
+    // Get Windows system information
+    SYSTEM_INFO sysinfo;
+    OSVERSIONINFO osvi;
+    char platform_buf[256] = "N/A";
+    char release_buf[256] = "N/A";
+    
+    GetSystemInfo (&sysinfo);
+    
+    // Initialize version info structure
+    ZeroMemory (&osvi, sizeof (OSVERSIONINFO));
+    osvi.dwOSVersionInfoSize = sizeof (OSVERSIONINFO);
+    
+    bool rc_version = (GetVersionEx (&osvi) != 0);
+    
+    // Get processor architecture string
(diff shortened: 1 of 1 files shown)


src/terminal.c
```diff
@@ -1254,21 +1254,60 @@ void backend_info (hashcat_ctx_t *hashcat_ctx)
...
+      snprintf (release_buf, sizeof (release_buf), "%lu.%lu.%lu",
+               osvi.dwMajorVersion, osvi.dwMinorVersion, osvi.dwBuildNumber);
+    }
+    
     if (user_options->machine_readable == false)
     {
       event_log_info (hashcat_ctx, "OS.Name......: Windows");
-      event_log_info (hashcat_ctx, "OS.Release...: N/A");
-      event_log_info (hashcat_ctx, "HW.Platform..: N/A");
+      event_log_info (hashcat_ctx, "OS.Release...: %s", rc_version ? release_buf : "N/A");
```
ventaquil: Is this really necessary? Since you fill `release_buf` with `N/A` maybe this check is pointless.

Oblivionsage: Thanks for the feedback @ventaquil you're right since `release_buf` is already initialized with "N/A", the ternary check is indeed pointless

Should I update it to simply use `release_buf` directly like this?

```c
event_log_info (hashcat_ctx, "OS.Release...: %s", release_buf);
```

And same for the JSON output:

```c
printf ("\"Release\": \"%s\" }, ", release_buf);
```

I appreciate you taking the time to review the code let me know if you'd like me to push an update with this fix

Oblivionsage:
I have  updated the code to remove the redundant ternary checks as suggested the fix has been pushed and the build still passes thanks for taking the time to review @ventaquil

ventaquil:
You're welcome!

PR #4314, opened 2025-07-11
Add CPU support for argon2 (34000)

fse-a:
**Argon2 support for CPU**
This resolves issues running the current argon2 implementation on CPU backends.

Some performance numbers on various CPUs using the self-test hash of 64 MB and 3 iterations:
```
Intel Xeon W-2235 @ 3.8GHz    : 48 H/s
Intel Xeon Gold 5317 @ 3.0GHz : 65 H/s
AMD Ryzen 7950X               : 75 H/s
Intel i7-14700K               : 91 H/s
```

**Changes**
 - Used `blake2b_transform()` instead of `blake2b_update()` to avoid compiler problems on Intel OpenCL and segfaults on POCL (still unsure of exact cause but possibly related to the shuffle functions in combination with these OpenCL drivers).
 - Remove 'bug' comments (these are resolved now).
 - Added implementation of `argon2_hash_block()` for non-warped (CPU) case.
 - Introduced `LBLOCKSIZE` for the size of an argon2 block per thread in u64.
 
Except for `argon2_hash_block()`, the implementation should be able to support any warp/wavefront/simd size in the future.

Kind regards,

Pelle & Ewald
Netherlands Forensic Institute

### DIFF
--- a/OpenCL/inc_hash_argon2.cl
+++ b/OpenCL/inc_hash_argon2.cl
@@ -12,24 +12,26 @@
 #include "inc_hash_blake2b.h"
 #include "inc_hash_argon2.h"
 
+#define LBLOCKSIZE      (128 / THREADS_PER_LANE)
+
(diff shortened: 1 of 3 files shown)


jsteube:
Top, thanks!

PR #4333, opened 2025-07-13
Fixed bug in inc_rp_optimized.cl on Apple Intel with Metal

matrix:
### Fixed old/critical bug on Apple Intel with Metal by patching inc_rp_optimized.cl.
Tested on Apple Intel and Silicon with Metal/OpenCL and on Linux with CUDA, HIP, OpenCL GPU/CPU

### Metal Backend: parallelize pipeline state object (PSO) compilation internally
Set unexported setting, setShouldMaximizeConcurrentCompilation, to boost kernel build process on Apple Metal (only >= 3)

### Other
Fixed warning messages during build md5 kernel

### DIFF
--- a/OpenCL/inc_common.h
+++ b/OpenCL/inc_common.h
@@ -304,6 +304,8 @@ DECLSPEC void hc_enc_init (PRIVATE_AS hc_enc_t *hc_enc);
 DECLSPEC int hc_enc_has_next (PRIVATE_AS hc_enc_t *hc_enc, const int sz);
 DECLSPEC int hc_enc_next (PRIVATE_AS hc_enc_t *hc_enc, PRIVATE_AS const u32 *src_buf, const int src_len, const int src_sz, PRIVATE_AS u32 *dst_buf, const int dst_sz);
 DECLSPEC int hc_enc_next_global (PRIVATE_AS hc_enc_t *hc_enc, GLOBAL_AS const u32 *src_buf, const int src_len, const int src_sz, PRIVATE_AS u32 *dst_buf, const int dst_sz);
+DECLSPEC int hc_enc_validate_utf8 (PRIVATE_AS const u32 *src_buf, const int src_pos, const int extraBytesToRead);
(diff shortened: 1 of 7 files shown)


jsteube:
Benchmark with 50m random pws, single hash ntlm, passwordspro rule, nv 4090

optimized:
real    0m10,549s
user    0m1,614s
sys     0m0,354s

pure:
real    1m17,226s
user    0m1,831s
sys     0m0,251s

After patch:

optimized:
real    0m12,556s
user    0m1,600s
sys     0m0,360s

pure:
real    1m17,228s
user    0m2,902s
sys     0m0,251s

I ran all commands multiple times to make sure all caching is in place, and fixed -n -T and -u.

-> Speed loss around 18%

We can see no change in pure kernel, makes sense, since pure kernels have not been changed.

I tried another GPU, here a 1080Ti, now in MH/s

Original: 490 MH/s
Patched: 420 MH/s

matrix:
Hi, 

> Benchmark with 50m random pws, single hash ntlm, passwordspro rule, nv 4090
> 
> optimized: real 0m10,549s user 0m1,614s sys 0m0,354s
> 
> pure: real 1m17,226s user 0m1,831s sys 0m0,251s
> 
> After patch:
> 
> optimized: real 0m12,556s user 0m1,600s sys 0m0,360s
> 
> pure: real 1m17,228s user 0m2,902s sys 0m0,251s
> 
> I ran all commands multiple times to make sure all caching is in place, and fixed -n -T and -u.
> 
> -> Speed loss around 18%
> 
> We can see no change in pure kernel, makes sense, since pure kernels have not been changed.
> 
> I tried another GPU, here a 1080Ti, now in MH/s
> 
> Original: 490 MH/s Patched: 420 MH/s

Hi
there are two points we can verify:
- find_bit_occurrence: by putting an ifdef we return to the code that was there before and the performance should be the same as before
- apply_rules_vect_optimized: I saw that for CUDA and HIP GLOBAL_AS is used instead of COSTANT_AS. There are two commented sections that might be useful for CUDA and HIP to improve performance.

Could you do these tests to see if it improves the benchmark?

matrix:
I answer to myself: I found a compromise to avoid complications.

jsteube:
Works, thanks!

PR #4343, opened 2025-07-18
Improve small base loop communication

PenguinKeeper7:
Instead of just printing a generic "Base loop isn't big enough" advisory message when Hashcat can't properly parallelise, this would tell the user exactly how many base words Hashcat is expecting, so they don't have to continually guess and check several combinations to see how big their wordlists/masks should be.

Totally flexible on wording.

Previous:
```
```

New:
```
Hashcat is expecting at least 917504 base words but only got 7.1% of that.
```


jsteube:
Nice idea, thanks!

PR #4351, opened 2025-07-20
Automatically generate docs using Github actions 

thatux:
Often enough I find myself using https://hashcat.net/wiki/doku.php?id=hashcat#Options and https://hashcat.net/wiki/doku.php?id=example_hashes because I don't have access to a machine that runs hashcat (master) at that moment. However, those docs are not up to date with master, so lack all cool new features.

By incorporating the generation of those docs in the Github action we always have up-to-date docs 🥳

You can find this Github-action in action here: https://github.com/thatux/hashcat/actions/runs/16404623224/job/46348763311?pr=1

PR #4353, opened 2025-07-21
Automatically generate docs using Github actions 3

thatux:
Follow up on https://github.com/hashcat/hashcat/pull/4351 and https://github.com/hashcat/hashcat/pull/4352

It seems to be intentional that Github Actions cannot add a commit to a PR (because the fork is not owned by the hashcat repo).. So we can't add the commit to the PR itself (as a 'suggestion')..

However, it does work having the check run automatically when something is merged into master (e.g. https://github.com/hashcat/hashcat/commit/423f7ef745a2551e7668260c76c922d54cac87ef).

Sorry for the bit of mess.. Happy with the result :)


jsteube:
Yeah it's indeed cool, but right now it create too much noise. It's a bit annoying on discord. Maybe we can reduce it somehow, so that it only triggers if there's  really a change to parameters, or a new mode?

PR #4357, opened 2025-07-23
Metal Backend: improved memory management

matrix:
Until now, support for Metal has been written using an Apple M1 and a 10-year-old Apple Intel as a basis to verify that the changes were also compatible with very old devices. Recently, the code has been tested on an Apple M4 Pro, with a performance increase of about 3,7 times compared to the M1. The code has also been sporadically tested on an Apple device with a discrete AMD GPU, but the performance was very low.

With this patch, I revisited memory management on Metal, initially creating an easily configurable array mapped 1 to 1 with the buffers allocated by hashcat. The configuration refers to the Storage Mode associated with the buffers, as well as an ad hoc modification that transforms buffers with a SHARED Storage Mode to MANAGED if the device is a discrete GPU and not an M* (Silicon) or integrated (Intel) one.

The result was excellent, as some very quick tests showed, for example, **argon2** going from **10 H/s (1330.58ms)** to **465 H/s (57.26ms)**! A **4550% increase in computational performance** and a **2223% increase in execution timing performance** on the GPU!

In addition to the array for configuring the buffer storage modes, a macro, HC_MTL_CREATEBUFFER, has also been created. This is used to generate the code that calls hc_mtlCreateBuffer, making the code much more readable than before.

In summary, this patch lays the groundwork for further improvements to the hashcat core, both on Metal itself and also for other runtimes, particularly OpenCL.

Thanks ;)


jsteube:
Absolutely amazing improvement on Metal here. Many thanks. We need to update all benchmark :)

Issue #4367, opened 2025-07-28
Issues with Argon2


matrix:
Hi, I found some issues on Argon2.

- selftest fail if the input hash have different t and/or p value. All runtimes are affected by this bug, but on Linux (tested with CUDA/HIP/OpenCL NV/OpenCL AMD/OpenCL CPU/PoCL) using --self-test-disable the password is found, on Apple no. Following a test case

```
$ echo 80 | ./hashcat -m 34000 '$argon2id$v=19$m=65536,t=2,p=5$AQclYXgIVFMCYWIXR3EpdA$JIzQTnb+bCGlrKpiweN2+RxHJr5bLax6NNOg3aWVy8k' -d 3 --potfile-disable 
hashcat (v6.2.6-1629-gbfb0d04d4) starting

CUDA API (CUDA 12.9)
====================
* Device #01: NVIDIA GeForce RTX 4080, skipped

HIP API (HIP 6.3.42134)
=======================
* Device #02: AMD Radeon RX 6900 XT, skipped

OpenCL API (OpenCL 3.0 CUDA 12.9.76) - Platform #1 [NVIDIA Corporation]
=======================================================================
* Device #03: NVIDIA GeForce RTX 4080, 15943/15943 MB (3985 MB allocatable), 76MCU

OpenCL API (OpenCL 2.1 AMD-APP (3635.0)) - Platform #2 [Advanced Micro Devices, Inc.]
=====================================================================================
* Device #04: AMD Radeon RX 6900 XT, skipped

OpenCL API (OpenCL 3.0 LINUX) - Platform #3 [Intel(R) Corporation]
==================================================================
* Device #05: 12th Gen Intel(R) Core(TM) i7-12700K, skipped

OpenCL API (OpenCL 3.0 PoCL 5.0+debian  Linux, None+Asserts, RELOC, SPIR, LLVM 16.0.6, SLEEF, DISTRO, POCL_DEBUG) - Platform #4 [The pocl project]
==================================================================================================================================================
* Device #06: cpu-haswell-12th Gen Intel(R) Core(TM) i7-12700K, skipped

Minimum password length supported by kernel: 0
Maximum password length supported by kernel: 256
Minimum salt length supported by kernel: 0
Maximum salt length supported by kernel: 256

Hashes: 1 digests; 1 unique digests, 1 unique salts
Bitmaps: 16 bits, 65536 entries, 0x0000ffff mask, 262144 bytes, 5/13 rotates
Rules: 1

[...]

PR #4370, opened 2025-07-29
Fixed bug in 34000 with Metal and OpenCL on Apple

matrix:
Unfortunately, the latest changes to Argon2 have not resolved the issue described in #4367. The only solution on Apple is this one at the moment.


jsteube:
The fixes have them solved, see Linux. The issue that remained is a completely different one, and exist only on Metal.

PR #4373, opened 2025-07-30
Support for LUKS2

fse-a:
**LUKS v2 support for Hashcat**

This plugin introduces support for LUKS v2 to Hashcat.

For now, there is support for argon2id as KDF, SHA-256 as hash and AES as cipher type. LUKS2 primarily uses Argon2 as KDF and performance depends mainly on the parameters used for Argon2. The plugin is based on hash mode 29521 (LUKSv1) in hashcat.

**Changes**
- Added hash mode 34100 for LUKS2 support.
- Added Python extracting script for generating a hash from a LUKS2 partition.

Future work may consist of adding support for other hash and cipher types.

Kind regards,

Pelle & Ewald
Netherlands Forensic Institute


tools/luks2_to_hashcat.py
ventaquil, line 1: Can you modify my script? User just will use one script to extract LUKS regardless of the version.

Here is the path - [tools/luks2hashcat.py](https://github.com/hashcat/hashcat/blob/master/tools/luks2hashcat.py).

If you have any questions just pick me up.

fse-a: Thank you for the suggestion, we will try to merge our code into your script. The script seems to be already prepared to add more LUKS versions by mapping.

fse-a: @ventaquil : We merged the relevant part of our script into yours. At the moment, we only process the first key slot, that's still something to improve upon later.

Could you please review our changes?

ventaquil: I am on vacations right now. I will look into it and try to merge in my spare time.

PR #4377, opened 2025-07-31
Docker runtime

dropdeadfu:


jsteube:
Super nice, thanks for the contribution!

I was following your updated instructions from BUILD_Docker.md. Check it out:

OK:

```
$ docker build --no-cache -f docker/BinaryPackage.ubuntu20 -t hashcat-binaries .
[+] Building 238.6s (37/37) FINISHED                                                                                                                                                 
```

OK (long runtime is only due to slow download speed)

```
$ docker build -f docker/runtime.cuda.ubuntu24 -t hashcat .
[+] Building 487.9s (10/10) FINISHED                                                                                                                                                  
                                                                                                                                                                                                                                                                
 2 warnings found (use docker --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 4)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 5)

```

OK:

```
$ docker run --rm --gpus=all -it hashcat bash

==========
== CUDA ==
==========

CUDA Version 12.9.1

Container image Copyright (c) 2016-2023, NVIDIA CORPORATION & AFFILIATES. All rights reserved.

This container image and its contents are governed by the NVIDIA Deep Learning Container License.
By pulling and using the container, you accept the terms and conditions of this license:
https://developer.nvidia.com/ngc/nvidia-deep-learning-container-license

A copy of this license is made available in this container at /NGC-DL-CONTAINER-LICENSE for your convenience.
```

That looks good so far, but now:

```
root@0e100a1e85cb:~/hashcat-6.2.6# ./hashcat.bin -V
v6.2.6
```

The issue might be user error, as I am not a Docker expert. However, that outdated release build of v6.2.6 should not be present. It may be left over from an older experiment. If so, the manual should include instructions to prevent this, since other users might encounter the same issue. Alternatively, the build script may need to be updated to handle this case.

dropdeadfu:
This is supposed to be on purpose. The build uses a dedicated release except if you chain it as the readme specifies.
To change the the release you want in your runtime change the version in the dockerfile:

ENV HASHCAT_VERSION=6.2.6

ENV HASHCAT_VERSION=7.0.0

jsteube:
I have revisited the PR from scratch, rebuilding everything and following the steps exactly as described in the BUILD_Docker.md.

In particular, I think there are some errors in this section:

```
docker build -f docker/BinaryPackage.ubuntu20 -t hashcat-binaries .
docker build -f docker/runtime.cuda.ubuntu24.withbuild -t hashcat-binaries .
docker run --rm --gpus=all -it hashcat bash
root@docker:~/# cd hashcat-6.2.6
root@docker:~/hashcat-6.2.6# ./hashcat.bin --help
```

Here is some log:

```
docker build -f docker/BinaryPackage.ubuntu20 -t hashcat-binaries . 
docker build -f docker/runtime.cuda.ubuntu24.withbuild -t hashcat-binaries .
docker run --rm --gpus=all -it hashcat bash
root@d742da35c640:~/hashcat-6.2.6# cd hashcat-6.2.6
bash: cd: hashcat-6.2.6: No such file or directory

OK, the issue is clear. Still, it should be fixed in the documentation.

root@docker:~/hashcat-6.2.6# ./hashcat.bin --help

root@d742da35c640:~/hashcat-6.2.6# ./hashcat.bin -V    
v6.2.6
root@d742da35c640:~/hashcat-6.2.6#
```

Here's the main issue: why is this showing as 6.2.6 instead of 7.0.0? But before you think about how to fix it, please read on.

I don't think that this needs fixing. Instead, all you need to do is build hashcat from the master branch, which makes the process very straightforward. 

There will be no need to package anything, update version numbers, install third-party libraries for Windows, or even cross-compile, is all not needed. All you need is to install the relevant apt packages. Since only the Linux binary is required inside the container, the overall package becomes much smaller.

This also resolves the version issue automatically, because when using master, there are no versioned folder names involved.

One more thing. Your changes to the existing file begin at this section:

```
### Runtime ###
```

But this does not match the original document name, "BUILD_Docker.md", which suggests the document is only about building. Users looking for runtime information might skip it, assuming it is not relevant. At the same time, users who are building may find the runtime sections confusing and out of place.

I think the simplest solution is to move the runtime content into a new file, for example, RUNTIME_Docker.md, and remove it from BUILD_Docker.md.

dropdeadfu:
Your issue when doing

docker build -f docker/BinaryPackage.ubuntu20 -t hashcat-binaries . 
docker build -f docker/runtime.cuda.ubuntu24.withbuild -t hashcat-binaries .
docker run --rm --gpus=all -it hashcat bash
root@d742da35c640:~/hashcat-6.2.6# cd hashcat-6.2.6

is that you are building the runtime with the same docker tag as the binaries but still revering to the old docker tag you build a few days ago when actually starting the container.

i agree with you on splitting the readme.

jsteube:
Many thanks!

PR #4385, opened 2025-08-04
Remove default -march=native and -march=avx2 flags

Xeonacid:
Compiling with flags like `-mavx*`, `-march=native`, or `/arch:AVX*` will make the resulting binary incompatible with cpus not supporting the requested instruction set.

This created problems for maintainers on various distributions.

And architecures like RISC-V don't support -march=native.

Anyone who needs the flags can feed them manually.

### DIFF
--- a/deps/yescrypt-master/Makefile
+++ b/deps/yescrypt-master/Makefile
@@ -21,8 +21,8 @@ LD = $(CC)
 RM = rm -f
 OMPFLAGS = -fopenmp
 OMPFLAGS_MAYBE = $(OMPFLAGS)
-#CFLAGS = -Wall -O2 -fomit-frame-pointer $(OMPFLAGS_MAYBE) -DSKIP_MEMZERO
-CFLAGS = -Wall -O2 -march=native -fomit-frame-pointer $(OMPFLAGS_MAYBE) -DSKIP_MEMZERO
+CFLAGS = -Wall -O2 -fomit-frame-pointer $(OMPFLAGS_MAYBE) -DSKIP_MEMZERO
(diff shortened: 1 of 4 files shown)


jsteube:
These flags are intended for our official binaries. I am fine with removing them for special cases, but by default they should always remain enabled.

I suggest introducing a new parameter at the top of the Makefile, maybe called MAINTAINER_MODE or something similar. Set it to 0 by default, and wrap the sections you removed in a conditional check based on that flag.

Then, when building packages for distributions, you can set MAINTAINER_MODE to 1 to disable those flags as needed.

Xeonacid:
> These flags are intended for our official binaries.

This means that users who expect to use the official binary must have hardware that supports at least the instruction set used by the build machine. If the build machine supports AVX512, then any CPU that does not support AVX512 will not be able to run this binary. I don't think it makes sense. 😕

jsteube:
The official binary we provide uses AVX2 rather than AVX512. That is perfectly fine, since we generally expect a typical hashcat user to run it on a system that supports AVX2. It was introduced back in 2013, so it is not exactly new anymore.

Your case is a bit different, and that is completely understandable. If you are distributing a binary to your users, it is possible that the machine used for compiling had AVX512 support enabled. That is why I suggested adding a MAINTAINER_MODE macro to prevent AVX512 from being used if it is available. 

I will take care of adding it and let you know.

jsteube:
Here you go:

https://github.com/hashcat/hashcat/commit/46fa39747f2cc1ce7a7ed014c27db3e37ea8c568

```
$ make | grep -e march -e mavx | wc -l
3
$ make MAINTAINER_MODE=1 | grep -e march -e mavx | wc -l
0
```

If you need other features to make packaging easier, let me know.

PR #4392, opened 2025-08-05
Updated *BSD support (OpenBSD, NetBSD, FreeBSD, DragonflyBSD)

matrix:
Hi,

*BSD don't have /proc/meminfo, so we can use sysctlbyname. **Must be tested on OpenBSD, FreeBSD and NetBSD**

Thanks


matrix:
I started by modifying one function, but in the end I built hashcat on all four *BSD OSes and fixed the errors that were popping up :)

PR #4398, opened 2025-08-06
keepass v4 initial code

### LINKED ISSUE #4214: Keepass v4.0 Cracking Support
**Describe the bug**
JTR/Hashcat does not support keepass v4.0 at this time.

```sh
keepass2john ./recovery.kdbx                                 
! ./recovery.kdbx : File version '40000' is currently not supported!
```

**Known Workarounds:**

CPU cracking Bash script: https://github.com/r3nt0n/keepass4brute

thatux:
initial work on supporting keepass v4 databases (should fix https://github.com/hashcat/hashcat/issues/4214).

use https://github.com/openwall/john/blob/bleeding-jumbo/src/keepass2john.c to generate hash.

only tested with keepass4.kdbx attached (argon2d), no support for keyfile.

1. how to go about the change in inc_hash_argon2.cl and inc_hash_argon2.h?
2. does argon2_tmp really need to exist?
3. what modenumber to use? 34200? I've kept it 34000 now to highlight the changes to the argon2 kernel


matrix:
You are modifying an existing hash mode, and that's not good. Run all the tests you want, but do it locally. There's no need to show that you're working on something. Once it works, then make a PR. You definitely can't modify -m 34000.

jsteube:
Keepass v4 is a very important algorithm to add to hashcat, and we really appreciate your work on it. You can use mode 34300, and we will reserve it for you.

I noticed that you initially tried to remove the byteswap from the Argon2 crypto library. That is the only change we cannot accept. However, the solution you implemented afterward is excellent. There is no need to worry about the additional byteswap overhead either.

We standardize all input to little-endian format across all interfaces. If an algorithm like yours needs to perform a byteswap before calling the library, it can result in a double byteswap. We allow this intentionally, because compilers detect and remove both swaps during optimization. This approach helps make development easier and more consistent for everyone.

jsteube:
Top! Many thanks

PR #4407, opened 2025-08-11
$HEX plains that end in whitespace

PenguinKeeper7:
For a good while, there have been multiple conversations about this topic - "If a plain ends in a whitespace, should it be $HEXed for improved readability?" and there are people who agree and disagree, so feel free to reject this PR. At very least, it can be an archive of the decision made.

The pro is that it'd be possible to see in a terminal that a plain has a whitespace character at the end of it and would make seeing these whitespaces in text editors more obvious.

The con is that it increases the number of semi-annoying $HEX conversions Hashcat does, forcing the user to manually unhex the plain.

Old behaviour:
```
a631788bc60c8bfdcc4edef7a318d18c:penguin 
```

New behaviour:
```
a631788bc60c8bfdcc4edef7a318d18c:$HEX[70656e6775696e20]
```

Example command:
```
> echo "penguin " | ./hashcat -m 0 a631788bc60c8bfdcc4edef7a318d18c --potfile-disable --quiet
```

(Also cleans up some slightly unsightly code in the hex checking function)

### DIFF
--- a/src/convert.c
+++ b/src/convert.c
@@ -189,42 +189,41 @@ size_t exec_unhexify (const u8 *in_buf, const size_t in_len, u8 *out_buf, const 
 
 bool need_hexify (const u8 *buf, const size_t len, const char separator, bool always_ascii)
 {
-  bool rc = false;
-
   if (always_ascii == true)
   {
     if (printable_ascii (buf, len) == false)
     {
-      rc = true;
+      return true;
     }
   }
   else
   {
     if (printable_utf8 (buf, len) == false)
     {
-      rc = true;
+      return true;
     }
   }
 
-  if (rc == false)
+  if (matches_separator (buf, len, separator) == true)
   {
-    if (matches_separator (buf, len, separator) == true)
-    {
-      rc = true;
-    }
+    return true;
   }
 
   // also test if the password is of the format $HEX[]:
 
-  if (rc == false)
+  if (is_hexify (buf, len))
   {
-    if (is_hexify (buf, len))
-    {
-      rc = true;
-    }
+    return true;
   }
 
-  return rc;
+  // check if the password ends in whitespace
+
+  if (len > 0 && isspace(buf[len - 1]))
+  {
+    return true;
+  }
+
+  return false;
 }
 
 void exec_hexify (const u8 *buf, const size_t len, u8 *out)
--- a/src/convert.c
+++ b/src/convert.c
@@ -218,7 +218,7 @@ bool need_hexify (const u8 *buf, const size_t len, const char separator, bool al
 
   // check if the password ends in whitespace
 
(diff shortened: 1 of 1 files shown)


roycewilliams:
Feature request (just me personally): make it a flag, so that the default behavior is not changed, but people can elect to opt into this approach.

PenguinKeeper7:
> Feature request (just me personally): make it a flag, so that the default behavior is not changed, but people can elect to opt into this approach.

For me personally, @roycewilliams I'd always want this on and if I didn't want the $HEX, I'd just universally disable hexxing through `--outfile-autohex-disable`
When would you want these whitespaces to be disabled in hexxing but also keep the others hexxed like `pass:word`?

roycewilliams:
@PenguinKeeper7 I'm spitballing here, but the goal is to make an orderly, backward-compatible transition from the current default behavior. Given the diversity of deployments that depend in a variety of ways on hashcat output, making a change to the output be optional, rather than the default, avoids unforeseen impact. Of course, in _theory_, anything that is "speaking potfiles" out there should already be able to handle de-HEXing as needed. But it's the unknown difference between theory and practice that I'm worried about.

roycewilliams:
@PenguinKeeper7 But I also don't want to be creating busywork for you. Might be best to wait to hear from @atom on his perspective (maybe even making it optional isn't wanted).
Just speaking personally, I would find benefit from being able to toggle it as needed.

PenguinKeeper7:
Yeah, I don't see any possible way that it'd break an existing deployment unless the code was very weak and didn't support $HEX. The number of plains that would be affected by this is absolutely dwarfed by the number that would be HEXed by default previous to this PR. Just using (a rather old) version of the HM wordlist, 415,657 more plains would be affected by this change, which is 0.04% of the total wordlist, compared to 11,444,708 already $HEXed plains, which is 1.17% of the whole. 

The only (fairly strange and probably unrealistic) situation I can think of, is if a program has a unit test or similar that expects a hash to be cracked and return in a hyper-specific way, without supporting the decoding of $HEX but I don't know why this would exist (especially given self-tests already exist in Hashcat) and it seems exceedingly rare.

As I mentioned in the original message, I'm more than happy for this to be rejected (or changed), as per Atom's wisdom, I'd just like to have the rejection on the record for when this inevitably comes up in the future, as it has in the past a few times

jsteube:
If it were just about me, I would leave it as it is, because I really do not like the HEX[] format (and it was not my idea). However, I can imagine that users who are less familiar with hashcat, especially in special cases like this, could run into serious trouble.

On the other hand, if there is an automated system that reads passwords from potfile, outfile, or stdout (and I think this PR will affect all of them), it does not really matter, because such systems should already support HEX[] format. In that case, it would not cause any problems.

This change is so minor that I do not think it is worth making a big deal out of it.

PR #4408, opened 2025-08-13
Python bridge improvements

thatux:
Allow Python bridge debugging with correct salt.
As discussed in https://hashcat.net/forum/thread-13346-post-63634.html#pid63634
Current implementation doesn't take the salt into account (outputting unsalted hash)

On the one hand this implementation is a bit dirty: you need to edit generic_hash_mp.py to enable dumping hashcat's ctx, edit generic_hash_mp.py again to disable it and add the dumped ctx to the file.

On the other hand it's quite clean: as we don't need to re-implement hashcat functionality i.e. build the ctx ourselves in Python. IMHO developers should be able to follow this train-of-thought, but maybe there is a better way of implementing this.

Even though pprint_bytes_runs() is a bit convoluted I do like it as it improves readability of the ctx struct that is passed by hashcat and pasted into the .py file.


jsteube:
That is a very important change, many thanks for the contribution! Could you please also add some entries docs/changes.txt (also don't forget the change of potfile handling), so it's not forgotten to mention in the release notes of the next release.

thatux:
Looks good to me

PR #4429, opened 2025-08-16
Improve install_modules.sh: add root and cpanm checks

thatux:
Improves experience for first time users


matrix:
We don't want to use root to install dependencies

PR #4438, opened 2025-08-19
Add variation of dahua Auth working with Besder cameras

zajc3w:
Module 24900, (dahua web Auth) has one minor difference to password hashing used by Besder cameras, module24901 works correctly with besder cameras that use xmeye sdk

Tested by discovering that firmware dump hash tzC42FmD:n51R6o[p
m24000 gives "({ZlIgk" for the same hash (first hit)


jsteube:
Super nice, many thanks for the contribution! I've added the missing unit test with commit: https://github.com/hashcat/hashcat/commit/035157b17cffb104b4694b5aa83111718b7d5ec3

PR #4444, opened 2025-08-20
Rust bridge

fukc-gihtub:
The Rust bridge.

To implement your own hashing algorithm, edit `Rust/generic_hash/src/generic_hash.rs`.


jsteube:
This is an amazing contribution, thank you very much. Motivated me also to do more Rust :)

PR #4449, opened 2025-08-22
test.pl enable --help before installing modules

### LINKED ISSUE #4413: test.pl early help output
As a "nice to have", `test.pl --help` should show help/usage even if many hashing-specific dependencies are not present.

```
$ test.pl --help
Can't locate Digest/MD4.pm in @INC (you may need to install the Digest::MD4 module) (@INC contains:  [...]
```

thatux:
Fixes https://github.com/hashcat/hashcat/issues/4413

### test.pl used to be
```
echo aa | tools/test.pl passthrough 0                                             
Can't locate Data/Types.pm in @INC (you may need to install the Data::Types module) (@INC entries checked: /home/k/.perl5/lib/perl5/5.38.2/x86_64-linux-gnu-thread-multi /home/k/.perl5/lib/perl5/5.38.2 /home/k/.perl5/lib/perl5/x86_64-linux-gnu-thread-multi /home/k/.perl5/lib/perl5 /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.38.2 /usr/local/share/perl/5.38.2 /usr/lib/x86_64-linux-gnu/perl5/5.38 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.38 /usr/share/perl/5.38 /usr/local/lib/site_perl) at tools/test.pl line 11.
BEGIN failed--compilation aborted at tools/test.pl line 11.
```

### test.pl improved

```
echo aa | tools/test.pl passthrough 0      
Missing Perl modules, read: docs/hashcat-plugin-development-guide.md (search test.pl), and run: ./tools/install_modules.sh
```

Nice thing is that also test.sh warns about installing modules now:

### test.sh used to be
```
./tools/test.sh -m 34000-34300 -D1 
Can't locate Data/Types.pm in @INC (you may need to install the Data::Types module) (@INC entries checked: /home/k/.perl5/lib/perl5/5.38.2/x86_64-linux-gnu-thread-multi /home/k/.perl5/lib/perl5/5.38.2 /home/k/.perl5/lib/perl5/x86_64-linux-gnu-thread-multi /home/k/.perl5/lib/perl5 /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.38.2 /usr/local/share/perl/5.38.2 /usr/lib/x86_64-linux-gnu/perl5/5.38 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl-base /usr/lib/x86_64-linux-gnu/perl/5.38 /usr/share/perl/5.38 /usr/local/lib/site_perl) at tools/test.pl line 11.
```

### test.sh improved
```
./tools/test.sh -m 34000-34300 -D1   
Missing Perl modules, read: docs/hashcat-plugin-development-guide.md (search test.pl), and run: ./tools/install_modules.sh
```

### DIFF
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -8,11 +8,22 @@
 use strict;
 use warnings;
 
-use Data::Types qw (is_count is_whole);
 use File::Basename;
 use FindBin;
 use List::Util 'shuffle';
-use Digest::MD4 qw (md4_hex);
+
+my $TYPES = [ 'edge', 'single', 'passthrough', 'potthrough', 'verify' ];
+
+my $TYPE = shift @ARGV;
+my $MODE = shift @ARGV;
+
+is_in_array ($TYPE, $TYPES) or usage_exit ();
+
+eval {
+    require Data::Types;  Data::Types->import(qw(is_count is_whole));
+    require Digest::MD4;  Digest::MD4->import('md4_hex');
+    1;
+} or die "Missing Perl modules, read: docs/hashcat-plugin-development-guide.md (search test.pl), and run: ./tools/install_modules.sh\n";
 
 # allows require by filename
 use lib "$FindBin::Bin/test_modules";
@@ -23,13 +34,6 @@ if (exists $ENV{"IS_OPTIMIZED"} && defined $ENV{"IS_OPTIMIZED"})
 {
   $IS_OPTIMIZED = $ENV{"IS_OPTIMIZED"};
 }
-
-my $TYPES = [ 'edge', 'single', 'passthrough', 'potthrough', 'verify' ];
-
-my $TYPE = shift @ARGV;
-my $MODE = shift @ARGV;
-
-is_in_array ($TYPE, $TYPES) or usage_exit ();
 
 is_whole ($MODE) or die "Mode must be a number\n";
 


jsteube:
Do I understand this correct it will only check for MD4 to be installed, and otherwise print the error? That is because MD4 is not installed by default?

thatux:
Yes, checks for md4 and datatypes. Making sure we have all dependencies is much more of a hassle I think, this will cover 99% of new user I expect

roycewilliams:
I apologize, I should have been more clear in my original specification of the problem. 

I like that it tells users what they can do to fix the problem. But part of my goal was to not just be helpful but to show the actual command usage as well. 

I also think that perl's natural behavior (to show which module is trying to be used, but missing) is good.

thatux:
I think it does what you want? It prints
`and run: ./tools/install_modules.sh`

That'd install all required perl modules

I don't think it'd make a lot of sense to show which specific modules people need to install, if that's what you mean

roycewilliams:
Let's say there's a problem with the module installer, script, and not all of the modules that are needed get installed. Now how will the user determine which modules are missing?

Perl's natural behavior should be preserved here. What I think is the best solution: 

1. Show the general usage, as if the script was run with no parameters. 
2. Show the feedback message that is now shown, telling them what to do to fix the problem. 
3. Instead of exiting, allow the script to continue so that it will print the standard Perl message about exactly which module failed to load.

roycewilliams:
My apologies, it looks like the full behavior looks like I was expecting. Thanks, and sorry for the confusion!

PR #4453, opened 2025-08-25
Make Python plugin development more robust

AmmoniaAvenue:
Using the Python plugin development was somewhat inconvenient. I tried to improve this for faster PoC development by:

- calling  `term()` after dumping ctx such that hashcat CLI will terminate gracefully
- dump hashcat `ctx`  to a Python pickle file which is loaded automatically instead of copy-paste std-out to your python script
- remove `pprint` helper functions in `hcshared.py` , this is not needed anymore because the Python pickle object can be written and loaded with standard libs and function calls
- dump hashcat `ctx` only when the Python plugin is called through hashcat CLI (otherwise the `ctx` is incorrect) and not when it is in debug mode
- include `example.ctx` file such that the example Python plugin runs automatically
- test if loaded hashcat `ctx` is correct by checking the input selftest hash against selftest hash in hashcat `ctx`  

I think this makes the use and development of Python plugins a lot easier and cleaner.


thatux:
Looks good 👍

Great idea writing it to a file!

AmmoniaAvenue:
Changed PR to draft because the lossy verification does not work as intended.

jsteube:
Let me know when to merge

AmmoniaAvenue:
Resolved review comments from thatux.  Ready to merge and thanks for checking!

jsteube:
Many thanks for contribution!

PR #4454, opened 2025-08-25
Small cleanup of 18600, 22800 and 34100 kernel calls with new KERNEL_FA

fse-a:
Small cleanup of the 22800 and 34100 kernels where the new FERNEL_FA define was missing.

Reverted Blowfish generalization for 18600 because only the `comp()` kernel should use the `FIXED_LOCAL_SIZE_COMP` (it is slightly different from other Blowfish plugins). This can probably still use the generalization if some changes are made.

Kind regards,

Ewald & Pelle


OpenCL/m18600-pure.cl
```diff
@@ -11,7 +11,6 @@
 #include M2S(INCLUDE_PATH/inc_platform.cl)
 #include M2S(INCLUDE_PATH/inc_common.cl)
 #include M2S(INCLUDE_PATH/inc_simd.cl)
-#include M2S(INCLUDE_PATH/inc_cipher_blowfish.cl)
```
matrix: @fse-a why?

src/modules/module_18600.c
```diff
@@ -330,7 +421,7 @@ void module_init (module_ctx_t *module_ctx)
   module_ctx->module_hook23                   = MODULE_DEFAULT;
   module_ctx->module_hook_salt_size           = MODULE_DEFAULT;
   module_ctx->module_hook_size                = MODULE_DEFAULT;
-  module_ctx->module_jit_build_options        = blowfish_module_jit_build_options;
```
matrix, line 333: @fse-a same why

OpenCL/m18600-pure.cl
fse-a: While looking at recent changes (i.e. the KERNEL_FA addition), we noticed that some kernels where missing this define or used it differently (the 18600). After looking at the changes to the 18600, we noticed that the generalization changed the original behavior, i.e. the `FIXED_LOCAL_SIZE` should only apply to the `comp()` kernel. After consulting with @jsteube we reverted the change.

matrix: Okay, but at this point, why not try updating the code inside inc_cipher_blowfish.cl to handle this change?

fse-a: We still need to test this, but we could add the following to the 18600 kernel (just after the includes):

```
#ifdef FIXED_LOCAL_SIZE
#define FIXED_LOCAL_SIZE_COMP FIXED_LOCAL_SIZE
#undef FIXED_LOCAL_SIZE
#undef KERNEL_FA
#define KERNEL_FA
#endif
```

In the `comp()` function we can then change `FIXED_LOCAL_SIZE` to `FIXED_LOCAL_SIZE_COMP`. The generalization in the plugin and kernel can then remain as-is except for these small changes.

Maybe not that pretty, but at least a minimal change. What do you think?

jsteube:
The change in the module is fine as proposed. This is an exceptional use case which only occurs for this hash mode. If it really is just one case, then it is acceptable to handle it inside the module, as it also helps keep the generalized library more consistent.

As for the kernel, I do not think much needs to be changed besides the kernel declaration line and reverting it to how it was before the generalization. The main change required in the kernel is this:

```
KERNEL_FQ KERNEL_FA FIXED_THREAD_COUNT(FIXED_LOCAL_SIZE_COMP) void m18600_comp (KERN_ATTR_TMPS_ESALT (odf11_tmp_t, odf11_t))
```

And then also lines which depend on FIXED_LOCAL_SIZE and must be rewritten to FIXED_LOCAL_SIZE_COMP. Other than that, there should be no change needed in the kernel.

matrix:
> The change in the module is fine as proposed. This is an exceptional use case which only occurs for this hash mode. If it really is just one case, then it is acceptable to handle it inside the module, as it also helps keep the generalized library more consistent.
> 
> As for the kernel, I do not think much needs to be changed besides the kernel declaration line and reverting it to how it was before the generalization. The main change required in the kernel is this:
> 
> ```
> KERNEL_FQ KERNEL_FA FIXED_THREAD_COUNT(FIXED_LOCAL_SIZE_COMP) void m18600_comp (KERN_ATTR_TMPS_ESALT (odf11_tmp_t, odf11_t))
> ```
> 
> And then also lines which depend on FIXED_LOCAL_SIZE and must be rewritten to FIXED_LOCAL_SIZE_COMP. Other than that, there should be no change needed in the kernel.

Since an ifdef is sufficient to reuse the generic code, I would have used it instead of restoring all the code. However, if you prefer this approach, that is acceptable to me.

jsteube:
I've updated the 18600 kernel, should be fine now

PR #4459, opened 2025-08-27
Add vectorized line scanning (AVX2/AVX512/NEON) with runtime dispatch

matrix:
- Introduced vectorized newline scanning routines using AVX2, AVX512, or NEON intrinsics for faster wordlist parsing compared to memchr().
- thread_next() and thread_seek() now transparently use the optimized implementation when available:
    * On x86/x86_64: AVX512 or AVX2 selected at runtime.
    * On Apple Silicon (arm64): NEON via sse2neon.
    * Fallback to generic memchr() if no SIMD path is available.
- Ensures portability across Linux, Windows, and macOS (Intel + Apple Silicon) without requiring build-time AVX feature flags.

During development, I noticed that compiling hashcat using clang/clang++ even on Linux resulted in a performance increase on -a 8.

Thanks

### DIFF
--- a/include/shared.h
+++ b/include/shared.h
@@ -131,12 +131,12 @@ bool get_free_memory (u64 *free_mem);
 u32 previous_power_of_two (const u32 x);
 u32 next_power_of_two (const u32 x);
 
-typedef size_t (*vector_nl_scan_t) (const u8 *ptr, size_t max_len);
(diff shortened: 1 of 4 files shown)


jsteube:
Please add support to specify a character, and afterwards rename to hc_memchr() or so

PR #4460, opened 2025-08-27
Make Argon2 more robust

thatux:
Prevent argon2 hashes with parameters all 0 to crash hashcat.

```
[ test_1756335826 ] > Init test for hash type 34000.
[ test_1756335826 ] [ Type 34000, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1756335826 ] [ Type 34000, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1756336403 ] > Init test for hash type 34000.
[ test_1756336403 ] [ Type 34000, Attack 0, Mode multi,  Device-Type Cpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1756336403 ] [ Type 34000, Attack 0, Mode multi,  Device-Type Cpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1756335826 ] > Init test for hash type 34300.
[ test_1756335826 ] [ Type 34300, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1756335826 ] [ Type 34300, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1756336445 ] > Init test for hash type 34300.
[ test_1756336445 ] [ Type 34300, Attack 0, Mode multi,  Device-Type Cpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1756336445 ] [ Type 34300, Attack 0, Mode multi,  Device-Type Cpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

### DIFF
--- a/src/modules/argon2_common.c
+++ b/src/modules/argon2_common.c
@@ -64,7 +64,12 @@ const char *argon2_module_extra_tuningdb_block (MAYBE_UNUSED const hashconfig_t 
 
   const u64 available_mem = MIN (device_param->device_available_mem, (device_param->device_maxmem_alloc * 4)) - (fixed_mem + spill_mem);
 
-  const u32 kernel_accel_max = (device_param->device_host_unified_memory == true) ? (available_mem / 2) / size_per_accel : available_mem / size_per_accel;
+  u32 kernel_accel_max = (kernel_accel_user>0) ? kernel_accel_user : 1;
+
+  if (size_per_accel>0)
+  {
+     kernel_accel_max = (device_param->device_host_unified_memory == true) ? (available_mem / 2) / size_per_accel : available_mem / size_per_accel;
+  }
 
   u32 kernel_accel_new = device_processors;
 
--- a/OpenCL/m34000-pure.cl
+++ b/OpenCL/m34000-pure.cl
@@ -14,6 +14,10 @@
 #define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
 #define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
 
+#ifndef ARGON2_PARALLELISM
+#define ARGON2_PARALLELISM 0
+#endif
+
 typedef struct argon2_tmp
 {
   u32 state[4];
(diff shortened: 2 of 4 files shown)


src/modules/argon2_common.c
```diff
@@ -64,7 +64,12 @@ const char *argon2_module_extra_tuningdb_block (MAYBE_UNUSED const hashconfig_t
 
   const u64 available_mem = MIN (device_param->device_available_mem, (device_param->device_maxmem_alloc * 4)) - (fixed_mem + spill_mem);
 
-  const u32 kernel_accel_max = (device_param->device_host_unified_memory == true) ? (available_mem / 2) / size_per_accel : available_mem / size_per_accel;
+  u32 kernel_accel_max = (kernel_accel_user>0) ? kernel_accel_user : 1;
```
thatux, line 67: value makes sense, but not 100% sure this is the best default value

jsteube:
I think is not the right approach. We should prevent the modules from accepting zero values and reject the hash in such a case.

thatux:
Ok, agreed. The most sensible check is to check the argon2 config in every decode() where argon2 is used, but that's already the case so it's ok as is :+1: 

34000 and 34300
```
  if (argon2_options->version != 19 && argon2_options->version != 16) return (PARSER_HASH_VALUE);
  if (argon2_options->memory_usage_in_kib < 1) return (PARSER_HASH_VALUE);
  if (argon2_options->iterations < 1) return (PARSER_HASH_VALUE);
  if (argon2_options->parallelism < 1 || argon2_options->parallelism > 32) return (PARSER_HASH_VALUE);
```

34100
```
  // argon2 - memory usage in KB (m)
  const u8 *mem_pos = token.buf[3];
  const u32 mem = hc_strtoul ((const char *) mem_pos + 2, NULL, 10);
  if (mem < 1) return (PARSER_HASH_VALUE);
```

PR #4465, opened 2025-08-30
Add docs/hashcat-rust-plugin-quickstart.md

fukc-gihtub:


jsteube:
Looks very good, many thanks already!

The parameter `--rounds 5` is the default, right? So why mention?

fukc-gihtub:
So that the user could try different costs.

PR #4468, opened 2025-08-30
Dynamic hash plugin

fukc-gihtub:
The plugin takes an algorithm description as `--bridge-parameter2`.

```
$ echo codebreaker | \
./hashcat -m 74000 \
    '$2b$12$iyvX.kDndyq/YiUvv7J.NePApguNMJG65lr2k7H0A7Y3d7LLc1tOS*12*iyvX.kDndyq/YiUvv7J.Ne' \
    --bridge-parameter1 ./bridges/subs/dynamic_hash.so \
    --bridge-parameter2 'bcrypt2b(cost=$s1,salt=$s2,hmac_sha256:b64(key=$s2,$p))'
```

where `$s1` is 12:

`$2b$12$iyvX.kDndyq/YiUvv7J.NePApguNMJG65lr2k7H0A7Y3d7LLc1tOS*`**12**`*iyvX.kDndyq/YiUvv7J.Ne`

and `$s2` is `iyvX.kDndyq/YiUvv7J.Ne`:

`$2b$12$iyvX.kDndyq/YiUvv7J.NePApguNMJG65lr2k7H0A7Y3d7LLc1tOS*12*`**iyvX.kDndyq/YiUvv7J.Ne**

Use `dynamic_hash.dll` and double quotes on Windows. If there's a single 'salt', use `$s`.

---

Other things you can do:

- NTLM:
    `md4(utf16le($p))`
- MySQL5:
    `upper(sha1(sha1:bin($p)))`
- Half-MD5:
    `cut(0,16,md5($p))`
- MD5($pass.$salt) with auto-decoding hex-encoded salt:
    `md5($p.$s:unhex)`
- SHA256(Base64($pass)):
    `sha256(b64($p))`
- Base64(SHA256_raw($pass)):
    `b64(sha256:bin($p))` or just `sha256:b64($p))`
- Peppers as string literals:
    `md4("--".$p."--".$s."--")`
    
See `tests.rs` for more examples.


jsteube:
Just applied the PR. When started: 

`ERROR: ./Rust/dynamic_hash/target/release/libdynamic_hash.so: No such file or directory`

So I build it manually using `cargo build --release`. 

Then it loads, but the command in the PR description:

```
$ echo codebreaker | \
./hashcat -m 74000 \
    '$2b$12$iyvX.kDndyq/YiUvv7J.NePApguNMJG65lr2k7H0A7Y3d7LLc1tOS*12*iyvX.kDndyq/YiUvv7J.Ne' \
    --bridge-parameter1 ./Rust/dynamic_hash/target/release/libdynamic_hash.so \
    --bridge-parameter2 'bcrypt2b(cost=$s1,salt=$s2,hmac_sha256:b64(key=$s,$p))'
```

It doesn't crack: `Status...........: Exhausted`

jsteube:
I think the change needs to be `key=$s,$p` -> `key=$s2,$p`

fukc-gihtub:
Fixed. And yes, you should build the plugin manually for now. There are no changes to the Makefiles.

fukc-gihtub:
It would be nice if plugins had a way to tell Hashcat not to run self-tests.
Right now, this is a bit of a hack:
```
   if password == b"__dynamic_hash__" {
        return vec![String::from("__dynamic_hash__")];
    }
```

jsteube:
> It would be nice if plugins had a way to tell Hashcat not to run self-tests. Right now, this is a bit of a hack:
> 
> ```
>    if password == b"__dynamic_hash__" {
>         return vec![String::from("__dynamic_hash__")];
>     }
> ```

If both are set to NULL, self-test is not performed.  Probably easiest way would be to patch bridge code to transport back NULL if for instance it's set to None?

fukc-gihtub:
I only had to change my Rust code a bit. No self-tests now.

fukc-gihtub:
`make` should build everything now.

jsteube:
I think we should make:

```
        $(RM) -rf Rust/generic_hash/target
        $(RM) -rf Rust/dynamic_hash/target
        $(RM) -rf Rust/feeds/dummy/target
```

like this:

```
        $(RM) -rf Rust/bridges/*/target
        $(RM) -rf Rust/feeds/*/target
```

And then we move the two bridge plugins to `bridges/`. Here's more detail:

-----

But I understand that you want to separate the "final product" somehow. That makes sense.

Let's rethink folder structure. Right now we 3 types of plugins:

- Feeds
- Bridges
- Modules

Modules end up in `modules/` and are final - they don't have sub-plugins.
Feeds end up in `feeds/` and are final - they don't have sub-plugins.
Bridges end up in `bridges/` and can be final, but some of them have sub-plugins, for instance the Python bridge and the Rust bridge. 

When I created the Python bridge, that is most similar to the Rust bridge, I created the `Python/` folder to place the sub-plugins, mostly because users can edit them directly.

But Rust bridge added another layer of complexity. We need to build these sub-plugins before they can be used. Therefore we need to some extra integration. I think the best way to keep the sources for these sub-plugins in `Rust/` but we move them to `Rust/bridges/` so they align better with the feeds sources. Once they are compiled, we move the output of them to `bridges/subs/`.

My suggestion would be:

1. move the two existing rust bridge sub sources from `Rust/` to `Rust/bridges/` 
2. Create the folder `bridges/subs/`
3. move the final product from the rust bridge subs (the *.so files) to `bridges/subs/`
4. Remove `plugins/` which is a too generic name (because we have 3 types of plugins)

fukc-gihtub:
I've made the proposed changes and verified that BinaryPackage.ubuntu20 works with the new code.

jsteube:
Fantastic!

jsteube:
This has grown into one of the most capable thing in the bridge tree and it is written down nowhere. 

Could you add a short page for it? You wrote "docs/hashcat-rust-plugin-quickstart.md" in #4465 and that is the shape I have in mind, 100 lines or so and one worked example that runs.

Four things carry the most for a reader who has only seen the changelog line:

- the function list, so the reach is visible without reading eval.rs
- $p and $s, and that a salt with * separators also binds $s1, $s2 and so on
- the named parameters, cost=, rounds=, dklen= and key=, which is what makes a bcrypt or PBKDF2 construction expressible rather than only the fast ones
- the output suffix, :hex, :b64 and :raw, since a digest stored in base64 needs it, and it is the one piece a reader cannot guess from an example

Issue #4467, opened 2025-08-30
Example hashes update


braindamagedjustnow:
When you have time can you update:
https://hashcat.net/wiki/doku.php?id=example_hashes

to match the recent version:
https://github.com/hashcat/hashcat/blob/master/docs/hashcat-example-hashes.md

roycewilliams:
This is actually harder than it looks - the formats don't directly map in a way that lends to clean conversion. We are discussing options. In the meantime, I did make a pass to update the easy ones.

thatux:
@roycewilliams you've seen this right? hopefully it helps https://github.com/hashcat/hashcat/blob/master/.github/workflows/hashcat-example-hashes_md2docuwiki.sh

roycewilliams:
Oh, definitely, and thanks! Once the other metadata is available (or deemed superfluous), this is exactly the path to take.

Issue #3958, opened 2024-02-28
add way to encode and display per-kernel advice/warnings


roycewilliams:
Kernel developers should have a way to specify a note/warning field for their module, that will be displayed to the user, perhaps near where the min/max supported password length is shown at startup.

Reasoning:

The concept of *length* trade-offs in optimized vs pure kernels is relatively well-revealed in the hashcat interface.

But some kernels have other trade-offs that are implemented for speed. For example, some byte encoding is simplified in the optimized kernel for NTLM that would make some non-ASCII passwords never be cracked, even if they are under the optimized byte limit.

Any existing trade-offs could then be retroactively documented and communicated to users dynamically.

alex343342:
01020407080b0d0e10131516191a1c1f20232526292a2c2f31323437383b3d3e40434546494a4c4f51525457585b5d5e61626467686b6d6e70737576797a7c7f80838586898a8c8f91929497989b9d9ea1a2a4a7a8abadaeb0b3b5b6b9babcbfc1c2c4c7c8cbcdced0d3d5d6d9dadcdfe0e3e5e6e9eaeceff1f2f4f7f8fbfdfen ojibj io jboi

PR #4483, opened 2025-09-09
Add a usage and advice notice to each module

thatux:
Added optional possibility to add additional info this allows module authors to add usage instructions/how to generate hash to their module.

Furthermore, it allows us to include the footnotes from https://hashcat.net/wiki/doku.php?id=example_hashes in to the modules themselves such that they do not need to be manually administered by @roycewilliams (https://github.com/hashcat/hashcat/issues/4467).

Example passwords and hashes are already maintained within each module, so makes sense to also keep some instructions there; it is very similar to `deprecated_notice`.

This implementation adds module_comment as an optional callback. Not every module needs to declare it in it's source: defaults of `N/A`. The only thing I think this breaks are pre-compiled modules (without any .c source), but I think those are very rare.

Example of footnotes:
```
4 The hash used here is not the one sent via e.g. the web interface to LastPass servers (pbkdf2_sha256_hex (pbkdf2_sha256 ($pass, $email, $iterations), $pass, 1) but instead the one stored (by e.g. your browser or the pocket version) to disk. For instance, Opera and Chrome store the hash in local SQLite databases; Firefox uses files ending with “lpall.slps” - for Linux: 2nd line is interesting / base64 decode it; for Windows, see here - and_key.itr
5 You can consider the second part as a “salt”. If it is equal to 00000000, the CRC32 code will be considered as “not salted”
6 The raw sha256 output is used for base64() encoding (not the hexadecimal output)
7 The format is hash:salt:id
10 This file actually contains several examples of the different hash+cipher combinations. The password is stored in the pw file.
11 You can use itunes_backup2hashcat to extract the hashes from the Manifest.plist file
12 Password: “hashcat!!!”. Min/max password length is exactly 10 characters/bytes.
13 You can use AxSuite by Fist0urs to retrieve the hashes.
```
Example output
```
$ ./hashcat -HH -m34300 | grep -i usage                                                          
  Usage.Notice........: To extract hash from KDBX use https://github.com/openwall/john/blob/bleeding-jumbo/src/keepass2john.c
```

```
$ ./hashcat -a3 -m1000 'b4b9b02e6f09a9bd760f388b67351e2b' '?a?a?a?a?a?a?a'                       
hashcat (v7.0.0-beta.1-306-g972f67ba0+) starting

[snip]

Module 1000 advice notice: When using the optimized-kernel: some byte encoding is simplified that would make some non-ASCII passwords never be cracked, even if they are under the optimized byte limit. The pure-kernel implements true UTF16 handling so doesn't have this problem and should crack all passwords.
```

<details>

TODO
- [x] Might be good to limit the character limit on the comment, to 140 characters. I'm not sure whether we really want this, maybe manual check per commit is fine. If we want to do it automatically we could use a linter? --> too preliminar to implement, an URL like this is already 71 characters 'https://github.com/openwall/john/blob/bleeding-jumbo/src/keepass2john.c'

</details>


PenguinKeeper7:
Related: https://github.com/hashcat/hashcat/issues/3958

roycewilliams:
> Related: #3958

Good callback! Maybe two different fields - general note, vs caveats/warnings?

thatux:
I don't have a very strong opinion, but I'm a bit hesitant to split between warning and a  note: not sure how much both will be used. Is it worthwhile to make that distinction?

Are there many more modules where we specifically like a warning to be printed? @roycewilliams 

I think most modules would at least benefit a note describing how to generate the hash i.e. what hashcat2x or john2x script to use.



Implementation could look like this:
- Warnings are always printed when running a attack.
- Usage notes are printed upon request -HH -m, or when the attack is run without a hash

I like usage_note better than comment so I'll change that.

roycewilliams:
> Yeah, that makes sense!

thatux:
Maybe call the warning a 'module advice' instead? I now use event_log_advice() instead of event_log_warning(), which makes sense to me

thatux:
@roycewilliams I've added all footnotes from the table at https://hashcat.net/wiki/doku.php?id=example_hashes to the relevant files; I don't think these are currently in use:
```
<sup>3</sup> Same format as in <sup>2</sup> but the number of rounds <strong>must</strong> be specified<br/>
<sup>6</sup> The raw sha256 output is used for base64() encoding (not the hexadecimal output)<br/>
<sup>21</sup> you can extract the hashes with <a href="[https://github.com/0x6470/bitwarden2hashcat](view-source:https://github.com/0x6470/bitwarden2hashcat)" class="urlextern" title="https://github.com/0x6470/bitwarden2hashcat" rel="ugc nofollow">https://github.com/0x6470/bitwarden2hashcat</a><br/>
```

I've added `<sup>21</sup>` to `23400`, the other two I'm not sure about..

thatux:
Looks good I think: https://github.com/hashcat/hashcat/blob/401acd32be2d3fa63adcdf81c3424ac3c4b0f2fe/docs/hashcat-example-hashes.md

include/common.h
```diff
@@ -145,3 +145,6 @@ but this is needed for VS compiler which doesn't have inline keyword but has __i
 
 #endif // HC_COMMON_H
 
+
+#define USAGE_NOTICE_NUM_ROUNDS "rounds=[# of iterations] is optional here, after signature, e.g. $5$rounds=5000"
```
jsteube, line 149: Cool example!

src/interface.c
```diff
@@ -122,6 +122,8 @@ int hashconfig_init (hashcat_ctx_t *hashcat_ctx)
 
   if (rc_load == false) return -1;
 
+  module_ctx->module_usage_notice = MODULE_DEFAULT; // set all module to have usage_notice by empty default; such that this property doesn't have to be declared explicitly by all modules (such that we don't break private plugins without this options)
```
jsteube, line 125: I kind of like the idea that we no longer need to update all existing plugins because of a newly added callback, however it's the first time we would do this. Did you check if this works on windows, too?

src/modules/module_01000.c
```diff
@@ -52,6 +52,12 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig,
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+const char *module_advice_notice (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  return "Some byte encoding is simplified in the optimized kernel for NTLM that would make some non-ASCII passwords never be cracked, even if they are under the optimized byte limit.";
```
jsteube: Only in -O mode. Pure mode implements true UTF16 handling.

src/modules/module_14800.c
```diff
@@ -49,6 +49,11 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig,
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+const char *module_usage_notice (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  return "You can use https://github.com/philsmd/itunes_backup2hashcat/ to extract the hashes from the Manifest.plist file";
```
jsteube, line 54: Phil also has https://github.com/philsmd/7z2hashcat

src/interface.c
thatux: Thanks! Just tested on Windows and there it also works  (built for Windows with [WSL](https://github.com/hashcat/hashcat/blob/master/BUILD_WSL.md):
```
$ file hashcat.exe
hashcat.exe: PE32+ executable (console) x86-64 (stripped to external PDB), for MS Windows

$ ./hashcat.exe -HH -m34300 | grep -i usage
  Usage.Notice........: To extract hash from KDBX use https://github.com/openwall/john/blob/bleeding-jumbo/src/keepass2john.c

$ ./hashcat.exe -a3 -m1000 'b4b9b02e6f09a9bd760f388b67351e2b' '?a?a?a?a?a?a?a'
Module advice notice: Some byte encoding is simplified in the optimized kernel for NTLM that would make some non-ASCII passwords never be cracked, even if they are under the optimized byte limit.
```

src/modules/module_01000.c
thatux: Thanks, tried to make this more clear by starting with this distinction.

src/modules/module_14800.c
thatux: Thanks! I've added that to 11600. I'm sure there are many more examples :smile: let me know!

jsteube:
OK, let's merge and see if there's anything we didn't think of..

Issue #3936, opened 2024-01-23
ssh modes: Hash 'hast.txt': Separator unmatched


noraj:
**Describe the bug**

`hash.txt` (from hashcat example)

```
[...]

PR #4485, opened 2025-09-09
Improve hash parse errors

thatux:
Accidentally providing hashcat a path for a non-existent hashfile results in a somewhat unexpected/unclear/unrelated error:
```
./hashcat -m34301 -a3 '/tmp/thisfiledoesntexist.hash' '?a?a?a?a?a?a' --potfile-disable
Hash '/tmp/keepass4_keepassxc_2.7.9_aeskdf_default_settings.hash': Separator unmatched
No hashes loaded.
```
Especially for new users this might be an unexpected error e.g. https://github.com/hashcat/hashcat/issues/3936 `You misspelled "hash.txt" in your command so Hashcat can't find your file.`

So provide some guidance on what is happening.

```
Hash was parsed as a commandline argument (not as a file, maybe the file doesn't exist?)
Hash parsing error: '/tmp/keepass4_keepassxc_2.7.9_aeskdf_default_settings.hash': Separator unmatched
No hashes loaded.
```

I think that it's difficult to solve this problem (in a different way), we're not sure whether users are passing a hash or an hashfile as argument.

### DIFF
--- a/src/hashes.c
+++ b/src/hashes.c
@@ -1376,6 +1376,7 @@ int hashes_init_stage1 (hashcat_ctx_t *hashcat_ctx)
           }
           else
           {
+            event_log_warning (hashcat_ctx, "Hash was parsed as a commandline argument (not as a file, maybe the hashfile doesn't exist?)");
             event_log_warning (hashcat_ctx, "Hash '%s': %s", input_buf, strparser (parser_status));
           }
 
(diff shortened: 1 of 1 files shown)


jsteube:
Exactly, we can't really know what the user wants a file or a hash, but the changed error message might improve the situation for new users.

PR #4493, opened 2025-09-15
add modulenumbers to 2hashcat scripts, and add 'how to extract' advice to modules

thatux:


thatux:
Let's merge this, when I find more, I'll make a new PR.

PR #4507, opened 2025-09-22
Improve GPG CAST5 salt iteration handling and eliminate TODO

Oblivionsage:
- Replace magic number 8 with type-safe GPG_S2K_MIN_ITERATIONS constant
- Add comprehensive documentation explaining OpenPGP RFC 4880 S2K Type 1 behavior
- Clarify that protocol requires 0 iterations but kernel needs positive loop count
- Eliminate TODO confusion about salt_iter logic

Resolves TODO in module_17040.c about salt iteration handling

### DIFF
--- a/src/modules/module_17040.c
+++ b/src/modules/module_17040.c
@@ -10,6 +10,9 @@
 #include "convert.h"
 #include "shared.h"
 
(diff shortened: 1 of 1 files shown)


thatux:
Cool 👍

FYI yesterday I worked on supporting iterated hashes with salts, hope to push it soon. Needs some more testing and I hope to include tests.

PR #4512, opened 2025-09-25
Add Wordpress bcrypt

PenguinKeeper7:
Apparently Wordpress have changed their hashing algorithms, with the new main choice being `bcrypt(base64(hmac-sha384_raw($plain)))`, using `wp-sha384` as the hmac key and a `$wp` prefix signature.
https://github.com/WordPress/WordPress/blob/6.8.1/wp-includes/pluggable.php#L2706-L2709
```
$ ./tools/test.sh -m 35500 -a all -t all -o win -P
[ test_1758841868 ] > Init test for hash type 35500.
[ test_1758841868 ] [ Type 35500, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1758841868 ] [ Type 35500, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1758841868 ] [ Type 35500, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1758841868 ] [ Type 35500, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

Satisfies: https://hashcat.net/forum/thread-13369-post-63813.html and additional user reports

### DIFF
--- a/OpenCL/m35500-pure.cl
+++ b/OpenCL/m35500-pure.cl
@@ -63,7 +63,6 @@ KERNEL_FQ KERNEL_FA void m35500_init (KERN_ATTR_TMPS (bcrypt_tmp_t))
   if (gid >= GID_CNT) return;
 
   // "wp-sha384" key
-  // Only works with 32-length array, unsure why
   const u32 key[32] = { 0x732D7077, 0x38336168, 0x00000034, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0};
 
   sha384_hmac_ctx_t ctx0;
--- a/OpenCL/m35500-pure.cl
+++ b/OpenCL/m35500-pure.cl
@@ -1,0 +1,502 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_blowfish.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha384.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+CONSTANT_VK u32 bin2base64[0x40] =
+{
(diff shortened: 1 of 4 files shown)


fukc-gihtub:
`const u32 key[32] = { 0x732D7077, 0x38336168, 0x00000034 };` would be valid, too.

See § 6.7.9 of the C17 standard, paragraphs 21 and 10.

PenguinKeeper7:
> `const u32 key[32] = { 0x732D7077, 0x38336168, 0x00000034 };` would be valid, too.

Sure, but that's not the precedent in Hashcat for declaring partially filled arrays. See:
https://github.com/hashcat/hashcat/blob/master/OpenCL/m29000_a0-pure.cl#L46
https://github.com/hashcat/hashcat/blob/master/OpenCL/m27200_a0-pure.cl#L30
https://github.com/hashcat/hashcat/blob/master/OpenCL/m19500_a0-pure.cl#L83
https://github.com/hashcat/hashcat/blob/master/OpenCL/m05800-optimized.cl#L329-L331
https://github.com/hashcat/hashcat/blob/master/OpenCL//inc_zip_inflate.cl#L504-L506
https://github.com/hashcat/hashcat/blob/master/OpenCL/inc_luks_twofish.cl#L46
https://github.com/hashcat/hashcat/blob/master/OpenCL/inc_diskcryptor_xts.cl#L126
https://github.com/hashcat/hashcat/blob/master/OpenCL/inc_luks_aes.cl#L50

fukc-gihtub:
`const int s_dist_extra[32] = { 0, 0, 0, 0, 1, 1, 2, 2, 3, 3, 4, 4, 5, 5, 6, 6, 7, 7, 8, 8, 9, 9, 10, 10, 11, 11, 12, 12, 13, 13 };`

Length 32, but only 30 values.

https://github.com/hashcat/hashcat/blob/4bd63ffa2b3a46cb93dec7bf49347e3a00c2d26e/OpenCL/inc_zip_inflate.cl#L507

PenguinKeeper7:
Sure, there are definitely counter-examples, it's something that's easy to overlook, 32 vs 30 and it's not a contribution rule or anything but if Atom wants to change it to your more implicit format, I'm welcome to changing, as well as the kernels above.

matrix:
@fukc-gihtub the file you are using "as an example" is not exactly the best. Considering that it also uses “goto", which is one of those issues that do not allow porting to Metal

jsteube:
> @fukc-gihtub the file you are using "as an example" is not exactly the best. Considering that it also uses “goto", which is one of those issues that do not allow porting to Metal

Yeah indeed we should fix the other ones.

Issue #4173, opened 2025-03-22
phpbb3(md5($pass))


Talented-1:
**Is your request based on a publicly known cryptographic schema and where can we find information about?**
Saw some topics regarding that on hashcat forum.

**Describe alternatives you've considered**
Cuting md5 foudns from rainbox table to get phpass:md5 as found then joining.

**Provide an example hash/database/file and a known correct password**
$H$9BhmdGtWgtB.H32r6WfA6ZRkYUhc6L/:eoseos
$H$94TJZI06OxMiGMkXngFhPP0ezHcgiZ0:ratchet
$H$9KlRe8iyydyIAighns4rYlCtrgM67F.:bikers
$H$9kX4dGOi2Fa8z4BofqmYK4JZVIEi811:angie1
$H$98vw3xbprwI5z6RJKKAwlhH9NnZ.xv1:ratchet
$H$9eA9HUaE0T8eTC/KU5kWCLrNR.WN4/1:120192
$H$9oF8R8uKC6tzcIheWr01sHk9TahA9U1:ratchet
$H$907Rvaj1QfnZKNBxZzWYyK8mU/PBOL.:ratchet

**Explain how the cryptographic schema is implemented**
phpbb3(md5($pass))

**Describe the known limitations of the algorithm**
No limitation since you can turn anything into md5 hash even few GB of file.

jsteube:
For anyone who want to add this plugin, some correction. 

Hash-Mode phpbb3() doesn't exist, I think what OP means is phpass (mode 400). So it's super simple, basically just md5 to _init kernel and that's it.

PR #4516, opened 2025-09-28
Add phpass(md5) -m 410

PenguinKeeper7:
Satisfy: https://github.com/hashcat/hashcat/issues/4173
Submitted on behalf of and full credit to @JustGuardian

```
$ ./tools/test.sh -m 410 -a all -t all -o win -P
[ test_1759069934 ] > Init test for hash type 410.
[ test_1759069934 ] [ Type 410, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1759069934 ] [ Type 410, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1759069934 ] [ Type 410, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1759069934 ] [ Type 410, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```


OpenCL/m00410-pure.cl
```diff
@@ -0,0 +1,330 @@
...
+
+  const u64 gid = get_global_id (0);
+  const u64 lid = get_local_id (0);
+  const u64 lsz = get_local_size (0);
+
+  /**
+  * bin2asc table
+  */
+
+  LOCAL_VK u32 l_bin2asc[256];
```
jsteube: Did you try to store the  md5 (in hex) in a new structure in the tmps variable in the _init? That would save the shared memory use in _loop, which translates (on NV) to additional registers.

PenguinKeeper7: As mentioned, this is a requested PR of Just's work, not my own other than the test but yeah, looks like it's duplicating work a little. Because the plaintext password (in this case, md5'ed) is constantly reused throughout the hashing process in _loop, the plaintext hashing can either be done inside the loop (which may cause the plain to be hashed multiple times uselessly) or done in the _init once but force shared mem use through tmps, which would you prefer to be implemented? The standard -m 400 kernel appears to do it the tmps way

fukc-gihtub: `tmps` isn't shared memory I think.

jsteube: Indirectly it does. The runtime uses shared memory when more registers are needed than available and swaps data there. This is better than swapping to global memory, but it only works if enough shared memory is available. That is why it is a good idea to keep shared memory free, to give the runtime some room.

PenguinKeeper7: Take a look at: https://github.com/hashcat/hashcat/pull/4516/commits/510df1edfae5a53675f555183d20078fa4d8d748. It has removed the secondary hash and bin2asc table, passing the md5(plain) through the tmp structure and removes some of the functions in favour of hardcoding buffers for speed

src/modules/module_00410.c
```diff
@@ -0,0 +1,315 @@
...
+
+static const u32   ATTACK_EXEC    = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32   DGST_POS0      = 0;
+static const u32   DGST_POS1      = 1;
+static const u32   DGST_POS2      = 2;
+static const u32   DGST_POS3      = 3;
+static const u32   DGST_SIZE      = DGST_SIZE_4_4;
+static const u32   HASH_CATEGORY  = HASH_CATEGORY_GENERIC_KDF;
+static const char *HASH_NAME      = "phpass(md5($pass))";
+static const u64   KERN_TYPE      = 410;
```
jsteube: While -m 410 is actually not too bad, other "custom" kdf's such as bcrypt(md5($pass)) use their own hashing mode. I think we should follow that rule, and make this also a dedicated mode. Can you rename to the next free?

PenguinKeeper7: See https://github.com/hashcat/hashcat/pull/4516/commits/aeed7200222ddf7a386300380cf33328724e6a3e

jsteube:
Many thanks!

PR #4522, opened 2025-09-30
Add BIP39 cryptocurrency wallet recovery (mode 36000)

### LINKED ISSUE #1546: [Feature Request] Crypto hardwallet (BIP39 passphrase recovery) - using PBKDF2-HMAC-SHA512
Hardware wallets for cryptocurrencies generally use BIP39 / BIP44 / BIP49 for the deterministic generation of keys given a master key (stored as a mnemonic for easy restoration).  They also have an option for the usage of passphrases - which each generate a new key - for plausible deniability.  However, if you don't use a passphrase in a while, you are liable to forget it.  Like me.

If you have the seed words (the mnemonic), the target address, and knowledge of the path, using a program like hashcat with a password candidate list could recover the password. I've been working to modify hashcat myself, but figured I'd try asking here as a feature request.  

[EDIT: No longer available] I'm happy to donate a significant amount to this project to help make it worth your time.

The required changes can be broken down as follows:
1.) Modify the PBKDF2-HMAC-SHA512 function to input the password candidates as the salt and data from the mnemonic as the password.
2.) Convert the the resulting private key (along with the path) into a public address and compare this against the target.  

Additional info:
1.) BIP39 info can be found here:
https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki
with the key information being:  "To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string "mnemonic" + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the derived key is 512 bits (= 64 bytes)."

2.) For my specific use case the target address is m / 49' / 0' / 0' / 0 / 0.  You could allow it to be more general by allowing people to specify a path or even a range of paths.

the format of the path is "m / purpose' / coin_type' / account' / change / address_index"
The user should generally know the exact values of purpose, coin_type, and account.  They may wish to cycle through change and address_index (as accounts will increment these during normal use) with an upper limit.

Read more here: https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki

Example test vector:
Target Address: 33747aCmUp8PkWmWWY8epR1Cph8Tf9Aozt (m/49'/0'/0'/0/0)
BIP39 mneumonic: abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about 
passphrase: testpass

Extra (not visible to user): 
BIP32 root key: yprvABrGsX5C9janudxvqdKg47twvnZoNjbmqgMt8GCDpx2xYN2351agtWysXhZ44tV92JDvZU2ppimnoJzF2A79r58KLA7c2aNV65zwZrph67B
BIP32 Extended Private Key: yprvALaobegiKdgMpUaWNgK12feBKeewQ72CM2wwtcmi1CYvuCjKmn2SgmVJkFN9KURybbqiA8U7fN1JEsPRZUhJNWtbWEddEMk93QXD6DEuB1U
Address public key: 03eabf118ca2918b3c1fa16d675ffb6382d6f40bae63c9e096548dd2b3dd6bfd5a
Address private key: KwxGq13B9B4BX6tjNc9pvVzMpKQ72SjEJ4R1ZLtSAC5qRuFzUFWW

A new set of test vectors can be generated here:
https://www.ledgerwallet.com/support/bip39-standalone.html

Lyapsus:
GPU-accelerated BIP39 passphrase recovery via PBKDF2 + BIP32.

Supports mode `36000` for ASCII passphrases with P2SH, P2PKH, and P2WPKH targets.

Resolves: #1546

Changes:
- Adds `src/modules/module_36000.c`, `OpenCL/m36000-pure.cl`, and `tools/test_modules/m36000.pm`.
- Supports same-salt multihash cracking by deriving the hashcat salt key from the static BIP39 work (`target_type`, mnemonic, and path).
- Uses `OPTS_TYPE_DEEP_COMP_KERNEL` / `KERN_RUN_3` so multiple digests sharing the same static work are compared in the deep compare kernel.
- Keeps 5-word digest comparison for address/hash160 targets.

Verification:
- `tools/test.sh --hash-mode 36000` single and multi tests passed on CPU OpenCL.
- Same-salt multihash reproducer passed on CPU OpenCL.
- CUDA self-test vector passed on NVIDIA GTX 1050 and NVIDIA RTX 4060 Laptop GPU.
- Same-salt multihash reproducer passed on NVIDIA GTX 1050 and NVIDIA RTX 4060 Laptop GPU.


jsteube:
Could you please provide a unit-test so we can test the implementation using tools/test.sh or tools/test_edge.sh ?

Lyapsus:
> Could you please provide a unit-test so we can test the implementation using tools/test.sh or tools/test_edge.sh ?

Sorry for delay, added tools/test_modules/m32001.pm

src/modules/module_32001.c
```diff
@@ -0,0 +1,1776 @@
...
+  module_ctx->module_salt_max = MODULE_DEFAULT;
+  module_ctx->module_salt_min = MODULE_DEFAULT;
+  module_ctx->module_salt_type = module_salt_type;
+  module_ctx->module_separator = MODULE_DEFAULT;
+  module_ctx->module_st_hash = module_st_hash;
+  module_ctx->module_st_pass = module_st_pass;
+  module_ctx->module_tmp_size = module_tmp_size;
+  module_ctx->module_unstable_warning = MODULE_DEFAULT;
+  module_ctx->module_warmup_disable = MODULE_DEFAULT;
+  module_ctx->module_iterations_count = module_iterations_count;
```
jsteube: This doesn't exist in hashcat

Lyapsus: Yeah, this one was accidentally left off after some local attempts to augment progress reporting. See, it seems no module before 32001 needed 64 bit number for all combinations - but I think it'll be too bold to propose infra modifications now and per your rules you'll prefer to see it as another PR anyway, so I removed it for now - shouldn't break anything except pretty progress reporting. Gotta propose infra changes later - or find a good way to solve this within 32001. Sorry for this oversight, I'm kinda new

jsteube:
> See, it seems no module before 32001 needed 64 bit number for all combinations - but I think it'll be too bold to propose infra modifications now and per your rules you'll prefer to see it as another PR anyway, so I removed it for now - shouldn't break anything except pretty progress reporting. Gotta propose infra changes later - or find a good way to solve this within 32001. Sorry for this oversight, I'm kinda new

Can you explain a bit how a typical attack scenario on this mode would look like? It's very different then a typical password scenario, and for people (like me) who are not used on a daily basis with cryptocurrencies it would help to get a bit more background how this attack is used. Something like what's the original problem coming from, how to extract needed data to feed hashcat, and maybe an same command. I also understand you want to primarily focus on -a 1, than makes sense in many cases, but in theory we should leave it up to the user. We should be able to search through a 64 bit keyspace, but not larger.

sixTheDave:
> > See, it seems no module before 32001 needed 64 bit number for all combinations - but I think it'll be too bold to propose infra modifications now and per your rules you'll prefer to see it as another PR anyway, so I removed it for now - shouldn't break anything except pretty progress reporting. Gotta propose infra changes later - or find a good way to solve this within 32001. Sorry for this oversight, I'm kinda new
> 
> Can you explain a bit how a typical attack scenario on this mode would look like? It's very different then a typical password scenario, and for people (like me) who are not used on a daily basis with cryptocurrencies it would help to get a bit more background how this attack is used. Something like what's the original problem coming from, how to extract needed data to feed hashcat, and maybe an same command. I also understand you want to primarily focus on -a 1, than makes sense in many cases, but in theory we should leave it up to the user. We should be able to search through a 64 bit keyspace, but not larger.

In https://github.com/hashcat/hashcat/issues/1546 we discuss the details, but tl;dr is when you set a Passphrase on a 24 word mnemonic (where the cryptocurrency wallet binary seed/private keys are derived from), thats an extra layer of security, however many users forget it or lose it as time passes.

To create a binary seed from the mnemonic, we use the PBKDF2 function with a mnemonic sentence (in UTF-8 NFKD) used as the password and the string “mnemonic” + passphrase (again in UTF-8 NFKD) used as the salt. The iteration count is set to 2048 and HMAC-SHA512 is used as the pseudo-random function. The length of the derived key is 512 bits (= 64 bytes).

Once you have a valid binary seed with the passphrase, you use HD wallet standard (BIP-32 (Hierarchical deterministic wallet)) and derive the cryptocurrency address, eg. m/84'/0'/0' for the bc1q addresses.

To crack a wallet we need to do the same as long and match the right address. That is the value we need to check against - otherwise we'd just have billions of private keys, not knowing which has balance on it.

example32001.hash
```diff
@@ -0,0 +1 @@
+abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:33747aCmUp8PkWmWWY8epR1Cph8Tf9Aozt:m/49'/0'/0'/0/0
```
jsteube: I think we can remove example32001.* files since we have unit tests now. These files are there only for hashcat demonstration examples.

OpenCL/m32001-pure.cl
```diff
@@ -0,0 +1,3119 @@
...
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha512.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha256.cl)
+#include M2S(INCLUDE_PATH/inc_hash_ripemd160.cl)
+#include M2S(INCLUDE_PATH/inc_ecc_secp256k1.cl)
+#endif
+
+#if BIP39_MINIMAL_KERNEL
```
jsteube: Could you please explain, what is a minimal kernel?

src/modules/module_32001.c
```diff
@@ -0,0 +1,1762 @@
...
+{
+  return KERNEL_ACCEL_MAX;
+}
+
+u32 module_kernel_loops_min (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  return 1;
+}
+
+void module_hook23 (hc_device_param_t *device_param, MAYBE_UNUSED const void *hook_extra_param, MAYBE_UNUSED const void *hook_salts_buf, MAYBE_UNUSED const u32 salt_pos, const u64 pw_pos)
```
jsteube: I don't see what this hook is doing. If it's not needed, please remove it. It makes the code more complicated than it needs to be in general.

src/modules/module_32001.c
```diff
@@ -0,0 +1,1762 @@
...
+
+static const u32 ATTACK_EXEC = ATTACK_EXEC_OUTSIDE_KERNEL;
+static const u32 DGST_POS0 = 0;
+static const u32 DGST_POS1 = 1;
+static const u32 DGST_POS2 = 2;
+static const u32 DGST_POS3 = 3;
+static const u32 DGST_SIZE = DGST_SIZE_4_5;
+static const u32 HASH_CATEGORY = HASH_CATEGORY_CRYPTOCURRENCY_WALLET;
+static const char *HASH_NAME = "BIP39 Passphrase Recovery (ASCII, P2SH/P2PKH/P2WPKH)";
+static const u64 KERN_TYPE = 32001;
```
jsteube: I think 32001 is an inappropriate mode for this very relevant hash-mode. Please rename it to 36000 we'll keep it free for this mode.

sixTheDave:
I decided to award [Lyapsus](https://github.com/Lyapsus) with the bounty I wrote on https://github.com/hashcat/hashcat/issues/1546#issuecomment-3247699887.

Only thing left: @Lyapsus please answer the comments above as I requested so the PR can be concluded.

example32001.hash
Lyapsus: Done, coming soon

src/modules/module_32001.c
Lyapsus: Basically debugging infra for some obscure crashes investigation some time ago. Sorry for leaking into PR branch. Removal is done on my side, coming soon

src/modules/module_32001.c
Lyapsus: Done, coming soon

OpenCL/m36000-pure.cl
```diff
@@ -0,0 +1,2860 @@
...
+
+typedef struct bip39_sha512_ctx
+{
+  u64 h[8];
+  u64 total_len;
+  u32 buffer_len;
+  u8 buffer[128];
+} bip39_sha512_ctx_t;
+
+DECLSPEC u64 bip39_rotr64 (const u64 x, const u32 n)
```
jsteube: I think you can replace with hc_rotr64_S()

jsteube:
I just checked the multihash support, and it seems incorrectly handled. Here's all details to reproduce:

```
This hash-mode plugin cannot crack multiple hashes with the same salt, please select one of the hashes.
! unhandled return code 255, cmdline : cat test_1762610722/36000_passwords.txt | ./hashcat --quiet --potfile-disable --logfile-disable -O --runtime 400 -D 2 --backend-vector-width 4 -a 0 -m 36000 test_1762610722/36000_hashes.txt
$ cat test_1762610722/36000_passwords.txt

6149676292
6314228317240566
2930502613140005514
728548665193475221279663589
94045176751733200762481555594
505047214583050265390840376280
7528519758250712386714547256387
$ cat test_1762610722/36000_hashes.txt
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:1837c1be8e2995ec11cda2b066151be2:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:0db3d112e2c3feae1d684b6b5bc1181f:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:0c7c4ad0bd7913a0214cb5fe772488b7:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:fea650363e4e37b9e3e1d388c30566c7:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:7127d38a4c7614b30d4f5ccac21b71e7:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:9020c010fa08a8ed1c6c8edf6b620969:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:110c3b9550d4ace72873f60b6e2c94f5:m/49
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about:fdb9fb6720f792563385fa0b6c2d3a04:m/49
```

Basically what hashcat doesn't like is that the same salt is used for all the hashes, which is possible, but requires special handling (OPTS_TYPE_DEEP_COMP_KERNEL). You have two options. 

1. What you can do is to add all the data into the salt_struct (salt_buf and salt_buf_pc for extra data), which makes the entry unique, but then depending on the algorithm all keymaterial is recalculated
2. If the algorithm is somewhat broken (like in WPA/WPA2) where you have two parts of salts and where the heavy lifting depends only on the static one, it's better to use OPTS_TYPE_DEEP_COMP_KERNEL

OpenCL/m36000-pure.cl
Lyapsus: Done

ipsbruno3:
Guys, I’m debating whether it’s better to keep this SHA-512/PBKDF2 OpenCL kernel in 64-bit (ulong) or migrate to a 32-bit (uint) implementation using split-limb operations (e.g., implementing the Σ/σ functions as s0/s1/s2/s3 style sequences on 32-bit halves).

From your experience on modern GPUs (NVIDIA/AMD):

Is SHA-512 generally faster with native 64-bit rotates/adds (ulong), or does a 32-bit split implementation typically win due to higher uint throughput?

How do the tradeoffs look in practice regarding register pressure, instruction count, and occupancy when emulating 64-bit ops via 32-bit halves?

Are there architectures where ulong is clearly the right choice (e.g., recent NVIDIA) versus ones where 32-bit is strongly preferred (e.g., some AMD)?

In OpenCL specifically, do rotate() / bitwise ops on ulong map efficiently, or do drivers often lower them in ways that hurt?

If you’ve benchmarked both approaches, what’s the typical crossover point (if any) for PBKDF2-HMAC-SHA512?

Any guidance or rules-of-thumb would be appreciated.

Repo (kernels):

* [https://github.com/ipsbruno3/bitcoin_electrum_cracking/blob/main/kernel/sha512.cl](https://github.com/ipsbruno3/bitcoin_electrum_cracking/blob/main/kernel/sha512.cl)
* [https://github.com/ipsbruno3/bitcoin_electrum_cracking/blob/main/kernel/pbkdf2_hmac_sha512.cl](https://github.com/ipsbruno3/bitcoin_electrum_cracking/blob/main/kernel/pbkdf2_hmac_sha512.cl)


Right now, on an RTX 5090, the kernel can exceed **800 kH/s (BIP39) and 65,000 KH/s (Electrum "BIP-39"(** and is already fairly optimized, but I’m looking for further improvements. I’ve been reviewing Hashcat for ideas. Suggestions are welcome — and if it’s acceptable, I’d be happy to help with development/testing.

ipsbruno3:
image

Electrum BIP-39 Scan:
It was able to generate 66 million seeds per second in PBKDF/PUBLICKEY + BECH32 in a **RTX NVIDIA 5090**

jsteube:
The multihash case works now. Three P2SH targets sharing one salt come back 3/3, each paired with the right passphrase and no refusal, and the IL_HEX form does the same. OPTS_TYPE_DEEP_COMP_KERNEL with module_deep_comp_kernel() is the right answer to what I raised in November, and it clears it.

A dynamic path still loses hashes. With a range segment, m/49'/0'/0'/0/{0-3}, each of two targets cracks on its own but together in one salt the run ends Exhausted at 1/2. In "m36000-pure.cl" the fixed path branch sets derived_ready unconditionally, which is why that case works, while the dynamic branch gates derived_ok on a comparison against esalt_bufs[DIGESTS_OFFSET_HOST].target_hash, which is the first digest of the salt and not the one being tested. m36000_comp then returns early wherever derived_ready stayed 0.

Before any of that, the branch does not build against master. Github reports it mergeable and the text merge is clean, but four things have moved underneath it:

```
  module_36000.c  calls input_tokenizer(), now declared in include/parser.h, not included here
  module_36000.c  assigns module_ctx->module_dictstat_disable, which no longer exists
  module_36000.c  never assigns module_advice_notice, module_hash_hints, module_usage_notice
  m36000-pure.cl  calls check() with BITMAP_MASK, BITMAP_SHIFT1, BITMAP_SHIFT2; check() takes the mask alone now
```

The first two are compile errors, the third kills every run with "Module context missing field", and the fourth is 6 JIT errors. With those four fixed it builds clean with no warnings and benchmarks at 605.6 kH/s on a 4090, which is about what the derivation on top of 2048 PBKDF2 iterations should cost.

The kernel is still carrying its development scaffolding: 26 BIP39_DISABLE_* bisection switches, BIP39_PROFILE, BIP39_DEBUG_PRINT, BIP39_TEST_KERNEL, 82 printf() calls and five alternative m36000_init bodies selected between them. None can be reached, since module_jit_build_options() returns NULL. Could that come out before this goes in? Related, is_testpass compares every candidate against the literal "testpass" to feed those printfs, so the self test password is compiled into the production kernel.

One gap in the test module. In "m36000.pm" only the IL_HEX form is generated, so P2SH, P2PKH and P2WPKH, which the mode is named for, and the dynamic path syntax have no vector at all. The self test covers one P2SH target, and no vector covers the path that just failed above.

Two smaller things for the same pass. No file under docs/ is touched, so "docs/changes.txt" wants a New Algorithms line. And the rebase pays for itself: secp256k1 gained a safegcd modular inverse in #4763 on 14 August, which point_mul_xy() reaches through inv_mod(), so the speed above understates what the mode will do once this sits on current master.

jsteube:
One note attached to the test vector request above, so you do not spend time chasing it. Running "tools/test_edge.sh" with -m 36000 reports 120 errors over 90 combinations today, and none of them come from the mode.

The suite puts the hash through a shell eval twice, once to unwrap the interchange format "tools/test.pl" writes and once on the hashcat command line. The hardened segment markers in m/49'/0'/0'/0/0 make both evals die with "unexpected EOF while looking for matching". Substituting h for the apostrophe in the same vector gives 0 errors, so that is the whole of it. No mode shipped so far has an apostrophe in its self test hash, which is why 36000 is the first to reach this.

I am fixing that on our side, so this pull request needs no change for it. The new vectors are still worth adding.

jsteube:
The harness side of that is in master now, in 8e5c16946. The three fields come out of "tools/test.pl" hex encoded and "tools/test_edge.sh" decodes each one where it reads it, so a derivation path with hardened segments in it no longer ends the line early. A rebase picks it up, and -m 36000 then runs the edge suite for real rather than reporting 120 errors it never reached.

PR #4525, opened 2025-10-06
Fix typos in makefiles for bridge plugins

maede97:
In the two python binding plugins, the makefile error (when not supported) shows the wrong module number.
In the lines below, the correct number is used. This PR simply changes the plugin number in the error string.


jsteube:
Thanks!

PR #4530, opened 2025-10-09
Add 2 new rules, 'convert to lower hex' and 'convert to upper hex'

Chick3nman:
This PR adds 2 new rule functions:

'h' converts the entire password buffer to lowercase hex
'H' converts the entire password buffer to uppercase hex

### DIFF
--- a/OpenCL/inc_rp_common.h
+++ b/OpenCL/inc_rp_common.h
@@ -66,6 +66,9 @@
 #define RULE_OP_MANGLE_DUPEBLOCK_FIRST   'y'
 #define RULE_OP_MANGLE_DUPEBLOCK_LAST    'Y'
 #define RULE_OP_MANGLE_TITLE             'E'
+#define RULE_OP_MANGLE_TO_HEX_LOWER      'h'
+#define RULE_OP_MANGLE_TO_HEX_UPPER      'H'
+
 
 /* using character classes */
 #define RULE_OP_MANGLE_REPLACE_CLASS     0x01
--- a/OpenCL/inc_rp.cl
+++ b/OpenCL/inc_rp.cl
@@ -486,7 +486,7 @@ DECLSPEC int mangle_to_hex_lower (MAYBE_UNUSED const u8 p0, MAYBE_UNUSED const u
 
   if (out_len >= RP_PASSWORD_SIZE) return len;
 
-  for (int pos = len + 1; pos >= 0; pos--)
+  for (int pos = len - 1; pos >= 0; pos--)
   {
     const u8 c = buf[pos];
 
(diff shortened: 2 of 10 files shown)


fukc-gihtub:
How useful is this?

PenguinKeeper7:
> How useful is this?

I wondered the same - certainly seems like a strange one. I tried it against a popular distributed cracking platform's 50 million MD5 leftlist with `h` and `H` being the outer layer of stacked rules and only got 6 cracks, 2 of which were probably accidental digit-only plains so unless there's a specific algorithm/framework where this hexing is more popular than I know, it doesn't seem particularly effective

Chick3nman:
> How useful is this?

It's not overly useful against the average password list, but it does solve a use case I had and it adds a bit of modularity for performing attacks on odd constructions without needing to mod any kernels. A base64 version of this would do the same, but that seems like it requires more discussion about default behavior regarding which base64 dict and w/ or w/o padding.

PenguinKeeper7:
Might be nice to exclude it from `-g` (if that's easily possible?)

Chick3nman:
> Might be nice to exclude it from `-g` (if that's easily possible?)

Looks like -g is already picking from a subset of values for it's rules, it could be made to pick from a tighter subset of ops as well
https://github.com/hashcat/hashcat/blob/f5a5bd83c24c8b59b461d299f3d506dce64b122d/src/rp.c#L145

OpenCL/inc_rp_common.h
```diff
@@ -66,6 +66,9 @@
 #define RULE_OP_MANGLE_DUPEBLOCK_FIRST   'y'
 #define RULE_OP_MANGLE_DUPEBLOCK_LAST    'Y'
 #define RULE_OP_MANGLE_TITLE             'E'
+#define RULE_OP_MANGLE_TO_HEX_LOWER      'h'
```
jsteube, line 69: Did we check if this collides with jtr 'h' or 'H'?

Chick3nman: I did, according to @roycewilliams rosetta page for rule functions, no one else appears to be using 'h' or 'H' 

https://docs.google.com/spreadsheets/d/1vDQWg-eZplEScs7sF4m3eJCEGq3bmmifHsLGL73n4Ig/edit

roycewilliams:
My Rosetta might have been out of date but I just double-checked JtR directly:

https://www.openwall.com/john/doc/RULES.shtml
https://github.com/openwall/john/blob/bleeding-jumbo/doc/RULES

... and it looks clear.

magnumripper:
Do I understand correctly that `h` would be "hello -> 68656c6c6f"?

fukc-gihtub:
@magnumripper Yes.

PR #4531, opened 2025-10-10
Add a github action to run checks and tests for Rust code

pwrbob:
Since using Rust in hashcat is quite new, it's still easy to start running unit tests automatically without generating a lot of work for previous contributors. This pull request adds a github action to automatically execute common formatting checks using `rustfmt` and to run unit tests in all crates (currently `hashcat-sys`, all bridges and all feeds).

Various Rust source files were changed slightly to meet the `rustfmt` specification.

Currently only works on mac and linux.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -575,7 +575,6 @@ clean:
 	$(RM) -rf kernels
 	$(RM) -rf Rust/bridges/*/target
 	$(RM) -rf Rust/feeds/*/target
-	$(RM) -rf Rust/hashcat-sys/target
 	$(RM) -rf Rust/hashcat-sys/src/bindings.rs
 
 .PHONY: distclean
--- a/src/Makefile
+++ b/src/Makefile
@@ -575,6 +575,8 @@ clean:
 	$(RM) -rf kernels
 	$(RM) -rf Rust/bridges/*/target
 	$(RM) -rf Rust/feeds/*/target
+	$(RM) -rf Rust/hashcat-sys/target
+	$(RM) -rf Rust/hashcat-sys/src/bindings.rs
 
 .PHONY: distclean
 distclean: clean
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -15,7 +15,6 @@ on:
   push:
     branches:
       - "master"
-      - "feat/*"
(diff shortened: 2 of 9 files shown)


src/Makefile
```diff
@@ -575,6 +575,8 @@ clean:
 	$(RM) -rf kernels
 	$(RM) -rf Rust/bridges/*/target
 	$(RM) -rf Rust/feeds/*/target
+	$(RM) -rf Rust/hashcat-sys/target
```
matrix: this directory is not created on my side

matrix (changes requested):

src/Makefile
pwrbob: You could manually run the unit tests (currently only one, but that might change in the future) in that crate, which would create the target directory. It seems cleaner to ensure that it's removed.

matrix: so we have a problem: i'm testing on linux and macos and Rust/hashcat-sys/target is never created

matrix: it's created on that create because you run "cargo check" & "cargo test". These commands are not used in Makefile, so thats why

pwrbob: Of course this line can be removed if `make clean` should only delete things created by other `make` commands. However, since the `-f` flag is used,  removing something that doesn't exist isn't an error. Thus, this line would additionally clean up things created during manual development, which seems nice. Rust's `target` directories tend to be very large, so it would be nice if `make clean` really removed all those that might exit.
Apart from this detail, I don't really have an opinion on this. The line could be removed or left- it seems to be more of a discussion on `Makefile` conventions.

An alternative would be to include building and testing of `hashcat-sys` in the Makefile. Then the directory would exist and the process would be consistent.

Another solution would be to put all Rust code into one big workspace, allowing the compilation process to use one single `target` directory. I'm not sure this is better in this case.

matrix: There is no point in complicating the Makefile by adding checks and tests or other commands that are not useful in real cases.

matrix:
@pwrbob after #4535 is merged, you can try to enable Rust tests also on windows

pwrbob:
windows support added here: https://github.com/hashcat/hashcat/pull/4537

PR #4540, opened 2025-10-15
Increase max iteration count length to 8 digits for PBKDF2

### LINKED ISSUE #4539: PBKDF2 (10900) round count limited to 6 digits
Hello,

I noticed that the number of rounds of `PBKDF2-HMAC-SHA256` (`10900`) is limited to a 6 digit number making it impossible to attempt to crack hashes with 1M rounds or more.

mimi89999:
Fixes https://github.com/hashcat/hashcat/issues/4539

### DIFF
--- a/src/modules/module_10000.c
+++ b/src/modules/module_10000.c
@@ -96,7 +96,7 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
 
   token.sep[1]     = '$';
   token.len_min[1] = 1;
-  token.len_max[1] = 6;
+  token.len_max[1] = 8;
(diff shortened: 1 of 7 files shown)


mimi89999:
@jsteube I agree that 8 digits is still low, but I don't think that it makes sense to go up to 252. That's a value that is not testable. If somebody ends up with such a high iteration count, there must be an issue somewhere. However, I'm not sure about what a reasonable number of digits would be.

From my [simple benchmark](https://gist.github.com/mimi89999/c4230180104de3de4c441953fee22ea3#file-l40s-1-48g_hashcat_benchmark-txt) I can infer that `10 000 000 000` rounds would be below one iteration per second on a high end GPU. That might be an upper limit that makes sense.

PR #4552, opened 2025-10-23
Add 1 new rule op, "insert every"

Chick3nman:
This new rule inserts the desired byte X every N bytes in the password buffer.

The rule is written `vNX` where `v` is our operator, `N` is how many bytes between insertions, and `X` is the byte to insert. 

The behavior of the rule is such that if the last position in the buffer is equal to N bytes after the previous insertion, X will be inserted at the end as well. If the value of N is 0, this rule cannot operate and returns the initial buffer. If the value of N is larger than the length of the original buffer, again we cannot operate and must return the original buffer.

```
Password1

v0_
v1_
v2_
v3_
v4_
v5_
v6_
v7_
v8_
v9_
vA_

Password1
P_a_s_s_w_o_r_d_1_
Pa_ss_wo_rd_1
Pas_swo_rd1_
Pass_word_1
Passw_ord1
Passwo_rd1
Passwor_d1
Password_1
Password1_
Password1
```


OpenCL/inc_rp_common.h
```diff
@@ -68,6 +68,7 @@
 #define RULE_OP_MANGLE_TITLE             'E'
 #define RULE_OP_MANGLE_TO_HEX_LOWER      'h'
 #define RULE_OP_MANGLE_TO_HEX_UPPER      'H'
+#define RULE_OP_MANGLE_INSERT_EVERY      'v'
```
jsteube, line 73: Did you check JtR compatibility?

Chick3nman: Yes, `v` is currently unused by JTR. I would have preferred to use `I` but JTR implements it as adding `ing` to the end of the password. No other tools use `v` or `I`.

fukc-gihtub:
Could this rule op also insert the byte at the beginning? It's easier to truncate than to prepend the same byte.

`v1_ [` / `v1_ ^_`

Chick3nman:
> Could this rule op also insert the byte at the beginning? It's easier to truncate than to prepend the same byte.
> 
> `v1_ [` / `v1_ ^_`

I had considered this but it doesn't fit the operational definition of how I intended for the rule to work. I'm inserting a character "every N characters from the start of the string". Which is just a bit different from "interleave X character every N bytes" in my opinion.

Chick3nman:
> Could this rule op also insert the byte at the beginning? It's easier to truncate than to prepend the same byte.
> 
> `v1_ [` / `v1_ ^_`

Having thought about this some more, I think I may actually be coming around to the idea of changing the behavior to include the preceding position. It's not how I had intended the rule originally, but I think it may be more useful as in your example, a value agnostic `[` is "easier" than a value specific prepend. I'd like some more input on this from others before I change the behavior, but this does seem like it may be good change overall.

fukc-gihtub:
As the author of a [Markov rule generator](https://git.launchpad.net/rulechief), I think it would be more useful.

fukc-gihtub:
In a small survey I ran, 6 people voted for the original behavior and 1 for the change.

Chick3nman:
> In a small survey I ran, 6 people voted for the original behavior and 1 for the change.

To clarify, original behavior is how it is in the PR now, and change is including the character in the first index (0)?

fukc-gihtub:
Yes.

Chick3nman:
In that case, I will leave it as it is and have it merged. Thanks for the input :)

fukc-gihtub:
There are conflicts currently that prevent merging.

PR #4558, opened 2025-10-29
Add support for LoongArch, PowerPC and MIPS

Suyun114:


Suyun114:
It has been built successfully on the following platforms:

- LoongArch: Loongson 3C6000/D
- PowerPC: IBM POWER8
- MIPS: Loongson 3B4000

The benchmark results will be uploaded soon.

src/cpu_features.c
```diff
@@ -16,7 +16,7 @@ int cpu_supports_avx2 ()     { return 0; }
 int cpu_supports_avx512f ()  { return 0; }
 int cpu_supports_avx512vl () { return 0; }
 
-#elif defined(__riscv)
+#elif defined(__riscv) || defined(__loongarch__) || defined(__powerpc__) || defined(__mips__)
 
 // TODO: Support RVV
```
ziyao233: This comment should be updated as well.

src/cpu_features.c
```diff
@@ -16,7 +16,7 @@ int cpu_supports_avx2 ()     { return 0; }
 int cpu_supports_avx512f ()  { return 0; }
 int cpu_supports_avx512vl () { return 0; }
 
-#elif defined(__riscv)
+#elif defined(__riscv) || defined(__loongarch__) || defined(__powerpc__) || defined(__mips__)
```
xen0n: I think this is better described as `#else`, as it's clearly providing the fallback code.

src/bridges/bridge_argon2id_reference.c
```diff
@@ -18,7 +18,7 @@
 #include "core.c"
 #include "blake2/blake2b.c"
 
-#if defined (__riscv)
+#if defined (__riscv) || defined(__loongarch__) || defined(__powerpc__) || defined(__mips__)
```
xen0n: So is this -- we should selectively enable the optimized implementation for the ported architectures, and leave the rest as fallback.

xen0n (commented):
Just some drive-by comments ;-)

src/bridges/bridge_argon2id_reference.mk
```diff
@@ -11,11 +11,16 @@ ifeq ($(IS_APPLE_SILICON),0)
 ARGON2_REFERENCE_CFLAGS += -mavx2
 endif
 else
+ifeq ($(IS_PPC),1)
+ARGON2_REFERENCE_CFLAGS += -mcpu=native
```
jsteube, line 15: I haven't used -mcpu so far, but isn't -mcpu not already covering -mtune? Maybe it's better to stick to -march and -mtune instead.

MingcongBai: > I haven't used -mcpu so far, but isn't -mcpu not already covering -mtune? Maybe it's better to stick to -march and -mtune instead.

Indeed (and so does -march), I think the equivalent here should be -mcpu=native.

Suyun114:
> It has been built successfully on the following platforms:
> 
> * LoongArch: Loongson 3C6000/D
> * PowerPC: IBM POWER8
> * MIPS: Loongson 3B4000
> 
> The benchmark results will be uploaded soon.

I have tested with [this script](https://gist.github.com/Suyun114/019fe7c9ae52a6ff25532046a2cc448e) to ensure that a crash in a single hash node does not halt the entire process.

- LoongArch: https://gist.github.com/Suyun114/e6022385c7692b796bfd08a1364bf2e1
- PowerPC: https://gist.github.com/Suyun114/5984e62c0dd6f441e82ddac32538fb36
- MIPS: https://gist.github.com/Suyun114/fb3317359de3f60c7ea485e33a4f4fc5
- As reference, x86_64: https://gist.github.com/Suyun114/97a173b8f5d8af1de3d8de065ea1a124

src/bridges/bridge_argon2id_reference.mk
Suyun114: > I haven't used -mcpu so far, but isn't -mcpu not already covering -mtune? Maybe it's better to stick to -march and -mtune instead.

PowerPC does not support -march.

jsteube: OK, I'll merge, but I'd be nice to have some comment about this in the Makefile so it's not forgotten, or we can revisit this in the future and check again if support for march has been added. Could you create a new PR for that please?

PR #4566, opened 2025-11-05
Replace deprecated Bitcoin::Crypto symbols with supported counterparts

### LINKED ISSUE #3894: Bitcoin::Crypto deprecation warnings
**Describe the issue**

Hashcat distribution uses Bitcoin::Crypto perl module as a dependency for some tests. There was a major release yesterday which deprecated some symbols. Hashcat is affected and will now be issuing deprecation warnings (as it looks like Bitcoin::Crypto is not locked to a specific version - the newest one will be installed in `install_modules.sh`)

The deprecated methods are not going away for a long time, but you may want to modify the code accordingly. In this case, seems like it would involve changing `$key->to_hex` to `to_format [hex => $key->to_serialized]` (https://metacpan.org/pod/Bitcoin::Crypto::Util#to_format) and `->from_hex($hex_string)` to `->from_serialized([hex => $hex_string])`. There may be some others - warnings issued during test runtime should point you to all the places needing attention.

You can read more about new style of handling formats in Bitcoin::Crypto [in the manual](https://metacpan.org/dist/Bitcoin-Crypto/view/lib/Bitcoin/Crypto/Manual.pod#How-to-pass-data-to-functions?)

Cheers.

bbrtj:
Resolves #3894

I am the author of Bitcoin::Crypto. I am planning to remove the deprecated symbols from the module in the very near future. They've been deprecated for around 24 months.

I don't intend to break compatibility of any public code that depends on my module, and so merging this PR will ensure using syntax that will be supported for the foreseeable future. I've searched the repository for other code that may use deprecated symbols and found none.

I didn't run these tests myself since I don't know much about hashcat, but I have sanity-checked the code with `perl -c`.

### DIFF
--- a/tools/test_modules/m30901.pm
+++ b/tools/test_modules/m30901.pm
@@ -9,6 +9,7 @@ use strict;
 use warnings;
 
 use Bitcoin::Crypto         qw (btc_prv btc_extprv);
+use Bitcoin::Crypto::Util   qw (to_format);
 use Bitcoin::Crypto::Base58 qw (decode_base58check);
 
 sub module_constraints { [[64, 64], [-1, -1], [-1, -1], [-1, -1], [-1, -1]] }
@@ -27,7 +28,7 @@ sub module_generate_hash
 
   my @is_valid_hex = eval
   {
-    $priv = btc_prv->from_hex ($word);
+    $priv = btc_prv->from_serialized ([hex => $word]);
   };
 
   return if (! @is_valid_hex);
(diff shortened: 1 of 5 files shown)


bbrtj:
New Bitcoin::Crypto major version has been released. Since hashcat does not download a fixed version number, I suggest merging this PR as soon as possible.

ahmedaljwkr2012-ux:
أنا مش عارف أجيب حساب التيك توك دهarwa.khaled157

PR #4567, opened 2025-11-06
Add hash mode management functions and user options for listing hash …

sol0lvl:
…modes

- Added functionality to use  -m alias in addition to -m code.
- Added functionality to list available hash modes and their respeciv aliases.
- Modified command-line argument parsing to handle the new option.

### DIFF
--- a/include/shared.h
+++ b/include/shared.h
@@ -139,4 +139,9 @@ size_t hc_memchr_avx512       (const u8 *ptr, int ch, size_t max_len);
 
 hc_memchr_t hc_memchr_get     (void);
 
+int build_hash_mode_map       (hashcat_ctx_t *hashcat_ctx);
+void destroy_hash_mode_map    (hashcat_ctx_t *hashcat_ctx);
+u32  hash_mode_from_string    (hashcat_ctx_t *hashcat_ctx, const char *str);
+void list_hash_modes          (hashcat_ctx_t *hashcat_ctx);
+
 #endif // HC_SHARED_H
--- a/include/types.h
+++ b/include/types.h
@@ -867,6 +867,7 @@ typedef enum user_options_map
   IDX_ENCODING_TO               = 0xff16,
   IDX_HASH_COPY                 = 0xff62,
   IDX_HASH_INFO                 = 'H', // 0xff17
+  IDX_LIST_HASH_MODES           = 0xff63,
   IDX_FORCE                     = 0xff18,
   IDX_HWMON_DISABLE             = 0xff19,
   IDX_HWMON_TEMP_ABORT          = 0xff1a,
@@ -2584,6 +2585,7 @@ typedef struct user_options
   u32          debug_mode;
   u32          hwmon_temp_abort;
   u32          hash_info;
+  bool         list_hash_modes;
   int          hash_mode;
   u32          hccapx_message_pair;
   u32          hook_threads;
--- a/src/hashcat.c
+++ b/src/hashcat.c
@@ -1202,22 +1202,15 @@ int hashcat_session_init (hashcat_ctx_t *hashcat_ctx, const char *install_folder
   user_options_session_auto (hashcat_ctx);
 
   /**
-   * event init (needed for logging so should be first)
-   */
-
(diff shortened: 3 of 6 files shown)


jsteube:
In "shared.c" the part of build_hash_mode_map() that handles a repeated alias keeps whichever mode number is lower and drops the other, with no report. Running generate_friendly_alias() over the 598 hash names in master gives 8 pairs that land on the same string. The comparison characters are dropped, which merges three version splits, and the closing parenthesis is dropped while the dot becomes a hyphen, which erases the nesting in the other five. Both iTunes backup modes reduce to itunes-backup-10-0, and the left number in each row is the one that shared alias resolves to:

```
  141  1441   Episerver 6.x < .NET 4              Episerver 6.x >= .NET 4
 2611  2711   vBulletin < v3.8.5                  vBulletin >= v3.8.5
14700 14800   iTunes backup < 10.0                iTunes backup >= 10.0
 4410  4420   md5(sha1($pass).$salt)              md5(sha1($pass.$salt))
 4710 21100   sha1(md5($pass).$salt)              sha1(md5($pass.$salt))
20710 20730   sha256(sha256($pass).$salt)         sha256(sha256($pass.$salt))
 4110 33100   md5($salt.md5($pass.$salt))         md5($salt.md5($pass).$salt)
21900 31700   md5(md5(md5($pass.$salt1)).$salt2)  md5(md5(md5($pass).$salt1).$salt2)
```

The listing in list_hash_modes() generates its alias with a second call to generate_friendly_alias() rather than reading the map back, so itunes-backup-10-0 prints against 14800 as well, and a user who copies that alias out of the listing is given 14700. This is the part that blocks for me, and I do not think it can be repaired inside the derivation, because what separates each pair is exactly what the transformation removes.

Where the same idea works without that problem is the shell. In "extra/tab_completion/hashcat.sh" the case on the previous word carries a branch for each option that takes a fixed set, and -m has none today. A branch for -m could match on the name and put the mode number in COMPREPLY when one name matches, or put the names back when several do. That leaves a number on the command line, so the resolution never happens inside hashcat, and an ambiguous name costs a user one more keystroke rather than the wrong mode. The generated list has somewhere to live too, since "generate_docs.sh" already rebuilds "docs/hashcat-help.md" from --help on every commit to master. Would you consider taking it there instead, given that --help already prints the mode and name table from the sweep in "usage.c" and --hash-info --machine-readable already emits the same per mode as JSON? The shape to copy sits next to it:

```
    -a|--attack-mode)
      COMPREPLY=($(compgen -W "${ATTACK_MODES}" -- ${cur}))
      return 0
      ;;
```

One other thing, whichever way that goes. Now folder_config_init() runs in main() before user_options_getopt(), and it reads user_options->cache_path to place cache_root. At that point the value is still the NULL that user_options_init() set, so --cache-path stops having any effect. The same move takes folder_config_init() and event_ctx_init() out of hashcat_session_init(), which leaves install_folder and shared_folder unused in a function "include/hashcat.h" exports, so an external caller of libhashcat is left with a zeroed folder_config from the call that used to set it up. I would want --cache-path to keep working and hashcat_session_init() to stay self-contained for an external caller.

jsteube:
Rules now run in every attack-mode, which landed in "user_options.c" and the startup path this PR rewrites. The sanity checks it edits and the init ordering it moves are both in that work, so what is here no longer applies to the tree it was written against.

The asks from the review are still open underneath that. Deriving an alias from HASH_NAME collides for 8 pairs of modes, and the resolver keeps whichever number is lower without saying so, while the listing prints the losing alias against the mode it does not select. Moving folder_config_init() ahead of user_options_getopt() also leaves --cache-path reading the NULL that user_options_init() set, and takes folder_config_init() and event_ctx_init() out of hashcat_session_init(), so an external caller of libhashcat gets a zeroed folder_config from the call that used to build it.

So I am closing this one rather than asking for a rebase onto a moving target. The idea of reaching a mode by name is worth having, and tab completion is where I would put it: the shell can offer the names and leave a number on the command line, so an ambiguous name costs a keystroke instead of selecting the wrong mode. "extra/tab_completion/hashcat.sh" has a case per option that takes a fixed set and none for -m today.

PR #4584, opened 2025-12-12
Update metamask2hashcat.py

b8vr:
The vault extraction routine has been updated. This was needed because sometimes wallet_data could contain data like:

`{"cipher":"<the cipher>","iv":"<the iv>","keyMetadata":{"algorithm":"PBKDF2","params":{"iterations":5000}`

where the salt part is left out, hence no hash is extracted. With the update, wallet_data now contains entire vault data needed to create the hash:

`{"cipher":"<the cipher>","iv":"<the iv>","keyMetadata":{"algorithm":"PBKDF2","params":{"iterations":5000}},"lib":"original","salt":"<the salt>"}
`

This issue seems to have come due to new versions of metamask.

### DIFF
--- a/tools/metamask2hashcat.py
+++ b/tools/metamask2hashcat.py
@@ -6,6 +6,16 @@
 # Date: Thu 28 Aug 2023 05:12:40 PM CEST
 # License: MIT
 
+# Update: b8vr
+# Version: 2.2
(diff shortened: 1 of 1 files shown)


b8vr:
I have other changes. Dropping this request.

PR #4586, opened 2025-12-15
Include hash extraction from persist-KeyringController

b8vr:
The JSON from newer versions of metamask on mobiles is located in persist-KeyringController. The old version of metamask2hashcat failed when extracting a hash from this file.

### DIFF
--- a/tools/metamask2hashcat.py
+++ b/tools/metamask2hashcat.py
@@ -11,39 +11,40 @@
 # From version 2.0 works also for Metamask Mobile
 # reference: https://github.com/3rdIteration/btcrecover/pull/346
 
+# Update: b8vr
+# Version: 2.3
+# Date: Mon 15 Dec 2025 09:50
+# The vault extraction routine has been updated to accomodate both persist-root and persist-KeyringController
+
 import json
 import argparse
 import base64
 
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks for this, and sorry it has been sitting for so long. The brace counting is a real improvement and I want to keep it. There are three things to sort out before this can go in:

1. The main one is that desktop vaults no longer work: the old code only text-scanned mobile vaults, while the new path scans everything and then expects `lib.original`. A normal desktop vault has neither, so it exits before reaching the existing `$metamask$` output path. That also makes the `isMobile is False` branch unreachable.
2. There are two smaller edge cases in the parser. The vault guard is inverted because find() returns `-1` when the string is absent, so that comparison should be made explicitly. And if the braces never balance, `j` is never assigned and the code raises `UnboundLocalError` instead of reaching the `ValueError` handler.
3. Merge ordering. The new `$metamaskMobile$rounds=600000$<salt>$<iv>$<cipher>` format is not accepted by module 31900 on master yet; your #4601 adds the variable salt length and rounds field it needs. So these two changes really need to land together, otherwise the tool can produce hashes that hashcat does not yet accept.

PR #4589, opened 2025-12-15
Create module_31910.c

b8vr:
Metamask Mobile with 44 bytes salt


b8vr:
Will make changes to 31900 instead to support both fashions of metamaskMobile. PoC now exists. Also looking at future proofing with variable rounds.

PR #4590, opened 2025-12-16
feat: Cardano Eternl Wallet (PBKDF2-HMAC-SHA512-ChaCha20Poly1305)

AndrewWestberg:
The older version of Cardano Eternl wallet used a less-secure password derivation scheme. This has been patched for several years now. There are still a few users who have forgotten their wallet passwords for the old wallet and need to recover them. 

The hash is defined as the hex of the encrypted private key.

### DIFF
--- a/OpenCL/m01852-optimized.cl
+++ b/OpenCL/m01852-optimized.cl
@@ -1,0 +1,746 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_rp_optimized.h)
+#include M2S(INCLUDE_PATH/inc_rp_optimized.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha512.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct eternl
+{
+    u32 salt_buf[64];
+    u32 nonce[3];
+    u32 tag[4];
+    u32 encrypted[42];
+} eternl_t;
+
+typedef struct pbkdf2_sha512_tmp
+{
+    u64 ipad[8];
+    u64 opad[8];
+    u64 dgst[16];
+    u64 out[16];
+} pbkdf2_sha512_tmp_t;
+
+DECLSPEC void chacha20_block (PRIVATE_AS u32 *state, PRIVATE_AS u32 *out)
+{
+  u32 x[16];
+  for (int i = 0; i < 16; i++) x[i] = state[i];
+
+  for (int i = 0; i < 10; i++)
+  {
+    x[0] += x[4]; x[12] = hc_rotl32(x[12] ^ x[0], 16);
+    x[8] += x[12]; x[4] = hc_rotl32(x[4] ^ x[8], 12);
+    x[0] += x[4]; x[12] = hc_rotl32(x[12] ^ x[0], 8);
+    x[8] += x[12]; x[4] = hc_rotl32(x[4] ^ x[8], 7);
+
+    x[1] += x[5]; x[13] = hc_rotl32(x[13] ^ x[1], 16);
+    x[9] += x[13]; x[5] = hc_rotl32(x[5] ^ x[9], 12);
+    x[1] += x[5]; x[13] = hc_rotl32(x[13] ^ x[1], 8);
+    x[9] += x[13]; x[5] = hc_rotl32(x[5] ^ x[9], 7);
+
+    x[2] += x[6]; x[14] = hc_rotl32(x[14] ^ x[2], 16);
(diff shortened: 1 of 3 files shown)


hashscherlock761:
Quick question: do you know if there were any variations of this legacy format in the wild (e.g. different PBKDF2 iteration counts or salt lengths), or was it consistent across all older Eternl releases?

AndrewWestberg:
@hashscherlock761 There were no variations on the legacy format.  It was just v1 and now we have v2 in the wallet.

jsteube:
In "module_01852.c" the token table gives a separator to the signature alone, so the salt, nonce, tag and ciphertext after it are read back to back as fixed lengths. ST_HASH matches that, at 457 characters with one colon. The encoder writes a colon between all five fields instead, which comes out at 460 characters and four colons, and the tokenizer then rejects it, so a line written by --left cannot be given back to hashcat. Could you have the encoder emit the shape the tokenizer reads?

```
    const int out_len = snprintf (line_buf, line_size, "%s:%s:%s:%s:%s",
```

The module also asks for OPTI_TYPE_SLOW_HASH_SIMD_LOOP, and in "backend.c" that divides the loop kernel's work item count by the device vector width. That width drops to 1 for a pure kernel only on a GPU, so a CPU device keeps the width it reports, while the loop kernel in "m01852-pure.cl" guards on gid by itself and carries one candidate per work item. The candidates after the first of each vector then never run their iterations. A pure kernel that sets the flag takes the width into the guard and works in the vector types throughout, as "m07100-pure.cl" does:

```
  if ((gid * VECT_SIZE) >= GID_CNT) return;
```

Either removing the flag or moving the loop kernel to that shape closes it, and I would rather have the second. On the optimized kernel beside it, 1 of the 44 slow hash modes added between 30000 and 70000 carries an -optimized.cl, so a new one is normally pure only, and this one is a second implementation of ChaCha20, Poly1305 and the PBKDF2 loop that has to stay in step with the first. Would you take it out? There is no test module either. It can be Perl or Python, since "test.sh" builds its module list from both m*.pm and m*.py, so either "m36800.pm" or "m01000.py" is a shape to copy.

The number to use is 37700. Two small ones in the module as well: it is indented at 4 spaces where the tree uses 2, and the include of <stdio.h> can go, since no other module carries it and they call snprintf() all the same.

AndrewWestberg:
Rebased onto current master and updated mode 37700 for the latest module interface.

Changes include:
- registered the assigned mode 37700 and removed the temporary 01852 implementation
- retained only the pure kernel, with `-O` correctly falling back to it
- implemented genuine SIMD PBKDF2 processing
- added the independent Perl reference module
- fixed host serialization and scalar ChaCha20 rotation typing
- updated `module_init()` and tokenizer declarations for current master

Post-rebase verification:
- full build: passed
- exact hash decode/encode round trip: passed
- malformed hash rejection: passed
- Perl generator, authentication, verifier, `$HEX[]`, and rejection checks: passed
- CPU OpenCL SIMD widths `1`, `2`, `4`, `8`, and `16`: exact `17/17` recovery at every width
- GPU OpenCL pure kernel: exact `17/17` recovery
- optimized-kernel request: correctly fell back to pure and recovered the test hash
- built-in self-tests: passed on CPU and GPU
- `tools/test.sh` CPU/GPU, single/multi: zero missing, mismatched, timed-out, or skipped hashes
- `git diff --check`: passed

jsteube:
In "src/modules/module_37700.c", the decoder reads the salt with a loop of its own, while the nonce, tag and ciphertext beside it each go through "hex_decode()":

```
  for (int i = 0; i < salt_len; i += 2)
  {
    const u8 p0 = salt_pos[i + 0];
    const u8 p1 = salt_pos[i + 1];
```

Token 1 is FIXED_LENGTH at 64, so the pair is always whole here. The shape still reads two characters per step while its bound counts one, and that is being taken out of the modules that carry it. Could you use "hex_decode()" for the salt as well, the way the three fields after it already do?

In "OpenCL/m37700-pure.cl" the init and loop kernels carry "KERNEL_FA" and "m37700_comp" does not. "m07100-pure.cl", which this kernel follows for its loop guard and its "i += 8" block, declares it on all three entry points. Was leaving it off the comp kernel deliberate?

Two buffers are wider than what goes into them. "eternl_t" reserves "salt_buf[64]" for a 32 byte salt, and DGST_SIZE is DGST_SIZE_8_16, which is 128 bytes, where the decoder writes 4 words and "m37700_comp" compares those same 4 at DGST_R0 to DGST_R3. Would DGST_SIZE_4_4 and a "salt_buf[8]" fit what the mode actually carries?

AndrewWestberg:
Thank you. I made the requested changes:

- Replaced the salt decode loop with `hex_decode()`.
- Added `KERNEL_FA` to `m37700_comp`. The omission was not deliberate.
- Changed `DGST_SIZE_8_16` to `DGST_SIZE_4_4`.
- Changed `eternl_t.salt_buf` from 64 words to 8 words in the host and OpenCL structures.

`make -j32` completed without errors. A direct CPU OpenCL test recovered the test hash: `Recovered 1/1`.

I ran the mode 37700 edge tests on CPU OpenCL:

```text
$ ./tools/test_edge.sh -m 37700 -D 1 -f
[ test_edge_1790016431 ] > All tests done in 0d:00h:08m:24s
[ test_edge_1790016431 ] > Errors detected: 0
```

jsteube:
In "src/modules/module_37700.c" the self test password is the one place 01852 survives:

```
  static const char *ST_PASS        = "hashcat01852";
```

The ".pm" you added regenerates both halves through the passthrough entry point:

```
  echo -n hashcat | perl tools/test.pl passthrough 37700
```

Could you take ST_PASS and ST_HASH from a run of that?

AndrewWestberg:
@jsteube I'm not changing the example data. I generated the data when I had access to the old wallet. That wallet is no longer available to generate new stuff. It's fine if the example data doesn't match the new number.

I'd prefer not to regenerate from something that isn't the source of the wallet itself.

PR #4591, opened 2025-12-18
Metamask update to accept both 24 and 44 byte salt + variable amount of rounds

b8vr:
metamaskMobile updated to also accept 44 byte salt as well as the existing 24 byte salt. It also accept future implementations with rounds different from the default 5000. Minimum 1000 and maximum 1000000 rounds are accepted. Extraction of hash from newer versions of metamask produce 44 byte salt. PoC can be tested by cracking known hashes, cracking the same hashes with rounds=5000 inserted and then with different rounds to see them fail.


b8vr:
The code is too different from the existing metamask modules and there seem to be an issue when the compiled version of it is used in hashtopolis. Replaced by pull request #4601.

Issue #4594, opened 2025-12-23
PDF 10500 tokenizer does not accept hashes with empty ID


aurabirb:
**Describe the bug**
The following hash is crackable (the password is `5001685`), but it is reported as invalid due to a 0-length ID:
<pre>$pdf$4*4*128*-1852*1*<b>0**</b>32*10c0a03e14d09d6d9ced7fa73cbc137c6afd6e2adb64e55c15ef36beae718a0b*32*8ccdd43f9b50e7951a159d1912836d6df735dd882c17051cec184fee239fa653</pre>

The PDF I got it from had the following trailer: `trailer<</Size 37/Root 1 0 R/ID [() ()]/Encrypt 36 0 R>>` 
It was most likely produced by AFAS software.
See https://github.com/sighook/pdf2hashcat/pull/3 for extra details.

**To Reproduce**

Save this hash to a file named `pdf.hashcat` and test the cracking using a 7-digit pattern. It fails with the following error:

`echo '$pdf$4*4*128*-1852*1*0**32*10c0a03e14d09d6d9ced7fa73cbc137c6afd6e2adb64e55c15ef36beae718a0b*32*8ccdd43f9b50e7951a159d1912836d6df735dd882c17051cec184fee239fa653' > pdf.hashcat`
`./hashcat.bin --potfile-disable -m 10500 pdf.hashcat -a 3 ?d?d?d?d?d?d?d`

```
hashcat (v7.1.2) starting
...

Hashfile 'pdf.hashcat' on line 1 ($pdf$4...f735dd882c17051cec184fee239fa653): Token length exception

* Token length exception: 1/1 hashes
  This error happens if the wrong hash type is specified, if the hashes are
  malformed, or if input is otherwise not as expected (for example, if the
  --username or --dynamic-x option is used but no username or dynamic-tag is present)

No hashes loaded.
...
```

**Expected behavior**
I expect there to be no token length exception about the 0-length ID.
This hash type should be supported and crackable with a hash produced from [pdf2hashcat](https://github.com/sighook/pdf2hashcat/blob/master/pdf2hashcat.py) or [pdf2john](https://github.com/openwall/john/blob/bleeding-jumbo/run/pdf2john.py).
In fact, I have implemented the changes in the hashcat module 10500 in https://github.com/hashcat/hashcat/pull/4595

<details><summary>Output after a fix:</summary>

[...]

PR #4595, opened 2025-12-24
PDF 10500 Accept hashes with empty ID #4594

aurabirb:
This MR:
- Fixes issue #4594 about the pdf hash tokenizer not supporting hashes from PDFs with empty IDs (`**` in position 7):
`$pdf$4*4*128*-1852*1*0**32*10c0a03e14d09d6d9ced7fa73cbc137c6afd6e2adb64e55c15ef36beae718a0b*32*8ccdd43f9b50e7951a159d1912836d6df735dd882c17051cec184fee239fa653`
- The changes adjust the tokenizer to accept 0-length IDs and prime the salt correctly.

### DIFF
--- a/src/modules/module_10500.c
+++ b/src/modules/module_10500.c
@@ -190,12 +190,13 @@ int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSE
                    | TOKEN_ATTR_VERIFY_DIGIT;
 
   token.sep[6]     = '*';
-  token.len[6]     = 2;
-  token.attr[6]    = TOKEN_ATTR_FIXED_LENGTH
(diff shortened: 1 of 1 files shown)


jsteube:
A few things for a follow-up when you have time:

Formatting: else { needs its brace on its own line, and the blank line before salt->salt_iter got lost.

Maybe more important: Unit test have not been updated to reflect your changes. 

This vector works, password is hashcat:

```
$pdf$2*3*128*-4*1*0**32*8dd8a5f9997c3c564b2e793094055fc200000000000000000000000000000000*32*000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f
```

Would be good to have in tools/test_modules/m10500.pm.

PR #4600, opened 2026-01-03
Python Fernet HMAC support

Hamid-K:
This update adds support for cracking [Fernet](https://cryptography.io/en/latest/fernet/) encrypted tokens by focusing on the HMAC part of the message.


jsteube:
In "docs/fernet.md" the mode is built around a key derived as SHA256(passphrase), and the kernel takes the first 16 bytes of that digest as the Fernet signing key. A Fernet key in the cryptography library is 32 random bytes, or PBKDF2HMAC over a password where a password is the source, so a token keyed on a bare SHA256 comes from an application that chose that for itself. Which application is it? Whether a mode number fits here turns on that answer, and the answer is not in the branch.

A doc for one mode is not how hashcat carries this. The files in "docs/" cover features and project reference, and none of them is about a single mode, so what a user needs about one arrives through the generated example hash table and through the module. The place for what "docs/fernet.md" holds is module_usage_notice(), which --hash-info prints. Would you move the key derivation and the 256 byte token limit into one of those? A note of that size fits, and "module_06800.c" is already doing the same job of separating the construction hashcat takes from a close one:

```
  return "The hash used here is not the one sent via e.g. the web interface to LastPass servers (pbkdf2_sha256_hex (pbkdf2_sha256 ($pass, $email, $iterations), $pass, 1) but instead the one stored (by e.g. your browser or the pocket version) to disk. ...";
```

The change also carries the fork it came from. In "docs/hashcat-example-hashes.md" the edits are by hand, although "generate_docs.sh" rebuilds that file from --example-hashes on every commit to master and it opens by warning that manual changes are overwritten. The example set is the same kind of thing, since 0, 400 and 500 are the three modes that carry one and "example99998.dict" repeats "example.dict" as well. Could those come out, along with the lines added to "README.md"?

```
Custom additions in this fork:
- Fernet hash mode `99998` (HMAC-SHA256, key = SHA256(pass)) in `docs/fernet.md`
```

A test module is missing on the other side. It can be Perl or Python, since "test.sh" builds its module list from both m*.pm and m*.py, so either "m36800.pm" or "m01000.py" is a shape to copy. On the number, 99998 sits directly under 99999, which is the Plaintext utility mode, so it reads as a placeholder picked to stay clear of the sequence rather than an assignment. A real one comes with the answer to the first question.

PR #4601, opened 2026-01-07
Metamask update to accept both 24 and 44 byte salt + variable amount of rounds of salt

b8vr:
This replaces pull request #4591. 
metamaskMobile updated to also accept 44 byte salt as well as the existing 24 byte salt. It also accept future implementations with rounds different from the default 5000. Minimum 1000 and maximum 1000000 rounds are accepted. Extraction of hash from newer versions of metamask produce 44 byte salt. PoC can be tested by cracking known hashes, cracking the same hashes with rounds=5000 inserted and then with different rounds to see them fail.

### DIFF
--- a/src/modules/module_31900.c
+++ b/src/modules/module_31900.c
@@ -63,6 +63,7 @@ typedef struct pbkdf2_sha512_aes_cbc
 
 } pbkdf2_sha512_aes_cbc_t;
 
+static const u32   ROUNDS_METAMASK_WALLET    = 5000;
 static const char *SIGNATURE_METAMASK_WALLET = "$metamaskMobile$";
 
 char *module_jit_build_options (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra, MAYBE_UNUSED const hashes_t *hashes, MAYBE_UNUSED const hc_device_param_t *device_param)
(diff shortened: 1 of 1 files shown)


jsteube:
In "module_31900.c" the second tokenizer pass cannot succeed. The two specs differ in one value:

```
    token.len_min[1] = 0;
    token.len_max[1] = 60;   //includes 44 bytes salt
```

```
    token.len_min[1] = 0;
    token.len_max[1] = 40;   //includes 24 bytes salt
```

Everything else in the retry is identical, so its range sits inside the first one's. A line the retry would take was already taken by the first pass, and a line the first pass turned down is turned down again, so the block is unreachable and one spec covers both salt lengths on its own.

The rounds value underflows on the way in:

```
    salt->salt_iter = hc_strtoul ((const char *) token.opt_buf + 7, NULL, 10) - 1; // 7 = "rounds="
```

A hash carrying rounds=0, or a rounds= that is not a number at all, gives salt_iter 0xffffffff, and the run then sits in the loop kernel for as long as anyone waits. The module this follows, "module_07400.c", has no minus one on that line for the same reason. Your description mentions a floor of 1000 and a ceiling of 1000000, and neither appears in the code.

The bound is looser than the goal as well. Two salt lengths are wanted here, 24 and 44, and 0 to 60 takes everything between them, an empty salt included.

Smaller things, all in the same file: 9 of the added lines carry trailing whitespace, token.len[0] is 16 now where it was strlen(SIGNATURE_METAMASK_WALLET), the new snprintf prints a u32 with %d, the if and the else in module_hash_encode() have multi line bodies and no braces, and ST_HASH has no rounds= form, so the self test never reaches the new path.

PR #4602, opened 2026-01-08
Don't regen bindings.rs every time

fukc-gihtub:
It causes intermittent build failures like this one:

```
warning: unused import: `bindings::*`
  --> /home/chicken/testhashcat/hashcat/Rust/hashcat-sys/src/lib.rs:10:9
   |
10 | pub use bindings::*;
   |         ^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

gcc    -std=gnu99 -flto=auto -march=native -mtune=native -W -Wall -Wextra -O2 -fomit-frame-pointer -fno-plt -pipe -Iinclude/ -IOpenCL/ -Ideps/LZMA-SDK/C -Ideps/zlib -Ideps/zlib/contrib -Ideps/OpenCL-Headers -DWITH_BRAIN -Ideps/xxHash -DWITH_CUBIN -Ideps/unrar -DWITH_HWMON src/modules/module_00121.c obj/combined.NATIVE.a -o modules/module_00121.so -flto=auto -Wno-lto-type-mismatch -s -lstdc++ -lpthread -ldl -lrt -lm -shared -fPIC -D MODULE_INTERFACE_VERSION_CURRENT=700
error[E0432]: unresolved imports `hashcat_sys::bridge_context_t`, `hashcat_sys::generic_io_t`, `hashcat_sys::generic_io_tmp_t`, `hashcat_sys::salt_t`
  --> src/interop.rs:14:19
   |
14 | use hashcat_sys::{bridge_context_t, generic_io_t, generic_io_tmp_t, salt_t};
   |                   ^^^^^^^^^^^^^^^^  ^^^^^^^^^^^^  ^^^^^^^^^^^^^^^^  ^^^^^^ no `salt_t` in the root
   |                   |                 |             |
   |                   |                 |             no `generic_io_tmp_t` in the root
   |                   |                 no `generic_io_t` in the root
   |                   no `bridge_context_t` in the root

For more information about this error, try `rustc --explain E0432`.
error: could not compile `dynamic_hash` (lib) due to 1 previous error
make: *** [src/bridges/bridge_rust_generic_hash.mk:64: bridges/subs/dynamic_hash.so] Error 101
make: *** Waiting for unfinished jobs....
cp Rust/bridges/generic_hash/target/release/libgeneric_hash.so bridges/subs/generic_hash.so
```


pwrbob:
This can lead to problems if architecture-dependent `#ifdef`s are ever used. According to the `bindgen` user manual, when doing this in a `build.rs` script, the file should always be regenerated. I'm working on a Rust bridge that will have this property.
Source: https://rust-lang.github.io/rust-bindgen/library-usage.html

Concerning the error: I don't quite understand how you are getting these errors. Are there maybe multiple instances of `cargo` compiling different crates at the same time, with one overwriting the `bindings.rs` file while another is trying to read it? If so, a better solution may be to use `.NOTPARALLEL: $(RUST_SUBS_DIR)/%.so` or similar in `bridge_rust_generic_hash.mk`. Cargo parallelizes compilation internally, so this will not introduce a lot of slowdown.

fukc-gihtub:
There are no architecture-dependent ifdefs in `hashcat_types.h` currently, so I don't think we need to run bindgen. It also introduces a libclang dependency: https://github.com/hashcat/hashcat/pull/4603

matrix:
Hi, @fukc-gihtub @pwrbob maybe I found a solution. Could you please review it to see if it looks correct?

jsteube:
Thanks for reporting this, and sorry it sat for so long. The diagnosis was correct and the race is fixed on master, so I am closing it.

The fix removed the shared file rather than changing the regeneration logic. `Rust/hashcat-sys/build.rs` now writes the bindings into `OUT_DIR`, which belongs to one build, instead of `src/bindings.rs` in the shared source directory. The two bridge crates are built by separate Cargo invocations that make can run in parallel, so that shared path was the whole problem. One build script could rewrite the file while the other crate was still reading it, and the unresolved import errors you posted were caused by a partially written file rather than by the bindings themselves.

This approach also addresses @pwrbob's concern, which I think was valid. The bindings are still generated for every build that needs them, with `cargo:rerun-if-changed` on `src/hashcat_types.h` and `build.rs`, so if an architecture dependent `#ifdef` is ever added to that header, the generated output follows it. Your point that there are none today was correct, but committing the generated file would make that remain correct only as long as nobody adds one later, which is difficult to guarantee.

The libclang dependency, which was your other concern, was handled where it belonged by installing it in the containers that were missing it rather than avoiding bindgen entirely.

#4652 reached the same race from the Makefile side and is being closed alongside this one. You found it first, and the reproduction you posted made the problem obvious, so thanks for that.

PR #4603, opened 2026-01-12
Update installation command in BUILD_WSL.md

JustGuardian:
Added 'clang' in installation command.


jsteube:
Thanks for looking into this. Adding clang moves the instructions in the right direction, but this line is still incomplete for the current Windows build.

The Rust bindings need libclang-dev, and the Windows Rust plugins are linked with -fuse-ld=lld, so lld is also required. I think the docs also list wget and curl, but neither is currently included in the package list.

Could you please update the dependency list to include at least: libclang-dev lld wget curl

Using libclang-dev here is preferable to relying on the clang compiler package to provide a compatible library indirectly. This also matches the dependencies used by the repository's Docker build.

PR #4604, opened 2026-01-13
Refactor process_json for userId handling and fallbacks

Svenja240602:
Updated process_json to handle new userId extraction and added fallback for old account formats.

### DIFF
--- a/tools/bitwarden2hashcat.py
+++ b/tools/bitwarden2hashcat.py
@@ -92,22 +92,63 @@ def process_leveldb(path):
 
 def process_json(data):
     data = json.loads(data)
-
+    out = []
+    
+    # new version: userId from global_account_activeAccountId
     try:
-        out = []
+        user_id = data.get("global_account_activeAccountId", None)
+        if user_id:
+            out.append(extract_json_profile_new(user_id, data))
+            if out:
+                return out
+    except (KeyError, TypeError):
+        pass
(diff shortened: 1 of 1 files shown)


jsteube:
In "bitwarden2hashcat.py" the new branch returns whether or not it found anything:

```
        user_id = data.get("global_account_activeAccountId", None)
        if user_id:
            out.append(extract_json_profile_new(user_id, data))
            if out:
                return out
```

Inside extract_json_profile_new() the hash and the iteration count are read with .get() and a default of None, so a vault that carries global_account_activeAccountId but keeps the master key hash somewhere else still gets a tuple appended. The list is never empty after that append, so the two older formats below are unreachable whenever that key is present.

Both outcomes reach the caller. A missing masterKeyHash prints a line that looks like a hash and is not one:

```
$bitwarden$2*600000*2*Tm8gZW1haWwgZm91bmQ=*None
```

A missing kdfConfig instead ends the run on "TypeError: %d format: a real number is required, not NoneType", which lands outside the try that would have caught it.

Could you have extract_json_profile_new() return None when the hash or the iteration count is missing, and append only a complete tuple, so the older formats still get their turn? The "No email found" default has the same shape, since it travels into the output rather than stopping it, and that is the base64 in the line above.

8 of the added lines carry trailing whitespace.

PR #4605, opened 2026-01-14
simplification of Rust bridge implementations

pwrbob:
These changes move duplicate code from the crates `generic_hash` and `dynamic_hash` into the `hashcat-sys` crate, allowing its reuse in both bridges (and all future ones).

Additionally, the structure of `hashcat-sys` was slightly modified to allow new Rust bridges to be implemented more easily.

Various other minor changes and small doc updates were made.


Rust/hashcat-sys/Cargo.toml
```diff
@@ -3,5 +3,8 @@ name = "hashcat-sys"
 version = "0.1.0"
 edition = "2024"
 
+[dependencies]
+anyhow = "1.0.100"
```
fukc-gihtub, line 7: I'd avoid third-party crates in dependencies.

pwrbob: The goal of this PR is to simplify the code, and avoid having to implement a custom error enum for (currently) one situation.

Why should third-party crates be avoided? A big advantage of Rust is its crate ecosystem. All bridges that currently exist depend on many third-party crates anyways.

fukc-gihtub: Adding a dependency just to create an error that will be ignored anyway? Looks very much like bloat.

jsteube:
In "hashcat-sys/src/generic.rs" the two helpers that moved out of the bridges lost the unsafe from their signatures. A pointer that gets past the null check can still be dangling or misaligned, and a length that does not match the allocation is undefined behaviour whatever the pointer is, so the obligation the unsafe marked is still on the caller. Master carries both as unsafe fn, and the point of the move is that future bridges build on this crate, so could the marker stay on them? That one blocks for me. The function vec_from_raw() shows it:

```
pub fn vec_from_raw<T: Clone>(data: *const T, length: c_int) -> Result<Vec<T>> {
    if data.is_null() {
        Err(anyhow!("null pointer encountered in conversion to Vec<T>"))
    } else {
        Ok(Vec::from(unsafe {
            slice::from_raw_parts(data, length as usize)
        }))
    }
}
```

The same file no longer asserts on two of the pointers new_context() is handed. A null st_salts_buf or st_esalts_buf now gives an Err that .unwrap_or_default() turns into an empty Vec, and the first get_raw_esalt() of a self test indexes that and panics, where the assert used to stop it at the boundary and name the pointer. Would you put the two back? At the branch point all five were guarded:

```
    assert!(!module_name.is_null());
    assert!(!salts_buf.is_null());
    assert!(!esalts_buf.is_null());
    assert!(!st_salts_buf.is_null());
    assert!(!st_esalts_buf.is_null());
```

The Result those helpers now return is not read anywhere. All 12 call sites end in .unwrap_or_default(), and anyhow enters "hashcat-sys" as a dependency to supply an error that no caller acts on. Could it come back out, unless one of them is going to handle it?

A rebase comes before any of this can be looked at in place, since the attack mode 9 work added salt_per_pw to ThreadContext and to the new_context() signature after this branched, and that is what both conflicts are. One line stands apart from the refactor though. In ".github/workflows/rust.yml" the rustfmt check runs with --edition 2021 while every crate under "Rust/" declares edition 2024, and this sets it to 2024, so it could go in without waiting on the rest. Would you rather send that as its own pull request, or shall we take it from here?

PR #4608, opened 2026-01-17
Added new attack-mode: Probabilistic Context-Free Grammar Attack (aka "PCFG") + Multiple Improvements/Bugfixes

matrix:
Patch contents:

-  Hashcat PCFG Attack Mode (-a 10)

Attack mode 10 (-a 10) is a new hashcat attack mode that generates password candidates using PCFG
(Probabilistic Context-Free Grammar) models. Instead of generating combinations blindly (like brute-
force -a 3) or reading from a file (like dictionary -a 0), it learns the patterns and building blocks of real
passwords from training data, then generates candidates that follow those patterns.

More info on docs/hashcat-pcfg-attack-mode.md

- Metal 4 support (full)
Apple has introduced new APIs that can be used from macOS 26 onwards. It seemed right to me to fully support them within hashcat

- Multiple Improvements

> Docker: update hashcat_binaries to version 2.1
> Enable SSE2 intrinsics path for scrypt-jane bridge on ARM via sse2neon
> Exclude tools directory from github action build triggers
> Improved performance by skip unnecessary memset of pws_comp/pws_idx in dispatch loops
> Replaced qsort with in-place MSD radix sort for non-salted hash deduplication
> Replaced scalar line counting with SIMD-accelerated implementation (AVX2/SSE2/NEON)
> Updated benchmark_deep.pl: now support --attack-mode (only 0, 3 and 10) and other options
> Updated github workflow: add OpenMP support, fix NetBSD job
> Updated install_modules.sh: add missing py sm3utils, fix py cryptocplus, fix perl DES, handle RPI
> Updated makefile: add OpenMP support, switch from -O2 to -O3
> Updated test_edge.sh: fix 33700 handling, skip luks, add more stats
> Updated BUILD.md: add missing Rustup stage and required packages for Linux/MacOS

- Multiple bugfixes

> Fixed bug on 35500 kernel with Metal
> Fixed bugs on 13731, 13732 and 13733 with AMD: add missing module_jit_build_options to force native_threads to 32 with GPU devices
> Fixed memory leak in loopback context (missing filename deallocation)
> Fixed module_constraints for 14400, 15000, 34400, 34500, 35300 and 35400
> Fixed module_generate_hash for 26900 by strip trailing zero bytes from engineID
> Fixed setShouldMaximizeConcurrentCompilation on Metal backend for parallel PSO compilation
> Fixed test.pl indirect method call syntax for is_count
> Fixed test.sh attack mode 3 and 6 for hash type 3000, 8500, 8501, 7701, 11500, 14000, 14100, 14400, 16000 and 27900
> Fixed test.sh attack mode 7 for hash type 33501 and 33502
> Fixed test.sh cryptoloop_test bug for hash type 14553
> Fixed unsupported get_segwit_address call for modules 28504 and 30904
> Fixed u32 multiplication overflow in pointer arithmetic for large hash/salt counts
> Fixed signed integer issues (int to u32) in find_hash(), hashcat_status_t, and sort_by_salt()
> Fixed AUX2/AUX3/AUX4 kernel dispatch returning AUX1 pipeline on both Metal and OpenCL backends
> Fixed Python reference counting bugs in bridge (crash with Python 3.14+)
> Fixed --runtime usage togheter with --stdout


fukc-gihtub:
@lakiw

matrix:
pcfg models must be retrained after "updated PCFG (2)" commit

fukc-gihtub:
You could consider adding an option to enumerate the start token in a burst instead of the end token.

L3 D2 C1
cat 12 !
dog 12 !
rat 12 !
etc

instead of
cat 12 !
cat 12 ?
cat 12 #

matrix:
> You could consider adding an option to enumerate the start token in a burst instead of the end token.

Hi, I've added lots of things, including second-order Markov models and OMEN
Anyway, I'll try to understand, since you're talking about another mode ;)

Thanks!

matrix:
@fukc-gihtub 

Input
```
$ cat example.dict.test2
apple
grape
lemon
apple1
grape1
lemon1
apple2
grape2
lemon2
```

output using the new parameter
```
$ ./hashcat -a10 --pcfg-train example.dict.test2 --pcfg-order-mode 0 --stdout --pcfg-burst-first

PCFG: Exporting model to hashcat.model.pcfg ...
PCFG: Export complete - 2 structures, 5 terminals written

PCFG: Loaded filtered structures (2 of 2) into model ...
PCFG: Loaded 5 filtered terminals into model ...

apple1
grape1
lemon1
apple
grape
lemon
apple2
grape2
lemon2
```

and output without it

```
$ ./hashcat -a10 --pcfg-train example.dict.test2 --pcfg-order-mode 0 --stdout

PCFG: Exporting model to hashcat.model.pcfg ...
PCFG: Export complete - 2 structures, 5 terminals written

PCFG: Loaded filtered structures (2 of 2) into model ...
PCFG: Loaded 5 filtered terminals into model ...

apple1
apple2
apple
grape
lemon
grape1
grape2
lemon1
lemon2
```

;)

matrix:
Complete test.sh session using NVIDIA GeForce RTX 4080
[test_d4.log](https://github.com/user-attachments/files/25920997/test_d4.log)

Got the same result with Apple Metal (also Metal 4)

matrix:
Complete test_edge.sh session using NVIDIA GeForce RTX 4080
[test_edge_d4.log](https://github.com/user-attachments/files/25963245/test_edge_d4.log)

fukc-gihtub:
This PR fails to build with the current master.

`pcfg_trainer.c:(.text+0x3f3): undefined reference to XXH64`

matrix:
> This PR fails to build with the current master.
> 
> `pcfg_trainer.c:(.text+0x3f3): undefined reference to XXH64`



> This PR fails to build with the current master.
> 
> `pcfg_trainer.c:(.text+0x3f3): undefined reference to XXH64`

Hi, if you could give me a little more information to put this into context, I could look into it further

TovStalin:
Hi. I try to merge this PR with current stable branch, compile via 'make', and run `hashcat --help`.
But I got a segmentation fault.
I use Fedora Linux 44 (KDE), x86_64.

matrix:
> Hi. I try to merge this PR with current stable branch, compile via 'make', and run `hashcat --help`. But I got a segmentation fault. I use Fedora Linux 44 (KDE), x86_64.

Hi, I just tried it again and I don't see any segmentation faults. Try running “make clean” first.

dalmatele:
I've worked with this algorithm and think it has 2 big big problems:

- It consumes too much RAM.  If you monitor the tool from @lakiw you will see this tool will consume all the memory you have and even all the SDD write bandwidth if you make swap.
- Its restore mechanic is bad, it takes too much RAM to restore and too much time to find the restore point.
Have you tested above problems?
And one more question, do you have bechmark for this new option?

matrix:
> I've worked with this algorithm and think it has 2 big big problems:
> 
> * It consumes too much RAM.  If you monitor the tool from @lakiw you will see this tool will consume all the memory you have and even all the SDD write bandwidth if you make swap.
> * Its restore mechanic is bad, it takes too much RAM to restore and too much time to find the restore point.
>   Have you tested above problems?
>   And one more question, do you have bechmark for this new option?

Hi @dalmatele 
it uses up all the RAM and all the swap space if you're working with large input files. What exactly is your use case and your hardware? As for the restore, I've made it as fast as possible (I have a private one, but I don't have time to benchmark the two right now). Again, please specify your use case in detail.

Generally speaking, if you have any concrete suggestions on how to optimize these two aspects... please share them; any suggestion is welcome.

Thanks!

[edit] I've found a way to optimize memory more effectively

jsteube:
First of all, thanks a lot for this. PCFG is something I've wanted in hashcat for a long time, and it's great to finally see it implemented. I spent quite a while going through the PR. I've merged it on top of current master and played around with it a bit. It works and it actually works really well. I trained a model using only example.dict and recovered about 20% example0.hash in a short time.

So everything below is really just about architecture. I definitely want PCFG in hashcat. I just think we should integrate it a bit differently.

My feeling is that this should go into an -a 8 plugin instead of becoming a new -a 10. I know you mentioned that this isn't possible right now, and that's completely true. My answer is simply: let's make it possible. That's basically why I started -a 8 in the first place. It's still only on master, nothing depends on it yet, so this is exactly the right time to improve the design.

To me, attack modes are about connecting backend logic with kernel logic to enable amplification. -a0 goes through _a0, -a1, -a6, and -a7 through _a1, -a3 through _a3, and even -a9 exists because it changes backend behavior and passes additional kernel parameters. Your -a 10 doesn't really do that. It runs on _a0 without introducing any new amplification. What it really provides is an excellent candidate generator, and that's exactly what I had in mind when I started working on -a 8.

The nice thing is that -a8 already supports --keyspace, -s, and -l, and it also supports generators with an unknown keyspace. Rules work on top of a generator as well. Every compute device gets its own generator, so there is no central feeding bottleneck, and the new double buffered candidate pipes keep generating while the GPU is busy. The interface is also pretty simple. hashcat assigns a keyspace range, calls seek(), and then repeatedly calls next(). So the main question is really whether the generator can seek, because that's basically the equivalent of --restore.

From what I can see, modes 0, 2, 4, and 6 already support this. Actually, mode 0 already contains most of what -a8 needs. pcfg_cpu_random_gen_skip() already splits the work between generators. Under -a8, hashcat would simply do that job instead, so this actually removes code instead of adding more.

AHF is the one I looked at hardest, and I think it comes out better in -a8 than in -a10. Having no keyspace is fine, that is the unknown keyspace case and -a8 handles it, threaded, seeking included. Seeking there means replaying, so it is slow, but it works. The only thing in the way is that the AHF rng is seeded from time, thread id, and the generator pointer, which is exactly why restore has to be refused for it today. If we derive that seed deterministically then replay becomes possible, and then AHF gets --restore, which it doesn't have right now. So I don't think you lose AHF by moving it over. I think it ends up better than it is today.

There are really only two things I'd like to improve in -a8 to support PCFG properly: native parameter mapping, so generators feel much more like native hashcat functionality, and a way for plugins to add status lines. Information such as the current PCFG structure, probability, count, and length is useful, and users will definitely want to see it. I'll work on this next.

There are also two things I'd probably move out of the attack mode. The first one is training. Right now it still needs a hash argument, which feels a bit strange to me. I think training deserves its own little tool. Then it can train, inspect, merge, update, and diff models without flooding hashcat option space, while the -a8 generator simply loads a model and generates candidates.

The second one is the GPU generator. This is really the only thing I'd give up by moving it to -a8, and I don't want to pretend otherwise. There's no amplification in it anyway, so it's not that we lose performance. That's why I think it's a good trade. PCIe is just so fast these days that feeding candidates from the host is fast enough, while being much simpler to understand and maintain. The host generator reaches around 80 MH/s. But 80 MH/s is already enough to feed a GPU, and with -a8 every GPU naturally gets its own host generator. The only case where the device generator really wins is PCFG without rules on a fast hash, and I don't think that's the configuration users should normally run anyway. For slow hashes it makes almost no difference. In return we remove roughly 5600 lines of GPU specific code together with most of the backend integration.

The good news is that almost all of your work stays exactly as it is. The model format, tokenizer, CPU generators, statistics, and everything around them would remain untouched. What mostly disappears is infrastructure that hashcat already has.

One unrelated thing. The PR also contains quite a few changes that have nothing to do with PCFG: the Metal 4 fixes, radix sort, SIMD line counting, swap detection, automatic bitmap sizing, old_apply_rule, the find_hash cleanup, and the -O3 Makefile change. Several of those actually look really nice, but inside a 32k line PCFG PR they are almost impossible to review properly. I'd split those into separate PRs. 

Finally, I think it would be nice to add the original academic work to credits.txt. PCFG comes from Weir, Aggarwal, and de Medeiros, while OMEN comes from Duermuth and others. There is no licensing issue, but I think giving proper credit is simply the right thing to do. It would also help to mention that what is called OMEN here is actually a cost ordering over PCFG terminals rather than the original character level enumerator from the paper. That would save some confusion for people familiar with the original work.

matrix:
> First of all, thanks a lot for this. PCFG is something I've wanted in hashcat for a long time, and it's great to finally see it implemented. I spent quite a while going through the PR. I've merged it on top of current master and played around with it a bit. It works and it actually works really well. I trained a model using only example.dict and recovered about 20% example0.hash in a short time.
> 
> So everything below is really just about architecture. I definitely want PCFG in hashcat. I just think we should integrate it a bit differently.
> 
> My feeling is that this should go into an -a 8 plugin instead of becoming a new -a 10. I know you mentioned that this isn't possible right now, and that's completely true. My answer is simply: let's make it possible. That's basically why I started -a 8 in the first place. It's still only on master, nothing depends on it yet, so this is exactly the right time to improve the design.
> 
> To me, attack modes are about connecting backend logic with kernel logic to enable amplification. -a0 goes through _a0, -a1, -a6, and -a7 through _a1, -a3 through _a3, and even -a9 exists because it changes backend behavior and passes additional kernel parameters. Your -a 10 doesn't really do that. It runs on _a0 without introducing any new amplification. What it really provides is an excellent candidate generator, and that's exactly what I had in mind when I started working on -a 8.
> 
> The nice thing is that -a8 already supports --keyspace, -s, and -l, and it also supports generators with an unknown keyspace. Rules work on top of a generator as well. Every compute device gets its own generator, so there is no central feeding bottleneck, and the new double buffered candidate pipes keep generating while the GPU is busy. The interface is also pretty simple. hashcat assigns a keyspace range, calls seek(), and then repeatedly calls next(). So the main question is really whether the generator can seek, because that's basically the equivalent of --restore.
> 
> From what I can see, modes 0, 2, 4, and 6 already support this. Actually, mode 0 already contains most of what -a8 needs. pcfg_cpu_random_gen_skip() already splits the work between generators. Under -a8, hashcat would simply do that job instead, so this actually removes code instead of adding more.
> 
> AHF is the one I looked at hardest, and I think it comes out better in -a8 than in -a10. Having no keyspace is fine, that is the unknown keyspace case and -a8 handles it, threaded, seeking included. Seeking there means replaying, so it is slow, but it works. The only thing in the way is that the AHF rng is seeded from time, thread id, and the generator pointer, which is exactly why restore has to be refused for it today. If we derive that seed deterministically then replay becomes possible, and then AHF gets --restore, which it doesn't have right now. So I don't think you lose AHF by moving it over. I think it ends up better than it is today.
> 
> There are really only two things I'd like to improve in -a8 to support PCFG properly: native parameter mapping, so generators feel much more like native hashcat functionality, and a way for plugins to add status lines. Information such as the current PCFG structure, probability, count, and length is useful, and users will definitely want to see it. I'll work on this next.
> 
> There are also two things I'd probably move out of the attack mode. The first one is training. Right now it still needs a hash argument, which feels a bit strange to me. I think training deserves its own little tool. Then it can train, inspect, merge, update, and diff models without flooding hashcat option space, while the -a8 generator simply loads a model and generates candidates.
> 
> The second one is the GPU generator. This is really the only thing I'd give up by moving it to -a8, and I don't want to pretend otherwise. There's no amplification in it anyway, so it's not that we lose performance. That's why I think it's a good trade. PCIe is just so fast these days that feeding candidates from the host is fast enough, while being much simpler to understand and maintain. The host generator reaches around 80 MH/s. But 80 MH/s is already enough to feed a GPU, and with -a8 every GPU naturally gets its own host generator. The only case where the device generator really wins is PCFG without rules on a fast hash, and I don't think that's the configuration users should normally run anyway. For slow hashes it makes almost no difference. In return we remove roughly 5600 lines of GPU specific code together with most of the backend integration.
> 
> The good news is that almost all of your work stays exactly as it is. The model format, tokenizer, CPU generators, statistics, and everything around them would remain untouched. What mostly disappears is infrastructure that hashcat already has.
> 
> One unrelated thing. The PR also contains quite a few changes that have nothing to do with PCFG: the Metal 4 fixes, radix sort, SIMD line counting, swap detection, automatic bitmap sizing, old_apply_rule, the find_hash cleanup, and the -O3 Makefile change. Several of those actually look really nice, but inside a 32k line PCFG PR they are almost impossible to review properly. I'd split those into separate PRs.
> 
> Finally, I think it would be nice to add the original academic work to credits.txt. PCFG comes from Weir, Aggarwal, and de Medeiros, while OMEN comes from Duermuth and others. There is no licensing issue, but I think giving proper credit is simply the right thing to do. It would also help to mention that what is called OMEN here is actually a cost ordering over PCFG terminals rather than the original character level enumerator from the paper. That would save some confusion for people familiar with the original work.

Keep in mind that the new version I'm working offline is much more compact in terms of code and add additional features into this new pcfg engine. As a result, I'm not sure how worthwhile it is to do the porting right now; it's definitely something we should work on together offline.

All the other changes included were necessary during development, which was very complicated without write access to the repository (in that case, I would have handled them differently). You’re the only one with permissions to submit code; if you stop… I’ll be stuck, and this is one of the consequences. I had to simplify things so as not to make my life more difficult :)
Nothing is impossible to recover, but it requires extra work and more time. In any case, over the past few months I’ve managed to split some features into separate PRs, which are already ready to be merged.

As for the credits, we can add whatever you want, but keep in mind that the code I wrote isn’t a port of someone else’s work, it’s completely new. I’ve obviously read the theory in academic papers. Consequently, following your line of reasoning, you should consider adding many more credits for various features/code added to Hashcat over the years (quite a task).

dalmatele:
@matrix : I've tested with both python and go version and got the same result: It consumes all my memory after a few day running
Go version:  https://github.com/cyclone-github/pcfg-go.git
Python version: https://github.com/lakiw/pcfg_cracker.git
My test hardware:
- CPU: 2 
- RAM: 6GB
- GPU: Tesla T4
My test data:  1 mils words from rockyou
Then I make a new improvement in this: https://github.com/dalmatele/pcfg-go.git
My improvement includes:
==========
Limit the ptChanSize = 1024
Limit the outputChanSize = 1024
Limit batchSize=32678 bytes
Limit writerBufSize = 1024 * 1024: this will limit the buffer output to 1MB
Limit defaultMaxQueueEntries = 250000 item
==========
The result, for now, is good. You can see below image:
image
image

jsteube:
Hey, we have been through the PCFG discussion plenty of times already, so I don't want to reopen that here. What I had not done until now was properly go through everything else in the PR, and there is actually quite a lot in there that has nothing to do with PCFG. That's what this comment is about.

Just to put the current state here for anyone following the PR who wasn't part of those discussions: PCFG is on master now as attack mode 4 in src/feeds/feed_pcfg.c, including OMEN support. The attack mode 10 implementation here makes up roughly 26k of the 32k lines in the PR, and the 2 implementations can't really coexist. I know that's not the outcome you were hoping for.

The trainer is the one PCFG part that has no equivalent on master, so I also want to give you a clear answer on that instead of leaving it hanging. I don't think we should have the trainer inside hashcat. Training is developing pretty quickly outside the project right now. There's the original pcfg_trainer, pcfg-go and cpcfg in particular already works with the new attack mode 4 support and has seen more improvements recently. Those projects can move much faster in that area than we can, so I'd rather keep hashcat on the consuming side and let the training projects focus on what they do best.

With that out of the way, here is where the rest of the PR stands against current master:

```
  already on master
  ---
  MSD radix sort for non-salted dedup   src/hashes.c
  sse2neon path for scrypt-jane on ARM  src/bridges/bridge_scrypt_jane.c:15

  master went a different way
  ---
  hash list line counting               hashlist_count_lines_threaded () counts
                                        on every core now, so the SIMD counter
                                        lands on different ground
  m28504.pm, m30904.pm                  changed by the test.sh -g work
  BUILD.md python version               #4780 set 3.13.9 on master, this
                                        branch moves it to 3.14.3

  still missing, still wanted
  ---
  Metal 4 support                       no MTL4 anywhere in src/ext_metal.m
  loopback filename leak                src/loopback.c:81 hcmalloc, then
                                        src/loopback.c:110 hc_asprintf
                                        replaces the pointer
  13731/13732/13733 on AMD              module_jit_build_options is still
                                        MODULE_DEFAULT in module_13731.c:350
  test.pm for 14400, 15000, 26900,
    34301, 34400, 34500
  BUILD.md rustup step                  no rustup in BUILD.md on master
  install_modules.sh                    see below
  exclude tools/** from build triggers  .github/workflows/build.yml:52 and :64

  not taking
  ---
  Makefile -O2 to -O3                   see below
  Makefile OpenMP, USE_OMP              only src/pcfg.c and
                                        src/pcfg_omen_stats.c used it
  BUILD.md libomp and llvm              part of the same OpenMP change
  PCFG trainer                          see above

  leave with me
  ---
  benchmark_deep.pl                     see below
```

There are a few smaller things I did not classify, like the memset skip in the dispatch loops, the Docker image bump and the test_edge.sh changes. You probably know immediately whether those are still relevant, so there is not much value in me trying to reconstruct all the history there.

On the Makefile, I would leave both parts out. I have been bitten by O3 too many times over the years. It may be perfectly fine these days, but the gain is small and I would rather keep O2 than spend time debugging some strange compiler issue later.

OpenMP also seems to be part of the PCFG work rather than something useful on its own. The only pragmas are in src/pcfg.c and src/pcfg_omen_stats.c, so without attack mode 10 nothing uses it anymore. In that case I would also drop the build support for now. If we need OpenMP later, we can add it together with whatever actually uses it.

install_modules.sh I would like, but not quite as it stands. The sm3utils addition and the Raspberry Pi handling are the parts I am after. Alongside those it switches pycryptoplus and Crypt::DES to forks under your own account, comments out pycryptodome, and adds a pip3 uninstall line that also removes setuptools. Could you split the first 2 out, and we can talk about the rest on its own?

benchmark_deep.pl is one I'd like to keep aside for the moment. It connects to some release work I want to look at myself, so please leave that one out of the split and I'll come back to you about it.

The loopback leak is probably the easiest one to pull out first. loopback_ctx->filename is allocated during init, then hc_asprintf replaces the pointer when a loopback file is opened, so the original allocation is lost. That's still on master and should make a nice small PR.

Metal 4 is probably the next one I'd most like to see separated. It's a real feature on its own, needs somebody with the hardware to properly test it, and right now it's buried inside a very large PR.

So I don't think rebasing the whole PR makes much sense anymore. I'd much rather see the remaining useful parts split into smaller PRs, and I'm happy to take them in whatever order works best for you.

PR #4609, opened 2026-01-17
Fix NetBSD build

matrix:
Using pkgin instead of pkg_add


Bakrsfff (approved):

matrix:
will be added to the PCFG PR to avoid conflicts

PR #4617, opened 2026-01-30
Fix for issue #4610

### LINKED ISSUE #4610: 3710 reports wrong salt length for -O
**Describe the bug**
If I run hashcat for -m 3710 (md5($salt.md5($pass))) with -O it reports me, that:
```
Minimum salt length supported by kernel: 0
Maximum salt length supported by kernel: 51
```

But if I use a longer salt, that is shorter than 52, e.g. length 30, hashcat is not able to crack the test hashes.

**To Reproduce**
Let's create test hashes:
```
$ echo -n "Hello" | md5sum
8b1a9953c4611296a827abf8c47804d7  -
$ sTmpText="123456789012345678901234567890123456789012345678901234567890"
$ for i in {1..51} ; do echo -n "${sTmpText:0:$i}" | sed -E 's/$/8b1a9953c4611296a827abf8c47804d7/' | md5sum | sed -E 's/^(.{32}).*$/\1:'${sTmpText:0:$i}'/' ; done >> TestHashes.hashes 
```
Thus we have a hash with every salt length from 1 to 51 bytes in TestHashes.hashes and with password "Hello".

Now let's crack it:
`hashcat-7.1.2>hashcat.exe -a 3 -m 3710 --potfile-disable -o ..\TestHashes_712O.potfile ..\TestHashes.hashes ?a?l?l?l?l -O`
and we get the result
```
Session..........: hashcat
Status...........: Exhausted
Hash.Mode........: 3710 (md5($salt.md5($pass)))
Hash.Target......: ..\TestHashes.hashes
[...]
Kernel.Feature...: Optimized Kernel (password length 0-55 bytes)
Guess.Mask.......: ?a?l?l?l?l [5]
Speed.#01........: [...]
Recovered........: 23/51 (45.10%) Digests (total), 23/51 (45.10%) Digests (new), 23/51 (45.10%) Salts
```
Thus hashcat accepts every hash but is not able to crack every one even though it is the same password for every hash. If we look into the output potfile we see, that hashcat was able to crack the hashes with salt length 1 to 23 but not with salt length from 24 to 51.
If we cross check and use pure kernel, hashcat is able to crack all 51 hashes.

**Expected behavior**
I would expect that hashcat is able to crack hashes that obey the constraints hashcat itself give.

I looked into the kernel code and see that it calculates two md5s, a inner one around the password and a outer one, both with only one block.  For md5 we have 64 byte each block, removing 9 byte for overhead thus 55 byte remains for payload. The inner hash in hexadecimal took up 32 byte thus only 23 remain for salt. This explains why only salts up to 23 are cracked.

Well of cause it is possible to calculate two blocks for the outer md5 but I think the number hashcat reports should be changed.

 - OS: Windows and Linux
 - hashcat Version: 6.2.6, 7.1.2 and current beta 7.1.2-338-g18c6aaf

**Additional notes**
I saw that the kernel copies and calculates with 64 byte of salt. I wonder if there are side effects that this is necessary or if more than half of the salt code can be removed from optimized kernel.

4RH1T3CT0R7:
Problem: Hash mode 3710 (md5($salt.md5($pass))) with the optimized kernel (-O) reported a maximum salt length of 51 bytes, but silently failed to crack hashes with salts longer than 23 bytes.                                                                                                                             
                                                                                                                                                                
Root cause: The optimized kernel uses single-block MD5 (64 bytes). With 9 bytes of overhead (0x80 padding + 8-byte length field), 55 bytes remain for payload.
The inner MD5 hex output consumes 32 bytes, leaving only 23 bytes for the salt. The module had no custom module_salt_max(), so it fell through to default_salt_max() which returned SALT_MAX_OLD = 51.                                                                                                          
                                                                                                                                                                
Fix in src/modules/module_03710.c:                                                                                                                            
 1. Added a module_salt_max() function (lines 48-58) that returns 23 when using optimized kernels, and the default hashconfig->salt_max otherwise              
2. Wired it up in module_init() at line 227: module_salt_max instead of MODULE_DEFAULT                                                                        
                                                                                                                                                                
This follows the same pattern used by mode 4710 (src/modules/module_04710.c:50-59), which also overrides module_salt_max() for its optimized kernel constraints.                                                                                                                                                  
                                                                                                                                                                
Effect: With -O, hashcat will now correctly report "Maximum salt length supported by kernel: 23" and reject hashes with salts > 23 bytes upfront, instead of silently failing to crack them.                           

fix #4610

### DIFF
--- a/src/modules/module_03710.c
+++ b/src/modules/module_03710.c
@@ -45,6 +45,18 @@ u32         module_salt_type      (MAYBE_UNUSED const hashconfig_t *hashconfig, 
 const char *module_st_hash        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_HASH;         }
 const char *module_st_pass        (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra) { return ST_PASS;         }
 
+u32 module_salt_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
+{
+  const bool optimized_kernel = (hashconfig->opti_type & OPTI_TYPE_OPTIMIZED_KERNEL);
+
+  // optimized kernel uses single-block MD5: 64 - 9 (overhead) = 55 bytes payload
+  // inner md5 hex output = 32 bytes, leaving 55 - 32 = 23 bytes for salt
+
+  const u32 salt_max = (optimized_kernel == true) ? 23 : hashconfig->salt_max;
+
+  return salt_max;
+}
+
 int module_hash_decode (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED void *digest_buf, MAYBE_UNUSED salt_t *salt, MAYBE_UNUSED void *esalt_buf, MAYBE_UNUSED void *hook_salt_buf, MAYBE_UNUSED hashinfo_t *hash_info, const char *line_buf, MAYBE_UNUSED const int line_len)
 {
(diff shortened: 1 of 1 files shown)


jsteube:

PR #4618, opened 2026-01-30
Fix stack smashing in rule parsing

Chick3nman:
An incorrect bounds check introduced by a previous addition to rp_cpu.c leads to stack smashing during rule parsing


Chick3nman:
`iZ" zZ f d H`

This rule will cause stack smashing in master without the above fix

matrix (approved):

matrix:
fix #4664

PR #3950, opened 2024-02-15
Improve performance of scrypt-based algorithms by code reordering.

fse-a:
Performance improvements to scrypt-based algorithms. In particular for `scrypt_mix_init()` and `salsa_r()`. Part of these methods is the reordering (demultiplexing) of elements in `X[]` and `TI[]`. By changing the moment of reordering, the GPU (or compiler) seems to be more efficient which improves performance.

Tested with the self-test hash on the following device: `Device #1: NVIDIA GeForce RTX 3070, 7113/8191 MB, 46MCU`
Core was: `1725 MHz`
 
| Plugin | Speed Before | Speed After |
| ------- | ------- | ------- |
| 8900|    1660 H/s|   1863 H/s|
|15700|       8 H/s|      9 H/s|
|22700|    1676 H/s|   1889 H/s|
|27700|    1667 H/s|   1898 H/s|
|28200|    1651 H/s|   1865 H/s|

PR #4631, opened 2026-02-22
Reapply shuffle-before-salsa optimization for scrypt kernels

npinto:
The v7.0.0 refactoring (c231b2ec5) rewrote `salsa_r()` from `uint4` to `u32` and extracted the scrypt kernel code into the shared `inc_hash_scrypt.cl`. During this rewrite, the shuffle/salsa reordering from PR #3950 was inadvertently lost. `scrypt_shuffle()` ended up back after `salsa_r()` at all three call sites.

This PR restores the optimization in the new shared code structure.

The approach operates in an "inverse-shuffle domain" throughout the SMix computation. 

### Cross-architecture benchmark (5 runs per mode, hashcat v7.1.2 master)

| GPU       | Arch   | Before (H/s) | After (H/s) | Change   |
|-----------|--------|---------------|--------------|----------|
| RTX 3070  | Ampere | 1743          | 2060         | **+18.2%** |
| RTX 4070  | Ada    | 3424          | 4040         | **+18.0%** |
| RTX 4090  | Ada    | 7986          | 8910         | **+11.6%** |

Averages across modes 8900, 22700, 27700, 28200 (all R=8).

The RTX 4090 shows slightly lower improvement (+11.6% vs +18%) likely because its larger L2 cache (72 MB) better absorbs stack spill traffic.

### Kernel resources (m08900_loop, same across all GPUs)

| Metric | Before | After  |
|--------|--------|--------|
| REG    | 255    | 255    |
| STACK  | 3216   | 2208 (-31%) |

The speedup compared to the original PR #3950 (+12-14%) is larger because the new `u32` salsa code benefits more from the reordering. The compiler achieves better register scheduling with reduced stack spill.

### Per-mode detail (RTX 3070)

| Mode  | Before (H/s) | After (H/s) | Change |
|-------|-------------|------------|--------|
| 8900  | 1744        | 2064       | +18.3% |
| 15700 | 6           | 6          | R=1    |
| 22700 | 1742        | 2057       | +18.1% |
| 27700 | 1743        | 2059       | +18.2% |
| 28200 | 1745        | 2059       | +18.0% |

### DIFF
--- a/OpenCL/inc_hash_scrypt.cl
+++ b/OpenCL/inc_hash_scrypt.cl
@@ -41,6 +41,23 @@ DECLSPEC void scrypt_shuffle (PRIVATE_AS u32 *TI)
     for (int j = 0; j < SALSA_CNT4; j++) TI[dst_off + j] = TT[src_off + j];
   }
 }
+
+DECLSPEC void scrypt_shuffle_inv (PRIVATE_AS u32 *TI)
+{
+  u32 TT[STATE_CNT4];
+
+  for (int j = 0; j < STATE_CNT4; j++) TT[j] = TI[j];
+
+  for (int dst_off = SALSA_CNT4 * 2, src_off = SALSA_CNT4; dst_off < STATE_CNT4; dst_off += SALSA_CNT4 * 2, src_off += SALSA_CNT4)
+  {
+    for (int j = 0; j < SALSA_CNT4; j++) TI[dst_off + j] = TT[src_off + j];
+  }
+
+  for (int dst_off = SALSA_CNT4, src_off = STATE_CNT4 / 2; dst_off < STATE_CNT4; dst_off += SALSA_CNT4 * 2, src_off += SALSA_CNT4)
+  {
+    for (int j = 0; j < SALSA_CNT4; j++) TI[dst_off + j] = TT[src_off + j];
+  }
+}
 #endif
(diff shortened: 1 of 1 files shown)


npinto:
Curious if there interest in more performance improvements like these? I have a few cooking and wondering how I should prioritize. Thanks!

jsteube:
Good stuff, and I just came across. And of course, we totally want more of these performance improvements.Doing some test now.

jsteube:
In "inc_hash_scrypt.cl" the inverse takes twice the scratch the forward one does:

```
DECLSPEC void scrypt_shuffle_inv (PRIVATE_AS u32 *TI)
{
  u32 TT[STATE_CNT4];
```

The forward function manages with STATE_CNT4 / 2, because it saves only the blocks it is about to overwrite and slides the rest down in place. The inverse can work the same way: save the blocks that will land on odd positions, then walk m down from R-1 and write block 2m from block m. Descending order is what makes that safe, since the only write that lands on block m comes from block m / 2, and that one happens later.

At r=16 this is the difference between 1 KiB and 2 KiB of private memory per work item, on top of X and T. Private memory is what bounds how many work items a launch can hold on a memory hard algorithm, so where it is already tight the surplus costs hashes rather than just space. I would want that changed before this goes in.

Issue #4615, opened 2026-01-26
Hashcat does not crack anything when too many hashes are loaded


Shooter3k:
When too many hashes are loaded into Hashcat, Hashcat does not crack a single one despite correct plains being used. The number seems to be somewhere around 100,000,000 hashes

steps to reproduce: 
step1. 
download found list here https://hashmob.net/api/v2/hashlist/12996/found/100

step 2. 
remove found results so you just have hashes:
> cut -d':' -f1 12996.100.found > 12996.100.found.hashes

step 3. 
start up hashcat with large hash list using any wordlist that contains plains found in step #1

> hashcat.exe -m100  -a0 12996.100.found.hashes 
> hashmob.net.found.v7 -O  -D1 

step 4. 
watch no hashes get cracked: 

> hashcat (v7.1.2-79-g96659b5be+) starting
> Session..........: hashcat
> Status...........: Exhausted
> Hash.Mode........: 100 (SHA1)
> Hash.Target......: 12996.100.found.hashes
> Time.Started.....: Mon Jan 26 10:12:04 2026 (3 mins, 33 secs)
> Time.Estimated...: Mon Jan 26 10:15:37 2026 (0 secs)
> Kernel.Feature...: Optimized Kernel (password length 0-31 bytes)
> Guess.Base.......: File (hashmob.net.found.v7)
> Guess.Queue......: 1/1 (100.00%)
> Speed.#25........: 13364.9 kH/s (1.34ms) @ Accel:1024 Loops:1 Thr:1 Vec:8
> Recovered........: 0/757767271 (0.00%) Digests (total), 0/757767271 (0.00%) Digests (new)
> Remaining........: 757767271 (100.00%) Digests
> Recovered/Time...: CUR:0,N/A,N/A AVG:0.00,N/A,N/A (Min,Hour,Day)
> Progress.........: 3102354347/3102354347 (100.00%)
> Rejected.........: 27544758/3102354347 (0.89%)
> Restore.Point....: 3102354347/3102354347 (100.00%)
> Restore.Sub.#25..: Salt:0 Amplifier:0-1 Iteration:0-1
> Candidate.Engine.: Device Generator
> Candidates.#25...: 
> Hardware.Mon.#25.: N/A

pure kernel :

> Session..........: hashcat
> Status...........: Exhausted
> Hash.Mode........: 100 (SHA1)
> Hash.Target......: 12996.100.found.hashes
[...]

PR #4634, opened 2026-02-26
Fixes multiple overflow issues

matrix:
Fixes multiple overflow issues in:
- src/hashes.c and src/brain.c: see description on issue #4615
- find_hash(): with an int return type, any index above INT_MAX would wrap negative
- hashcat_status_t: values were copied from u32 to int, so for values > INT_MAX the status output became negative
- sort_by_salt(): now using the same comparison pattern as sort_by_digest_p0p1, which prevents corrupting sort order


Shooter3k:
I tested this and verified it works with more than 286,331,153 hashes now

Recovered........: 367362/506631643 (0.07%) Digests (total), 367362/506631643 (0.07%) Digests (new)

matrix:
will be added to the PCFG PR to avoid conflicts

PR #4641, opened 2026-03-09
Docker build: Add clang dependency

maede97:
This PR just adds the clang dependency in the Dockerfile for building with the Ubuntu 20 container.

This dependency is needed for the rust bindings, otherwise the `make` step fails with errors stating that `libclang.so` is missing.

Also related: #4603

### DIFF
--- a/docker/BinaryPackage.ubuntu20
+++ b/docker/BinaryPackage.ubuntu20
@@ -31,7 +31,8 @@ RUN apt-get update && \
     unzip \
     zstd \
     curl \
-    tar
+    tar \
+    clang
 
 WORKDIR /root
 


matrix:
Hi, I recommend using hashcat_binaries.ubuntu

Let me know if there are any problems

jsteube:
Thanks for this, and sorry for delay. You were right about the cause, and it is fixed on master, so I am closing it.

The container now installs libclang-dev instead of clang. This is the same failure you reported. The Rust bridges stop during make because bindgen cannot find libclang.so. So libclang-dev is the package that actually provides the library bindgen loads. clang is only the compiler driver and happens to pull the library in as a dependency on some releases, so it can make the build work without being the dependency we actually need. There is now a comment above the package entry naming the error, so it is clear why it must stay.

The fix went in as part of the packaging cleanup in 22ff5ca6a, five months after you reported it, rather than through this pull request. That delay is on me. Thanks for the report and for including the exact error you saw.

Issue #3887, opened 2023-10-05
Requesting support for hash mode to support FT PSK additional to 22000 mode


Rozin14:
In certain situations dumptools are forced only to capture handshakes with FT using PSK (802.11r). Theres no hash mode to convert these captures. So requesting to add hash mode supporting this. 

A txt file explaining this attached. (Explained by a known hash dev)
[message (1).txt](https://github.com/hashcat/hashcat/files/12821312/message.1.txt)

ZerBea:
The formulas to recover the PSK:
PBKDF2 (the same as on WPA1, WPA2 and WPA2 key version 3):
`PMK[256] = PBKDF2(PSK, SSID, SSID-length, 4096).`

get PMK-R0:
```
PMK-R0 = R0-Key-Data = KDF-256(XXKey, "FT-R0", SSIDlength || SSID || MDID || R0KHlength || R0KH-ID || S0KH-ID)
sha256_prf(PMK, PMK_len, "FT-R0", buf, pos - buf, r0_key_data, r0_key_data_len)
PMKID-R0 = HMAC-SHA1-128(PMK-R0, "PMK Name" | MAC_AP | MAC_STA)
PTK = PRF(PMK-R0, "pairwise key expansion", MAC1 || MAC2 || Nonce1 || Nonce2)

```
get PMK-R1:
```
PMK-R1 = KDF-256(PMK-R0, "FT-R1", R1KH-ID || S1KH-ID)
sha256_prf(pmk_r0, pmk_r0_len, "FT-R1", buf, pos - buf, pmk_r1, pmk_r0_len);
PMKID-R1 = HMAC-SHA1-128(PMK-R1, "PMK Name" | MAC_AP | MAC_STA) 
```

An example (containing all calculated session keys) 
```
ESSID: test-ft-psk (746573742d66742d70736b)
MAC_AP: 020000000000
MAC_CLIENT: 020000000100
PMKID: 3378f874c1930b599405d3de4b6e05cc
ANONCE: eb131d608a197829340c645c3bf30df2c0c8e818e9e31c560af630664a21a009
MIC: dd411684bd6e123bce3a606b2be06de2 (same calculation as of WPA2 key version 3)
[...]

PR #4645, opened 2026-03-11
Add WPA 802.11r Fast BSS Transition (FT-PSK) hash cracking module

c0rs:
This PR adds module with support for cracking WPA Fast BSS Transition (802.11r / FT-PSK) hashes — both the PMKID (type 3) and EAPOL-MIC (type 4) variants.

WPA-FT derives the PMK identically to classic WPA-PSK:
```
PMK = PBKDF2-HMAC-SHA1(passphrase, SSID, 4096, 32 bytes)
```
However, the subsequent key derivation chain is different from classic WPA-PSK and is defined by IEEE 802.11r using an HMAC-SHA256-based KDF:
- Type 3 (PMKID): PMK → PMK-R0-Name-salt (KDF-384 counter=2) → PMK-R0-Name (SHA-256) → PMKID (SHA-256)
- Type 4 (EAPOL): PMK → PMK-R0 (KDF-384 counter=1) → PMK-R1 (KDF-256) → PTK (KDF-384) → KCK; MIC = AES-128-CMAC(KCK, EAPOL)

References

1. Existing open issue: https://github.com/hashcat/hashcat/issues/3887
2. Reference Python implementation: https://github.com/DarkWolf-Labs/ft-crack/

Implementation notes

1. Separate module: A dedicated module was created rather than extending m22000. The primary reason is a hard architectural constraint: m22000 already occupies all four available aux kernel slots (aux1: EAPOL keyver=1, aux2: EAPOL keyver=2, aux3: EAPOL keyver=3, aux4: PMKID). The FT implementation requires two additional aux kernels (FT-PMKID and FT-EAPOL), which would exceed that limit. Beyond the slot constraint, the FT key derivation chain is structurally different enough (three HMAC-SHA256 stages vs. one PRF-512 call; additional struct fields for mdid/r0khid/r1khid/pke_r0/pke_r1) that merging would significantly increase the complexity of an already large module.
  
2. Explicit FT-specific fields: Although MDID, R0KH-ID, and R1KH-ID can technically be extracted from FT tags (Mobility Domain (0x36), Fast BSS Transition (0x37)) inside the EAPOL data payload, they are included as explicit fields in the hash format because:
- For PMKID (type 3), no EAPOL field is present.
- Parsing the FT tags inside the EAPOL data field would introduce significant complexity into the module. This parsing is better handled by an external capture tool, analogous to how hcxpcapngtool processes classic WPA captures.


ZerBea:
Great improvement.
The detection of FT-PSK is already implemented to hcxpcapngtool.
Now I'm going to add the conversion to format 37100 to hcxpcapngtool and the handling of 37100 to hcxhashtool and hcxpsktool.

ZerBea:
This will be the new hash line (mode 37100):
```
WPA*03*3378f874c1930b599405d3de4b6e05cc*020000000000*020000000100*746573742d66742d70736b***01*0102*77697265736861726b2d66742d70736b*020000000000
WPA*04*dd411684bd6e123bce3a606b2be06de2*020000000000*020000000100*746573742d66742d70736b*eb131d608a197829340c645c3bf30df2c0c8e818e9e31c560af630664a21a009*010300fa02010b00000000000000000001b44919486bacb36befb4ab90c14e639bc5027cca16d62a1525c2424e6ac7d2fe000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000009b30260100000fac040100000fac040100000fac04000001003378f874c1930b599405d3de4b6e05cc3603010201376c000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000106020000000000031077697265736861726b2d66742d70736b*00*0102*77697265736861726b2d66742d70736b*020000000000
```

Hints:
Conversion to ancient hccap format is not possible to limited binary hccap fomat.
Conversion to ancient hccapx format is not possible to limited binaryhccapx  fomat.
Conversion to john format is not possible, because john internally still uses ancient hccap format.

We should no longer use these outdated formats (hccap and hccapx).

ZerBea:
I discovered an issue in 37100 module (and 22000 module).
It looks like it can't handle oversized EAPOL MESSAGEs with  len > 255, e.g.:
```
802.1X Authentication
    Version: 802.1X-2001 (1)
    Type: Key (3)
    Length: 263
    Key Descriptor Type: EAPOL RSN Key (2)
    [Message number: 2]
    Key Information: 0x010b
        .... .... .... .011 = Key Descriptor Version: AES Cipher, AES-128-CMAC MIC (3)
        .... .... .... 1... = Key Type: Pairwise Key
        .... .... ..00 .... = Key Index: 0
        .... .... .0.. .... = Install: Not set
        .... .... 0... .... = Key ACK: Not set
        .... ...1 .... .... = Key MIC: Set
        .... ..0. .... .... = Secure: Not set
        .... .0.. .... .... = Error: Not set
        .... 0... .... .... = Request: Not set
        ...0 .... .... .... = Encrypted Key Data: Not set
        ..0. .... .... .... = SMK Message: Not set
    Key Length: 0
    Replay Counter: 1
    WPA Key Nonce: 1ff2efe31c81abd3ca51f9730da78f8b40dd4ebb529ad69029447c6625c1c3c1
    Key IV: 00000000000000000000000000000000
    WPA Key RSC: 0000000000000000
    WPA Key ID: 0000000000000000
    WPA Key MIC: fff0efb551dc9b6c95dc44641ff19b9e
    WPA Key Data Length: 168
    WPA Key Data […]: 30260100000fac040100000fac040100000fac040000010069ae0ebd04bc8f5a1e4f4587cb4a776536034a6400377900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
        Tag: RSN Information
        Tag: Mobility Domain
        Tag: Fast BSS Transition
```
This leads to a "Token length exception".
Unfortunately nearly all FT-PSK EAPOL MESSAGES contain more than 255 Bytes.
The length field inside an EAPOL frame is uint16_t. Due to ancient hccap and hccapx file format, hashcat limted this to uint8_t, which is too small for this kind of EAPOL MESSAGEs.

How to use hcxpcapngtool in combination with hashcat is [here](https://github.com/ZerBea/hcxdumptool/discussions/546).
All mentioned limits are now coded to hcxpapngtool. It it detects an oversized MESSAGE, it shows a warning:
```
EAPOL M2 messages (oversized)............: 1 (not supported by hashcat/JtR)
```
To avoid a Token Len exception warning by hashcat, EAPOL MESSAGEs that exceed the limits are not converted.

ZerBea:
By latest git head of hcxtools, I've reactivated conversion of FT-PSK PMKID.
Also I've added an EAPOL MESSAGE size limitation according to hashcat's limitation.
pro: no token length exceptions
con: EAPOL information greater than hashcat's limits are not converted

RealEnder:
Another issue with the PR is that it ignores already cracked hashes in the potfile, and if, for example, you try several attacks, the cracking process starts from the beginning.

ZerBea:
Also [nonce-error-corrections](https://hashcat.net/forum/showthread.php?tid=6361&pid=33920) is not working/implemented.
In case of poor quality dump files with heavy packet loss recorded by passive dump tools, this is a nice feature.
Please consider to fix/implement it.

c0rs:
@ZerBea , @RealEnder — thank you for the feedback.

First of all, I tested hcxpcapngtool with FT-PSK on various pcap files, and it works as expected for both PMKID and EAPOL. @ZerBea , thank you for adding this implementation so quickly! Module 37100 also successfully processes the resulting hashes (both WPA*03 and WPA*04) and recovers the password. So I assume the issues you mentioned earlier have been resolved (FYI, based on your output, the R1KH-ID length in the resulting WPA*03 hash was odd — only 11 characters without the final 0).

Regarding the limitations on the EAPOL size and the M2 message: I’m not entirely sure, but it looks like increasing the EAPOL size inside m37100 is not a problem. However, I would need a sample pcap where the M2 size exceeds 256 bytes for checking. In this example (https://github.com/ZerBea/hcxdumptool/discussions/546), only M3 exceeds 256 bytes.

As for the potfile: @RealEnder , thanks for pointing that out. I missed the `module_potfile_custom_check()` function, but I’ve now added it to the PR.

Regarding `nonce_error_corrections`: I need a bit more time to look into this. I did a quick check, and it seems that the logic in m37100 is similar to m22000. Could you provide a sample hash (pcap) where this option does not work as expected (or where the result differs compared to m22000)?

ZerBea:
Thanks for you reply.

It is simple to check the NC feature. Therefor just change the last digit of the ANONCE inside the hash line.
An example running hash mode 22000

```
WPA*02*cd233ba67d241df2b64c28b0602aa72b*3481c4e7991b*b0c090467cab*574c414e2d32*aacdb63cdc22a2615420abf4fc1a001bf0f3705a452607dc18e64b4d7abcb50d*0103007502010a00000000000000000001756b0c976942934fb44f69b5af714007c63a36252ffabc2da954bfb732b31886000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001630140100000fac040100000fac040100000fac020000*82
```

Run hashcat with deactivated NC == hash cracked
```
$ hashcat -m 22000 --session=test --nonce-error-corrections=0 nctest.22000 cracked.txt.gz
hashcat (v7.1.2-382-g2d71af371) starting - session [test]
...
cd233ba67d241df2b64c28b0602aa72b:3481c4e7991b:b0c090467cab:WLAN-2:ABCDEFGH
                                                          
Session..........: test
Status...........: Cracked
```

modify last xdigit of ANONCE from 7abcb50d to 7abcb50e and run hashcat again
Run hashcat with deactivated NC == hash not cracked
```
$ hashcat -m 22000 --session=test --nonce-error-corrections=0 nctest.22000 cracked.txt.gz
hashcat (v7.1.2-382-g2d71af371) starting - session [test]
...
Session..........: test                                   
Status...........: Exhausted
```

Activate NC == hash cracked
```
$ hashcat -m 22000 --session=test --nonce-error-corrections=8 nctest.22000 cracked.txt.gz
hashcat (v7.1.2-382-g2d71af371) starting - session [test]
...
cd233ba67d241df2b64c28b0602aa72b:3481c4e7991b:b0c090467cab:WLAN-2:ABCDEFGH
                                                          
Session..........: test
Status...........: Cracked
```

The same test environment should work running 37100, too.

ZerBea:
This is an example of an oversized FT-PSK M2 frame:
```
802.1X Authentication
    Version: 802.1X-2001 (1)
    Type: Key (3)
    Length: 263
    Key Descriptor Type: EAPOL RSN Key (2)
    [Message number: 2]
    Key Information: 0x010b
    Key Length: 0
    Replay Counter: 1
    WPA Key Nonce: 1ff2efe31c81abd3ca51f9730da78f8b40dd4ebb529ad69029447c6625c1c3c1
    Key IV: 00000000000000000000000000000000
    WPA Key RSC: 0000000000000000
    WPA Key ID: 0000000000000000
    WPA Key MIC: fff0efb551dc9b6c95dc44641ff19b9e
    WPA Key Data Length: 168
    WPA Key Data […]: 30260100000fac040100000fac040100000fac040000010069ae0ebd04bc8f5a1e4f4587cb4a776536034a6400377900000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
        Tag: RSN Information
            Tag Number: RSN Information (48)
            Tag length: 38
            RSN Version: 1
            Group Cipher Suite: 00:0f:ac (Ieee 802.11) AES (CCM)
            Pairwise Cipher Suite Count: 1
            Pairwise Cipher Suite List 00:0f:ac (Ieee 802.11) AES (CCM)
            Auth Key Management (AKM) Suite Count: 1
            Auth Key Management (AKM) List 00:0f:ac (Ieee 802.11) FT using PSK
            RSN Capabilities: 0x0000
            PMKID Count: 1
            PMKID List
        Tag: Mobility Domain
            Tag Number: Mobility Domain (54)
            Tag length: 3
            Mobility Domain Identifier: 0x644a
            FT Capability and Policy: 0x00
        Tag: Fast BSS Transition
            Tag Number: Fast BSS Transition (55)
            Tag length: 121
            MIC Control: 0x0000
            MIC: 00000000000000000000000000000000
            ANonce: 0000000000000000000000000000000000000000000000000000000000000000
            SNonce: 0000000000000000000000000000000000000000000000000000000000000000
            Subelement: PMK-R1 key holder identifier (R1KH-ID)
            Subelement: PMK-R0 key holder identifier (R0KH-ID)
```

hcxpcapngtool convert the PMKID, but it does not use this EAPOL M2 MESSAGE to calculate a MESSAGE PAIR:
```
EAPOL M2 messages (oversized)............: 1 (not supported by hashcat/JtR)
```

Please note:
Even oversized M2 frames contain a PMKID we can use.
So there is no real need to calculate MESSAGE PAIRs from this oversized frames, but it will be a nice ad-on.
Since I'm running the same limitations like hashcat (22000, 22001 and 37100), Token length exceptions are gone.

ZerBea:
```
$ hcxpcapngtool -f xx complete.pcapng
hcxpcapngtool 7.1.2-39-g2bd2d84 reading from complete.pcapng...
maximum of supported interfaces reached: 255

summary capture file
--------------------
file name................................: complete.pcapng
version (pcapng).........................: 1.0
operating system.........................: linux
application..............................: gopacket
interface name...........................: intf0
interface vendor.........................: 000000
openSSL version..........................: 1.0
weak candidate...........................: N/A
MAC ACCESS POINT.........................: 000000000000 (incremented on every new client)
MAC CLIENT...............................: 000000000000
REPLAYCOUNT..............................: 0
ANONCE...................................: 0000000000000000000000000000000000000000000000000000000000000000
SNONCE...................................: 0000000000000000000000000000000000000000000000000000000000000000
timestamp minimum (timestamp)............: 31.01.2026 03:36:20 (1769830580)
timestamp maximum (timestamp)............: 31.01.2026 03:50:06 (1769831406)
duration (rounded).......................: 13m:46s
used capture interfaces..................: 255
link layer header type...................: DLT_IEEE802_11_RADIO (127)
endianness (capture system)..............: little endian
packets inside...........................: 3
frames with FCS (radiotap)...............: 3
frames with correct FCS (crc)............: 3
packets received on 2.4 GHz..............: 3
ESSID (total unique).....................: 1
BEACON (total)...........................: 1
BEACON on 2.4 GHz channel (from IE_TAG)..: 2 
EAPOL messages (total)...................: 2
EAPOL RSN messages.......................: 2
EAPOL ANONCE error corrections (NC)......: not detected
EAPOL M2 messages (total)................: 1
EAPOL M2 messages (oversized)............: 1 (not supported by hashcat/JtR)
EAPOL M2 messages (FT using PSK).........: 1 (not supported by hashcat/JtR)
EAPOL M3 messages (total)................: 1
EAPOL M3 messages (oversized)............: 1 (not supported by hashcat/JtR)
RSN PMKID (total)........................: 1
RSN PMKID FT-PSK ........................: 1
RSN PMKID (best).........................: 1
RSN PMKID written to 37100 hash file.....: 1
packet read error........................: 309

frequency statistics from radiotap header (frequency: received packets)
-----------------------------------------------------------------------
 2417: 3	
```

The packet errors are related to an outdated version of bettercap gopacket.
gopacket reports faulty interface counts. Don't worry about this errors.

Please let me know when you have downloaded the zip file so that I can remove it.

c0rs:
Thanks, downloaded

ZerBea:
Due to the limit to 255 bytes, we hcxpcapngtool only converts the following MESSAGE PAIRS:
```
M1M2E2  if M2 is not oversized
M2M3E2 if M2 and M3 is not oversized
M2M3E3 if M2 and M3 is not oversized
M3M4E3 if M3 and M4 is not oversized and anonce M4 is not zeroed
M3M4E4 if M3 and M4 is not oversized and anonce M4 is not zeroed
M1M4E4 if M4 is not oversized and anonce M4 is not zeroed
```
If you need to expand the limit in hcxpcapngtool, this is the line:
https://github.com/ZerBea/hcxtools/blob/2bd2d8453dfded5eeab5adeb4b7fcc0868b2d5bb/include/hcxpcapngtool.h#L54
Default is 255 bytes, but you can increase it up to uint16_t for testing purpose.

RealEnder:
Hi @c0rs, thanks for your work!
Pulled the last commit, but unfortunately it still doesn't filter the cracked hashes. Did a clean hashcat rebuild.

ZerBea:
By latest commit hcxpcspngtool EAPOL AUTH LIMIT for all ESPOL MESSSGES (m1, m2, m3, m4) is:
internal 1024 bytes
22000 255 bytes
37100 1024 bytes
john 255 bytes
hccapx 255 bytes
hccap 255 bytes

c0rs:
@ZerBea ,

Thanks for the detailed info on nonce correction. I’ve fixed the issues in m37100, and it should be working properly now.

Also, it turns out that the nonce correction logic in the m22000_aux3 function of m22000 (keyver=3) is probably implemented incorrectly as well. A tentative fix looks like this. If you have a WPA-PSK handshake with keyver=3, you can share it with me to verify the fix.

```
@@ -884,24 +884,13 @@ KERNEL_FQ KERNEL_FA void m22000_aux3 (KERN_ATTR_TMPS_ESALT (wpa_pbkdf2_tmp_t, wp

   u32 to;

-  u32 m0;
-  u32 m1;
-
   if (wpa->nonce_compare < 0)
   {
-    m0 = pke[15] & ~0x000000ff;
-    m1 = pke[16] & ~0xffffff00;
-
-    to = pke[15] << 24
-       | pke[16] >>  8;
+    to = pke[16];
   }
   else
   {
-    m0 = pke[23] & ~0x000000ff;
-    m1 = pke[24] & ~0xffffff00;
-
-    to = pke[23] << 24
-       | pke[24] >>  8;
+    to = pke[24];
   }

   u32 bo_loops = wpa->detected_le + wpa->detected_be;
@@ -953,13 +942,11 @@ KERNEL_FQ KERNEL_FA void m22000_aux3 (KERN_ATTR_TMPS_ESALT (wpa_pbkdf2_tmp_t, wp

       if (wpa->nonce_compare < 0)
       {
-        pke[15] = m0 | (t >> 24);
-        pke[16] = m1 | (t <<  8);
+        pke[16] = t;
       }
       else
       {
-        pke[23] = m0 | (t >> 24);
-        pke[24] = m1 | (t <<  8);
+        pke[24] = t;
       }
```

c0rs:
@ZerBea , I also increased the EAPOL buffer size to 512 bytes, so now it’s possible to brute-force such hashes in FT-PSK cases as well. As an example, I used a hash from your pcapng file.

ZerBea:
@c0rs great, thanks for the work.

hcxpcangtool now uses 512 bytes AUTHLEN too:
https://github.com/ZerBea/hcxtools/commit/a4045f368931d02a57519c8ce97b6051391db12a

ZerBea:
Please let me know if you downloaded it. Than I can delete it here.

c0rs:
downloaded

c0rs:
Indeed, nonce correction is implemented incorrectly for keyver=3 in m22000 (function `m22000_aux3()`). Fix above solves the problem.

c0rs:
@RealEnder,

Could you please clarify what exactly you’re doing and what result you’re expecting? Below is an example of how it works on my end. The recovered hashes are saved to / loaded from the hashcat.potfile.

```
~/hashcat % rm  hashcat.potfile
~/hashcat % echo -n '12345678' | ./hashcat -a 0 -m 37100 hash_example.txt --potfile-disable --quiet
dd411684bd6e123bce3a606b2be06de2:020000000000:020000000100:test-ft-psk:12345678
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 37100 (FT-PSK (802.11r Fast BSS Transition))
Hash.Target......: dd411684bd6e123bce3a606b2be06de2:020000000000:02000...ft-psk
Time.Started.....: Mon Apr 13 20:12:42 2026 (0 secs)
Time.Estimated...: Mon Apr 13 20:12:42 2026 (0 secs)
Kernel.Feature...: Pure Kernel (password length 8-63 bytes)
Guess.Base.......: Pipe
Speed.#02........:       19 H/s (0.21ms) @ Accel:64 Loops:1024 Thr:256 Vec:1
Progress.........: 1
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#02..: Salt:0 Amplifier:0-1 Iteration:0-1
Candidates.#02...: 12345678 -> 12345678
Hardware.Mon.SMC.: Fan0: 0%, Fan1: 0%
Hardware.Mon.#02.: Util: 99% Pwr:9770mW
~/hashcat % echo -n '12345678' | ./hashcat -a 0 -m 37100 hash_example.txt --quiet
dd411684bd6e123bce3a606b2be06de2:020000000000:020000000100:test-ft-psk:12345678
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 37100 (FT-PSK (802.11r Fast BSS Transition))
Hash.Target......: dd411684bd6e123bce3a606b2be06de2:020000000000:02000...ft-psk
Time.Started.....: Mon Apr 13 20:13:01 2026 (0 secs)
Time.Estimated...: Mon Apr 13 20:13:01 2026 (0 secs)
Kernel.Feature...: Pure Kernel (password length 8-63 bytes)
Guess.Base.......: Pipe
Speed.#02........:       20 H/s (0.21ms) @ Accel:64 Loops:1024 Thr:256 Vec:1
Progress.........: 1
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#02..: Salt:0 Amplifier:0-1 Iteration:0-1
Candidates.#02...: 12345678 -> 12345678
Hardware.Mon.SMC.: Fan0: 0%, Fan1: 0%
Hardware.Mon.#02.: Util: 99% Pwr:9790mW
~/hashcat % echo -n '12345678' | ./hashcat -a 0 -m 37100 hash_example.txt --quiet
~/hashcat % echo -n '12345678' | ./hashcat -a 0 -m 37100 hash_example.txt --quiet --show
dd411684bd6e123bce3a606b2be06de2:020000000000:020000000100:test-ft-psk:12345678
~/hashcat % echo -n '12345678' | ./hashcat -a 0 -m 37100 hash_example.txt --potfile-disable --quiet
dd411684bd6e123bce3a606b2be06de2:020000000000:020000000100:test-ft-psk:12345678
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 37100 (FT-PSK (802.11r Fast BSS Transition))
Hash.Target......: dd411684bd6e123bce3a606b2be06de2:020000000000:02000...ft-psk
Time.Started.....: Mon Apr 13 20:13:32 2026 (0 secs)
Time.Estimated...: Mon Apr 13 20:13:32 2026 (0 secs)
Kernel.Feature...: Pure Kernel (password length 8-63 bytes)
Guess.Base.......: Pipe
Speed.#02........:       17 H/s (0.21ms) @ Accel:64 Loops:1024 Thr:256 Vec:1
Progress.........: 1
Rejected.........: 0
Restore.Point....: 0
Restore.Sub.#02..: Salt:0 Amplifier:0-1 Iteration:0-1
Candidates.#02...: 12345678 -> 12345678
Hardware.Mon.SMC.: Fan0: 0%, Fan1: 0%
Hardware.Mon.#02.: Util: 99% Pwr:9762mW
~/hashcat % cat hashcat.potfile
f91fea0712af6e92192a51f92acc483e8184f528220fc02308b4102cf79373b2*746573742d66742d70736b:12345678
```

RealEnder:
@c0rs, noticed that when running a new attack, not all cracked hashes from the potfile were removed. Here is a minimal example:
Hash: `WPA*03*c2625d495598b06749480f4251a55882*50ff20595f1e*7c2eddb6f176*736f63683139***20*696f*4b65656e657469633a35303a66663a32303a35393a35663a31662d3030*50ff20595f1e`
cracks with the module successfully.
potfile entry:
`2b197bcbf934d0b96b4a87247d268fa6370539b93e52a879c615dbbba5858114*736f63683139:12345678`
Rerunning the attack doesn't identify and remove the hash and tries to crack it again.

@ZerBea, hunted after Salt-value exception for some hashlines. The offender was the keyver check in the module:
```
    // keyver

    const u16 key_information = byte_swap_16 (auth_packet->key_information);

    wpa->keyver = key_information & 3;

    if (wpa->keyver != 3) return (PARSER_SALT_VALUE); // support only keyver = 3
```
The hashes were with keyver = 2. Maybe we can skip converting those in hcxpcapngtool. I don't know how feasible it is to implement support for keyver 2 in the module, but PMKID for FT should be enough.

ZerBea:
I suspect a PMKID is mandatory. If that is the case, we can remove MESSAGEPAIR conversion completely.

ZerBea:
If the the option --remove is used, the newly saved hash list is destroyed. Every hash line is truncated in the ESSID field.

c0rs:
@RealEnder, fixed. Pls, check.

c0rs:
@ZerBea ,
About `--remove`. Looks like it's a general problem that is not specifically related to module 37100.

ZerBea:
Remove in combination with 22000 and 22001 is working as expected. Remove in combination with 37100 destroys the hash file.

c0rs:
fixed. pls check

ZerBea:
Remove is now working as expected. Thanks.

RealEnder:
Skipping cracked hashes in the beginning of the attack also works as expected.

jsteube:
> The primary reason is a hard architectural constraint: m22000 already occupies all four available aux kernel slots (aux1: EAPOL keyver=1, aux2: EAPOL keyver=2, aux3: EAPOL keyver=3, aux4: PMKID). The FT implementation requires two additional aux kernels (FT-PMKID and FT-EAPOL), which would exceed that limit.

You were right back then when you wrote it, but that has changed since: a fifth slot went in as f21753126, and FT-PSK is what it was added for. There will not be a sixth, because all 64 opts_type bits are in use and aux5 only exists by reclaiming the bit from the retired OPTS_TYPE_PT_ADD02.

Could you integrate this into 22000 instead, with FT-PMKID and FT-EAPOL sharing aux5 and branching on the type field you already added? In module_deep_comp_kernel() types 1 and 2 pick between AUX4 and AUX1 to AUX3 the same way. Mode 22001 needs the same, since it has those four aux kernels and that same wpa_t, and differs only in taking the PMK rather than deriving it.

That is a large rework to ask for, and what decides it is that the init, loop and comp in this branch are m22000's already, on the same wpa_pbkdf2_tmp_t and the same wpa_t. So the only thing a separate mode adds is a second PBKDF2-HMAC-SHA1 over 4096 iterations, and a capture holding a classic handshake beside an FT one derives the same PMK twice.

One thing the integration has to settle is the size of that shared wpa_t. It is the esalt both modes carry once per hash, and the FT fields here take it from 700 bytes to 1380, so every classic WPA hash in a capture would grow with it. A union would avoid most of that, because a hash of type 3 or 4 never carries the PMKID and EAPOL fields that types 1 and 2 use. The wider eapol buffer is a separate question, since that one came out of the oversized M2 frame ZerBea reported and 22000 may need it as well.

jsteube:
> Also, it turns out that the nonce correction logic in the m22000_aux3 function of m22000 (keyver=3) is probably implemented incorrectly as well. A tentative fix looks like this.

You were right. The two PKE layouts differ by one byte, and that is what moves the counter:

```
  keyver 1/2   "Pairwise key expansion\x00" at +0    nonces at +35 and +67
  keyver 3     "Pairwise key expansion"     at +2    nonces at +36 and +68
```

The counter is the last 4 bytes of the ANonce. Under keyver 1 and 2 that lands on bytes 63 to 66, straddling pke[15] and pke[16], which is the read m22000_aux1 does and the one aux3 inherited. Under keyver 3 it lands on bytes 64 to 67, which is pke[16] on its own, so the straddling read and the write back put the corrected counter in the wrong bits. Your to = pke[16] matches that layout.

Could you open it as a pull request of its own? It is a fault in master rather than anything to do with FT-PSK, and this branch is about to change shape, so it would be lost in the rework. The offsets above stand on their own, so it does not have to wait for the keyver=3 handshake you asked for in April, and that capture is still worth having on the new pull request to close it end to end.

PR #4646, opened 2026-03-12
Implement hc_lockfile and hc_unlockfile on Windows

### LINKED ISSUE #3941: Lock/unlock files properly for Windows
Implementing a proper fix will allow better feature parity between Linux and Windows and may reduce future bugs. This code is 8 years old and isn't causing any known bugs but would be good practice to fix.
See comments: (fcntl is Linux-only)
https://github.com/hashcat/hashcat/blob/c1a10518fd98535e3ef573d9cd5f8bab292c4609/src/locking.c#L49-L61

matrix:
`hc_lockfile` and `hc_unlockfile` do nothing on Windows. They return 0, so a caller that takes the lock holds nothing, and two hashcat processes appending to one potfile can interleave their lines.

## Reproduction

```
$ sed -n '/^#else/,/^}/p' src/locking.c

#else

int hc_lockfile (MAYBE_UNUSED HCFILE *fp)
{
  // we should put windows specific code here

  return 0;
}
```

The comment has been there since the function was written. This is the platform half that was left for later, and it is #3941.

## What it does now

`LockFileEx` and `UnlockFileEx` on the handle behind the descriptor, which is what the POSIX branch does with `fcntl` and `F_SETLKW`. `include/locking.h` takes `io.h` for `_get_osfhandle` on Windows and keeps `unistd.h` and `fcntl.h` everywhere else.

## Verification

This code compiles only for Windows, so it is checked by cross compiling on Linux:

```
$ make win

before: rc=0, 0 warnings, 0 errors
after:  rc=0, 0 warnings, 0 errors
```

A clean build does not say the calls are in the binary. The import table and the object file do:

```
$ x86_64-w64-mingw32-objdump -p hashcat.dll | grep -iE 'KERNEL32|LockFileEx'
        DLL Name: KERNEL32.dll
        1bf0f2   1013  LockFileEx
        1bf268   1458  UnlockFileEx

$ x86_64-w64-mingw32-nm obj/locking.WIN.o | grep __imp
         U __imp__get_osfhandle
         U __imp_LockFileEx
         U __imp_UnlockFileEx
```

I have no Windows machine, so the lock has not been observed to hold against a second process. What is shown above is that the code is compiled, linked and imported from kernel32.

## What this leaves out

An earlier version of this branch also added a check at nine call sites, across six files. That is a separate problem and it is not this platform's: those same nine of the seventeen places that take or release this lock discard the value on POSIX today as well, where the function has always been able to fail. Eight sites have been given checks already, in two different shapes, and what a failure should do is not settled between them.

So the callers are untouched here and the situation is written up in #4855 instead, which lists all seventeen sites and which of them check.

## tools/test_edge.sh

No hash mode is touched. The suite has no case that runs two processes against one file, so there is nothing in it that reaches this path.

### DIFF
--- a/include/locking.h
+++ b/include/locking.h
@@ -9,9 +9,14 @@
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
-#include <unistd.h>
 #include <errno.h>
+
+#if defined (_WIN)
+#include <io.h>
+#else
+#include <unistd.h>
 #include <fcntl.h>
+#endif
 
 int hc_lockfile   (HCFILE *fp);
 int hc_unlockfile (HCFILE *fp);
--- a/src/debugfile.c
+++ b/src/debugfile.c
@@ -69,16 +69,16 @@ void debugfile_write_append (hashcat_ctx_t *hashcat_ctx, const u8 *rule_buf, con
 
   const u32 debug_mode = debugfile_ctx->mode;
 
+  if (hc_lockfile (&debugfile_ctx->fp) == -1)
+  {
+    event_log_warning (hashcat_ctx, "%s: Failed to lock file.", debugfile_ctx->filename);
+  }
+
   if ((debug_mode == 2) || (debug_mode == 3) || (debug_mode == 4) || (debug_mode == 5))
   {
     debugfile_format_plain (hashcat_ctx, orig_plain_ptr, orig_plain_len);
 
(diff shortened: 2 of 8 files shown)


PenguinKeeper7:
(See: https://github.com/hashcat/hashcat/issues/3941#issuecomment-4190089900 for context)
I tested this PR and it does seem to have tentatively fixed potfiles, but `-o`utfiles are still capable of corruption when multiple simultaneously active Hashcat sessions attempt to access the same file.

(All plains included in this image are generated from rockyou and are therefore not sensitive)
image

PR #4648, opened 2026-03-12
Add --rules-concat to use the -r files as one list

matrix:
Several `-r` files are chained: the rules of one file are applied to the output of the next, so two files of 3 and 2 rules give 6 candidates per word. There is no way to ask for the rules of every file one after another.

## Reproduction

```
$ printf ':\nu\nl\n' > a.rule
$ printf '$1\n$2\n'  > b.rule
$ echo Pass > w.dict

$ ./hashcat --stdout -a 0 -r a.rule -r b.rule w.dict
Pass1
PASS1
pass1
Pass2
PASS2
pass2
        6 candidates, which is 3 * 2

$ ./hashcat --stdout -a 0 --rules-concat -r a.rule -r b.rule w.dict

before: hashcat: unrecognized option `--rules-concat'
        Invalid argument specified.
        rc=255
after:  Pass
        PASS
        pass
        Pass1
        Pass2
        5 candidates, which is 3 + 2
        rc=0
```

## How it works

`kernel_rules_load ()` reads every file into `all_kernel_rules_buf` and then combines them. The count of a chain is the product of the counts, and the count of a list is their sum, so the concatenating case is one allocation and one `memcpy` per file:

```c
if (user_options->rp_files_concat == true)
{
  u32 kernel_rules_cnt = 0;

  for (u32 i = 0; i < user_options->rp_files_cnt; i++)
  {
    if (overflow_check_u32_add (kernel_rules_cnt, all_kernel_rules_cnt[i]) == true)
    ...
```

It leaves early, which is what the single file case immediately above it already does, and for the same reason given in the comment there: there is nothing to combine, and going on would copy the array a second time.

`overflow_check_u32_add ()` is the same guard the chaining path uses in its multiplying form.

## What is refused

`--rules-concat` without `-r` has nothing to concatenate, so it is refused. There is no separate check against `-g`, because `-r` with `-g` is already refused and without `-r` the first check fires:

```
$ ./hashcat --stdout -a 0 --rules-concat w.dict
Parameter --rules-concat requires -r/--rules-file.
rc=255

$ ./hashcat --stdout -a 0 --rules-concat -r a.rule -g 10 w.dict
Combining -r/--rules-file and -g/--rules-generate is not supported.
rc=255
```

## tools/test_edge.sh

No hash mode is touched. The change is in rule loading, which every mode that takes `-r` shares, and the runs above exercise both paths of it.

### DIFF
--- a/include/types.h
+++ b/include/types.h
@@ -927,6 +927,7 @@ typedef enum user_options_map
   IDX_RP_GEN_FUNC_SEL           = 0xff41,
   IDX_RP_GEN                    = 'g',
   IDX_RP_GEN_SEED               = 0xff42,
+  IDX_RP_FILE_CONCAT            = 0xff88,
   IDX_RULE_BUF_L                = 'j',
   IDX_RULE_BUF_R                = 'k',
   IDX_RUNTIME                   = 0xff43,
@@ -2527,6 +2528,7 @@ typedef struct user_options
   bool         veracrypt_pim_stop_chgd;
   bool         version;
   bool         wordlist_autohex;
+  bool         rp_files_concat;
   #ifdef WITH_BRAIN
   char        *brain_host;
   char        *brain_password;
--- a/include/types.h
+++ b/include/types.h
@@ -1036,6 +1036,7 @@ typedef enum user_options_map
   IDX_RESTORE_FILE_PATH         = 0xff3e,
   IDX_RESTORE_POSITION          = 0xff87,
   IDX_RP_FILE                   = 'r',
+  IDX_RP_FILE_CONCAT            = 0xff8c,
   IDX_RP_GEN_FUNC_MAX           = 0xff3f,
   IDX_RP_GEN_FUNC_MIN           = 0xff40,
   IDX_RP_GEN_FUNC_SEL           = 0xff41,
@@ -2898,6 +2899,7 @@ typedef struct user_options
   bool         restore;
   bool         restore_enable;
   bool         restore_position;
+  bool         rp_files_concat;
   bool         self_test;
   bool         show;
   bool         slow_candidates;
--- a/src/rp.c
+++ b/src/rp.c
@@ -1022,98 +1022,147 @@ int kernel_rules_load (hashcat_ctx_t *hashcat_ctx, kernel_rule_t **out_buf, u32 
   hcfree (rule_buf);
 
   /**
-   * merge rules
+   * merge or concat rules
    */
 
-  u32 kernel_rules_cnt = 1;
+  u32 kernel_rules_cnt = 0;
 
-  u32 *repeats = (u32 *) hccalloc (user_options->rp_files_cnt + 1, sizeof (u32));
+  kernel_rule_t *kernel_rules_buf = NULL;
 
-  repeats[0] = kernel_rules_cnt;
-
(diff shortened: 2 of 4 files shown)


roycewilliams:
Am I reading right that this just concatenates them all into a single set of rules and then runs them? If so, won't that hit memory thresholds? My goal was to make it easy to run them one at a time, not to run them all at once.

matrix:
> Am I reading right that this just concatenates them all into a single set of rules and then runs them? If so, won't that hit memory thresholds? My goal was to make it easy to run them one at a time, not to run them all at once.

Have you seen what happens without this patch when two rules are loaded? The question you ask me about “memory thresholds” doesn't make sense.

PenguinKeeper7:
Forgive me if I'm wrong, Tycho but I believe the feedback was that the original issue wants each rule to be run sequentially instead of merged into one and effectively `cat`-ing them together, in a similar way that wordlists work when you do `Wordlists/*`. It doesn't just merge them all into a large blob, it makes a `Guess.Base` queue and runs each wordlist as its own attack.

What Tycho's probably referring to about “memory thresholds” is that with your solution of just merging them all together, VRAM limitations could easily occur with potentially tens or hundreds of millions of rule lines all being run at once. With the sequential approach, you'd only need to load in a single individual ruleset's contents at any time, cutting down the memory requirements drastically

matrix:
> Forgive me if I'm wrong, Tycho but I believe the feedback was that the original issue wants each rule to be run sequentially instead of merged into one and effectively `cat`-ing them together, in a similar way that wordlists work when you do `Wordlists/*`. It doesn't just merge them all into a large blob, it makes a `Guess.Base` queue and runs each wordlist as its own attack.
> 
> What Tycho's probably referring to about “memory thresholds” is that with your solution of just merging them all together, VRAM limitations could easily occur with potentially tens or hundreds of millions of rule lines all being run at once. With the sequential approach, you'd only need to load in a single individual ruleset's contents at any time, cutting down the memory requirements drastically

I fully understand both points, but it’s much easier to run an external script than to implement what you’re proposing to do in C. As a result, I simply modified the existing logic in Hashcat to change the way multiple rules are processed simultaneously, and this approach definitely saves memory compared to the original version.

roycewilliams:
I'm happy if it's an improvement, but it doesn't address my original request, hinging on the word "serial", meaning "one at a time". 

(I definitely understand that it can be worked around with a script. The point of my request was to make it easier for users to not have to write their own script, and that it is a common enough workflow that it would be reasonable to bring into hashcat itself, just like we do by walking through an entire directory of wordlists. Even though that could be scripted, it's more efficient to be in hashcat itself)

If that means that we need to remove the assertion that this PR addresses my issue, and then just close my issue as a WONTFIX ... that won't make me mad, just disappointed. 😉

matrix:
> I'm happy if it's an improvement, but it doesn't address my original request, hinging on the word "serial", meaning "one at a time".
> 
> (I definitely understand that it can be worked around with a script. The point of my request was to make it easier for users to not have to write their own script, and that it is a common enough workflow that it would be reasonable to bring into hashcat itself, just like we do by walking through an entire directory of wordlists. Even though that could be scripted, it's more efficient to be in hashcat itself)
> 
> If that means that we need to remove the assertion that this PR addresses my issue, and then just close my issue as a WONTFIX ... that won't make me mad, just disappointed. 😉

done

PR #4651, opened 2026-03-15
Add -M to test.sh, and fix five faults in the same file

matrix:
`-M` runs 24 hash modes chosen one per distinct code path, which is the list the first commit here put in. Running a wider set over it turned up three faults in this file, and those are what this branch now carries beyond `-M` itself. All three are reproduced below.

## `-M`

The list is the one the first commit had, and review settled that it stays: the benchmark array in `src/benchmark.c` answers which modes a user wants measured, while `-M` answers the other end, one mode per distinct code path for a short run before a full sweep. Reading the array instead cost six of those paths, `14511` and `14600` outright, and `20510`, `14900`, `15400` and `33500` each reached by their own number.

```
$ grep -cE '^[[:space:]]*(if|elif) .*20510|20510\) skip_type' tools/test.sh
12
$ grep -cE '^[[:space:]]*(if|elif) .*20510' tools/test_edge.sh
4
```

One thing does change. Six of the 24 are the first entry of a family list defined near the top of the same file, so they are read from there rather than typed twice:

```
-  MINIMAL_MODES="0 100 110 400 500 2600 3000 3200 6211 11600 12500 13711 14200 14511 14600 14900 15400 15700 20510 22000 29511 33000 33500 34100"
+  MINIMAL_MODES="0 100 110 400 500 2600 3000 3200 ${TC_MODES%% *} 11600 12500 ${VC_MODES%% *} 14200 ${CL_MODES%% *} ${LUKS1_LEGACY_MODE} 14900 15400 15700 20510 22000 ${LUKS1_MODES%% *} 33000 33500 ${LUKS2_MODES}"
```

The two give the same 24 modes, and the second follows the family when the family changes.

End to end, the same command over both forms of the list:

```
$ ./tools/test.sh -M -t single -a 0 -d 1

list                          modes   Error   time
the 24 typed out               24      1      1238s
six of them read back          24      1      1244s
```

The same 24 hash modes are initialised on either side. The one error is mode 14200 both times, 4 of 8 not found on one run and 3 of 8 on the other, which is the mode and not the list: `tools/test.pl` generates fresh passwords per run, so how much of it goes wrong moves and the fact that it does stays.

```
[ Type 14200, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > Error : 4/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
```

Nothing else in this file reaches a `-t single -a 0` run: the other two changes are inside `attack_3 ()`, and the four skips are for hash modes the minimal list does not name.

The assignment also stops needing the line the packaging mode carried for it. A package has no `src/`, which is why reading `src/benchmark.c` from a packaged run needed that line, and the family lists this one reads instead are literals in the same file.

## The four hash modes skipped for Apple

```
$ ./tools/test.sh -m 600 -t all -a 0 -d 1
```

Before the change that command prints nothing at all, not even the `Init test` line, because the mode is never initialised. After:

```
[ test_1789257260 ] > Init test for hash type 600.
[ test_1789257260 ] [ Type 600, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1789257260 ] [ Type 600, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1789257260 ] [ Type 600, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1789257260 ] [ Type 600, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

The same run on all four modes:

```
  mode    version   OK   Error   Skip   what the run printed
  600     before    0    0       0      nothing
  600     after     4    0       0      4 verdicts
  1800    before    0    0       0      nothing
  1800    after     4    0       0      4 verdicts
  6900    before    0    0       0      nothing
  6900    after     4    0       0      4 verdicts
  18600   before    0    0       0      nothing
  18600   after     4    0       0      4 verdicts
```

Each of the four skips carries a comment naming a cause, and all four causes are gone. 600 and 6900 blame `Digest::BLAKE2` and `Digest::GOST` on Apple Silicon, and `tools/install_modules.sh` installs the two patched forks:

```
$ grep -n 'p5-Digest-BLAKE2\|digest-gost' tools/install_modules.sh
177:install_module "https://github.com/matrix/p5-Digest-BLAKE2.git"
179:install_module "https://github.com/matrix/digest-gost.git"
```

18600 blames `Crypt::GCrypt`, which `1cc8ed779` took out of the suite on 2025-07-12, fourteen months ago. Nothing under `tools/` mentions it any more, and `tools/test_modules/m18600.pm` uses `Crypt::Blowfish` and `Crypt::Mode::CFB`.

```
$ git log -1 --format='%h %ci %s' 1cc8ed779
1cc8ed779 2025-07-12 20:00:43 +0200 get rid of Crypt::GCrypt
$ git grep -c GCrypt -- tools/
$ echo $?
1
```

No other platform is affected either way. Each of the four skips was guarded by the platform before it did anything, so on anything but a Mac the four blocks were already dead code:

```
-      if [[ "$OSTYPE" == "darwin"* ]]; then
-      if [ "${IS_APPLE_SILICON}" -eq 1 ]; then
-      if [ "${IS_APPLE_SILICON}" -eq 1 ]; then
-      if [[ "$OSTYPE" == "darwin"* ]]; then
```

1800's comment is the fourth, and says only "test.pl produce wrong hashes with Apple, would be necessary to investigate to understand why". The runs above are that investigation: the oracle's hashes are the ones hashcat cracks, on this machine, in every attack the suite covers.

## The multi byte test, and the mask it selects

These two are one change, because fixing the first is what exposes the second. Three versions of this file, differing only in the mask region, over `-m 600` five times each:

```
$ ./tools/test.sh -m 600 -t multi -a 3 -d 1 -O

version   run   OK  Error  Warning  branch  mask lines  ?d on one line  hashes  time
head       1    0   2      0        ?d      -           -               2       12s
head       2    0   2      0        ?d      -           -               3       12s
head       3    2   0      0        ?d      -           -               7        7s
head       4    2   0      0        ?d      -           -               1        5s
head       5    0   2      0        ?d      -           -               1       12s
+ tr -d    1    2   0      0        ?d      -           -               7        7s
+ tr -d    2    0   0      2        hcmask  8           12              8      812s
+ tr -d    3    0   0      2        hcmask  8           17              8      812s
+ tr -d    4    0   0      2        hcmask  8           12              8      809s
+ tr -d    5    -   -      -        hcmask  8           11              8       n/a
this patch 1    2   0      0        hcmask  1            7              1        8s
this patch 2    2   0      0        hcmask  1            2              1        4s
this patch 3    2   0      0        hcmask  1            2              1        5s
this patch 4    2   0      0        ?d      -           -               7        7s
this patch 5    2   0      0        hcmask  2            7              2        8s
```

Five runs per side because `tools/test.pl` generates fresh passwords each time, so which branch the run takes and whether the single mask happens to cover them is not fixed. The fifth run of the middle side has no verdict and no time: the machine went to sleep in the middle of it, and what it wrote before that is still its own.

On the head of this branch the multi byte branch never fires, so the run gets one `?d` mask. The suite's own reproduction line, from the `test_report.log` of run 1:

```
./hashcat --quiet --potfile-disable --logfile-disable -d 1 -O --runtime 400 -D 2 --backend-vector-width 1 -a 3 -m 600 --increment --increment-min 1 --increment-max 8 test_1789259898/600_multihash_bruteforce.txt ?d?d?d?d?d?d?d?d
```

and the two verdicts that run came back with:

```
[ Type 600, Attack 3, Mode multi,  Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > Error : 1/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ Type 600, Attack 3, Mode multi,  Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > Error : 1/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

The two passwords in that run's window carry a multi byte character each, so no `?d` mask spells either of them. With the test fixed and nothing else, the hcmask branch fires and takes every password in the file, digits still spelled `?d`, which is one line per password and no bound on how many `?d` a line gets. The Linux section below shows one of those files.

Eleven digit positions on one line is 10^11 candidates for that line alone, and there are several, so runs 2, 3 and 4 each timed out twice, once per vector width, at the 400 seconds a kernel the suite allows, and took 809 to 812 seconds to come back having found nothing new.

Taking the same length window the single mask path takes bounds a line at `increment_max` positions, which is the bound that path already lived with, and takes the same hashes. Five runs, no error and no warning, four to eight seconds.

The same one run per side on four other fast modes:

```
mode    version      OK  Error  branch  mask lines  ?d on one line  time
0       head         2   0      ?d      -           -                4s
0       this patch   2   0      hcmask  4            5               6s
100     head         2   0      ?d      -           -                5s
100     this patch   2   0      hcmask  2            7               6s
1400    head         2   0      ?d      -           -               10s
1400    this patch   2   0      hcmask  2            1                5s
1700    head         0   2      ?d      -           -               19s
1700    this patch   2   0      hcmask  1            2                6s
6900    head         0   2      ?d      -           -               23s
6900    this patch   2   0      hcmask  3            2               15s
```

1700 and 6900 fail on the head for the same reason 600 does. 0, 100 and 1400 pass there because that run's in-window passwords happened to be digits only, and move to the hcmask branch here, which is the path that does not depend on that.

### And the same file on Linux

`-P` is a GNU extension, so the branch that never fires on a Mac is the one Linux takes, and there the second fault is what master does today. Two files, `tools/test.sh` as master has it and this branch's, against one binary built from master, one set of oracles and an RTX 4080:

```
$ ./tools/test.sh -m 600 -t multi -a 3 -O -d 1

version     run   OK  Error  Warning  branch  mask lines  ?d on one line  hashes  time
master       1    0   1      1        hcmask  8           12              8       903s
this file    1    2   0      0        hcmask  2            7              2         9s
master       2    0   0      2        hcmask  8           17              8       888s
this file    2    2   0      0        hcmask  3            5              3         8s
master       3    2   0      0        ?d      -           -               7        10s
this file    3    2   0      0        hcmask  1            1              1         4s
```

Two of the three runs on master took the hcmask branch and came back a quarter of an hour later, one with an error and a warning and one with two warnings, having timed out at the 400 seconds a kernel the suite allows. The third got a window with no multi byte character in it, took the single mask path and passed in ten seconds. Which of the two happens is down to the passwords `tools/test.pl` made that minute. The `+ tr -d` rows of the first table are that same state reached on a Mac, by fixing the test and nothing else.

The two files the second run wrote, kept from it. Master's, eight lines including the empty one the zero length password makes, and the seventh of them is seventeen digit positions, which is 10^17 candidates for that line alone:

```

龍?d?d?d
?d?d?d龍?d?d?d
?dＡ?d码?d?d
龍😀?d?d?d?d
?d?d龍😀?d?d?d?d?d?d?d?d?d
?d?d?d?d?d?d?d?d?d码?d?d?d?d文?d?d?d?d
カ€?dＡ?d?d?dか?d?d?d?d😀中?d中
```

and this file's, from the same run, three lines and five positions at the widest:

```
?d?d
?d?d?d?d?d
龍?d?d?d
```

`-m 1700` there, one run a side: master took the single mask path that time and passed in 17 seconds, this file took the hcmask path with three lines and five digit positions and passed in five.

## On `tools/test_edge.sh`

This pull request changes `tools/test.sh` and adds no hash mode, so there is no mode whose `tools/test_edge.sh` output would differ. That script's own side of the same review comment is in #4852.


jsteube:
In "test.sh", and again in "test_edge.sh" in #4852:

```
MINIMAL_MODES="0 100 110 400 500 2600 3000 3200 6211 11600 12500 13711 14200 14511 14600 14900 15400 15700 20510 22000 29511 33000 33500 34100"
```

Six of those are the head of a list that sits a few lines above in the same file: 6211 is TC_MODES, 13711 is VC_MODES, 14511 is CL_MODES, 14600 is LUKS1_LEGACY_MODE, 29511 is LUKS1_MODES, 34100 is LUKS2_MODES. Add a container family and the -M run still passes, one family short of what it claims to cover.

Before, a list of the modes that matter was read from the source. In "benchmark_deep.pl" that is DEFAULT_BENCHMARK_ALGORITHMS_BUF out of "benchmark.c", and in "test_coverage.sh" it is the family lists out of "test.sh". Now the same 24 modes are typed out in two more files.

I would have -M read DEFAULT_BENCHMARK_ALGORITHMS_BUF the way "benchmark_deep.pl" does, and where a mode belongs in that set and is missing, add it there rather than keep a second list beside it. That is what I would want settled before this and #4852 go in.

The help text in both scripts and the section in #4854 all describe -M as 24 hash types covering every code path, so those three strings follow from it too.

Separately for #4852, the mode loop in "test_edge.sh" still globs .pm only, so 1000 and 5200 have been outside the edge suite since #4817 merged.

matrix:
Done, `-M` reads the array and the hardcoded list is gone, and the three strings that called it 24 hash types go with it: the help text here, the one in `tools/test_edge.sh` in #4852, and the section in #4854.

It parses `DEFAULT_BENCHMARK_ALGORITHMS_BUF` the same way `benchmark_deep.pl` does, and gives the same answer. Both parsers over the same file, over a copy with seven of the twelve stale modes added to the array, `110 2600 6211 13711 14511 14900 15400`, and with the file not there at all, so the agreement is not an artefact of one list:

```
                                 benchmark_deep.pl   test.sh -M
src/benchmark.c                       107 modes       107 modes, identical
the same plus those seven             114 modes       114 modes, identical
the file not there                     dies           empty list, the error path exits 1
```

`-M` orders them numerically where `benchmark_deep.pl` keeps file order, which changes nothing: `HASH_TYPES` is the list the loop walks. The assignment sits with the other lists this file reads out of `src/`, because the copy it writes into a package freezes all of them into literals and a packaged run has no `src/`.

On the second half, adding what belongs in the set, I have left `src/benchmark.c` alone, and here is what it would take so the choice is yours rather than mine. Your six heads against the array:

```
list                 head    head in array   members   members in the array
TC_MODES             6211    no              24        29341
VC_MODES             13711   no              48        29421
CL_MODES             14511   no              15        none
LUKS1_LEGACY_MODE    14600   no               1        none
LUKS1_MODES          29511   yes             12        29511
LUKS2_MODES          34100   yes              1        34100
```

Two of the six heads are in the array. Four are not, and of those, TC and VC are still reached through another member of the family while `CL_MODES` and the legacy LUKS1 mode are not reached at all. So `-M` covers four of the six container families today and two of them not.

The twelve modes the hardcoded list had that the array does not are the whole of the candidate set:

```
110 2600 6211 13711 14200 14511 14600 14900 15400 20510 33000 33500
```

Adding any of them changes what `hashcat -b` reports for every user, which is not a call I should make from the test side, and it is a one line edit in `src/benchmark.c` now that nothing keeps a second copy. If you put in what you want there, `-M` picks it up with no further change here.

On `test_coverage.sh`, which reads the family lists out of this file: its reader is `grep -hoE '^[A-Z0-9_]*MODES="[0-9 ]*"'`, and the new assignment is `MINIMAL_MODES=$(`, so it is not picked up as a container list. The packaged copy of `test.sh` does have it rewritten to a literal, along with `PM_MODES` and the rest, and `test_coverage.sh` is not shipped into a package.

The branch has grown since you looked, all of it in the same file:

* the `-a 3` multi hash path tested for a multi byte character with `grep -qP`, which is a GNU extension. The grep on a Mac answers `invalid option -- P` and exits 2, `2>/dev/null` hides it, and the branch was never taken there. It is `tr` deleting the ASCII bytes and asking whether anything is left.
* the same path took every generated password once the branch did fire, while spelling digits as `?d`. Over one generation pass across all 513 modes that have an oracle, 3010 of the 3630 passwords come out longer than 8 bytes, which is where the single mask path stops on a fast hash, and those carry 15 digits at the median. Mode 22400 produces a 128 byte password of nothing but digits, which would be one line of that mask file asking for 10^128 candidates. It now takes the same length window the single mask path takes, so a mask covers at most `increment_max` positions, which is the bound that path already lived with.
* the four hash modes skipped for Apple: 1800, 600, 6900 and 18600. Their oracles produce the right hashes here, `tools/install_modules.sh` installs the two patched forks the comments blame, and `Crypt::GCrypt` left the suite in `1cc8ed779` fourteen months ago. The reproduction and the runs on both sides are in the description.

jsteube:
The first version of the section in #4854 said what -M is for:

```
* Minimal mode (-M): Test only 24 hash types that cover all distinct code paths and logic branches. This is useful for quick validation without running the full test suite across all hash modes
```

Pointing at DEFAULT_BENCHMARK_ALGORITHMS_BUF was my mistake, and I was not clear about who -M is for. The array answers which modes a user wants measured. What -M is for is the opposite end: the most diverse setups the tree has, one per distinct code path, so that a change touching many modes at once can be checked in a single short run before anyone spends a night on a full sweep. The 24 were chosen for that, and the benchmark list cannot stand in for them, because it was chosen to answer the other question.

Against that purpose the new list moves backwards twice over. The count goes from 24 to 107, and 6 code paths drop out: 14511, the only CL_MODES entry present in either script, 14600, which is LUKS1_LEGACY_MODE, and 20510, 14900, 15400 and 33500, each reached by its number, with 11 branches for 20510 in "test_edge.sh" alone. No coverage is lost for 6211 and 13711, since 29341 and 29421 are in the array and take the same container handling.

So I would put the list back to your first commit, with the help string and the section in #4854 going back with it. The three faults the wider run turned up stay, and so does the glob fix in #4852, since both stand on their own.

One part of my earlier comment survives that, and it does not block anything. Six of the 24 are the first entry of a list defined a few lines above, so reading them from there keeps -M following the family when the family changes:

```
LUKS1_LEGACY_MODE="14600"
CL_MODES="14511 14512 14513 14521 14522 14523 14531 14532 14533 14541 14542 14543 14551 14552 14553"
```

matrix:
Reverted. The list is the one the first commit had, mode for mode, and the help string here goes back with it. `src/benchmark.c` was never touched either way.

Your six paths check out against the array as it stands:

```
mode    in DEFAULT_BENCHMARK_ALGORITHMS_BUF   what reaches it
14511   no                                    the only CL_MODES entry either script has
14600   no                                    LUKS1_LEGACY_MODE
20510   no                                    its own number
14900   no                                    its own number
15400   no                                    its own number
33500   no                                    its own number
6211    no, but 29341 is                      the same container handling
13711   no, but 29421 is                      the same container handling
```

One number in your comment is on the other file. `20510` is named in twelve conditions of `tools/test.sh` and four of `tools/test_edge.sh`, not eleven of the latter, which puts the weight of it on the file this pull request changes:

```
$ grep -cE '^[[:space:]]*(if|elif) .*20510|20510\) skip_type' tools/test.sh
12
$ grep -cE '^[[:space:]]*(if|elif) .*20510' tools/test_edge.sh
4
```

The part of your first comment that survives is in:

```
-  MINIMAL_MODES="0 100 110 400 500 2600 3000 3200 6211 11600 12500 13711 14200 14511 14600 14900 15400 15700 20510 22000 29511 33000 33500 34100"
+  MINIMAL_MODES="0 100 110 400 500 2600 3000 3200 ${TC_MODES%% *} 11600 12500 ${VC_MODES%% *} 14200 ${CL_MODES%% *} ${LUKS1_LEGACY_MODE} 14900 15400 15700 20510 22000 ${LUKS1_MODES%% *} 33000 33500 ${LUKS2_MODES}"
```

Same 24 modes, and now `-M` follows a family list when the family changes rather than pinning the entry that happened to be first when it was written. It needed no packaging rewrite: those lists are literals in this file, so a packaged copy has them, which the array read did not.

The three faults stay here, and the glob fix stays in #4852. That one has had its `-M` half taken back out the same way, so what is left there is the glob, the skip for the two modes whose oracle is a `.py`, and the 73000 message that was only printed on Darwin. The section in #4854 goes back to what it said.

PR #4652, opened 2026-03-15
fix Rust build race condition

matrix:
details in #4602


pwrbob:
Thanks for this change, it looks good to me!

jsteube:
Thanks for this, and sorry it sat for so long. You asked @fukc-gihtub and @pwrbob to look at it on 15 March and nobody answered, which is on us rather than on you. Closing it now because the race is gone from master, though not by the route taken here.

Rather than serialising the two Cargo invocations, `Rust/hashcat-sys/build.rs` now writes the generated bindings into `OUT_DIR` instead of `src/bindings.rs`. `OUT_DIR` belongs to one build, so the two bridge crates no longer share the file they were racing on, and make can go on running them in parallel. That leaves `src/rust_common.mk` and the rest of the make plumbing here with nothing left to do.

#4602 reported the same race in January and is being closed alongside this one. Thanks for picking it up from the build side and for trying to get it settled.

PR #4653, opened 2026-03-19
Add PKCS#12 (.pfx, .p12) cracking support (modes 24430, 24440)

### LINKED ISSUE #351: PFX Algorithm Request
Since John (CPU only) and Elcomsoft DPR have support, I'd like to suggest trying to achieve native support in oclHashcat. 

In the past I've had terrible luck with buggy Elcomsoft Distributed Password Recovery (wordlist and mutation support is weak, handling large list of files to crack is terrible, stability having to restart agents, Windows only, etc). 

Either using rules directly in John or feeding Hashcat output via standard out to John worked, but was slow and had it's own issues. pfx2john already exists so that may help.

Generating PFX files can be done through numerous tools on both Windows or Linux. Example using Openssl:
openssl genrsa 2048 > private.pem
openssl req -x509 -new -key private.pem -out public.pem
openssl pkcs12 -export -in public.pem -inkey private.pem -out mycert.pfx -password PASSWORD

Thank you

kozmer:
This resolves #351 which has been open since 2016. It adds PKCS#12 (.pfx, .p12) password cracking by going after the integrity MAC instead of the encrypted contents. Uses the PKCS#12 KDF from RFC 7292 Appendix B with id=3, which means only one key derivation is needed and no decryption at all.

- 24430: SHA-1 KDF + HMAC-SHA1
- 24440: SHA-256 KDF + HMAC-SHA256

Also includes extraction tool (`tools/pkcs12_to_hashcat.py`) and test modules for both modes.

```
$ ./tools/test.sh -m 24430 -a all -t all -P
[ test_1773946355 ] > Init test for hash type 24430.
[ test_1773946355 ] [ Type 24430, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1773946355 ] [ Type 24430, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1773946355 ] [ Type 24430, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1773946355 ] [ Type 24430, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

```
$ ./tools/test.sh -m 24440 -a all -t all -P
[ test_1773946384 ] > Init test for hash type 24440.
[ test_1773946384 ] [ Type 24440, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1773946384 ] [ Type 24440, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
[ test_1773946384 ] [ Type 24440, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1773946384 ] [ Type 24440, Attack 0, Mode multi,  Device-Type Gpu, Kernel-Type Pure, Vector-Width 4 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

### DIFF
--- a/OpenCL/m24430-pure.cl
+++ b/OpenCL/m24430-pure.cl
@@ -1,0 +1,263 @@
+/**
+ * Author......: kozmer
+ * License.....: MIT
+ */
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha1.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct pkcs12_tmp
+{
+  u32  dgst[5];
+  u32  out[5];
+
+} pkcs12_tmp_t;
+
+typedef struct pkcs12
+{
+  u32 data_buf[16384];
+  int data_len;
+
+  u32 mac_digest[5];
+
+} pkcs12_t;
+
+KERNEL_FQ KERNEL_FA void m24430_init (KERN_ATTR_TMPS_ESALT (pkcs12_tmp_t, pkcs12_t))
+{
+  const u64 gid = get_global_id (0);
+
+  if (gid >= GID_CNT) return;
+
+  // PKCS#12 KDF (RFC 7292 Appendix B), id=3 for MAC key derivation
+  // Compute SHA-1(D || S || P) where:
+  //   D = 64 bytes of 0x03
+  //   S = salt repeated to 64 bytes
+  //   P = UTF-16BE password (with null terminator) repeated to ceil(len/64)*64
+
+  const int pw_len = pws[gid].pw_len;
+  const int salt_len = salt_bufs[SALT_POS_HOST].salt_len;
+  const int utf16_len = (pw_len + 1) * 2;
+  const int v2 = ((utf16_len + 63) / 64) * 64;
+
+  sha1_ctx_t ctx;
+
+  sha1_init (&ctx);
+
+  // diversifier
+
+  {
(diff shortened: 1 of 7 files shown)


jsteube:
In both modules the iteration count goes from the token straight into the salt with no floor under it. A line carrying $pkcs12$1$0$ gives iter of 0, and salt_iter = iter - 1 wraps that to 4294967295, so the loop kernel is handed a count no run will finish. The PKCS#8 module next door has the guard these are otherwise shaped after, and it is one line in the same place:

```
  const u32 iter = hc_strtoul ((const char *) iter_pos, NULL, 10);

  if (iter < 1) return (PARSER_SALT_ITERATION);

  salt->salt_iter = iter - 1;
```

The algorithm field in "module_24440.c" is checked a character at a time rather than as a value, so $pkcs12$299$ and $pkcs12$200$ both pass as SHA-256. Could that compare the whole field?

```
  if (token.len[1] != 3) return (PARSER_SIGNATURE_UNMATCHED);
  if (mac_algo_pos[0] != '2') return (PARSER_SIGNATURE_UNMATCHED);
```

The numbers to use are 37900 for the SHA1 mode and 37901 for the SHA256 one. One other thing, on the example hashes: both carry a run of filler bytes behind a zero length DER header, with 0123456789abcdef twice over as the salt, and that pair is what "docs/hashcat-example-hashes.md" publishes for the two modes once they land. Since "pkcs12_to_hashcat.py" comes with the branch, could they be taken from a real .pfx instead?

PR #4657, opened 2026-03-30
New Algorithm: GPU yescrypt

Chick3nman:
Added GPU yescrypt, current mode is 67000, before merge we will select the next available hash mode.

```
$y$j9T$oJqQoBLMgF5$77xERSZazujSxPgj6nYmzH5r0O2HelZJUyfW.ta7vxD:hashcat

Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 67000 (yescrypt)
Hash.Target......: $y$j9T$oJqQoBLMgF5$77xERSZazujSxPgj6nYmzH5r0O2HelZJ...ta7vxD
Time.Started.....: Sun Mar 29 14:25:57 2026 (3 secs)
Time.Estimated...: Sun Mar 29 14:26:00 2026 (0 secs)
Speed.#01........:      419 H/s (204.82ms) @ Accel:528 Loops:1024 Thr:1 Vec:1
Progress.........: 1056/9025 (11.70%)
Rejected.........: 0/1056 (0.00%)
Restore.Point....: 528/9025 (5.85%)
Restore.Sub.#01..: Salt:0 Amplifier:0-1 Iteration:5120-5462
Candidates.#01...: hashcRS -> hashc1s
Hardware.Mon.#01.: Temp: 40c Fan: 90% Util:100% Core:2790MHz Mem:10251MHz Bus:16
```


jmvermeulen:
Nice work getting yescrypt on the GPU!

I ran some benchmarks on an RTX 3080 (10 GB) and noticed a quick optimization that roughly doubles throughput. The `LOCAL_AS u32 sbox[Swords]` in the init and loop kernels allocates 12 KB of shared memory per workgroup, but since `kernel_threads_max = 1` there's no inter-thread sharing benefit. It just caps SM occupancy at ~2-5 workgroups.

Since the S-box data already lives in global memory (`tmps[gid].S`), you can point directly at it instead of copying back and forth:

```diff
--- init kernel ---
- LOCAL_AS u32 sbox[Swords];
+ GLOBAL_AS u32 *sbox = tmps[gid].S;

  // ... later, remove the copy-out:
- for (u32 i = 0; i < Swords; i++) tmps[gid].S[i] = sbox[i];
+ // sbox already points to tmps[gid].S, no copy needed

--- loop kernel ---
- LOCAL_AS u32 sbox[Swords];
- u32 X[YESCRYPT_STATE_CNT4];
- for (u32 i = 0; i < YESCRYPT_STATE_CNT4; i++) X[i] = tmps[gid].P[i];
- for (u32 i = 0; i < Swords; i++) sbox[i] = tmps[gid].S[i];
+ GLOBAL_AS u32 *sbox = tmps[gid].S;
+ u32 X[YESCRYPT_STATE_CNT4];
+ for (u32 i = 0; i < YESCRYPT_STATE_CNT4; i++) X[i] = tmps[gid].P[i];

  // ... later, remove the copy-back:
- for (u32 i = 0; i < Swords; i++) tmps[gid].S[i] = sbox[i];
+ // sbox already points to tmps[gid].S, no copy needed
```

Same change in `inc_hash_yescrypt.h` and `inc_hash_yescrypt.cl` - replace `LOCAL_AS u32 *sbox` with `GLOBAL_AS u32 *sbox` in the function signatures.

Benchmark results on RTX 3080 (10 GB, yescrypt $y$j9T$ default params):

| Version | Max Accel | Speed |
|---------|-----------|-------|
| Original (LOCAL sbox) | 384 | 68 H/s |
| Global sbox, same accel | 384 | 116 H/s (+70%) |
| Global sbox, higher accel | 512 | 135 H/s (+99%) |

The freed shared memory also allows higher accel values - 448 and 512 work where they OOM'd before. Correctness verified by cracking the self-test hash.

The 12 KB sbox fits comfortably in L2 cache (RTX 3080 has 5 MB), so the global memory latency penalty is minimal while the occupancy improvement is significant.

Two other minor things I noticed while reviewing:
- `OPTS_TYPE` is missing `OPTS_TYPE_MULTIHASH_DESPITE_ESALT`. Without it, loading multiple hashes with the same salt will error at runtime (`hashes.c:2292`). The bridge-based yescrypt modules (72000-74000) all include this flag.
- The Perl test module doesn't generate a random salt when `$salt` is undef, which could cause issues with `test.sh` in single mode.

Chick3nman:
@dev-goalz Thanks for the note, took that into consideration and then pushed it quite a bit further :)

New benchmark:
```
-----------------------------------------------
* Hash-Mode 67000 (yescrypt) [Iterations: 5462]
-----------------------------------------------

Speed.#01........:     1408 H/s (94.71ms) @ Accel:890 Loops:1024 Thr:32 Vec:1
```

solardiz:
Thank you @Chick3nman and @dev-goalz, this is good work and was much needed to have reference GPU speeds for yescrypt proper (as opposed to the older 0.5'ish revisions used during PHC and by miners).

Now that the latest revision here implements groups of 4 cooperating threads, is the preference to use global rather than local memory still right? I think this change should have improved occupancy when using local memory for S-boxes by a factor of 4 (compared to this PR's original code), did it?

Chick3nman:
@solardiz To make sure I'm understanding and we are talking about the same thing, each hash is now processed by a full 32-thread warp (unless I have messed all of this up and it magically worked out). All 32 threads share the load of moving the big V blocks around and 4 of them, as with PWXgather, do the sbox lookups. So the sbox is shared across the whole group, not just 4 threads.

I did try keeping the sbox in LDS with the cooperative threads, and global still came out ahead for the default `$y$j9T$` params. On my laptop RTX 4060 8 GB 24SM: LDS sbox came in at 655 H/s, global at 721 H/s, so I discounted the LDS method and moved forward with cleaning up the global arrangement, though I will do some more testing if you think there's room for improvement there as my implementation is still certainly not perfect.

You're right that sharing the sbox improves occupancy, but in testing it seems that local memory isn't what is holding back the performance, the limit seems to be the 16 MB of V each hash needs. With the sbox in LDS we can only fit ~6 hashes at once per SM. With it in global we can pull ~19 hashes per SM now, limited by VRAM instead of LDS. yescrypt is latency limited so the sbox lookups are the slowest part of the inner loop and having more hashes running at once hides that latency better than the faster local memory lookups do in practice. So global ended up both faster and with a shorter loop/kernel time. This is for the default params on a GPU with relatively limited VRAM, especially given how large cards have gotten with AI workloads driving massive VRAM. With more VRAM, there's room for more concurrent hashes, at least until we hit some other bottleneck (I haven't tested the scaling to find where this falls off but the 4070TiS w/ 16GB was almost exactly 2x the speed of 8GB so it doesn't seem like we are compute bound on cards from the same generation/class). For smaller settings (lower N or r), where VRAM isn't necessary for many concurrent threads, local memory should win but I assumed (semi-naively, I will admit) that real world hashes would rarely have settings low enough for it to make sense.

solardiz:
@Chick3nman This makes sense. The performance difference you show is small, so micro-optimizations (such as how exactly the 64-bit S-box loads and mul-add-xor are done) could possibly make LDS win or not.

But more importantly:

Maybe having the current block `X` in global while S-boxes in local will work better, because access to the block is sequential.

> With the sbox in LDS we can only fit ~6 hashes at once per SM. With it in global we can pull ~19 hashes per SM now, limited by VRAM

> For smaller settings (lower N or r), where VRAM isn't necessary for many concurrent threads, local memory should win but I assumed (semi-naively, I will admit) that real world hashes would rarely have settings low enough for it to make sense.

What about larger settings? For example, in CIQ's RLC Pro Hardened, we upgraded to 128 MB by default (higher N). That's 8x larger, so your ~19 hashes per SM will become 2 or 3 per SM and it will be same number regardless of where the S-boxes are, right? If so, when occupation is similarly low _anyway_, you'd want to optimize latency - meaning put S-boxes in local.

Chick3nman:
@solardiz Unsurprisingly, it looks like you are right after some quick tests, but I also ran into some unexpected results just now testing variants of mixed LDS/Global configs. I'll put some thorough testing together and publish what I find but it looks like an "adaptive" approach may be feasible where we move things around for best performance at different hash params/configs. Give me a few hours to get something put together and cleaned up :)

Chick3nman:
@solardiz I've made some more changes to allow adaptive placement and the explanation/data will hopefully shed some light on what is going on:

First thing to note is that placement is... finnicky and it seems as though placing both together out performs across my full set of tests.

All four placement options 4060, default N, accel 454:

X | S-box | H/s
-- | -- | --
LDS | global | 722
LDS | LDS | 652
global | LDS | 652
global | global | 759

The gains are small here but consistent and remain consistent across my other test cards so far(except for some behavior explained below). This is something I'm not sure I fully understand the reasoning for and it's possible there's still an optimal placement strategy that mixes LDS/Global. There's some oddities in my test data which you will see below. To be clear, despite it showing global/global being the best in this table, what the rest of my data shows and what I'm trying to represent here is that the speeds are best when both are placed together. The global/global thing seems to be an artifact of the test being run on the 4060 specifically, which I can't explain.

Global vs LDS across N sizes, on both of my local cards and both backends(CUDA vs OpenCL):

N (MB) | 4060 OpenCL (hashes/SM) global / LDS | 4070 Ti S OpenCL (hashes/SM) global / LDS | 4070 Ti S CUDA global / LDS
-- | -- | -- | --
16 | (18.9) 759 / 652 | (13.5) 1472 / 1784 | 1468 / 1804
32 | (9.5) 224 / 284 | (6.7) 484 / 651 | 481 / 654
64 | (4.7) 74 / 121 | (3.3) 211 / 273 | 210 / 273
128 | (2.3) 29 / 35 | (1.7) 66 / 77 | 65 / 77

Here we see some interesting behaviors and this is the unexpected result I referred to earlier. On the 4060 we see Global win when occupancy can be pushed higher, then LDS wins as N climbs. However, on the 4070Ti Super 16GB, we see LDS wins across the board in OpenCL and CUDA (where we can alter the LDS allocation slightly, though we seemingly didn't need to much like how bcrypt seems to behave). This is where we find the threshold, >15 hashes in flight leans towards Global, <15 leans LDS and by a noticeable margin, at least on my local dev systems. But as we see from the next test, this is not so clear cut. I've chosen to leave this threshold for now as for smaller GPUs like my dev systems it appears to be beneficial. 

Now on to pushing things a bit and chasing occupancy on big cards to see what happens. Here I used an RTX Pro 6000 Max-Q w/ 96GB of VRAM. It has a lower power limit than the server or full power Pro 6000 but should still give us the result we want as we aren't bouncing off the power limit in these tests as far as I can tell.

Running my test sweep of the different N values:
```
N(MB)  config   speed
-------------------------------------------------------------
4      global   6597 H/s (76.84ms) @ Accel:1024
4      lds      8761 H/s (57.67ms) @ Accel:1024
8      global   3299 H/s (102.87ms) @ Accel:1024
8      lds      4377 H/s (77.39ms) @ Accel:1024
16     global   1041 H/s (66.82ms) @ Accel:467
16     lds      1967 H/s (78.69ms) @ Accel:1024
32     global   805 H/s (107.34ms) @ Accel:1024
32     lds      983 H/s (86.13ms) @ Accel:1024
64     global   1 H/s (55.74ms) @ Accel:1
64     lds      491 H/s (86.30ms) @ Accel:1024
128    global   201 H/s (73.47ms) @ Accel:704
128    lds      384 H/s (34.50ms) @ Accel:704
```

This is... not the intended behavior we want to see here and the cause appears to be hashcat's default clamp on Accel to 1024 max across the board, which drastically reduces the amount of hashes we can have in flight despite there being plenty of VRAM or other capacity left on the card. This is not a cap I can safely increase in this PR but for the sake of experimentation, I did it anyways to see what would happen.

```
N(MB)  config   speed
-------------------------------------------------------------
4      global   13824 H/s (776.01ms) @ Accel:21495
4      lds      18097 H/s (592.56ms) @ Accel:21495
8      global   6662 H/s (536.91ms) @ Accel:10747
8      lds      8971 H/s (398.52ms) @ Accel:10747
16     global   2789 H/s (312.34ms) @ Accel:5373
16     lds      3564 H/s (242.45ms) @ Accel:5373
32     global   1452 H/s (159.68ms) @ Accel:2686
32     lds      2068 H/s (109.61ms) @ Accel:2686
64     global   451 H/s (126.84ms) @ Accel:1342
64     lds      643 H/s (86.43ms) @ Accel:1342
128    global   197 H/s (71.30ms) @ Accel:671
128    lds      296 H/s (44.47ms) @ Accel:671
```

Now this result is extremely interesting. Not only does LDS maintain a healthy lead across the full range of N sizes, but the speeds far exceed what hashcat is currently capable of under the default accel cap, at least in the smaller N size tests. There's a very real amount of performance left on the table here. Note that 128 is lower in this set of tests due to my hacky hard coded values throwing the Accel calculation down to 671 instead of 704, but we see it is perfectly capable of the higher accel with the shipped code in the PR. 

For completeness, I reran the "sbox/X block global/LDS" mix test cases as well on the RTX Pro 6000 Max-Q w/ the accel limit hack:
```
N(MB)  config(sbox + X, global/LDS)   speed
-------------------------------------------------------------
4      gg       14240 H/s (753.35ms) @ Accel:21495
4      gl       12925 H/s (830.13ms) @ Accel:21495
4      lg       14928 H/s (718.58ms) @ Accel:21495
4      ll       18105 H/s (592.25ms) @ Accel:21495
8      gg       6810 H/s (525.27ms) @ Accel:10747
8      gl       7206 H/s (496.38ms) @ Accel:10747
8      lg       7787 H/s (459.22ms) @ Accel:10747
8      ll       7510 H/s (476.21ms) @ Accel:10747
16     gg       2811 H/s (309.81ms) @ Accel:5373
16     gl       3067 H/s (283.23ms) @ Accel:5373
16     lg       3651 H/s (236.77ms) @ Accel:5373
16     ll       3564 H/s (242.51ms) @ Accel:5373
32     gg       1462 H/s (158.43ms) @ Accel:2686
32     gl       1544 H/s (149.50ms) @ Accel:2686
32     lg       1309 H/s (177.91ms) @ Accel:2686
32     ll       2067 H/s (109.59ms) @ Accel:2686
64     gg       466 H/s (122.79ms) @ Accel:1342
64     gl       510 H/s (111.55ms) @ Accel:1342
64     lg       505 H/s (112.32ms) @ Accel:1342
64     ll       643 H/s (86.42ms) @ Accel:1342
128    gg       195 H/s (71.86ms) @ Accel:671
128    gl       134 H/s (106.91ms) @ Accel:671
128    lg       264 H/s (50.90ms) @ Accel:671
128    ll       366 H/s (34.46ms) @ Accel:671
```

With the cap hacked to allow higher values we start to see some inconsistency in the placements. It looks like sbox in LDS wins in every case, but for some reason X blocks in global wins for N=16 and N=8 specifically. It's not by much but it does appear to remain this way when rerunning the tests. I don't think this is great test data, the accel hack is definitely a hack, but it's interesting nonetheless.

So this leaves us in a bit of a weird position. The 4060 in my dev laptop shows a crossover with Global/Global being ahead consistently in a few of my tests, but only at smaller sizes of N. Additionally, we can see that on extremely large GPUs we could be going a lot faster for some hashes but the artificial Accel cap kills our ability to do this specific attack at high enough occupancy to actually achieve those speeds. Honestly, I'm not 100% sure where to go from here. The code in the PR is updated to be as fast as I can get it to go on my 4060/4070TiS test cards, but when testing on very large GPUs we see that there may be larger changes needed to achieve full speed here and this could potentially be impacting other modes in a similar way. 

TL;DR: @solardiz your assumptions seem validated and LDS does seem to win but there's some interesting results in the test data that I can't fully explain. I think the PR is in about as good of a state as I alone can get it, but it seems like there's still some room for improvements though they require larger test hardware(I briefly rented the Pro 6000 as I don't own one) and potentially larger changes to hashcat given the results of my accel hack.

solardiz:
Thank you @Chick3nman, these are very helpful results, including for my decision-making on (not) bumping the S-box size in defensive uses of yescrypt yet.

> micro-optimizations (such as how exactly the 64-bit S-box loads and mul-add-xor are done

I guess you have seen https://lists.openwall.net/phc-discussions/2018/04/29/1? Also, what we do in Argon2-opencl in JtR? This is not necessarily applicable or beneficial here, but just ideas to consider.

Chick3nman:
I don't think that thread has come up or if so I didn't look too deeply at it so thanks for the link, I'll look into that and the argon2 implementation and see if there's anything left I can employ here :)

PR #4662, opened 2026-04-09
Add module 38000: Mikrotik RouterOS EC-SRP5 (Curve25519)

dev-goalz:
## Summary

Mikrotik RouterOS is the operating system running on all Mikrotik network devices (routers, switches, access points). Starting with RouterOS 7.x, user passwords are stored as EC-SRP5 verifiers based on Curve25519 scalar multiplication, replacing the older MD5-based scheme.

The verifiers are stored in the device configuration and included in `.backup` files. The hash extraction tool (`tools/routeros2hashcat.py`) supports three input types:
- `.backup` files (packed binary, plaintext only)
- Unpacked backup directories (`.dat`/`.idx` pairs)
- Monolithic config store files (`cfg.*` from NAND flash)

**Algorithm:**
- `inner = SHA256(username + ":" + password)`
- `scalar = SHA256(salt_16bytes + inner)`
- `point = scalar * G` on Curve25519 (Weierstrass form)
- `x_mont = convert point.x to Montgomery form`
- `verifier = first 28 bytes of x_mont (big-endian)`

**Hash format:** `$mikrotik$<username>$<salt_hex_32>$<verifier_hex_56>`

```
$mikrotik$hashcat$3069ece5c30baced372fb3d64647d42f$25772ddb1fa2e0f13511d7d7dc8c611b6529dd915d10482243dfca38:hashcat

Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 33200 (Mikrotik RouterOS EC-SRP5 (Curve25519))
Hash.Target......: $mikrotik$hashcat$3069ece5c30baced372fb3d64647d42f$...dfca38
Speed.#01........:   221.9 kH/s (9.08ms) @ Accel:1 Loops:1 Thr:32 Vec:1
Hardware.Mon.#01.: Temp: 35c Util: 99% Core:1980MHz Mem:9251MHz Bus:16
```


jsteube:
In "m33200-pure.cl" the init kernel assembles the username, the colon and the password into one buffer before hashing it, and that buffer holds 256 bytes where the string reaches 512. The username token in "module_33200.c" accepts 255 bytes, module_pw_max is left at the default of 256, so total_len runs to 255 + 1 + 256. The username alone fills w_tmp to its last word, and the first password byte then writes one past the end:

```
  u32 w_tmp[64];
  ...
  for (u32 i = 0; i < pw_len; i++)
  {
    const u32 dst_pos  = username_len + 1 + i;
    const u32 dst_word = dst_pos / 4;
    ...
    w_tmp[dst_word] |= (b << (dst_byte * 8));
  }
```

A long username is not needed to reach it. An "admin" account and a 251 character candidate come to 257 bytes, and sha256_update_swap() then reads the same range back. Passing the username and colon to one sha256_update() and the password to a second removes the buffer from the picture, because the context keeps the partial block between the two calls. This one blocks for me.

The number to use is 38000. The branch also conflicts in "docs/changes.txt", where the entry opens a "## New Algorithms" section that the changelog rework replaced with "## New Hash Modes". Could you drop that hunk? The changelog entries are written here at merge.

dev-goalz:
Fixed the init kernel. The username + ":" + password assembly could indeed run past the 256 byte buffer, so I split it into two sha256_update_swap calls, "username:" first and the password second. The context keeps the partial block between the calls, so no buffer has to hold the whole string. Self-test still passes.

Renumbered the module to 38000.

Dropped the docs/changes.txt hunk as well. The branch no longer touches that file, so the changelog entry is yours to add at merge. I also squashed the branch down to a single commit while I was at it.

matrix:
@dev-goalz 

```
  src/modules/module_38000.c:103:28: error: call to undeclared function 'input_tokenizer'; ISO C99 and later do not support implicit function declarations
  src/modules/module_38000.c:194:15: error: no member named 'module_dictstat_disable' in 'struct module_ctx'

```

dev-goalz:
Thanks for the heads up @matrix. Build fix is up in #4894: added the parser.h include for input_tokenizer, registered module_advice_notice and module_hash_hints, and dropped the removed module_dictstat_disable field. Builds clean and self-test passes here. @jsteube

PR #4663, opened 2026-04-10
Add hash-mode: Mozilla key4.db SHA384 (m26150)

### LINKED ISSUE #4585: Change in Firefox 146.0 Credential Manager. Migration SHA1 -> SHA384
**Is your request based on a publicly known cryptographic schema and where can we find information about?**
- As of version 146.0, Mozilla Firefox has changed the way the primary password is used to protect the Masterkey used to encrypt the credentials storage
- This hashing mode is similar to the one implemented currently as -m26100
- The major changes are
       - global_salt length changes from 20 bytes to 48 bytes
       - instead of SHA1(global_salt || primary_password), the hashing algorithm changed to SHA384(global_salt || primary password)
       - the rest is identical to m26100 
 - I have made a post on StackExchange highlighting the change: https://superuser.com/questions/1931365/changes-in-firefox-credential-storage-version-146-0

**Describe alternatives you've considered**
I have attempted to copy and modify the m26100 module myself, but lack the C knowledge to do so adequately.

**Provide an example hash/database/file and a known correct password**
[dollarSign]mozilla[dollarSign]\*AES\*95f6acda22fbfb543226eca01b2e65099730d20cb3114c47536bc33a96b9137ea4a1dcebbcb8d924d20f5784be1fb1f3\*b3edd4ec861f2e7a16c72771302a55f11b1504ac1792a4c408c44aa82023ef7f\*10000\*040eaff904bd66d4f3ba86cbb13ced52\*a0b908a87ba816a5cb4d9632f015a972
note the longer global_salt (third parameter)
plaintext: hashcat

**Explain how the cryptographic schema is implemented**
- Schema similar to -m26100.
- I have made this sketch that visualizes m26100: https://i.sstatic.net/8lNMyGTK.png. In the new algorithm SHA1 needs to be replaced by SHA384
- Pseudocode:

```
SIGNATURE_MOZILLA = hash[0]      // mozilla
SIGNATURE_MOZILLA_AES = hash[1]        // AES
global_salt = hash[2]        // 95f6acda22fbfb543226eca01b2e65099730d20cb3114c47536bc33a96b9137ea4a1dcebbcb8d924d20f5784be1fb1f3
entry_salt = hash[3]       // b3edd4ec861f2e7a16c72771302a55f11b1504ac1792a4c408c44aa82023ef7f
iterations = hash[4]       // 10000
iv  = hash[5]       // 040eaff904bd66d4f3ba86cbb13ced52
encrypted_value = hash[6]      // a0b908a87ba816a5cb4d9632f015a972

sha384hash = SHA384.ComputeHash(primary_password_to_check.EncodeAsUTF8Bytes + global_salt).first_20_bytes
master_key = PBKDF_HMAC(PASSWORD = sha384hash, SALT= entry_salt, ITERATIONS = iterations, ALGORITHM = Sha256)         // master_key length = 32
plain_value = AES-CBCDecrypt(KEY = master_key, IV = iv, CIPHERTEXT = encrypted_value)
if (plain_value.EncodeAsUTF8String == "password-check") {correct_password = primary_password_to_check}
```

**Describe the known limitations of the algorithm**
- global_salt length needs to be 48 bytes

**Where do you typically find the algorithm**
- Mozilla Firefox above version 146.0

dev-goalz:
# Add hash-mode: Mozilla key4.db SHA384 (m26150)

## Summary

Adds hash-mode `26150` for cracking the password of Mozilla `key4.db` files
as written by Firefox 146 and later. Starting with Firefox 146 (released March
2025), Mozilla replaced the SHA1-based key derivation from the legacy NSS
`SDR` container with a SHA384-based variant. The existing `26100` plugin no
longer works against these new profiles.

Tracks issue #4585.

## Algorithm

```
sha384_hash = SHA384(global_salt || password)                         # full 48 bytes
master_key  = PBKDF2-HMAC-SHA256(sha384_hash, entry_salt, iter, 32)
plaintext   = AES-256-CBC-Decrypt(master_key, iv, ct)
verify      : plaintext == b"password-check\x02\x02"
```

Differences vs. m26100:

- Inner hash is SHA384 instead of SHA1.
- `global_salt` is 48 bytes (was 20 bytes for SHA1-based profiles).
- The *full* 48-byte SHA384 output is used as the PBKDF2 password. The
  original issue description suggested truncating to 20 bytes - this is
  incorrect; verified against a real Firefox 146 key4.db and independently
  with Python before any kernel work.

The rest of the pipeline (PBKDF2-HMAC-SHA256, AES-256-CBC decrypt, PKCS#7
padding check against `"password-check\x02\x02"`) is identical to m26100.

## Hash Format

```
$mozilla$*AES*<global_salt_96hex>*<entry_salt_64hex>*<iterations>*<iv_32hex>*<ct_32hex>
```

- `global_salt`: 48 bytes, 96 hex chars
- `entry_salt` : 32 bytes, 64 hex chars
- `iterations` : decimal, typically 10000
- `iv`         : 16 bytes, 32 hex chars
- `ct`         : 16 bytes, 32 hex chars (encrypted `password-check` string)

Example (self-test vector, password `hashcat`):

```
$mozilla$*AES*95f6acda22fbfb543226eca01b2e65099730d20cb3114c47536bc33a96b9137ea4a1dcebbcb8d924d20f5784be1fb1f3*b3edd4ec861f2e7a16c72771302a55f11b1504ac1792a4c408c44aa82023ef7f*10000*040eaff904bd66d4f3ba86cbb13ced52*a0b908a87ba816a5cb4d9632f015a972
```

## Files

- `src/modules/module_26150.c` - host module (based on m26100)
- `OpenCL/m26150-pure.cl` - GPU kernel (SHA384 init, SIMD PBKDF2-HMAC-SHA256 loop, AES-256-CBC comp)
- `tools/test_modules/m26150.pm` - Perl unit test module
- `tools/mozilla2hashcat_sha384.py` - extractor for Firefox 146+ `key4.db`
- `docs/changes.txt` - changelog entry

## Extracting from a real profile

`tools/mozilla2hashcat.py` only produces SHA1-variant (`m26000`) hashes. The
new `tools/mozilla2hashcat_sha384.py` handles the SHA384 variant. It is
stdlib-only (sqlite3 + a tiny hand-rolled DER parser, no pyasn1/PyCryptodome
dependency), opens `key4.db` read-only, and auto-detects the variant by the
`metaData.item1` length (48 bytes = SHA384 / m26150, 20 bytes = SHA1 /
m26000). Legacy profiles are rejected with a pointer to the existing tool so
the two scripts stay cleanly separated.

Verified end-to-end against a local Firefox 146 profile with master password
`hashcat`: the extracted hash decrypts to `password-check\x02\x02` in the
Python reference and cracks with the m26150 kernel on an RTX 3080.

### test.sh (single)

```
$ ./tools/test.sh -m 26150 -a 0 -t single
[ test_1775802750 ] > Init test for hash type 26150.
[ test_1775802750 ] [ Type 26150, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
```

### Benchmark (self-test inside hashcat)

```
$ ./hashcat -m 26150 -b -d 1
-------------------------------------------------------------
* Hash-Mode 26150 (Mozilla key4.db SHA384) [Iterations: 9999]
-------------------------------------------------------------

Speed.#01........:   332.1 kH/s (93.84ms) @ Accel:6 Loops:1000 Thr:768 Vec:1
```

GPU: NVIDIA RTX 3080. Performance is in the same ballpark as m26100 since
SHA384 only runs once per candidate in the init kernel; the hot path is the
PBKDF2-HMAC-SHA256 loop which is identical to m26100.

### Mask attack self-test

```
$ ./hashcat -m 26150 -a 3 '$mozilla$*AES*95f6acda...' 'hashc?a?a'
$mozilla$*AES*95f6acda...:hashcat
```

## Real-world Use Case

Firefox 146 was released on 4 March 2025. Any Firefox profile created or
password-migrated on Firefox 146 or later uses the new SHA384-based key
derivation for `key4.db`. `m26100` does not work against Firefox 146+ profiles.

The algorithm is also used by Thunderbird once it picks up the same NSS
change.

## Notes

- Based on m26100; kernel structure is a drop-in replacement with SHA384
  instead of SHA1 in the init kernel and a longer global salt in the host
  decoder. The PBKDF2 loop and AES-CBC comp kernels are unchanged.
- `module_hashes_count_max` behavior matches m26100 (multi-hash with
  distinct salts works; multi-hash with the same salt follows the same
  constraint as m26100 via the esalt path).
- `OPTS_TYPE_PT_GENERATE_LE` matches m26100.
- A Python reference implementation (`verify_m26150.py`) was used to verify
  the test vector before any kernel code was written, per the lesson learned
  in issue #4585 where the spec description was incorrect about truncation.

Closes #4585.

### DIFF
--- a/OpenCL/m26150-pure.cl
+++ b/OpenCL/m26150-pure.cl
@@ -1,0 +1,447 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha384.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha256.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_aes.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct mozilla_aes_tmp
+{
+  u32  ipad[8];
+  u32  opad[8];
+
+  u32  dgst[8];
+  u32  out[8];
+
+} mozilla_aes_tmp_t;
+
+typedef struct mozilla_aes
+{
+  u32 iv_buf[4];
+  u32 ct_buf[4];
+
+} mozilla_aes_t;
+
+DECLSPEC void hmac_sha256_run_V (PRIVATE_AS u32x *w0, PRIVATE_AS u32x *w1, PRIVATE_AS u32x *w2, PRIVATE_AS u32x *w3, PRIVATE_AS u32x *ipad, PRIVATE_AS u32x *opad, PRIVATE_AS u32x *digest)
+{
+  digest[0] = ipad[0];
+  digest[1] = ipad[1];
+  digest[2] = ipad[2];
+  digest[3] = ipad[3];
+  digest[4] = ipad[4];
+  digest[5] = ipad[5];
+  digest[6] = ipad[6];
+  digest[7] = ipad[7];
+
+  sha256_transform_vector (w0, w1, w2, w3, digest);
+
+  w0[0] = digest[0];
+  w0[1] = digest[1];
+  w0[2] = digest[2];
+  w0[3] = digest[3];
+  w1[0] = digest[4];
+  w1[1] = digest[5];
+  w1[2] = digest[6];
+  w1[3] = digest[7];
+  w2[0] = 0x80000000;
+  w2[1] = 0;
+  w2[2] = 0;
+  w2[3] = 0;
+  w3[0] = 0;
+  w3[1] = 0;
+  w3[2] = 0;
+  w3[3] = (64 + 32) * 8;
+
+  digest[0] = opad[0];
+  digest[1] = opad[1];
+  digest[2] = opad[2];
+  digest[3] = opad[3];
+  digest[4] = opad[4];
+  digest[5] = opad[5];
+  digest[6] = opad[6];
+  digest[7] = opad[7];
+
+  sha256_transform_vector (w0, w1, w2, w3, digest);
+}
+
+KERNEL_FQ KERNEL_FA void m26150_init (KERN_ATTR_TMPS_ESALT (mozilla_aes_tmp_t, mozilla_aes_t))
+{
+  /**
+   * base
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= GID_CNT) return;
+
+  // SHA384(global_salt + password) -- Firefox 146+ uses SHA384 instead of SHA1
+  // global_salt is 48 bytes (12 x u32) stored in salt_buf[0..11]
+
+  sha384_ctx_t ctx;
+
+  sha384_init (&ctx);
+
+  u32 gs[32] = { 0 };
+
+  gs[ 0] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 0];
+  gs[ 1] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 1];
+  gs[ 2] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 2];
+  gs[ 3] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 3];
+  gs[ 4] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 4];
+  gs[ 5] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 5];
+  gs[ 6] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 6];
+  gs[ 7] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 7];
+  gs[ 8] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 8];
+  gs[ 9] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[ 9];
+  gs[10] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[10];
+  gs[11] = salt_bufs[DIGESTS_OFFSET_HOST].salt_buf[11];
+
+  sha384_update_swap (&ctx, gs, 48);
+  sha384_update_global_swap (&ctx, pws[gid].i, pws[gid].pw_len);
+
+  sha384_final (&ctx);
+
+  // Full 48-byte SHA384 output used as PBKDF2 password
+  // SHA384 ctx.h[] are u64 (big-endian), extract to u32
+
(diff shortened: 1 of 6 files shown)


jsteube:
In "tools/mozilla2hashcat_sha384.py" the new script reads the same structure as the existing "mozilla2hashcat.py". Both decode PBES2 1.2.840.113549.1.5.13, PBKDF2 1.2.840.113549.1.5.12 and hmacWithSHA256 1.2.840.113549.2.9 out of metaData.item2, and what separates the two hash modes is a field both of them already read:

```
        sys.exit(f"unexpected global_salt length {gs_len}; expected 48 (SHA384) or 20 (SHA1)")
```

Could you fold this into "mozilla2hashcat.py" and have the one script print whichever mode it finds? Someone holding a key4.db has to choose between two scripts otherwise, and the salt length already settles which it is.

The mode named on the other branch is the wrong one:

```
            "global_salt is 20 bytes — this is the legacy SHA1 variant (m26000).\n"
```

Mode 26000 is Mozilla key3.db and 26100 is Mozilla key4.db, so a key4.db carrying a 20 byte global salt belongs to 26100. That reads three times in the script, in two docstring lines and here. Your description has 26100 throughout, so it looks confined to the tool.

The real line also carries an em dash, as does the metaData lookup failure above it. Those two are the only non ASCII in the diff, and CONTRIBUTING.md asks for plain ASCII in comments and documentation with en and em dashes named directly.

jsteube:
In "backend.c" the field that should already carry this is recorded one query short:

```
        if (hc_clGetDeviceInfo (hashcat_ctx, device_param->opencl_device, CL_DEVICE_MAX_WORK_GROUP_SIZE, sizeof (device_maxworkgroup_size), &device_maxworkgroup_size, NULL) == -1)
        {
          device_skip (device_param, "clGetDeviceInfo() failed");

          continue;
        }

        device_param->device_maxworkgroup_size = device_maxworkgroup_size;
```

Hashcat launches one dimensional kernels only, so the usable ceiling for it is the smaller of that value and work_item_sizes[0]. Those cross on the Adreno, and what gets written down is a capability the driver will not honour.

Could you clamp it there, at enumeration, rather than in get_opencl_kernel_wgs()? Four places read that field and each inherits the high number today: the thread cap in "autotune.c", the device residency behind the spill estimate in "backend.c", and the spill sizing in "scrypt_common.c" and "argon2_common.c". A clamp in the kernel helper leaves all four as they are, and it repeats two device queries inside a function that runs 33 times per device.

Keeping a clamp in get_opencl_kernel_wgs() still makes sense, since CL_KERNEL_WORK_GROUP_SIZE is evidently not bounded by the dimension limit on that runtime, but it can read the stored field instead of querying the device itself.

One smaller thing:

```
  if (max_dimensions > 16) return -1; // Should never happen
```

A device reporting more than 16 dimensions is refused rather than run. The spec minimum is 3 so it should never fire, but its failure mode is worse than the case it guards against.

dev-goalz:
> In "backend.c" the field that should already carry this is recorded one query short:
> 
> ```
>         if (hc_clGetDeviceInfo (hashcat_ctx, device_param->opencl_device, CL_DEVICE_MAX_WORK_GROUP_SIZE, sizeof (device_maxworkgroup_size), &device_maxworkgroup_size, NULL) == -1)
>         {
>           device_skip (device_param, "clGetDeviceInfo() failed");
>

Sorry @jsteube but i think this comment is related to #4729 instead of this PR?

jsteube:
You are right, that one was meant for #4729 and landed here by mistake. Please ignore it, and sorry for the noise. The comment above it, about the global salt length separating the two modes, is the one for this pull request.

dev-goalz:
@jsteube could you please review again :)

jsteube:
Was already in the pipe, see squash message

PR #4665, opened 2026-04-11
tunings: add Apple Silicon Vec:2 tuning for 7 hash modes

bandrel:
## Motivation

Apple M-series GPUs have no entries in the hashcat tuning database, so all hash modes default to Vec:1. Testing on M3 Max shows Vec:2 provides a statistically significant throughput improvement for several hash modes in attack mode 3 (brute-force/mask).

## Changes

- Add `ALIAS_Apple_M` device alias group covering M1-M4 variants (base, Pro, Max, Ultra) to `tunings/Alias.hctune`
- Add Vec:2 tuning entries for 7 hash modes on `ALIAS_Apple_M` in `tunings/Modules_default.hctune`, scoped to attack mode 3

## Benchmarks

Tested on Apple M3 Max (40 GPU cores), macOS, OpenCL backend, optimized kernel. **30 interleaved trials per configuration**, Welch's t-test for significance (all p < 0.01).

| Mode | Algorithm | Vec:1 Mean (MH/s) | Vec:1 SD | Vec:2 Mean (MH/s) | Vec:2 SD | Improvement | t-stat |
|------|-----------|-------------------|----------|-------------------|----------|-------------|--------|
| 18700 | Java Object hashCode() | 84,544 | 2,600 | 109,639 | 4,158 | **+29.7%** | 28.03 |
| 11 | Joomla < 2.5.18 | 18,043 | 73 | 20,300 | 169 | **+12.5%** | 67.35 |
| 22 | Juniper NetScreen/SSG (ScreenOS) | 10,428 | 88 | 11,261 | 474 | **+8.0%** | 9.15 |
| 900 | MD4 | 31,976 | 109 | 34,135 | 174 | **+6.8%** | 57.56 |
| 0 | MD5 | 18,135 | 141 | 19,775 | 403 | **+6.6%** | 21.03 |
| 1000 | NTLM | 32,139 | 203 | 34,121 | 214 | **+6.6%** | 36.86 |
| 1100 | Domain Cached Credentials (DCC), MS Cache | 10,054 | 51 | 10,571 | 111 | **+5.1%** | 23.08 |

Additional modes tested with 30 interleaved trials but excluded (improvement <3% or not statistically significant): modes 10, 20, 30, 40, 200, 2400, 2410, 3800, 4800. Modes without existing NVIDIA Vec entries were also excluded per developer feedback to avoid fake vectorization issues.

Tuning entries are scoped to attack mode 3 (brute-force/mask) to match how existing NVIDIA entries are structured.

## Trade-offs

- Benchmarked on M3 Max only. The M1-M4 family shares the same GPU architecture, so Vec:2 is expected to benefit all variants, but only M3 Max is confirmed.
- Accel and Loops are left at `A` (auto) so the autotune system still controls those parameters.
- Scoped to attack mode 3 only. Other attack modes are unaffected.
- Modes not listed above remain at Vec:1.

## Test plan

- `hashcat -b -m 1000` on Apple M-series should show `Vec:2`
- `hashcat -b -m 100` (SHA1) should still show `Vec:1` (no tuning entry added for modes that regressed)

### DIFF
--- a/tunings/Alias.hctune
+++ b/tunings/Alias.hctune
@@ -459,3 +459,24 @@ Intel(R)_Arc(TM)_Pro_B50_Graphics               ALIAS_INTEL
 Intel(R)_Arc(TM)_Pro_B60_Graphics               ALIAS_INTEL
 
 Intel(R)_Data_Center_GPU_Max_1100               ALIAS_INTEL
+
+##
+## Apple Silicon
+##
+
+Apple_M1                                        ALIAS_Apple_M
+Apple_M1_Pro                                    ALIAS_Apple_M
+Apple_M1_Max                                    ALIAS_Apple_M
+Apple_M1_Ultra                                  ALIAS_Apple_M
+Apple_M2                                        ALIAS_Apple_M
+Apple_M2_Pro                                    ALIAS_Apple_M
+Apple_M2_Max                                    ALIAS_Apple_M
+Apple_M2_Ultra                                  ALIAS_Apple_M
+Apple_M3                                        ALIAS_Apple_M
+Apple_M3_Pro                                    ALIAS_Apple_M
+Apple_M3_Max                                    ALIAS_Apple_M
+Apple_M3_Ultra                                  ALIAS_Apple_M
+Apple_M4                                        ALIAS_Apple_M
+Apple_M4_Pro                                    ALIAS_Apple_M
+Apple_M4_Max                                    ALIAS_Apple_M
+Apple_M4_Ultra                                  ALIAS_Apple_M
--- a/tunings/Modules_default.hctune
+++ b/tunings/Modules_default.hctune
@@ -190,6 +190,8 @@ ALIAS_INTEL                                     3       23700   2       A       
 ALIAS_INTEL                                     3       24700   2       A       A
 ALIAS_INTEL                                     3       99999   4       A       A
 
+ALIAS_Apple_M                                   *       1000    2       A       A
+
 ##
 ## CryptoAPI
 ##
--- a/tunings/Modules_default.hctune
+++ b/tunings/Modules_default.hctune
@@ -190,7 +190,13 @@ ALIAS_INTEL                                     3       23700   2       A       
 ALIAS_INTEL                                     3       24700   2       A       A
 ALIAS_INTEL                                     3       99999   4       A       A
 
+ALIAS_Apple_M                                   *       0       2       A       A
+ALIAS_Apple_M                                   *       11      2       A       A
+ALIAS_Apple_M                                   *       22      2       A       A
+ALIAS_Apple_M                                   *       900     2       A       A
 ALIAS_Apple_M                                   *       1000    2       A       A
+ALIAS_Apple_M                                   *       1100    2       A       A
+ALIAS_Apple_M                                   *       2600    2       A       A
 
 ##
 ## CryptoAPI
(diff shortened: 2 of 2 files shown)


bandrel:
Feel free to drop the MD files. I included them for transparency on the process.

jsteube:
In your "Alias.hctune" the new group names every chip one at a time:

```
Apple_M1                                        ALIAS_Apple_M
Apple_M1_Pro                                    ALIAS_Apple_M
Apple_M1_Max                                    ALIAS_Apple_M
Apple_M1_Ultra                                  ALIAS_Apple_M
```

But master moved in the meanwhile and reaches the same alias from the device name instead, in tuning_db_vendor_alias() in "tuningdb.c", and the comment above it gives the reason:

```
  // Two runtimes do not say whose silicon they are driving. Metal reports Apple for every device it
  // has, including a Radeon in an Intel Mac, and Mesa reports itself. Both still put the vendor in the
  // device name, and a handful of vendor prefixes is a far smaller thing to keep current than a list
  // of every card that vendor ever shipped.

  if (strncmp (device_name, "NVIDIA ", 7) == 0) return "ALIAS_NV";
  if (strncmp (device_name, "AMD ",    4) == 0) return "ALIAS_AMD";
  if (strncmp (device_name, "Intel",   5) == 0) return "ALIAS_INTEL";
```

Would you take the same route here? One line beside those three covers the 16 entries and the next chip as well, and a Radeon in an Intel Mac still misses it, because Metal names that one AMD Radeon rather than Apple M:

```
  if (strncmp (device_name, "Apple M", 7) == 0) return "ALIAS_Apple_M";
```

The entries in "Modules_default.hctune" then stay exactly as they are. That is the one thing I would want changed before this goes in.

bandrel:
Done, that's a better fit. Dropped the `Alias.hctune` block entirely and added the one line next to the other three in `tuning_db_vendor_alias()`:

```c
if (strncmp (device_name, "NVIDIA ", 7) == 0) return "ALIAS_NV";
if (strncmp (device_name, "AMD ",    4) == 0) return "ALIAS_AMD";
if (strncmp (device_name, "Intel",   5) == 0) return "ALIAS_INTEL";
if (strncmp (device_name, "Apple M", 7) == 0) return "ALIAS_Apple_M";
```

The `Modules_default.hctune` entries are unchanged apart from where they sit, since master moved that file in the meantime. The MD files are gone too. Rebased onto current master and squashed to one commit: +1 in `tuningdb.c`, +8 in `Modules_default.hctune`.

I re-measured everything from scratch rather than reusing the original numbers. M3 Max, 15 rounds per mode, vector width forced to 1 then 2 back to back inside each round so drift cancels out, `-b -O` throughout since like the existing vendor rows these only apply to optimized kernels. Both backends measured separately: Metal auto-disables as a duplicate of the OpenCL device, so it needs `--backend-ignore-opencl` to get a look in. Median of 15, run-to-run CV under 3%:

| mode | | OpenCL | Metal |
|---|---|---|---|
| 0 | MD5 | +8.7% | +10.2% |
| 11 | Joomla < 2.5.18 | +13.1% | +14.0% |
| 22 | Juniper NetScreen/SSG | +11.8% | +12.8% |
| 900 | MD4 | +6.9% | +5.6% |
| 1000 | NTLM | +9.5% | +9.1% |
| 1100 | DCC, MS Cache | +4.0% | +3.0% |
| 18700 | Java Object hashCode() | +30.9% | +35.6% |

Vec:2 won all 15 rounds in every mode on both backends.

I also checked the alias really does resolve from the device name with nothing in `Alias.hctune`: mode 1400, which has no row, stays at `Vec:1` on the same device, so the rows above are what moves these seven rather than a wildcard fallback.

Issue #3816, opened 2023-08-04
multiple same-salt descrypt hashes with impossible time and speed


roycewilliams:
Symptom:

For example, running -a 3 -i ?a?a?a?a?a?a?a?a against 2 or more artificial/test descrypt hashes, all with the same salt such as '00', the x8 ?a sequence claims to have completed in less than a minute.

In this example, we are artificially using a very high descrypt hash count (20 million hashes), but the same symptom happens with only 2 hashes. With only a single hash, the speed/behavior is normal as expected.

```
Session..........: desdigit00000.20230803_193719_DDm2NY
Status...........: Running
Hash.Mode........: 1500 (descrypt, DES (Unix), Traditional DES)
Hash.Target......: /var/local/hashbench/targets//desdigit00000/unfound.hashes
Time.Started.....: Thu Aug  3 19:40:25 2023 (42 secs)
Time.Estimated...: Thu Aug  3 19:41:32 2023 (25 secs)
Kernel.Feature...: Pure Kernel
Guess.Mask.......: ?a?a?a?a?a?a?a?a [8]
Guess.Queue......: 8/8 (100.00%)
Speed.#1.........: 20293.3 GH/s (0.03ms) @ Accel:512 Loops:1024 Thr:512 Vec:1
Speed.#2.........: 18843.5 GH/s (0.03ms) @ Accel:512 Loops:1024 Thr:512 Vec:1
Speed.#3.........: 18630.8 GH/s (0.03ms) @ Accel:512 Loops:1024 Thr:512 Vec:1
Speed.#4.........: 20530.8 GH/s (0.03ms) @ Accel:512 Loops:1024 Thr:512 Vec:1
Speed.#5.........: 20376.2 GH/s (0.03ms) @ Accel:512 Loops:1024 Thr:512 Vec:1
Speed.#*.........: 98674.6 GH/s
Recovered........: 0/20000000 (0.00%) Digests (total), 0/20000000 (0.00%) Digests (new)
Remaining........: 20000000 (100.00%) Digests
Recovered/Time...: CUR:N/A,N/A,N/A AVG:N/A,N/A,N/A (Min,Hour,Day)
Progress.........: 4060518594641920/6634204312890625 (61.21%)
Rejected.........: 0/4060518594641920 (0.00%)
Restore.Point....: 4702863360/7737809375 (60.78%)
Restore.Sub.#1...: Salt:0 Amplifier:691200-692224 Iteration:0-1024
[...]

PR #4666, opened 2026-04-11
fix(OpenCL): implement m01500_mxx for DESCRYPT_SALT path

raulperdomo:
When brute-force attacking mode 1500 (descrypt) with two or more hashes that share the same DES salt, hashcat takes the DESCRYPT_SALT bitslice path (salt baked into kernel at JIT time) and dispatches the mxx kernel for multi-digest comparison. That kernel was an empty stub, so no hashes were ever recovered.

Fill in m01500_mxx in the #else DESCRYPT_SALT block using the same key setup and DESCrypt call as the adjacent sxx kernel, then reuse the dynamic multi-digest comparison loop already present in the #ifndef DESCRYPT_SALT mxx kernel (out[64] + per-digest bitslice XOR).

## Bug

When brute-force/mask attacking mode 1500 (descrypt) with two or more
hashes that share the same DES salt, hashcat recovers 0 hashes.

### Root cause

`OpenCL/m01500_a3-pure.cl` has two kernel variants gated on whether
`DESCRYPT_SALT` is defined at JIT time. `DESCRYPT_SALT` is set in
`src/modules/module_01500.c` when all three conditions hold:

  - attack mode is brute-force / mask (-a 3)
  - all hashes share exactly one salt  (salts_cnt == 1)
  - --slow-candidates is off

When those conditions are met and the hash file contains **two or more
digests** under that single salt, the multi-digest kernel `m01500_mxx`
is dispatched, but it was an empty stub, so no comparison was ever
performed and no hashes were ever recovered.

## Fix

Implemented `m01500_mxx` inside the `#else DESCRYPT_SALT` block using:

- The same key expansion and `DESCrypt()` call as the adjacent `sxx`
  kernel (compile-time-baked salt, bitslice key bits from w0/w1).
- The same dynamic multi-digest comparison loop already present in the
  `#ifndef DESCRYPT_SALT` `mxx` kernel: DES output stored in `out[64]`,
  then for each target digest a bitslice XOR is performed, with a fast
  `DIGESTS_CNT < 16` branch and a `transpose32c` path for larger sets.

## Reproduction

Test vectors:
Two passwords hashed with the same DES salt `ab`
(generated with Python 3 `crypt.crypt('<password>', 'ab')`):

```
abJnggxhB/yWI   (password: password)
ab9j1qG1p7TMk   (password: hashcat)
```

**Before fix** — correct password in candidate list, 0 recovered:
```
$ hashcat -m 1500 hashes.txt -a 3 password --potfile-disable
Recovered........: 0/2 (0.00%) Digests (total)
```

**After fix** — both hashes cracked:
```
$ hashcat -m 1500 hashes.txt -a 0 wordlist.txt --potfile-disable
abJnggxhB/yWI:password
ab9j1qG1p7TMk:hashcat
Recovered........: 2/2 (100.00%) Digests (total)
```

### DIFF
--- a/OpenCL/m01500_a3-pure.cl
+++ b/OpenCL/m01500_a3-pure.cl
@@ -2690,6 +2690,375 @@ KERNEL_FQ KERNEL_FA void m01500_sxx (KERN_ATTR_BITSLICE ())
 
 KERNEL_FQ KERNEL_FA void m01500_mxx (KERN_ATTR_BITSLICE ())
 {
+  /**
+   * base
+   */
+
+  const u64 gid = get_global_id (0);
+  const u64 lid = get_local_id (0);
+
+  if (gid >= GID_CNT) return;
+
+  /**
+   * salt
+   */
+
+  const u32 salt = salt_bufs[SALT_POS_HOST].salt_buf[0];
+
+  /**
+   * base
+   */
+
+  const u32 w0 = pws[gid].i[0];
+  const u32 w1 = pws[gid].i[1];
+
+  const u32 w0s = (w0 << 1) & 0xfefefefe;
+  const u32 w1s = (w1 << 1) & 0xfefefefe;
+
+  #define K00 (((w0s >> ( 0 + 7)) & 1) ? -1 : 0)
+  #define K01 (((w0s >> ( 0 + 6)) & 1) ? -1 : 0)
+  #define K02 (((w0s >> ( 0 + 5)) & 1) ? -1 : 0)
+  #define K03 (((w0s >> ( 0 + 4)) & 1) ? -1 : 0)
+  #define K04 (((w0s >> ( 0 + 3)) & 1) ? -1 : 0)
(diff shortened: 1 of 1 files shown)


raulperdomo:
I just realized this bug was also mentioned here:
https://github.com/hashcat/hashcat/issues/3816

jsteube:
Merged, thanks, and sorry this sat for so long.

The stub was mine. I split the kernels into two variants there and stubbed out the one I thought was unreachable in each branch. That is true for sxx under #ifndef DESCRYPT_SALT, but not for mxx, because DESCRYPT_SALT only requires salts_cnt == 1 and does not require digests_cnt == 1.

Confirmed on an RX 9070 XT before merging. Two hashes sharing one salt gave 0 of 2 and exhausted with the old kernel, and 2 of 2 with your fix.

I will commit a small follow up to remove the duplicate, so you are not surprised when the code disappears again. The mxx implementation you added is identical to the one already in the #ifndef branch, and DESCrypt handles the baked salt behind its own #ifdef, so a single copy outside the guard serves both builds.

PR #4667, opened 2026-04-14
Add module 1732: MSSQL (2025)

VladDBA:
# Add hash mode for Microsoft SQL Server 2025 (MSSQL 2025) sql login hashes

Module 1732 adds support for cracking SQL Server 2025's new sql login password hashes 
SQL Server 2025 introduces a new login password algorithm: PBKDF2-HMAC-SHA512 with 100k iterations (fixed iteration count).

More info can be found in [Secure by default: What’s new in SQL Server 2025 security](https://techcommunity.microsoft.com/blog/sqlserver/secure-by-default-what%E2%80%99s-new-in-sql-server-2025-security/4424340).

Pure T-SQL implementation of the hashing algorithm can be found [here](https://github.com/VladDBA/Blog/tree/main/SQL_Server_2025_PBKDF2_T-SQL_implementation) with an in-depth explanation and test posted [here](https://vladdba.com/2025/11/02/replicating-sql-server-2025-pbkdf2-hashing-algorithm-using-t-sql/).

## Algorithm parameters:

- PRF: HMAC-SHA512 (produces a 64-byte/512-bit output)
- Iterations: Fixed at 100,000 (not stored in the hash, hard-coded in the module as MSSQL2025_ITER)
- Salt: 4 bytes, embedded in the hash string
- Output (DK): 64 bytes

## Hash format

```
0x0300 <4-byte hex salt> <64-byte hex derived key>
```
- token[0] = "0x0300" (version signature)
- token[1] = salt     (8 hex chars = 4 bytes)
-  token[2] = hash     (128 hex chars = 64 bytes)

Example (self-test vector, password "hashcat"):

```
0x03004DB403DDD3A508F81058E6AD92A6A20E5C72A06AD0B00C0A733EAD6582BE2443AFA70C9516C2DFDC60EA927B40BD732D7763477484FC014435AAC4D57EA88237B39806C1
```

### test.sh single 

```
./tools/test.sh -m 1732 -a 0 -t single
[ test_1776152573 ] > Init test for hash type 1732.
[ test_1776152573 ] [ Type 1732, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1776152573 ] [ Type 1732, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0
```

### Benchmark (self-test inside hashcat)

```
./hashcat -m 1732 -b -d 1
---------------------------------------------------
* Hash-Mode 1732 (MSSQL (2025)) [Iterations: 99999]
---------------------------------------------------

Speed.#01........:    11914 H/s (79.84ms) @ Accel:4 Loops:500 Thr:1024 Vec:1
```

### Mask attack self-test

```
./hashcat -m 1732 -a 3 '0x03004DB403DDD3A508F81...39806C1' 'hashc?a?a'
Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 1732 (MSSQL (2025))
Hash.Target......: 0x03004DB403DDD3A508F81058E6AD92A6A20E5C72A06AD0B00...9806C1
Time.Started.....: Tue Apr 14 11:54:05 2026 (2 secs)
Time.Estimated...: Tue Apr 14 11:54:07 2026 (0 secs)
Speed.#01........:     4553 H/s (2.38ms) @ Accel:4 Loops:124 Thr:512 Vec:1
Speed.#02........:        0 H/s (0.00ms) @ Accel:45 Loops:1000 Thr:1 Vec:4
Speed.#*.........:     4553 H/s
Progress.........: 9025/9025 (100.00%)
```

### Complied binaries

Module 1732 binaries compiled for hashcat 7.1.2 can be found [here](https://github.com/VladDBA/Blog/tree/main/hashcat_mssql2025)

### DIFF
--- a/OpenCL/m01732-pure.cl
+++ b/OpenCL/m01732-pure.cl
@@ -1,0 +1,357 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha512.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct pbkdf2_sha512_tmp
+{
+  u64  ipad[8];
+  u64  opad[8];
+
+  u64  dgst[16];
+  u64  out[16];
+
+} pbkdf2_sha512_tmp_t;
+
+typedef struct pbkdf2_sha512
+{
+  u32 salt_buf[64];
+
+} pbkdf2_sha512_t;
+
+DECLSPEC void hmac_sha512_run_V (PRIVATE_AS u32x *w0, PRIVATE_AS u32x *w1, PRIVATE_AS u32x *w2, PRIVATE_AS u32x *w3, PRIVATE_AS u32x *w4, PRIVATE_AS u32x *w5, PRIVATE_AS u32x *w6, PRIVATE_AS u32x *w7, PRIVATE_AS u64x *ipad, PRIVATE_AS u64x *opad, PRIVATE_AS u64x *digest)
+{
+  digest[0] = ipad[0];
+  digest[1] = ipad[1];
+  digest[2] = ipad[2];
+  digest[3] = ipad[3];
+  digest[4] = ipad[4];
+  digest[5] = ipad[5];
+  digest[6] = ipad[6];
+  digest[7] = ipad[7];
+
+  sha512_transform_vector (w0, w1, w2, w3, w4, w5, w6, w7, digest);
+
+  w0[0] = h32_from_64 (digest[0]);
+  w0[1] = l32_from_64 (digest[0]);
+  w0[2] = h32_from_64 (digest[1]);
+  w0[3] = l32_from_64 (digest[1]);
+  w1[0] = h32_from_64 (digest[2]);
+  w1[1] = l32_from_64 (digest[2]);
+  w1[2] = h32_from_64 (digest[3]);
+  w1[3] = l32_from_64 (digest[3]);
+  w2[0] = h32_from_64 (digest[4]);
+  w2[1] = l32_from_64 (digest[4]);
+  w2[2] = h32_from_64 (digest[5]);
+  w2[3] = l32_from_64 (digest[5]);
+  w3[0] = h32_from_64 (digest[6]);
+  w3[1] = l32_from_64 (digest[6]);
+  w3[2] = h32_from_64 (digest[7]);
+  w3[3] = l32_from_64 (digest[7]);
+  w4[0] = 0x80000000;
+  w4[1] = 0;
+  w4[2] = 0;
+  w4[3] = 0;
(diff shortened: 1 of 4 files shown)


jmvermeulen:
Small thing on the `tmps` struct:

```c
typedef struct pbkdf2_sha512_tmp {
  u64 ipad[8]; u64 opad[8];
  u64 dgst[16]; u64 out[16];
} pbkdf2_sha512_tmp_t;
```

`dgst` and `out` are sized for two PBKDF2 output blocks (2 × 64 bytes), but the MSSQL 2025 derived key is fixed at 64 bytes, which is exactly one HMAC SHA512 output. That means only the first 8 entries of each array ever get written or read. The remaining 8 entries sit in the tmps buffer unused for the entire run.

The tmps buffer is allocated per concurrent GPU thread and lives in global memory for the full duration of the slow hash loop. With 128 bytes of dead space per thread and typical accel times loops counts on modern GPUs, that adds up to tens of megabytes of VRAM that the auto tuner has to account for when picking accel and loops values. On memory bound PBKDF2 loops this can quietly cost a step of headroom in the tuner, even if the kernel itself runs identically.

Trimming both arrays to `[8]` is a one line change and lines up exactly with what the format guarantees.

I know this shape is cloned from the generic PBKDF2 SHA512 tmps used by m7100 and m20011, where longer DKs are allowed and the extra space is actually needed. Since MSSQL 2025 bakes the 64 byte output into the hash format, there is no scenario where the second block gets used, so the extra space is pure overhead here.

VladDBA:
Hi @jmvermeulen,

Thanks for pointing that out.
I've updated `dgst` and `out` in both the module and the kernel.

jsteube:
The algorithm itself is not in question here, only where it sits. MSSQL modes hang off the primitive rather than off each other, and they carry no kernel of their own:

```
  130   sha1(utf16le($pass).$salt)      owns the kernel
  131   MSSQL (2000)                    KERN_TYPE=130, hash loader only
  132   MSSQL (2005)                    KERN_TYPE=130, hash loader only
  1730  sha512(utf16le($pass).$salt)    owns the kernel
  1731  MSSQL (2012, 2014)              KERN_TYPE=1730, hash loader only
```

1732 would break that, because the 17xx block is raw SHA-512 and everything in it runs inside the kernel, where this is a slow KDF. What is missing is the primitive itself: 12100 is PBKDF2-HMAC-SHA512 with no utf16le, and there is no utf16le counterpart for MSSQL 2025 to hang off.

Could you split it in two, the way 1730 and 1731 are split?

```
  36600  PBKDF2-HMAC-SHA512(utf16le($pass))   your kernel, renamed m36600-pure.cl
  36601  MSSQL (2025)                         KERN_TYPE=36600, your parser, no kernel
```

Both numbers are free. The kernel carries over unchanged, and 36600 then exists for everyone else who meets this construction.

That leaves 36600 without a hash format, since the only format in the PR is the MSSQL one and that belongs to 36601. Take the shape from 12100, which is the same construction minus the utf16le:

```
  12100  sha512:1000:NzY2:DNWohLbdIWIt4Npk9gpTvA==
```

Signature, iterations, base64 salt, base64 digest. The signature has to be a new one because 12100 already holds "sha512". 1730 is the model for how the two modes sit next to each other, but not for the format: it is a raw hash with nowhere to put an iteration count, and a PBKDF2 mode needs one in the hash.

A test module for each mode finishes the pair, "tools/test_modules/m36600.pm" and "m36601.pm", with m12100.pm and m01731.pm as the nearest references. We are moving these from Perl to Python, so a .py is equally welcome if you prefer that.

One last thing: the module needs a pass over the API drift since 7.1.2, which is written up in "docs/hashcat-plugin-development-guide.md" under "Porting a plugin from 7.1.2". A line in "docs/changes.txt" for each of the two modes and it is done.

VladDBA:
Hi @jsteube,
Sure. I'll take care of it this upcoming week.

PR #4676, opened 2026-05-08
Added hash-mode: Meshtastic LoRa frame (AES-128-CTR PSK)

alphafox02:
Adds the Meshtastic LoRa protocol's encrypted-frame channel PSK as a hashcat hash-mode. Each frame is AES-128-CTR encrypted with a 16-byte channel PSK; the protocol provides no MAC, so verification is structural — decrypt the first AES block and check the protobuf shape of the resulting Meshtastic Data envelope (portnum/payload tags + a small trailing-tag whitelist for the optional fields that follow).

The kernel also iterates a small list of common channel names (LongFast, MediumFast, etc.) when the hash file's name field is empty, so unknown-name attacks work without the user having to specify the channel name.

Validated with the embedded self-test, a .pm round-trip on 200 random PSKs, and a real-radio capture cracked end-to-end on RTX 3060 Mobile. Sustained 17.7 GH/s on a 4.3-billion-candidate mask attack; false-positive rate ≈ 1 in 13B with the included verifier.

Hash-mode 99001 is a dev-range placeholder — happy to take whatever number you assign.

### DIFF
--- a/OpenCL/m99001_a0-pure.cl
+++ b/OpenCL/m99001_a0-pure.cl
@@ -16,19 +16,28 @@
 #include M2S(INCLUDE_PATH/inc_cipher_aes.cl)
 #endif
 
-#define MESHTASTIC_NAME_MAX 64
-#define MESHTASTIC_CT_MAX   256
+#define MESHTASTIC_NAME_MAX   64
+#define MESHTASTIC_CT_MAX     256
+#define MESHTASTIC_FRAMES_MAX 16
+
+typedef struct meshtastic_frame
+{
+  u32 packet_id;
+  u32 from_node;
+  u32 ct_len;
+  u32 ct_buf[MESHTASTIC_CT_MAX / 4];
+
+} meshtastic_frame_t;
 
 typedef struct meshtastic
 {
   u32 chash;
   u32 name_xor;
-  u32 packet_id;
-  u32 from_node;
   u32 name_len;
-  u32 ct_len;
+  u32 nframes;
   u32 name_buf[MESHTASTIC_NAME_MAX / 4];
-  u32 ct_buf[MESHTASTIC_CT_MAX / 4];
+
+  meshtastic_frame_t frames[MESHTASTIC_FRAMES_MAX];
 
 } meshtastic_t;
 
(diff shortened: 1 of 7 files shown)


alphafox02:
FYI: I'll be publishing the companion IQ-to-hash extraction tool soon as I test one other SDR besides the hackRF.

dev-goalz (commented):
Inline version of the AES-256 patch from point 1 (the collapsed comment was hard to read). These cover the shared `meshtastic_verify` and the straight (`a0`) + mask (`a3`) kernels (both the `mxx` and `sxx` paths). Combinator (`a1`) is left out — its 16-byte `switch_buffer_by_offset_le_S` assembly would need to widen to 8 words, and it's an unlikely mode against a 256-bit key.

Verified test vector (offline; I haven't GPU-built this) — PSK = the 32 bytes `0123456789abcdef0123456789abcdef`, channel `LongFast`, `packet_id = from_node = 0xdeadbeef`, payload `hashcat`:

```
$meshtastic$1*0a*efbeadde*efbeadde*4c6f6e6746617374*e9886af9b78d2bdd9c52f492cc999e04
```

A 32-byte-PSK self-test vector for `m99001.pm` (generalise `_pad16` to 16-or-32) would cover the new path, and the AES-256 schedule's larger register footprint is worth a quick before/after benchmark — gate it behind a `-D` JIT option if it costs AES-128 throughput.

alphafox02:
Thanks @j0b-dev, really appreciate the expert feedback, this is new
territory for me and your AES-256 test vector saved me a lot of grinding.

The pw_max/AES-128 mismatch is clearly a real bug. I'd also like to
do my best to address the false-positive concern with a multi-frame
$meshtastic$2 format. I'll research the cleanest way to do each piece
and push when it's tested.

alphafox02:
Pushed the two responses on top of v1:

ca09f37 - AES-256 across all three attack-mode kernels. Your test
vector is now the self-test ST_HASH so it runs on evey startup.

2d215ec - $meshtastic$2 multi-frame format. The kernel builds the AES
schedule once per candidate and requires every frame in the line to
decrypt before reporting a crack. v1 hot path is unchanged.

Sniffer side gained --hashcat-export-merge=N for v2 emission
(meshtastic-sniffer @ 4d796eb on main). Happy to revise based on
furhter feedback — thanks again.

jsteube:
> Meshtastic PSK length selects the cipher: <=16 byte candidate -> AES-128 (zero-padded to 16), 17..32 byte candidate -> AES-256 (zero-padded to 32). Anything longer is truncated to 32 by the kernel.

The zero padding is where I would start, because channel.proto rules it out:

```
Must be either 0 bytes (no crypto), 16 bytes (AES128), or 32 bytes (AES256).
A special shorthand is used for 1 byte long psks.
```

A 7 byte candidate padded out to 16 is a key no node can hold, and that goes for every length from 1 to 15 and from 17 to 31. It is almost the entire space the kernel accepts, so a mask attack over anything shorter than 16 characters covers no reachable key, whatever it reports along the way.

Could you set module_pw_min to 16 and reject every length except 16 and 32? What is left after that is the part of the keyspace that exists, a 16 or 32 character key somebody typed instead of generating. The 1 byte shorthand needs no GPU, being 11 fixed keys with the default one sitting in the protobuf source.

One more in the same area. The function meshtastic_chash_match() falls back to a list of 13 preset channel hashes when the name is absent. That is 13 times the false positives of the exact comparison, and it puts protocol data inside a kernel that then needs an edit whenever Meshtastic adds a preset. Your $meshtastic$2 format answers both, since every frame has to decrypt before anything is reported. Is the list still needed?

Then the placement. The number to use is 36700. The mode also needs an a4 kernel, "m36700_a4-pure.cl", which is what the PCFG device engine loads. In the meanwhile master moved, so this needs a rebase, and the module needs a pass over the API drift since 7.1.2, written up in "docs/hashcat-plugin-development-guide.md" under "Porting a plugin from 7.1.2".

PR #4689, opened 2026-05-25
Fix Kerberos etype 17/18 domain buffer overflow

Chick3nman:
Kerberos etype 17/18 modules use a local `u8 domain[128]` buffer to uppercase the domain name before copying to the esalt struct. The tokenizer allows domain fields up to 512 bytes, leading to 384-bytes of overflow. This can lead to crashes and other unexpected behavior for hashes with larger domain fields set. 

Crafted test hash:

```
$krb5tgs$17$srv_http$AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA$849e31b3db1c1f203fa20b85$948690f5875125348286ad3346d27b43eaabc71896b620c16de7ddcdbd561628c650c508856a3f574261948b6db4b48332d30536e978046a423ad4368f9a69b4dc4642dab4e0d475d8299be718fd6f98ac85a771b457b2453e78c9411dfce572b19660fe7a5a8246d9b2a91ea2f14d1986ea0a77ecf9b8330bc8fd9ab540bcf46b74c5aa7005cfccd89ec05f66aeab30c6b2bf8595cf6c9a1b68ad885258850c4b1dd9265f270fb2af52fd76c16246df51ea67efc58a65c345686c84e43642febe908a
```

In a linux build with DEBUG=2:

```
*** buffer overflow detected ***: terminated
Aborted (core dumped)
```

Ansible Vault encode buffer 2x undersized folded in from #4701 

Module 16900 allocates `hcmalloc(16384 + 1)` for hex-encoding `ct_data_buf[4096]`. Each u32 produces 8 hex chars, so max output is 32768 bytes, 2x larger than the buffer. Crashes on `--show` with ct_data > 8 KB. Fixed here as well.


jsteube:
Thanks, the diagnosis is right and the domain buffer fix is needed. Three things to change first:

module_32100.c and module_32200.c are converted from LF to CRLF, 441 lines each way in both files. That turns them into whole file rewrites when the real change is two lines, u8 domain[128] to u8 domain[512] and a plus one on an hcmalloc. Could you revert the line endings so the diff shows the fix. I expect this is your editor rather than anything intended.

Could you also drop the docs/changes.txt hunk, from this and from #4698. I will write the entries when the changes land.

Last, both need a rebase on current master. They conflict with master and with each other, so whichever lands first leaves the other needing another pass. Order is up to you.

PR #4691, opened 2026-05-25
Fix Kerberos etype 23 account_info overflow

Chick3nman:
Kerberos etype 23 modules compute `account_info_len` from the distance between two * delimiters in the hash string using strchr. No upper bound or limit check is enforced on this length before it's used in a memcpy into the fixed sized account_info[512] esalt field.

An input hash with an arbitrarily long `*user$realm$spn*` section causes a heap overflow. Small overflows (< ~20KB) land within the module's own struct fields and are partially overwritten by subsequent parsing. Large overflows (> ~22KB) write past the struct into adjacent allocations, corrupting heap metadata and causing a crash or unexpected behavior.

Behavior with a Kerberos etype 23 hash with 23,000-byte account_info section designed to overflow past entire 22KB struct on a native Linux build with DEBUG=2:

```
double free or corruption (out)
Aborted (core dumped)
```

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -10,6 +10,7 @@ Docker: Add hashcat-toolchain
 ## Bugs
 ##
 
+- Fixed heap overflow in Kerberos etype 23 modules from account_info field with crafted/corrupt hash input
 - Added workaround for CUDA memory leak in benchmark mode caused by register spilling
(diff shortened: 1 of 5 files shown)


Chick3nman:
Folded changes into #4698

Issue #3612, opened 2023-02-05
missing safety check allows loading of too large of lists in -m 99999


Shooter3k:
**Describe the bug**
When using -m 99999 and the "hash list" is far too large, Hashcat starts up and does not error. Instead, it runs the attack and finds 0 results 

**To Reproduce**
Run hashcat against a "hash list" with -m 99999 (literally any wordlist file that's too large for hashcat to handle) 

**Expected behavior**
Hashcat to error or tell the user something bad happened.

PR #4696, opened 2026-05-26
Fix u32 multiplication overflow causing missed cracks/false negatives on large hash lists

Chick3nman:
The digest and esalt buffer offset calculations in `hashes_init_stage2` use `u32 * u32` multiplication to compute byte offsets. When the total buffer size exceeds 2 ^ 32 bytes (4 GB), this multiplication silently overflows, causing digest entries to be written to incorrect positions in the buffer due to the wrapped value. The sorted digest array becomes corrupted and the GPU binary search silently fails to find entries at corrupted positions causing missed cracks with no error or warning. 

This is also the root cause of https://github.com/hashcat/hashcat/issues/3612 and this fixes that issue. Conveniently, 99999 is one of the easier places to spot this issue due to the larger dgst_size it uses.

The threshold number of hashes for triggering this issue is hash mode dependent as it depends on the specific mode's dgst_size. For mode 99999 the digest size is 128 which means it takes a bit over 33.5M hashes to cause the overflow. For SHA512 with dgst_size 64 it's ~67M hashes. For MD5 dgst_size is only 16 and it takes ~268M hashes. 

The wrapping behavior is seemingly deterministic and we can track the offsets as this bug corrupts the array. Hash entry at sorted position 33,554,432: offset 33554432 × 128 = 2 ^ 32 wraps to 0, overwriting position 0. Hash entries at positions 33,554,432 - 33,999,999 begin overwriting the array after position 0 and the corruption spreads until it consumes the entire array or your GPU hits an OOM due to the size of the hash list. Both the overwritten "victim" hashes and overflowed "perpetrator" hashes become uncrackable because the binary search done by the GPU fails to find them during cracking. This means the GPU will miss the hashes that "overflowed" as well as the hashes that got trampled by the overwrites.

Tested in mode 99999 with 100 test passwords distributed through the list to track the missed cracks from this issue:

Hashes loaded | Cracks with overflow | Cracks with overflow fixed
-- | -- | --
33,554,432 (exactly 2^32 bytes) | 100/100 | 100/100
34,000,000 (2^32 + 57MB) | 96/100 | 100/100
40,000,000 (2^32 + 825MB) | 47/100 | 100/100
45,000,000 (2^32 + 1.47GB) | 0/100 | 100/100

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -10,6 +10,7 @@ Docker: Add hashcat-toolchain
 ## Bugs
 ##
 
+- Fixed u32 overflow in digest/esalt buffer offset calculations causing silent missed cracks with large hashlists
 - Added workaround for CUDA memory leak in benchmark mode caused by register spilling
 - Fixed bugs on multiple DES based modules when used with CPU devices
 - Fixed MSYS2 build
--- a/src/backend.c
+++ b/src/backend.c
@@ -14176,7 +14176,7 @@ int backend_session_begin (hashcat_ctx_t *hashcat_ctx)
 
     u64 size_plains  = (u64) hashes->digests_cnt * sizeof (plain_t);
     u64 size_salts   = (u64) hashes->salts_cnt   * sizeof (salt_t);
-    u64 size_esalts  = (u64) hashes->digests_cnt * hashconfig->esalt_size;
+    u64 size_esalts  = (u64) hashes->digests_cnt * (u64) hashconfig->esalt_size;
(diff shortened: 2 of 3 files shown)


jsteube:
Closing as already fixed, but the finding here is yours: you filed this on 26 May, 71 days before #4747, and identified it as the root cause of #3612. #4747 landed first as 62dcbb453 and carries the same fix plus one line in src/brain.c. That was my sequencing mistake, not a judgement on this PR. The changelog entry for the release will use your description.

PR #4698, opened 2026-05-27
Fix 0x80 padding byte overflow causing crashes/corrupt output

Chick3nman:
Many hash modules append a 0x80 padding byte after decoded data for MD5/SHA padding requirements. When the data fills the destination buffer to exact maximum capacity, the 0x80 is written one byte past the buffer, overwriting the first byte of the adjacent struct field.

On most systems this corrupts the length field, causing hash_encode to read past the buffer during --show/--left operations. Depending on the module, this produces either a crash or corrupted output.

Test hash for module 13100 with edata2 at exactly 40960 hex chars, 20480 decoded bytes, fills edata2 exactly, linux build and DEBUG=2:

```
double free or corruption (!prev)
Aborted (core dumped)
```

Test hash for 5300 with nr field at exactly 128 hex chars, 64 decoded bytes, fills nr_buf, using --show produces output containing leaked data from adjacent msg_buf field:

```
...bbbbbbbbbbbbbbbb8000000050503326cac6e4bd...
```

The 80 is the padding byte, followed by data from the next struct field.

Also fixes account_info overflow, folded changes in from #4691

Kerberos etype 23 modules parse account_info from user-controlled input via strchr without bounds checking. A crafted or corrupt hash with a long `*user$realm$spn*` section overflows the `account_info[512]` u32 array.


jsteube:
Thanks, this one looks good and nothing needs changing in the fixes.

Two things, both also raised on #4689 so you only have to do them once. Could you drop the docs/changes.txt hunk. And this needs a rebase on current master, since it conflicts with master and with #4689.

PR #4701, opened 2026-05-29
Fix off by 1, encode buffer overflow causing crashes with --show

Chick3nman:
The encode functions in several modules allocate buffers that are too small for the hex encoded output they produce, causing buffer overflows during --show/--left/potfile operations with hashes that have large data fields.

Module 16900 allocates `hcmalloc(16384 + 1)` for hex encoding ct_data_buf[4096]. Each u32 produces 8 hex chars via u32_to_hex, so max output is 4096 * 8 = 32768 bytes. The allocation is 2x undersized. With ct_data > 8 KB, --show crashes:

```
==97236==ERROR: AddressSanitizer: heap-buffer-overflow in u32_to_hex
src/modules/module_16900.c:251
0x529000013201 is located 0 bytes to the right of 16385-byte region
```

Kerberos etype 17/18 modules allocate `hcmalloc(5120 * 4 * 2)` = 40960 bytes for hex encoding edata2. At max edata2 (20480 bytes = 40960 hex chars), snprintf writes a null terminator 1 byte past the buffer:

```
==97273==ERROR: AddressSanitizer: heap-buffer-overflow
src/modules/module_19600.c:271
0x52e00000a400 is located 0 bytes to the right of 40960-byte region
```

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -13,6 +13,8 @@ Docker: Add hashcat-toolchain
 - Added workaround for CUDA memory leak in benchmark mode caused by register spilling
 - Fixed bugs on multiple DES based modules when used with CPU devices
 - Fixed MSYS2 build
+- Fixed encode buffer overflow in module 16900 causing crashes when --show
+- Fixed encode off by one in Kerberos etype 17/18 modules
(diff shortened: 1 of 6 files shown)


Chick3nman:
Folded these changes into #4689

PR #4703, opened 2026-06-01
Add wordlist seek-index cache (.hcidx) to accelerate large --skip

thatux:
# Wordlist seek index (`.hcidx`) — accelerate large `--skip`

## Problem

For dictionary attacks, `-s` / `--skip` is a candidate-index skip, not a byte
offset. To reach skip position **X**, hashcat reads and counts lines from the
start of the file. With multi-GB wordlists and large skips (e.g.
`-s 4000000000000`), this pre-seek takes a long time before cracking even
begins.

This is especially costly under **hashtopolis**: the same wordlist is chunked
into sequential assignments (`-s 0 -l N`, `-s N -l N`, …) and handed to many
agents. The seek cost is paid repeatedly, per chunk, per agent, on the same
static file. Tracking issue: [hashcat#1644].

## Idea

Maintain a **sparse, on-disk index** mapping `word_index → byte_offset`, with
one checkpoint every `K` base words (`K = 1,000,000` by default). On a request
to skip to **X**, binary-search the index for the largest checkpoint `≤ X`,
`fseek()` directly to that byte offset, then walk forward at most `K − 1` lines
to land exactly on **X**.

Seek cost goes from **O(X)** to **O(K)**, bounded — independent of how large
the skip is.

The index stays tiny: a 1 G-word list at `K = 1 M` produces ~1000 entries,
around 30 KB of JSON.

## Results (measured)

100 M-word wordlist (~1.1 GiB), `-m 0 -a 0 -s 99999990 -l 20`, MD5, PoCL CPU
OpenCL, kernel cache pre-warmed:

| Run                  | Wall time | User    | Sys    |
| -------------------- | --------- | ------- | ------ |
| Without `.hcidx`     | 18.444 s  | 3.690 s | 1.391 s |
| With `.hcidx`        |  2.414 s  | 0.659 s | 0.234 s |
| **Speedup**          | **7.6×**  | 5.6×    | 5.9×   |

30 M-word list, `--slow-candidates` path: 13.559 s → 6.357 s (~2.1×). The
fast-candidates speedup is larger because the linear walk was the dominant
cost there; with slow candidates a CPU-side rule loop adds fixed overhead.

## CLI

* `--wordlist-index-cache` *(no path)*: **build mode** — build the `.hcidx` as
  a side effect of the run. Gated on:
  * file size ≥ `HCIDX_MIN_FILE_SIZE` (256 MiB),
  * `-s` or `-l` set (i.e. this is a chunked run where the index pays off),
  * straight attack (`-a 0`).
  Below the size gate, the linear seek is already cheap and the write would
  just be overhead.
* `--wordlist-index-cache=PATH`: **use mode** — load the index from `PATH`.
  No build.
* `--wordlist-index-cache-sha256`: include a SHA-256 of the wordlist in the
  `.hcidx` header. Optional; costs one full read of the wordlist at build time.

If no flag is given but a `.hcidx` exists next to the wordlist, hashcat loads
it transparently. This is the desired hashtopolis behaviour: ship the index
once with the wordlist, every agent benefits without a flag change.

## File format (JSON)

The `.hcidx` is written **next to the wordlist** as `<wordlist>.hcidx` — same
directory, same basename, plus the `.hcidx` suffix. Always a plain text JSON
document. Size scales sparsely:

| Wordlist size       | Entries | `.hcidx` size |
| ------------------- | ------- | ------------- |
| 30 M words / ~300 MB | 30      | ~ 900 B       |
| 100 M words / ~1.1 GB| 100     | ~ 2.8 KB      |
| 1 B words / ~10 GB  | 1 000   | ~ 30 KB       |
| 10 B words / ~100 GB| 10 000  | ~ 300 KB      |

Example (100 M-word list, abbreviated):

```json
{
  "version": 1,
  "flags": 2,
  "interval": 1000000,
  "word_count": 100000000,
  "file_size": 1100000000,
  "sha256": null,
  "entries": [
    [0, 0],
    [1000000, 11000000],
    [2000000, 22000000],
    ...
    [99000000, 1089000000]
  ]
}
```

### How to read `entries`

Each element of `entries` is a `[word_index, byte_offset]` pair — a single
**checkpoint** in the wordlist.

* `word_index` — the zero-based position of a base word in the wordlist. For
  example `[1000000, 11000000]` means: "word number 1,000,000 starts at byte
  11,000,000."
* `byte_offset` — the absolute byte offset in the wordlist file where the
  **first byte of the line** for that word lives. Suitable for direct
  `fseek(fp, byte_offset, SEEK_SET)`.

Checkpoints are spaced every `interval` (= 1 000 000) words. They are stored
in ascending order, so a binary search lands on the largest
`word_index ≤ target` in O(log N). After seeking there, the consumer walks at
most `interval − 1` lines forward to reach the exact target word.

The first entry is always `[0, 0]` (start of file). The last entry's
`word_index` is the largest multiple of `interval` that is ≤ `word_count`.
A 100 M-word list at `interval = 1 M` therefore has exactly 100 checkpoints,
plus the leading `[0, 0]`.

### Other fields

* `version` — file format version (currently `1`).
* `flags` — bitmask:
  * `1` (`HCIDX_FLAG_SHA256`) — `sha256` is populated.
  * `2` (`HCIDX_FLAG_WHOLE_FILE`) — index covers the entire file. If unset,
    the index is partial and may not have checkpoints past `word_count`.
* `interval` — checkpoint spacing in base words.
* `word_count` — total base words covered.
* `file_size` — wordlist size in bytes at build time. Used as the cheap
  validator: on load, hashcat compares this with `stat()` size; on mismatch
  the index is ignored and hashcat falls back to a linear seek. Stale offsets
  are never trusted.
* `sha256` — optional strong validator (`null` or 64-char hex). Computing
  it requires reading the entire wordlist, so callers must opt in via
  `--wordlist-index-cache-sha256`. Useful when shipping the index across hosts
  (hashtopolis) where `file_size` alone is not enough of a trust signal.

## How the build piggybacks on `count_words`

The most expensive part of the index build — a linear scan of the wordlist —
is the scan hashcat already does in `count_words()` on a dictstat cache miss.
The builder simply records checkpoints during that pass. **First chunk pays
its normal scan cost once; the index drops out as a side effect.** From the
second run onward (or on every other agent fetching the wordlist + `.hcidx`),
the seek is bounded.

## Code surface

Touched files:

| File                          | Change                                                                  |
| ----------------------------- | ----------------------------------------------------------------------- |
| `include/wordlist_index.h`    | New: index struct, builder API, constants.                              |
| `src/wordlist_index.c`        | New: JSON writer (atomic `tmp + rename`), JSON parser, builder, lookup. |
| `src/wordlist.c`              | Hook builder into `count_words()` (free piggyback).                     |
| `src/slow_candidates.c`       | Index-accelerated seek in `slow_candidates_seek`.                       |
| `src/dispatch.c`              | Index-accelerated seek in the fast-candidates dispatch path.            |
| `src/user_options.c`          | `--wordlist-index-cache[=FILE]`, `--wordlist-index-cache-sha256`.        |
| `src/usage.c`                 | Help text.                                                              |
| `include/types.h`             | `IDX_WORDLIST_INDEX`, fields on `user_options_t`.                       |
| `src/Makefile`                | Wire `wordlist_index.o`.                                                |

## Validation rules (load time, fast path)

* `file_size` must equal `stat()` size of the wordlist (O(1) check). Mismatch
  → ignore the index, fall back to linear seek.
* If `sha256` is present **and** the caller opts in to verify, it must match a
  fresh hash of the wordlist. This is a full-file read, so it is **not** done
  on every run — only at trust-boundary crossings (e.g. hashtopolis
  distribution). The default path uses size alone.
* `mtime` is **intentionally not** consulted. It is unreliable across copies,
  bind mounts, and clock skew.

## Caveats

* Wordlist-based attacks only. `-a 3` (mask) seek is keyspace math, not a file
  seek — unaffected.
* Compressed wordlists (`.gz`, `.zip`, `.xz`) are excluded at runtime — their
  byte offsets are not meaningful as `fseek()` targets. The patch detects this
  via the `HCFILE` handle and falls back to linear scan.
* Rules (`-r`) and amplifiers do not change base-word offsets, so the index
  remains valid.
* For combinator (`-a 1`) and hybrid (`-a 6` / `-a 7`) attacks, only the base
  wordlist is indexed; the second list goes via the existing linear path.
* Encoding options do not change byte offsets — safe.

ps used claude code for this, test.sh is ok.

### DIFF
--- a/include/slow_candidates.h
+++ b/include/slow_candidates.h
@@ -21,6 +21,11 @@ typedef struct extra_info_straight
   u8  out_buf[256];
   u32 out_len;
 
+  // wordlist seek index (.hcidx) state; idx_tried guards one-shot lazy load
+  void *idx;        // hcidx_t *, allocated lazily; NULL until first seek
+  bool  idx_tried;
+  u64   idx_pos;    // current base-word position of fp (for index accel)
+
 } extra_info_straight_t;
 
 typedef struct extra_info_combi
@@ -53,6 +58,10 @@ typedef struct extra_info_mask
 } extra_info_mask_t;
 
 void slow_candidates_seek (hashcat_ctx_t *hashcat_ctx, void *extra_info, const u64 cur, const u64 end);
+
+// Frees the lazily-loaded wordlist seek index attached to a straight-mode
+// extra_info, if any. Safe to call when none was loaded.
+void slow_candidates_free_index (void *extra_info);
 void slow_candidates_next (hashcat_ctx_t *hashcat_ctx, void *extra_info);
 
 #endif // HC_SLOW_CANDIDATES_H
--- a/include/types.h
+++ b/include/types.h
@@ -954,6 +954,8 @@ typedef enum user_options_map
   IDX_VERSION_LOWER             = 'v',
   IDX_VERSION                   = 'V',
   IDX_WORDLIST_AUTOHEX_DISABLE  = 0xff54,
+  IDX_WORDLIST_INDEX_CACHE        = 0xff70,
+  IDX_WORDLIST_INDEX_CACHE_SHA256 = 0xff71,
   IDX_WORKLOAD_PROFILE          = 'w',
 
 } user_options_map_t;
@@ -2619,6 +2621,8 @@ typedef struct user_options
   u32          workload_profile;
   u64          limit;
   u64          skip;
+  char        *wordlist_index_cache;
+  bool         wordlist_index_cache_sha256;
   bool         hash_copy;
 
 } user_options_t;
--- a/include/wordlist_index.h
+++ b/include/wordlist_index.h
@@ -1,0 +1,139 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#ifndef HC_WORDLIST_INDEX_H
+#define HC_WORDLIST_INDEX_H
+
+#include <stdio.h>
+#include <stdbool.h>
+
+// Wordlist seek index (.hcidx).
+//
+// Maps base-word positions to byte offsets in the wordlist so that a large
+// --skip can fseek() close to the target instead of reading and counting
+// every line from the start of the file. Entries are sparse: one checkpoint
+// every HCIDX_DEFAULT_INTERVAL base words. A lookup binary-searches for the
+// largest checkpoint <= the requested base-word, seeks there, and the caller
+// walks forward at most (interval - 1) words to land exactly on the target.
+//
+// On-disk format is JSON (inspectable with cat/jq), shape:
+//   { "version": 1, "flags": 0, "interval": 1000000, "word_count": N,
+//     "file_size": B, "sha256": null | "<64-hex>",
+//     "entries": [[idx, off], [idx, off], ...] }
+//
+// Validation is intentionally cheap: only the wordlist size is checked at load
+// time (an O(1) stat), never mtime (unreliable across copies / clock skew).
+// An optional SHA-256 of the wordlist may be stored for callers that cross a
+// trust boundary (e.g. an index built on one host and shipped to another by a
+// distribution layer such as hashtopolis); it is verified only on request, not
+// on every run, because hashing the whole file would reintroduce the very scan
+// this index removes.
+
+#define HCIDX_VERSION           1
+#define HCIDX_SUFFIX            ".hcidx"
+
+#define HCIDX_FLAG_SHA256       (1u << 0)         // sha256 field is populated
+#define HCIDX_FLAG_WHOLE_FILE   (1u << 1)         // index covers the entire file
+
+// One checkpoint every this many *base words*. Keeps the index tiny: a 1e9-word
+// list produces ~1000 entries.
+#define HCIDX_DEFAULT_INTERVAL  1000000
+
+// Do not bother building an index for files smaller than this; the linear seek
+// is already cheap and the .hcidx write would just be overhead.
+#define HCIDX_MIN_FILE_SIZE     (256 * 1024 * 1024)   // 256 MiB
+
+typedef struct hcidx_header
+{
+  u32 version;
+  u32 flags;
+  u64 interval;       // base words between checkpoints
+  u64 word_count;     // total base words (whole-file) or covered (partial)
+  u64 file_size;      // wordlist size in bytes - cheap validation
+  u8  sha256[32];     // optional (flags & HCIDX_FLAG_SHA256); else zeroed
+  u64 entry_count;
+
+} hcidx_header_t;
+
+typedef struct hcidx_entry
+{
+  u64 word_idx;       // base-word position of this checkpoint
+  u64 byte_off;       // byte offset in the wordlist of that word
+
+} hcidx_entry_t;
+
+typedef struct hcidx
+{
+  bool             loaded;
+  hcidx_header_t   header;
+  hcidx_entry_t   *entries;
+
+} hcidx_t;
+
+// Builder: accumulates checkpoints while the caller walks the file, then writes
+// the .hcidx atomically (tmp + rename) on finalize. The builder records a
+// checkpoint only at the start of a line, i.e. immediately after a base word
+// boundary, so the stored byte offset is safe to fseek() to.
+typedef struct hcidx_builder
+{
+  bool             active;
+  char            *path;          // final .hcidx path
+  char            *tmp_path;      // path + ".tmp"
+  u64              interval;
+  u64              next_checkpoint; // next base-word idx to record at
+  u64              entry_count;
+  u64              entry_avail;    // allocated capacity
+  hcidx_entry_t   *entries;       // buffered in memory; sparse, so tiny
+  u64              word_count;     // base words seen so far
+  u64              file_size;
+  bool             want_sha256;
+
+} hcidx_builder_t;
+
+// --- loading / lookup -------------------------------------------------------
+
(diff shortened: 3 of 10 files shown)


thatux:
PR currently contains a json-parser and sha256 implementation. I don't really like it, but it's functional though..

thatux:
this can be used to test the functionality / speed difference
[seek_test.sh](https://github.com/user-attachments/files/28522248/seek_test.sh)
The current `N=1000000000` results in a ~10gb file

thatux:
I missed that similar functionality was already implemented for -a8 https://github.com/hashcat/hashcat/commit/d67a863b685c37af7efc06f2a57630ec6569cd0a has made that the default for -a0 😎
introduced in https://github.com/hashcat/hashcat/commit/53df0340238ba19fb0a868a0f01eeb2af366976f

Maybe seekdb could benefit from having mtime optional, such that those files can be distributed via Hashtopolis (https://github.com/hashtopolis/server/issues/1692), also a `magic` and `version` could make sense.

jsteube:
Thanks for this, and sorry it sat so long. I am closing the pull request, but your comment from 11 August is the more useful half of this thread, and all three things in it have just landed in 758a517cf.

On the mechanism first, so the close makes sense. Wordlists are read through a feed now, and each one gets a seek database under `seekdbs` in the cache directory, built in src/feeds/seekdb.c. Same shape as your `.hcidx`: record the byte offset every so many words, binary search for the largest checkpoint at or below the target, seek there and walk the remainder. The interval is 8192 words rather than a million, so the walk after the seek is shorter. That is why the implementation here is not the one being taken.

The mtime was the important one and it is gone from the identity completely, rather than made optional as you suggested and as I first intended. Optional turned out to be the wrong shape once I worked out what the fast path would be for. The only thing keeping the mtime could do is confirm a file has not been touched since the database was built, and in the case this is all for the wordlist arrived by copy, so it never matches and the fast path never fires. It also only skipped checks that are free. So it is recorded in the header as provenance and nothing reads it.

What that gives up is smaller than it looks, which is worth writing down because it worried me for a while. A database maps a line number to a byte offset, and a line boundary only moves when a line's length changes. An edit that preserves lengths, `sed -i 's/password/passw0rd/'` for instance, leaves every offset in the database correct. Only an edit that moves a boundary invalidates one, and almost everything that moves a boundary changes the file size, which is still hashed. What is left is an edit that keeps the total size and moves a boundary anyway, and `thread_seek()` now checks for exactly that: a checkpoint is the first byte of a line, so the byte in front of it has to be the line ending that closed the one before, and where it is not the checkpoint is ignored and the position is scanned for instead. The mapping is already in memory so that costs one load. Net, it refuses more than the mtime did.

Magic and version are in, and you were right that they matter, though the reason turned out sharper than format hygiene. `SEEKDB_STEP` is 8192, it was a compile time constant, and it was not written to the file, while `thread_seek()` reconstructed the line number as `db_idx * SEEKDB_STEP`. Retuning that constant would have silently invalidated every database already on disk, with wrong offsets rather than a refusal to load, so candidates would be skipped or repeated and nothing would say so. The step is read from the header now, which also means a database built by an older hashcat stays usable under a different tuning and is merely coarser.

The header is 512 bytes of text, padded so it stays printable, carrying the magic and version, the byte order, the step, the line count, the size, the identity hash, a hash of the whole wordlist, the build time and the wordlist's base name. `head -c 512` on a database now says which wordlist it belongs to and whether it is current, which is the thing that was missing once these files started being copied by other people's tooling and arriving named only by a hash. Identity and size are both checked against the wordlist actually present, so one renamed or misrouted in transit is refused rather than used.

One you did not raise and which only became a problem once the files could travel: the offsets are written in native byte order, so a database copied between machines of different endianness would have been read as offsets rather than refused. The order is in the header now.

Two things I decided against, so you know where the edges are. Nothing is ever written next to a wordlist, so there is no `rockyou.txt.seekdb` beside the file. Wordlist directories are curated, often shared or mounted read only, and are not ours to write into. Databases stay in the cache directory and I will tell the hashtopolis developers where that is and how the name is derived, since that is now effectively public API.

And a correction on the issue you linked. hashtopolis/server#1692 is not asking for distributable indexes, it asks for pipelining in the agent, requesting two chunks so the skip on the second is hidden behind cracking the first. Worth reading again, because the seek database probably answers it outright: the index is built once per wordlist per agent and cached, so their premise that the skip is paid for every chunk stops being true. That is a better outcome for them than either pipelining or copying files, and I will say so over there.

So: closing the implementation, and thanks for the review, which was the part that actually changed the code.

thatux:
thanks, real nice! 

an afterthought does it make sense to add --seekdb-path=
that'd allow to also load the seekdbs from a mount?

if you want I can try and implement, but feel free.

jsteube:
Yes, that makes sense. The path is internal right now, so every machine that reads the same wordlist builds its own copy of the same database, and each of those builds costs a full read of the file. A shared mount turns that into one build for the whole cluster. Please do implement it. `--seekdb-path` fits the existing naming next to `--potfile-path` and `--restore-file-path`, and `seekdb_path()` takes the directory from one field, so the change should be small.

Two things that should make it easier. A read-only mount already behaves correctly without any extra handling, because hashcat only writes on a miss and a failed write leaves it running from the database it built in memory. And a shared database is already safe to load: the header carries the magic and version, the byte order, the step, the identity hash and the byte count, and all of them are checked against the wordlist actually present, so one that does not belong to the file in hand is refused rather than trusted.

In case you want to try, don't forget the bash completion :)

PR #4708, opened 2026-06-18
Add SECURITY.md describing threat model and disclosure process

xfscott:
Documents that user-supplied inputs (rule files, hash files, wordlists, masks) are trusted by definition: crashes from voluntarily loaded malicious files are out of the threat model and handled as robustness bugs, while networked/multi-user features (brain server) and privilege boundaries remain in scope. Adds private GitHub advisory and email disclosure channels.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,64 @@
+# Security Policy
+
+## Threat Model
+
+hashcat is a local password recovery tool. Whoever runs it provides all of
+its inputs: the hash file, wordlists, rule files (including inline `-j`/`-k`
+rules), masks and custom charsets, and the session/potfile/config files.
+
+All of that input is trusted. It comes from the same person running the
+process, and it is parsed with that person's own privileges. There is no
+remote attacker in the picture, nothing runs with elevated rights, and there
+is no sandbox to escape. Reading a file you chose to load does not cross a
+privilege boundary.
+
+## Out of Scope
+
+These are not treated as security vulnerabilities, because they require you
+to feed malicious input to your own process:
+
+- Crashes, out-of-bounds access, or memory corruption from a crafted rule
+  file or inline `-j`/`-k` rule.
+- The same from a crafted hash file, wordlist, mask, charset, or any other
+  input file you load.
+- Memory, disk, or CPU exhaustion caused by input you chose to load.
+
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks for this, and sorry it sat unanswered for two months.

You were first. Hashcat now has a security policy, and this pull request is why it happened now rather than later. The commit that adds it, c10ac3355, names you and the date. That date matters beyond credit: you opened this on 18 June, six weeks before the coordinated vulnerability reports arrived on 31 July. Nobody can reasonably claim the policy was invented in response to those reports.

What landed is not this text, though, and I want to be clear about where it differs rather than quietly merging something else.

The shipped policy is built around one question instead of a fixed list of trusted inputs: can somebody other than the person running hashcat use hashcat to harm that user? If yes, it may be a vulnerability. If the operator controls the input, it is a bug and still gets fixed, but it is not a security failure. Inputs then fall into three groups: instructions, code and data. Instructions are the command line and bridge parameters. Code is modules, bridges, feeds and kernels, all of which are loaded and executed, so trusting one is equivalent to trusting a program. Everything else is data, and new inputs are treated as data unless the policy explicitly says otherwise.

The main difference between our versions is hash files. Your text puts a crafted hash file out of scope because somebody who can convince a user to load one could instead convince them to run a malicious binary. That argument holds for much of what hashcat reads, but not for hash files. Users normally do not create the hashes they crack. In an audit they come from the client, in a contest from the organizer, and in research from a dump. Processing data produced by somebody else is the job, not a mistake, so those parsers have to survive hostile input. The same applies to captured authentication data, and wordlists and rule files are commonly downloaded rather than written by the operator.

Most of the rest of your proposal is still there: hashcat is a local tool with no privilege boundary to cross, bugs remain worth fixing even when they are not vulnerabilities, the brain is the real network exposed exception, and private advisories are the reporting channel. What changed is where the boundary sits, not the overall argument.

Your scope also includes one local user interfering with a hashcat process owned by another user. The shipped policy does not mention that explicitly, but it is inside the boundary it defines. Hashcat already protects against it deliberately: system.c sets umask 077, and profile, cache, session and kernel cache directories are created with mode 0700, with nothing written into shared locations. That probably deserves to be written down as well.

So I am closing this rather than merging it. That is not a judgement on the work. The policy would not exist yet without this pull request.

PR #4709, opened 2026-06-22
Added hash-mode: LastPass Vault (Chrome IndexedDB)

datarecovery-com:
This plugin recovers a LastPass master password from the Chrome extension's local IndexedDB SQLite file (`chrome-extension_hdokiejnpimakedhajhdlcegeplioahd_0`), without requiring server contact.

The verification format is:

```
$lpvault$0$<iterations>$<email>$<iv_hex>$<verify_ct_hex>
```

Where:
- `iterations`: PBKDF2 iteration count from the LastPass vault metadata
- `email`: Account email (used as the PBKDF2 salt, lowercased)
- `iv_hex`: 16-byte AES-CBC IV from the first ACCT chunk's username field (hex)
- `verify_ct_hex`: 32 bytes (2 AES blocks) of ciphertext from the same field (hex)

The kernel:
1. Derives a 32-byte key via PBKDF2-HMAC-SHA256(password, email, iterations)
2. Decrypts the two-block verification ciphertext via AES-256-CBC
3. Verifies via PKCS7 padding validity + ASCII-printable check on block 1

Existing mode 6800 (LastPass + LastPass sniffed) does not cover this case because it requires the server-side auth hash, which isn't stored locally.

Test vector included in `tools/test_modules/m99000.pm` (password: `hashcat`).

Speed on RTX 4090 at 5000 iterations: ~1.7 MH/s.

Note: the kernel deliberately places the T-table loading loop **before** the `if (gid >= GID_CNT) return;` early-return so all workgroup threads participate in loading. With small candidate counts (e.g. self-tests), placing the load after the early-return causes most threads to exit and only one entry of each table to be populated — producing silently wrong AES output.

### DIFF
--- a/OpenCL/m99000-pure.cl
+++ b/OpenCL/m99000-pure.cl
@@ -1,0 +1,575 @@
+/**
+ * Author......: Ben Carmitchel (Datarecovery.com)
+ * License.....: MIT
+ *
+ * LastPass Vault (Chrome IndexedDB) - kernel
+ *
+ *   key = PBKDF2-HMAC-SHA256(password, email.lower(), iterations, 32)
+ *   pt  = AES-256-CBC-Decrypt(verify_ct, key, iv)   // 2 blocks
+ *   match = PKCS7-valid(pt[16:32]) AND ASCII-printable(pt[0:16])
+ *
+ * On match, write the magic digest {0xdeadbeef, 0xfeedface,
+ * 0xcafebabe, 0xbaadf00d} via COMPARE_M_SCALAR.
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_simd.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha256.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_aes.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct lpvault
+{
+  u32 iv[4];
+  u32 ct[8];
+} lpvault_t;
+
+typedef struct pbkdf2_sha256_tmp
+{
+  u32 ipad[8];
+  u32 opad[8];
+
+  u32 dgst[8];
+  u32 out[8];
+
+} pbkdf2_sha256_tmp_t;
+
+// hmac_sha256_run_V - defined per-module in hashcat (not in a header).
+// Copied verbatim from m15600-pure.cl.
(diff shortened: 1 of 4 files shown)


Chick3nman:
I understand that AI helped write this code/PR but it would be a good idea to review what the AI told you while doing so as it seems there are some inconsistencies in your PR:

image

image

https://github.com/hashcat/hashcat/blob/2d71af3718ef76f69c675e933cb4900c40788c58/src/modules/module_06800.c#L46-L50

6800 _already includes_ the local PBKDF2 stored to disk by the lastpass plugin. I've used it to crack and recover lastpass passwords from Chrome and Firefox myself without having to contact the remote lastpass servers during the process. I'm not sure if perhaps lastpass changed how their plugin works very recently, but they stored this information locally from the last time I did this (~1 year ago). The kernel has a branch for handling the offline mode hash using the same format.

Please review and ensure that this is not already supported by mode 6800 as it appears to do the same thing at first glance and the PR message seems to incorrectly state that 6800 doesn't support this.

Chick3nman:
I set up a fresh Chrome install with the LastPass extension and checked the IndexedDB storage `chrome-extension_hdokiejnpimakedhajhdlcegeplioahd_0.indexeddb.leveldb`. The extension stores an encryptedUsername field at the account level in the existing/known LastPass format: `!<IV_base64>|<ciphertext_base64>`

To extract a hash from Chrome IndexedDB:

Parse the LevelDB log file for the encryptedUsername field
Decode the base64 IV and ciphertext
Extract the iterations value
Format as `hex(ciphertext_block1):iterations:email:hex(iv)`

I verified this with a test account and current LastPass default configuration:


`./hashcat -m 6800 lp_test.hash wordlist`

```
67b1b199e6078824794f18b737c0d2e4:600000:testlp@chick3nman.com:886c8e2609d96e64e8b2ac233c3cc311:Masswprldf1@

                                                          

Session..........: hashcat
Status...........: Cracked
Hash.Mode........: 6800 (LastPass + LastPass sniffed)
Hash.Target......: 67b1b199e6078824794f18b737c0d2e4:600000:testlp@chic...3cc311
```

What might still be useful is an extraction script in tools/ that parses Chrome's IndexedDB LevelDB format and outputs 6800 format hashes, since the existing tools only seem to handle the older SQL format. Otherwise the added kernel/module seem entirely redundant and, looking at the padding+ASCII printable verification check, inferior to the existing kernel.

jsteube:
> Existing mode 6800 (LastPass + LastPass sniffed) does not cover this case because it requires the server-side auth hash, which isn't stored locally.

This is the one thing the PR rests on, and it is the wrong way round. 6800 takes the locally stored hash, not the one that travels to the LastPass servers. The usage notice hashcat prints for the mode spells that out. Chick3nman then went and cracked a hash pulled from a fresh Chrome install at current LastPass defaults, straight out of the IndexedDB store, with -m 6800.

The two kernels also do the same work. PBKDF2-HMAC-SHA256 over the email, then AES-256-CBC over one block of encryptedUsername. They part company at the check. 6800 compares the decrypted block against the email, which it already holds as the salt, so a match is a match. This one tests the PKCS7 padding and whether the plaintext bytes fall in the printable range, which is a guess at what the plaintext looks like rather than a comparison against it.

What is missing is the extraction, not the mode. The scripts in tools/ only handle the older SQL format, and nothing in the tree reads Chrome's IndexedDB LevelDB. A script there that parses it and prints hex(ct_block1):iterations:email:hex(iv) would close the real gap, and I would be glad to take it.

Closing this one. Thanks for the work that went into it, and please do open the extraction script if you still have the parsing code.

PR #4712, opened 2026-07-03
perf(rc4): shorten S-box address dependency chain in KEY8 (+4.6% Kerberoasting, +8% Office)

jeffmcjunkin:
### What
Rewrites the RC4 S-box byte-address macro `KEY8` in `OpenCL/inc_cipher_rc4.cl` from `+`-composition to `|`-composition of its four sub-fields.

### Why
`KEY8` combines four **disjoint** bit-fields — `k&3` (bits 0-1), `(t&31)<<2` (2-6), `(k>>2)<<7` (7-12), `(t>>5)<<13` (13+). Composing them with integer `+` forces the compiler to assume inter-field carries, so it materialises the per-thread base with an extra add on the latency-critical *dynamic-index → `LDS`* dependency chain. Because the fields never overlap, `|` is bit-for-bit identical and lets the base fold into the final `LOP3`, removing one instruction from every S-box access (256 KSA swaps + PRGA per candidate). RC4-HMAC kernels are latency-bound on exactly this chain (ncu: ~69% compute, IPC ~1.5, occupancy shared-memory-capped), so the saving shows up directly.

### Measured (RTX 3090, GA102 sm_86, clock-locked 1710 MHz, median of 3, clean kernel-cache recompile)
| Mode | Before | After | Δ |
|---|---|---|---|
| 13100 Kerberos etype23 TGS-REP (Kerberoasting) | 1382 MH/s | 1447 MH/s | **+4.6%** |
| 18200 etype23 AS-REP roasting | 1380 | 1442 | +4.5% |
| 7500 etype23 AS-REQ Pre-Auth | 1423 | 1489 | +4.6% |
| 9700 MS Office ≤2003 MD5+RC4 | 1049 | 1134 | **+8.0%** |
| 9800 MS Office ≤2003 SHA1+RC4 | 1173 | 1227 | +4.6% |

All 18 modes that `#include inc_cipher_rc4.cl` benefit.

### Correctness / scope
- Value-identical transform (disjoint bit-fields); the `IS_CPU` linear-addressing branch is untouched.
- Self-test passes for m13100/18200/7500/9700/9800/10400/10500/25400; example hashes crack.
- Single-file change; no shared struct/ABI impact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Y7xSLzQxSeeNFducn2iwaC

### DIFF
--- a/OpenCL/inc_cipher_rc4.cl
+++ b/OpenCL/inc_cipher_rc4.cl
@@ -78,7 +78,12 @@ DECLSPEC void SET_KEY32 (LOCAL_AS u32 *S, const u8 k, const u32 v, MAYBE_UNUSED 
 // Finally we can select the actual target byte from (1 out of 4) from this chunk:
 //   (k & 3)
 
-#define KEY8(t,k) (((k) & 3) + (((k) / 4) * 128) + (((t) & 31) * 4) + (((t) / 32) * 8192))
+// Equivalent to (((k) & 3) + (((k) / 4) * 128) + (((t) & 31) * 4) + (((t) / 32) * 8192)),
+// but the four sub-fields occupy disjoint bit ranges (bits 0-1, 7-12, 2-6, 13+), so they can be
+// OR-composed instead of added. This lets the compiler fold the per-thread-invariant base term
+// into a single LOP3 at the combine step, shortening the (latency-critical) dynamic-index address
+// dependency chain by one instruction vs the carry-assuming '+' form.
+#define KEY8(t,k) ((((t) & 31) << 2) | (((t) >> 5) << 13) | ((k) & 3) | (((k) << 5) & 0x1f80))
 
 DECLSPEC u8 GET_KEY8 (LOCAL_AS u32 *S, const u8 k, const u64 lid)
 {


jeffmcjunkin:
### Correctness sweep (functional, independent reference)

Ran a vector-width × attack-kernel sweep against independently-generated reference hashes (hashcat's own `tools/test.pl` + `test_modules/m*.pm`, i.e. `Crypt::RC4` + `Digest::MD4` + `Digest::HMAC_MD5` / `Digest::MD5` / `Digest::SHA`), cracking with `-a 3` on an RTX 3090:

- Modes: **13100, 18200, 7500, 9700, 9800**
- Kernels: **pure and optimized** (`-O`)
- Vector widths: **1, 2, 4, 8**
- Password lengths: the 8 test.pl lengths per mode (0–27), all attack-mode-3 masks

**Result: 280/280 configurations recovered the correct password, 0 failures.** Combined with the value-identical bit-field argument (the four `KEY8` sub-fields are provably disjoint) and per-mode startup self-tests, the change is correct across the full pure/optimized × vector-width matrix.

jsteube:
Could you mark this ready for review? It is still a draft, so it cannot be merged as it stands.

The change reproduces here on two NVIDIA generations and is flat on AMD. Across 7500, 13100, 18200, 9700 and 9800, an RTX 4090 gains 3.6 to 6.6 percent and a 2080 Ti 2.6 to 5.3, which puts your Ampere result in the middle of a range rather than at the end of it. An RX 7900 XTX lands between -0.7 and 0.0 percent, and the reason is worth having: LLVM already folds the additive spelling into v_and_or_b32 on RDNA3, so mode 7500 compiles to the same 20148 instructions either way and there is no chain left to shorten.

jsteube:
A correction to what I wrote above. I went through these in number order and only saw afterwards that #4731 carries the same change, gated per mode behind RC4_USE_BITWISE_ADDRESSING and switched on for 9700, 9800, 10500, 13100 and 18200.

This is still the one I would take first. The numbers above cover all five of those modes and no mode lost, so the per-mode gate is guarding against something that did not turn up, and the other 13 RC4 modes get the same win at no extra cost. With it in, [#4731](https://github.com/hashcat/hashcat/issues/4731) rebases onto it and the RC4_USE_BITWISE_ADDRESSING blocks come out, which is 20 lines and a per-mode ifdef less to read there. Its KEY32 form is a separate change and this PR does not touch it.

jsteube:
This is in master now by way of #4731, so I am closing it.

The OR composition sits behind RC4_USE_BITWISE_ADDRESSING, which modes 9700 and 13100 switch on around their include of the header and switch off after it. Modes that do not take it keep the addition:

```
#ifdef RC4_USE_BITWISE_ADDRESSING
#define KEY8(t,k) (((k) & 3) | (((k) / 4) * 128) | (((t) & 31) * 4) | (((t) / 32) * 8192))
#else
#define KEY8(t,k) (((k) & 3) + (((k) / 4) * 128) + (((t) & 31) * 4) + (((t) / 32) * 8192))
#endif
```

There is a difference from what you wrote here. That form keeps the arithmetic and swaps the operator, where yours also restructured the fields into shifts and a mask. If the shift form buys anything beyond the disjoint-OR itself, it would be worth its own measurement, because what landed does not carry it.

On an RTX 4090 the merged result measures +17.9 percent on 13100 and +22.8 percent on 18200 against master.

PR #4713, opened 2026-07-03
perf(whirlpool): store one shared-memory table instead of eight (m06100, +7%)

jeffmcjunkin:
### What
Cuts Whirlpool's shared-memory table footprint from eight tables to one in `m06100` (a3-optimized + a3-pure single-block fast path).

### Why
Whirlpool's eight round lookup tables are byte-rotations of one another: `MTk[x] == ROTR64(MT0[x], 8*k)`. Keeping only `MT0` in shared memory (**2 KB vs 16.4 KB per block**) and deriving the other seven lookups via `hc_rotr64` raises the shared-memory occupancy ceiling (`Block-Limit-Shared` 5 → 10, theoretical occupancy 10.4% → 16.7%). The a3-optimized `m06100_s04` kernel is latency-bound (ncu: ~16% compute, occupancy shared-mem-capped), so the freed occupancy — which the autotuner picks up (Thr 384→256, Accel 8→14) — is the win; the extra rotates land on the otherwise-idle ALU. A zero-key round-key schedule for the single-block case (message-independent from the all-zero IV) is folded in too.

### Measured (RTX 3090, GA102 sm_86, clock-locked 1710 MHz, median of 3, clean kernel-cache recompile)
| Mode | Before | After | Δ |
|---|---|---|---|
| 6100 Whirlpool | 1464 MH/s | 1566 MH/s | **+7.0%** |

### Correctness / scope
- Scope is `m06100` a3-optimized (pw ≤ 27) and the a3-pure single-block fast path (`pw_len ≤ 31`, exactly one Whirlpool block). `pw_len > 31` falls through to the **unchanged** stock `whirlpool_final_vector` (multi-block), where the zero-key assumption would not hold.
- `whirlpool_transform_zerokey_vector` is added **additively** to `inc_hash_whirlpool.cl/.h`; existing shared functions are untouched, so other Whirlpool consumers (6231, 13731, …) are unaffected.
- Self-test passes for m6100 (optimized + pure), m6231, m13731; the example hash and multi-block (>31-char) inputs crack correctly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Y7xSLzQxSeeNFducn2iwaC


jeffmcjunkin:
### Correctness sweep (functional, independent reference)

Ran a vector-width × kernel-type × length sweep against independently-generated reference hashes (hashcat's own `tools/test.pl` + `test_modules/m06100.pm` via `Crypt::Digest::Whirlpool`), cracking with `-a 3` on an RTX 3090:

- Kernels: **pure and optimized** (`-O`) — i.e. both changed files
- Vector widths: **1, 2, 4, 8**
- Password lengths: **0, 1, 5, 8, 18, 24, 29, 31** — deliberately crossing the single-block boundary (`pw_len ≤ 31` uses the new zero-key/single-table transform; `> 31` falls through to the unchanged `whirlpool_final_vector`)

**Result: 48/48 configurations recovered the correct password, 0 failures.** Startup self-tests for m6100 (optimized + pure), m6231, and m13731 also pass, confirming the additive header change leaves other Whirlpool consumers unaffected.

PR #4718, opened 2026-07-06
Fix memory safety issues: use-after-free, fd leaks, integer overflow, missing null termination

lxcxjxhx:
## Summary

This PR fixes several memory safety issues in hashcat core code that could lead to crashes, resource leaks, or undefined behavior.

### 1. Use-after-free in src/shared.c (line 590)

**Problem**: After calling putenv(display), the code immediately calls hcfree(display). However, POSIX specifies that putenv() places the string directly into the environment (not a copy), so freeing it causes use-after-free when the environment variable is later accessed.

**Fix**: Remove the hcfree(display) call and add a comment explaining why the memory cannot be freed (similar to the existing comment on line 617 for 	mpdir).

### 2. File descriptor leaks in src/filehandling.c (lines 278, 356)

**Problem**: When dopen() fails and returns NULL, the code returns false without closing the already-opened file descriptor p->fd, causing resource leaks.

**Fix**: Add close(fp->fd) and reset p->fd = -1 in the error path before returning false.

### 3. Integer overflow in src/memory.c (line 127)

**Problem**: The calculation sz + align + sizeof(void*) in hcmalloc_bridge_aligned() can overflow for large values of sz, leading to a smaller-than-expected allocation and potential buffer overflow.

**Fix**: Add overflow check before allocation: if (total_size < sz) return NULL;

### 4. Missing null termination in src/modules/module_22000.c (line 851) and src/modules/module_13100.c (lines 245, 263)

**Problem**: After hex_decode() or memcpy() operations, the buffers are later used as strings (with %s format specifier) but lack explicit null termination. While the structures may be zero-initialized, this is unsafe practice and could lead to reading beyond buffer bounds.

**Fix**: Add explicit null terminators after the copy operations: ((u8 *) buffer)[len] = '\0';

## Impact

- **Security**: Prevents potential use-after-free, buffer overflows, and information disclosure
- **Stability**: Fixes resource leaks that could cause failures under heavy load
- **Performance**: No performance impact; changes are in error paths or one-time initialization

## Testing

Changes are defensive fixes in error handling and initialization paths. The fixes follow established patterns already present in the codebase (e.g., the comment on line 617 of shared.c already notes that 	mpdir cannot be freed after putenv()).

### DIFF
--- a/src/filehandling.c
+++ b/src/filehandling.c
@@ -275,7 +275,13 @@ bool hc_fopen (HCFILE *fp, const char *path, const char *mode)
   }
   else
   {
-    if ((fp->pfp = fdopen (fp->fd, mode)) == NULL) return false;
+    if ((fp->pfp = fdopen (fp->fd, mode)) == NULL)
+    {
+      // fdopen failed, close the file descriptor to avoid resource leak
+      close (fp->fd);
+      fp->fd = -1;
+      return false;
+    }
 
     if (fp->bom_size)
     {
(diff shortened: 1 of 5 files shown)


lxcxjxhx:
Closing: no maintainer activity since 2026-07-06. Will reopen if needed.

PR #4720, opened 2026-07-08
Fix -cl-std= build option to use per-device OpenCL version instead of platform version

### LINKED ISSUE #4670: Incorrect -cl-std= argument passed in OpenCL kernel build options
**Describe the bug**
Hashcat incorrectly uses the OpenCL platform's OpenCL version to determine build flags, rather than using the per-device supported OpenCL and OpenCL C versions. Most of the time these will be identical (or the platform's OpenCL version will at least be listed in `CL_DEVICE_OPENCL_C_ALL_VERSIONS` such as on NVIDIA), but sometimes this is not the case, e.g. with the PoCL-Remote driver which itself reports platform version 3.0 but can provide access to devices with any supported OpenCL API and OpenCL C versions.

**To Reproduce**
* Obtain a system with a vendor OpenCL driver that is known to work with hashcat but does not support OpenCL 3.0 (e.g. ROCm)
* Build PoCL from source with cmake options `-DENABLE_REMOTE_CLIENT=1 -DENABLE_REMOTE_SERVER=1 -DENABLE_LLVM=0 -DENABLE_HOST_CPU_DEVICES=0`.
* Launch the pocld server, e.g. `build/pocld/pocld -a localhost -v all`
* In another terminal, set the relevant environment variables to configure the client side of pocl-remote e.g.
    ```
    OCL_ICD_VENDORS=xxxx # point to ocl-vendors in the build directory, alternatively OCL_ICD_FILENAMES=/path/to/libpocl.so
    POCL_BUILDING=1 # if using libpocl from the build directory rather than a proper install
    POCL_DEVICES=remote
    POCL_REMOTE0_PARAMETERS="localhost/0"
    ```
* Running `hashcat -b -m 100` should fail to compile kernels due to an unsupported `-cl-std=3.0` flag.

**Expected behavior**
Hashcat honors the reported `CL_DEVICE_OPENCL_C_VERSION` (or a supported version from the `CL_DEVICE_OPENCL_C_ALL_VERSIONS` list when that is available).

 - OS: Linux
 - Distribution: SUSE Linux Enterprise Server 15 SP6
 - Version: v7.1.2-382-g2d71af371+

**Diagnostic output compute devices:**

```
$ clinfo || true
Number of platforms:				 1
  Platform Profile:				 FULL_PROFILE
  Platform Version:				 OpenCL 3.0 PoCL 7.2-pre experimental-0-gad49ebce9  Linux, RelWithDebInfo, RELOC, without LLVM, REMOTE, POCL_DEBUG
  Platform Name:				 Portable Computing Language
  Platform Vendor:				 The pocl project
  Platform Extensions:				 cl_khr_icd cl_khr_priority_hints cl_khr_throttle_hints cl_khr_create_command_queue cl_pocl_content_size cl_ext_buffer_device_address


  Platform Name:				 Portable Computing Language
Number of devices:				 1
  Device Type:					 CL_DEVICE_TYPE_GPU
  Vendor ID:					 1002h
  Max compute units:				 110
  Max work items dimensions:			 3
    Max work items[0]:				 1024
    Max work items[1]:				 1024
    Max work items[2]:				 1024
  Max work group size:				 256
  Preferred vector width char:			 4
  Preferred vector width short:			 2
  Preferred vector width int:			 1
  Preferred vector width long:			 1
  Preferred vector width float:			 1
  Preferred vector width double:		 1
  Native vector width char:			 4
  Native vector width short:			 2
  Native vector width int:			 1
  Native vector width long:			 1
  Native vector width float:			 1
  Native vector width double:			 1
  Max clock frequency:				 1700Mhz
  Address bits:					 64
  Max memory allocation:			 58397294592
  Image support:				 Yes
  Max number of images read arguments:		 128
  Max number of images write arguments:		 8
  Max image 2D width:				 16384
  Max image 2D height:				 16384
  Max image 3D width:				 16384
  Max image 3D height:				 16384
  Max image 3D depth:				 8192
  Max samplers within kernel:			 16
  Max size of kernel argument:			 1024
  Alignment (bits) of base address:		 2048
  Minimum alignment (bytes) for any datatype:	 128
  Single precision floating point capability
    Denorms:					 Yes
    Quiet NaNs:					 Yes
    Round to nearest even:			 Yes
    Round to zero:				 Yes
    Round to +ve and infinity:			 Yes
    IEEE754-2008 fused multiply-add:		 Yes
  Cache type:					 Read/Write
  Cache line size:				 128
  Cache size:					 16384
  Global memory size:				 68702699520
  Constant buffer size:				 58397294592
  Max number of constant args:			 8
  Local memory type:				 Local
  Local memory size:				 65536
  Max pipe arguments:				 0
  Max pipe active reservations:			 0
  Max pipe packet size:				 0
  Max global variable size:			 0
  Max global variable preferred total size:	 0
  Max read/write image args:			 0
  Max on device events:				 0
  Queue on device max size:			 0
  Max on device queues:				 0
  Queue on device preferred size:		 0
  SVM capabilities:				
    Coarse grain buffer:			 No
    Fine grain buffer:				 No
    Fine grain system:				 No
    Atomics:					 No
  Preferred platform atomic alignment:		 64
  Preferred global atomic alignment:		 64
  Preferred local atomic alignment:		 64
  Kernel Preferred work group size multiple:	 0
  Error correction support:			 0
  Unified memory for Host and Device:		 0
  Profiling timer resolution:			 1
  Device endianess:				 Little
  Available:					 Yes
  Compiler available:				 Yes
  Execution capabilities:				
    Execute OpenCL kernels:			 Yes
    Execute native function:			 No
  Queue on Host properties:				
    Out-of-Order:				 No
    Profiling :					 Yes
  Queue on Device properties:				
    Out-of-Order:				 No
    Profiling :					 No
  Platform ID:					 0x7f612a19c3d0
  Name:						 gfx90a:sramecc+:xnack-
  Vendor:					 Advanced Micro Devices, Inc.
  Device OpenCL C version:			 OpenCL C 1.2 PoCL
  Driver version:				 3635.0 (HSA1.1,LC)
  Profile:					 FULL_PROFILE
  Version:					 OpenCL 2.0  HSTR: pocl-remote: gfx90a:sramecc+:xnack-
  Extensions:					 cl_khr_fp64 cl_khr_global_int32_base_atomics cl_khr_global_int32_extended_atomics cl_khr_local_int32_base_atomics cl_khr_local_int32_extended_atomics cl_khr_int64_base_atomics cl_khr_int64_extended_atomics cl_khr_3d_image_writes cl_khr_byte_addressable_store cl_khr_fp16 cl_khr_gl_sharing cl_amd_device_attribute_query cl_amd_media_ops cl_amd_media_ops2 cl_khr_image2d_from_buffer cl_khr_subgroups cl_khr_depth_images cl_amd_copy_buffer_p2p cl_amd_assembly_program  cl_khr_command_buffer
```

```
$ ./hashcat -I --force || true
hashcat (v7.1.2-382-g2d71af371+) starting in backend information mode

hipInit(): no ROCm-capable device is detected

OpenCL Info:
============

OpenCL Platform ID #1
  Vendor..: The pocl project
  Name....: Portable Computing Language
  Version.: OpenCL 3.0 PoCL 7.2-pre experimental-0-gad49ebce9  Linux, RelWithDebInfo, RELOC, without LLVM, REMOTE, POCL_DEBUG

  Backend Device ID #01
    Type...........: GPU
    Vendor.ID......: 1
    Vendor.........: Advanced Micro Devices, Inc.
    Name...........: AMD Instinct MI250X
    Version........: OpenCL 2.0  HSTR: pocl-remote: gfx90a:sramecc+:xnack-
    Processor(s)...: 110
    Preferred.Thrd.: 32
    Clock..........: 1700
    Memory.Total...: 65520 MB (limited to 55692 MB allocatable in one block)
    Memory.Free....: 65520 MB
    Memory.Unified.: 0
    Local.Memory...: 64 KB
    OpenCL.Version.: OpenCL C 1.2 PoCL
    Driver.Version.: 3635.0 (HSA1.1,LC)
```

shadowshan125:
Fixes #4670.

## Problem

Hashcat derives the `-cl-std=` OpenCL kernel build flag from the **platform's** `CL_PLATFORM_VERSION`, not from each individual **device's** `CL_DEVICE_VERSION`. A single OpenCL platform can expose multiple devices that don't all support the same OpenCL version — this is the case with heterogeneous/remote platforms such as PoCL-Remote, where the platform itself reports version 3.0 but an individual device behind it may only support a lower version. Hashcat then passes `-cl-std=CL3.0` to that device's compiler and kernel compilation fails.

Root cause in `backend_ctx_devices_init_opencl` (`src/backend.c`): the `use_opencl11/12/20/30` flags that select the `-cl-std=` flag were computed from `opencl_platform_version` *before* the per-device `CL_DEVICE_VERSION` was even queried. The per-device version was already being queried into `opencl_device_version` a few lines later, but never used for this purpose.

## Fix

Move the flag derivation to run after `CL_DEVICE_VERSION` is queried, and parse that value instead of the platform string. No new OpenCL symbols or header changes — minimal, surgical diff (36 insertions / 26 deletions, all in one function).

A device's `CL_DEVICE_VERSION` is always ≤ its platform's version, so switching platform → device can only correct a mismatch, never introduce a new one on setups where the two already agree.

### Why not `CL_DEVICE_OPENCL_C_ALL_VERSIONS`?

That query would be the spec-precise way to pick an OpenCL *C language* version (as opposed to the API version `CL_DEVICE_VERSION` reports), and I considered it. Two things ruled it out for this PR:

1. hashcat pins `CL_TARGET_OPENCL_VERSION` to `120` in `include/ext_OpenCL.h`, which gates `CL_DEVICE_OPENCL_C_ALL_VERSIONS` / `cl_name_version` out of the vendored headers entirely. Using it would require either a local symbol-redefinition workaround or bumping that target globally — the latter risks tripping the project's `-W -Wall` clean-build requirement via new deprecation warnings, which is likely why it's pinned at 120 in the first place.
2. It's also risky as a straight swap: on both an NVIDIA RTX 5060 and an Intel iGPU tested here, `CL_DEVICE_OPENCL_C_VERSION` (the older, single-value field) reports `"OpenCL C 1.2"` even though both devices fully support 3.0 — that's only visible via `ALL_VERSIONS`. Deriving from that legacy field alone would have silently downgraded both vendors from `CL3.0` to `CL1.2`.

Given that, I went with the device-version fix as the correctly-scoped, low-risk change, and left the `ALL_VERSIONS` precedence question (raised in the issue thread) open for maintainers to weigh in on separately if a more precise fix is wanted later.

## Testing

Tested on Windows 11 (MSYS2/mingw-w64), with an NVIDIA GeForce RTX 5060 Laptop GPU (CUDA + OpenCL) and an Intel RaptorLake-S iGPU (OpenCL).

- Builds clean with `-std=gnu99 -W -Wall -Wextra`, no warnings.
- `hashcat -I` device enumeration unaffected.
- Instrumented build confirmed the flag computation itself: both devices report `CL_DEVICE_VERSION = "OpenCL 3.0 ..."`, correctly resolving to `use_opencl30=1` → `-cl-std=CL3.0`.
- Functional check: `hashcat -b -m 100` (SHA1) — kernels (`shared.cl`, `m00100_a3-optimized.cl`, `markov_be.cl`) compile and link successfully on both vendors, benchmarked without error.
- Confirmed **no regression**: on this hardware the fix produces byte-identical `-cl-std=` output to master, since there's no platform/device version mismatch to trigger a behavior change here.

**What I could not verify:** the exact reported scenario — a PoCL-Remote platform reporting a higher version than an individual device behind it — since I don't have that setup available, and neither did the person who diagnosed the root cause in the issue thread. If a maintainer or the original reporter can confirm against a real PoCL-Remote setup, that would close the loop.

### DIFF
--- a/src/backend.c
+++ b/src/backend.c
@@ -7070,37 +7070,17 @@ static void backend_ctx_devices_init_opencl (hashcat_ctx_t *hashcat_ctx, int *vi
         device_param->opencl_platform_id = opencl_platforms_idx;
 
         // check OpenCL version
+        //
+        // note: the actual flag derivation happens further below, once the
+        // per-device CL_DEVICE_VERSION has been queried. the platform version
+        // is not a reliable source since a single OpenCL platform can expose
+        // devices that individually support a lower OpenCL version than the
+        // platform as a whole (for example remote/heterogeneous platforms)
 
         device_param->use_opencl11 = false;
         device_param->use_opencl12 = false;
         device_param->use_opencl20 = false;
         device_param->use_opencl30 = false;
-
-        int opencl_version_maj = 0;
-        int opencl_version_min = 0;
-
-        if (sscanf (opencl_platform_version, "OpenCL %d.%d", &opencl_version_maj, &opencl_version_min) == 2)
-        {
-          if (opencl_version_maj == 1)
-          {
-            device_param->use_opencl11 = true;
-          }
-
-          if ((opencl_version_maj == 1) && (opencl_version_min == 2))
-          {
-            device_param->use_opencl12 = true;
-          }
-
-          if (opencl_version_maj == 2)
-          {
-            device_param->use_opencl20 = true;
-          }
-
-          if (opencl_version_maj == 3)
-          {
-            device_param->use_opencl30 = true;
-          }
-        }
 
         size_t param_value_size = 0;
 
@@ -7292,6 +7272,36 @@ static void backend_ctx_devices_init_opencl (hashcat_ctx_t *hashcat_ctx, int *vi
         }
 
         device_param->opencl_device_version = opencl_device_version;
+
+        // derive the -cl-std= build option from the device's own OpenCL
+        // version instead of the platform's, since a platform can expose
+        // devices with a lower individually supported OpenCL version
+
+        int opencl_version_maj = 0;
+        int opencl_version_min = 0;
+
+        if (sscanf (opencl_device_version, "OpenCL %d.%d", &opencl_version_maj, &opencl_version_min) == 2)
+        {
+          if (opencl_version_maj == 1)
+          {
+            device_param->use_opencl11 = true;
+          }
+
+          if ((opencl_version_maj == 1) && (opencl_version_min == 2))
+          {
+            device_param->use_opencl12 = true;
+          }
+
+          if (opencl_version_maj == 2)
+          {
+            device_param->use_opencl20 = true;
+          }
+
+          if (opencl_version_maj == 3)
+          {
+            device_param->use_opencl30 = true;
+          }
+        }
 
         // opencl_device_c_version
 


jsteube:
In "backend.c" master grew a device level check of its own in the meanwhile, which is where the conflict comes from:

```
        // The platform version was already checked further up, but that is the platform, not
        // this device. OpenCL 3.0 made most of 2.x optional and lets a device advertise a
        // lower OpenCL C than its platform, so a 3.0 platform can still expose a device whose
        // OpenCL C is 1.1. That device would be handed -cl-std from the platform version and
        // fail much later in the kernel build, so ask the device directly as well.
```

That one only skips a device below OpenCL C 1.2. The derivation is still the platform's, so #4670 is open and this is still wanted, on a rebase.

Could you read opencl_device_c_version rather than opencl_device_version? The build option selects the OpenCL C language version, and CL_DEVICE_VERSION does not report it. OpenCL 3.0 decoupled the two, so a 3.0 device may offer no more than OpenCL C 1.2, which is the argument this PR makes about the platform, one level down. The two do diverge in practice: under rusticl an RX 7900 XTX here reports OpenCL 3.0 against OpenCL C 1.2. The block quoted above already parses that field into device_c_version_maj and device_c_version_min, so the flags can come off those with no new query.

jsteube:
The change I asked for in my last comment is withdrawn, and I have no other change to ask for on this one.

I wanted the build option derived from opencl_device_c_version rather than from the device version. Both fields, read across the runtimes here:

```
runtime                        CL_DEVICE_VERSION    CL_DEVICE_OPENCL_C_VERSION
NVIDIA CUDA, RTX 4090          OpenCL 3.0 CUDA      OpenCL C 1.2
PoCL 6.0                       OpenCL 3.0 PoCL      OpenCL C 1.2 PoCL
rusticl, RX 7900 XTX           OpenCL 3.0           OpenCL C 1.2
Intel OpenCL, CPU              OpenCL 3.0           OpenCL C 3.0
AMD APP, RX 7900 XTX           OpenCL 2.0           OpenCL C 2.0
Apple OpenCL, Intel Mac        OpenCL 1.2           OpenCL C 1.2
```

Three of those six advertise OpenCL C 1.2 on a device reporting OpenCL 3.0, so deriving the option from the right hand column would hand -cl-std=CL1.2 to every NVIDIA device, to PoCL and to rusticl. The rusticl row is the example my comment gave for making the change, and it sits on the wrong side of it.

PR #4721, opened 2026-07-08
Add NVML power usage, power limit, and PCIe generation to hwmon status

### LINKED ISSUE #4369: HWMon/status: Add Pwr:Usage/Cap, PCI Gen
Hi Team. For now statistics doesn't show power consumption and some other info may help collect all data for benchmarks. Should be great to have few more options:

* Current power consumption (power.draw)
* Max allowed power consumption (in case it limited by GPU settings good to know what was the limit)

On top of that good to know not only PCI lanes (buslanes) but also generation. Thanks!

 - Linux
 - Ubuntu 24.04
 - Version: v6.2.6-1121-g853b14956

shadowshan125:
Fixes #4369 (NVML/NVIDIA scope — see notes below).

## Problem

The issue asks for three hardware-monitoring fields hashcat doesn't currently show: current power draw, the power *limit/cap* the GPU is capped at, and PCIe *link generation* (Gen3/4/5), not just the lane count already shown via `Bus:`.

A previous merged PR (#4372) added a `Pwr:` field, but scoped it to Apple Silicon only (via IOKit/SMC) — NVML (NVIDIA), ADL (AMD), and the Linux sysfs backends all still explicitly hard-code `power_get_supported = false`. PCI generation wasn't touched at all.

## Scope: NVIDIA/NVML only

I don't have AMD hardware, and investigation turned up genuine gaps in the AMD path that I can't verify:
- Legacy AMD Overdrive5 cards have no power field at all in their telemetry struct (`ADLPMActivity`).
- A power *cap* value (as opposed to instantaneous draw) isn't exposed via any ADL API binding present in this codebase — would need a wholly new binding.

Given that, this PR is scoped to NVIDIA via NVML only, fully testable on my hardware (NVIDIA RTX 5060 Laptop GPU + Intel iGPU). AMD is left as a follow-up for a contributor with AMD hardware to verify against — this mirrors how #4372 itself scoped to Apple-only rather than attempting every platform at once.

## What changed

- `hm_get_power_with_devices_idx` (`src/hwmon.c`) gets an NVML branch added alongside the existing Apple-only path (which is untouched).
- Two new functions, `hm_get_power_limit_with_devices_idx` and `hm_get_pcie_gen_with_devices_idx`, follow the same NVML-branch shape already used by `hm_get_buslanes_with_devices_idx`.
- Two new NVML bindings from scratch (`nvmlDeviceGetPowerUsage`, `nvmlDeviceGetPowerManagementLimit`) in `ext_nvml.c`/`ext_nvml.h`, following the existing wrapper pattern (see `hm_NVML_nvmlDeviceGetFanSpeed` as the template).
- `nvmlDeviceGetCurrPcieLinkGeneration` turned out to be already half-wired — the typedef, struct field, and `HC_LOAD_FUNC` load call all already existed (apparently left over from when `CurrPcieLinkWidth` was added), just missing the wrapper function. That one was a ~15-line addition.
- Two new capability flags (`power_limit_get_supported`, `pcie_gen_get_supported`) in `hm_attrs_t`, wired through both NVML init blocks (CUDA + OpenCL/NV paths) and both NVML merge blocks, following the existing `power_get_supported`/`buslanes_get_supported` pattern exactly.
- Display added in all three existing surfaces: the interactive status line (`Cap:`/`Gen:` after `Pwr:`), `--machine-readable` (`POWER_LIMIT`/`PCIE_GEN` rows), and `--status-json` (`power_limit`/`pcie_gen` keys).

All three new getters are independent — each gates on and mutates only its own capability flag, no shared helper — so a failure on one (e.g. power limit not being supported) can never suppress the others.

## Testing

Tested on Windows 11 (MSYS2/mingw-w64), NVIDIA GeForce RTX 5060 Laptop GPU (CUDA + OpenCL) + Intel RaptorLake-S iGPU (OpenCL).

- Builds clean with `-std=gnu99 -W -Wall -Wextra`, no warnings on any of the 4 changed `.c` files.
- Interactive status: `Pwr:`/`Gen:` render correctly and track real load (power draw climbs with clock/temp during a sustained attack, `Gen:5` stays constant as expected). `Cap:` is correctly absent on this mobile GPU — `nvmlDeviceGetPowerManagementLimit` isn't supported on it (same failure class as `nvmlDeviceGetFanSpeed` on laptops, which hashcat users have already hit). Confirmed graceful degradation: exactly one `nvmlDeviceGetPowerManagementLimit(): Not Supported` log line across a multi-refresh run, not a repeat per tick.
- Intel iGPU (no NVML) shows `-1`/absent for all three new fields across every surface, no crash, no garbage.
- `--machine-readable` and `--status-json` both include the new fields, correct column/key counts; JSON validated as well-formed.
- The pre-existing `#if defined (__APPLE__)` block in `hm_get_power_with_devices_idx` is byte-for-byte unchanged.

**What I could not verify:** whether `Cap:` actually renders a real value anywhere, since that needs a desktop NVIDIA GPU (where `nvmlDeviceGetPowerManagementLimit` is more likely to be supported, unlike this mobile part) — I only have a laptop GPU to test on. If a maintainer or anyone with a desktop NVIDIA card can confirm, that would close the loop on the one thing I couldn't positively verify.

### DIFF
--- a/docs/changes.txt
+++ b/docs/changes.txt
@@ -5,6 +5,7 @@
 ##
 
 Docker: Add hashcat-toolchain
+Hardware Monitor: Added power limit (Cap) and PCIe link generation (Gen) status fields for NVML (NVIDIA) devices
 
 ##
 ## Bugs
--- a/include/ext_nvml.h
+++ b/include/ext_nvml.h
@@ -198,6 +198,8 @@ typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_CLOCKINFO) (nvmlDevice_t, n
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_THRESHOLD) (nvmlDevice_t, nvmlTemperatureThresholds_t, unsigned int *);
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_CURRPCIELINKGENERATION) (nvmlDevice_t, unsigned int *);
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_CURRPCIELINKWIDTH) (nvmlDevice_t, unsigned int *);
+typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_POWER_USAGE) (nvmlDevice_t, unsigned int *);
+typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_POWER_MANAGEMENT_LIMIT) (nvmlDevice_t, unsigned int *);
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_CURRENTCLOCKSTHROTTLEREASONS) (nvmlDevice_t, unsigned long long *);
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_GET_SUPPORTEDCLOCKSTHROTTLEREASONS) (nvmlDevice_t, unsigned long long *);
 typedef nvmlReturn_t (*NVML_API_CALL NVML_DEVICE_SET_COMPUTEMODE) (nvmlDevice_t, nvmlComputeMode_t);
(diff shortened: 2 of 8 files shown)


jsteube:
In "hwmon.c" the position on a power cap is already written down, just above the throttle mask:

```
// A software power cap is deliberately not one of them. A GPU worth cracking on sits at its power
// limit for the whole run, so a warning about it would fire constantly and mean nothing.
```

That is why I would leave the two power fields out. A card held at its limit for the whole run reports a Pwr: pinned to its cap, and Cap: is a setting that does not move at all, so the status line would refresh two numbers that never change. The one power condition that does cost speed, HwPowerBrake, is already in NVML_THROTTLE_REASONS_THAT_COST_SPEED and reaches the user through the throttle indicator. The Apple Pwr: is not quite the precedent it looks like either: there is no clocks throttle reason API on that platform, so a power reading is the only signal of that kind available there.

The PCIe generation is worth having, but the status block has little room left for it. Its fixed width lines run to 83 at the longest, and Hardware.Mon is already second:

```
 83  Recovered........: 6491/6494 (99.95%) Digests (total), 1/6494 (0.02%) Digests (new)
 79  Hardware.Mon.#01.: Temp: 60c Fan: 40% Util:100% Core:1800MHz Mem:7000MHz Bus:16
 78  Speed.#01........: 12345.6 MH/s (95.21ms) @ Accel:512 Loops:1024 Thr:256 Vec:1
```

A separate Gen: field takes Hardware.Mon to 85 and makes it the longest line in the block, and 86 on a card whose memory clock runs to five digits. Folded into the neighbouring field as Bus:16x4 it costs 2 columns instead of 6 and the line stays second, which is the form I would take it in.

Two things that follow it wherever it lands. The vendor test cannot fire for HIP, because a HIP device is given opencl_device_vendor_id = VENDOR_ID_AMD_USE_HIP, so is_hip in that condition is dead and hm_get_utilization_with_devices_idx() shows the shape without it. And master's hm_get_power_with_devices_idx() opens with a bridge block now, which is where the conflict comes from.

PR #4722, opened 2026-07-11
Fix embedded NUL truncation in stdin mode

mohammedjhaish:
## Summary

The stdin path determines candidate length using in_superchop(), which
uses strlen(). This truncates candidates containing embedded NUL bytes.

This patch reads stdin while tracking the actual byte count and uses
superchop_with_length() instead, making stdin behave identically to
regular wordlist files.

## Reproducer

printf 'a\0a\n' > /tmp/word.txt

Before:

hashcat --stdout < /tmp/word.txt

Output:

61 0a

After:

61 00 61 0a

This also fixes the corresponding cracking behavior.

### DIFF
--- a/src/dispatch.c
+++ b/src/dispatch.c
@@ -145,6 +145,34 @@ static u64 get_work (hashcat_ctx_t *hashcat_ctx, hc_device_param_t *device_param
   return work;
 }
 
+static bool stdin_readline (char *buf, const size_t buf_sz, size_t *line_len)
+{
+  size_t len = 0;
+
+  int c = 0;
+
+  while (len < buf_sz)
+  {
+    c = fgetc (stdin);
+
+    if (c == EOF) break;
+
+    buf[len] = (char) c;
+
+    len++;
+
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks for this, and sorry it is landing as already fixed rather than merged.

The bug was real. What happened is that the base word readers were reworked in `24d67b5ee` which moved the stdin producer out of `src/dispatch.c` into `src/feeds/feed_stdin.c`. The new path is length-based, so in_superchop() and the strlen() behind it are gone, and the truncation went with them.

You filed this four weeks earlier, so this is the ground moving under the PR rather than anything you missed.

Your reproducer on current master now gives:

```
$ printf 'a\0a\n' | ./hashcat --stdout | xxd
00000000: 6100 610a                                a.a.
```

which is the `61 00 61 0a` this PR was asking for.

Closing as already fixed rather than merged, since the code it patches no longer exists. That stdin path is newly rewritten and has had few eyes on it, though, so if you feel like poking at it again you would be very welcome.

PR #4726, opened 2026-07-13
fix: remove per-digest 256 collision ceiling in --keep-guessing mode

### LINKED ISSUE #4169: Option --keep-guessing does not guess
### Problem
Here's the quote of the help, describing the **--keep-guessing** option:
> Keep guessing the hash after it has been cracked

And... It keeps running, but not guessing. No more collisions are going to be found.

### Expected behaviour
When option **--keep-guessing** is set, hashcat finds all the collisions, not just the first one only.

### System info
```bash
$ ./hashcat.exe -V
v6.2.6
```

### Reproducing
```bash
cd hashcat-dir;

echo -e 'W@h=nkx9\nevi89668818677' > my-dict.txt; # collisions of the hash

./hashcat.exe -a 0 --session=2025-03-17-collisions-test -m 24900 -d 1 "qLmA2MYp" my-dict.txt --keep-guessing --potfile-disable
```
Output (only the first collision):
> qLmA2MYp:W@h=nkx9

When I swap the passwords:
```bash
echo -e 'evi89668818677\nW@h=nkx9' > my-dict.txt; # collisions of the hash
```
**hashcat** gets another collision (but still not the both):
> qLmA2MYp:evi89668818677

Functionhx:
Report all password collisions with --keep-guessing instead of stopping at 256 per digest. Removes the per-digest plains ceiling, fixes association mode regression, and updates 367 kernel bypass paths to respect keep_guessing_limit.

Fixes: #4169

Signed-off-by: Functionhx <2994114386@qq.com>

### DIFF
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -3005,14 +3005,25 @@ DECLSPEC void mark_hash (GLOBAL_AS plain_t *plains_buf, GLOBAL_AS u32 *d_result,
 {
   const u32 idx = hc_atomic_inc (d_result);
 
-  if (idx >= digests_cnt)
+  // Association mode (ATTACK_MODE==9): when DIGESTS_CNT (always 1) is
+  // passed as capacity from a mode-specific path, skip the check — the
+  // per-digest hashes_shown gate already limits results, and the buffer
+  // is sized exactly (plains_cnt == digests_cnt).  When a larger value
+  // (plains_cnt) is passed by keep-guessing or common macros, the
+  // capacity check must apply to prevent global overflow.
+  #if ATTACK_MODE == 9
+  if (digests_cnt > 1)
+  #endif
   {
-    // this is kind of tricky: we *must* call hc_atomic_inc() to know about the current value from a multi-thread perspective
-    // this action creates a buffer overflow, so we need to fix it here
+    if (idx >= digests_cnt)
+    {
+      // this is kind of tricky: we *must* call hc_atomic_inc() to know about the current value from a multi-thread perspective
+      // this action creates a buffer overflow, so we need to fix it here
 
-    hc_atomic_dec (d_result);
+      hc_atomic_dec (d_result);
 
-    return;
+      return;
+    }
   }
 
   plains_buf[idx].salt_pos   = salt_pos;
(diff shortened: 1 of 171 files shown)


jsteube:
This PR raises the reporting limit rather than removing it. On the "master" branch, we allow one reported candidate per digest per kernel invocation through "mark_hash()". For "NEVERCRACK" modes, we clear "digests_shown" so the next invocation can report candidates again. The proposed cap of 256K candidates per invocation still drops anything beyond that count.

Reporting every candidate from an invocation still has the memory cost described in "docs/limits.txt". That is why we explicitly rule it out there. The estimate in that file is actually too low: it assumes 20 bytes per "plain_t" structure, while the structure has been 40 bytes since the "gidvid" and "il_pos" fields were widened to 64 bits.

We also need to keep the change limited to the modes that need it. The "OPTS_TYPE_SUGGEST_KG" flag already identifies the 19 modules where we expect additional valid candidates, but the PR changes shared comparison code that almost every mode builds against, and edits 155 kernels on top. Even runs without the "--keep-guessing" option take the new path.

Despite that scope, the change still misses the kernels "m36400", "m36410", and "m36800", which call "mark_hash()" directly. They were added after this branch was created and remain untouched. We would have to remember to handle new modes manually, so the same gap could appear again.

The branch also has leftovers from earlier revisions. The sizing guards in "backend.c" cannot fire because they assign the cap and then check whether it exceeds itself. A comment still describes the old "digests_cnt * 256" calculation, and "inc_comp_multi_iba.cl" is modified even though no kernel includes it.

The change we would want is narrower, though not confined to a single file either. The gate would read a macro that defaults to 1 and is set from "OPTS_TYPE_SUGGEST_KG" for those 19 modules. We would still write that macro at the 66 places the gate already appears: 62 in "inc_simd.h", where the SIMD comparison unrolls per lane, 2 in "inc_scalar.cl", and one each in "inc_comp_multi.cl" and "inc_comp_single.cl". We would not need to touch "mark_hash()" itself, because the capacity already reaches it as a parameter. In return, every mode outside those 19 expands to what it expands to now, and no kernel changes apart from the three that call "mark_hash()" directly.

I am closing this PR because that would be a different change rather than another revision of the current approach. I would rather start fresh than ask you to work through another round here. Thank you for the work you put into this and for taking on the limitation in the first place.

PR #4732, opened 2026-07-29
Optimize LM mask kernel with staged DES rejection

jeffmcjunkin:
Optimizes mode 3000's pure mask kernel by staging DES output checks and deferring costly S-box work. Nonmatching candidates can stop before the final half-round instead of completing every S-box.

## Performance

| Mode | Improvement |
|---:|---:|
| 3000 | +12.9% |

## Validation

- Mode 3000 passes startup/self-test on CUDA and NVIDIA OpenCL.
- The change is limited to `m03000_a3-pure.cl`.

### DIFF
--- a/OpenCL/m03000_a3-pure.cl
+++ b/OpenCL/m03000_a3-pure.cl
@@ -250,6 +250,90 @@ DECLSPEC void s6 (const u32 a1, const u32 a2, const u32 a3, const u32 a4, const 
   *out4 ^= x4;
 }
 
+DECLSPEC void s1_early (const u32 a1, const u32 a2, const u32 a3, const u32 a4, const u32 a5, const u32 a6, PRIVATE_AS u32 *out1)
+{
+  LUT(xAA55AA5500550055, a1, a4, a6, 0xC1)
+  LUT(xA55AA55AF0F5F0F5, a3, a6, xAA55AA5500550055, 0x9E)
+  LUT(x5F5F5F5FA5A5A5A5, a1, a3, a6, 0xD6)
+  LUT(xF5A0F5A0A55AA55A, a4, xAA55AA5500550055, x5F5F5F5FA5A5A5A5, 0x56)
+  LUT(x947A947AD1E7D1E7, a2, xA55AA55AF0F5F0F5, xF5A0F5A0A55AA55A, 0x6C)
+  LUT(x55EE55EE55EE55EE, a1, a2, a4, 0x7A)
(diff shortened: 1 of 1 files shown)


jsteube:
In "m03000_a3-pure.cl" the guard around the round loop comes off:

```
-  #ifdef _unroll
   #pragma unroll
-  #endif
   for (u32 i = 0; i < 2; i++)
```

That define is a per device decision, not a constant. In "module_03000.c" it is passed for Intel CPU, AMD GPU-PRO, HIP and ROCm. NVIDIA matches no branch there, and Apple returns before reaching any of them. So on the 3090 you measured on, that loop was rolled before this change and unrolled after it, and the same goes for every Apple device.

The +12.9% therefore covers two changes at once, and the staged rejection is only one of them. Could you put the guard back and measure again? If the unrolling is where the gain is on NVIDIA, that is a one line change in "module_03000.c" adding NVIDIA to the list, which reaches the other 65 kernels written against the same guard rather than this one alone.

One thing worth a line of comment. In the recovery block, s2 is followed immediately by s2_early on the same two outputs, and so is s4:

```
      s2       (*D35 ^ k06, ..., D12, D27, D01, D17);
      s2_early (*D35 ^ k06, ..., D12, D01);
```

Could you say above it why the repeat is there? Without that, it reads as a duplicated call, and the next person has to count the XORs before they can touch the block.

jsteube:
I ran the split myself rather than leave you waiting on it. 4 builds on an RTX 4090, pure kernel, 3 reps alternated, spread under 0.2 percent inside each build:

```
master                                148.6  148.7  148.8 GH/s
this PR as it stands                  179.7  179.9  179.9 GH/s   +21.0%
this PR with the #ifdef _unroll back   162.3  162.3  162.8 GH/s   +9.2%
master plus NVIDIA in module_03000.c   162.0  162.0  162.3 GH/s   +9.0%
```

The staged rejection and the unrolling are worth about the same and they stack. Your +12.9 percent was both at once, and the staged rejection carries +9.2 of it on its own.

So the guard can go back without costing you anything, and the unrolling is better had the way the other 36 modules have it, as a line in module_03000.c beside the Intel and AMD ones:

```
  if (device_param->opencl_device_vendor_id == VENDOR_ID_NV)
    hc_asprintf (&jit_build_options, "-D _unroll");
```

That keeps the decision per device, which matters for Apple: it returns before reaching any of those branches today, and with the guard gone it would unroll whether that suits it or not.

PR #4733, opened 2026-07-29
Flatten the fixed-record RAR3 SHA-1 loop

jeffmcjunkin:
Optimizes the fixed-record RAR3 SHA-1 loop by compacting large-block staging and deriving each record counter independently. This removes nested-loop control and serial counter induction from the optimized kernel.

## Performance

| Mode | Improvement |
|---:|---:|
| 12500 | +7.0% |

The reported result is the lower of two isolated-branch confirmations; the other measured +15.6%.

## Validation

- Mode 12500 passes startup/self-test on CUDA and NVIDIA OpenCL.
- An independent RAR3 vector cracked correctly on both backends.
- The change is limited to `m12500-optimized.cl`; no module behavior is changed.

### DIFF
--- a/OpenCL/m12500-optimized.cl
+++ b/OpenCL/m12500-optimized.cl
@@ -25,6 +25,24 @@
 
 #define MIN(a,b) (((a) < (b)) ? (a) : (b))
 
+#if defined IS_NV
+
+#define RAR3_COUNTER_0(i,w) hc_byte_perm_S ((i), (w), 0x0124)
+#define RAR3_COUNTER_1(i,w) hc_byte_perm_S ((i), (w), 0x7012)
+#define RAR3_COUNTER_2(i,w) hc_byte_perm_S ((i), (w), 0x7601)
+#define RAR3_COUNTER_3(i,w) hc_byte_perm_S ((i), (w), 0x7650)
+#define RAR3_COUNTER_4(i,w) hc_byte_perm_S ((i), (w), 0x1254)
+
+#else
+
+#define RAR3_COUNTER_0(i,w) ((w) | iter_s)
+#define RAR3_COUNTER_1(i,w) ((w) | (iter_s >>  8))
+#define RAR3_COUNTER_2(i,w) ((w) | (iter_s >> 16))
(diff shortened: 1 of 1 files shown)


jsteube:
The title describes a flattened SHA-1 loop, and the change reaches further than that. The counter insertion moves from mask and shift to byte permutes, behind a vendor split master did not have:

```
+#if defined IS_NV
+#define RAR3_COUNTER_0(i,w) hc_byte_perm_S ((i), (w), 0x0124)
    ...
+#else
+#define RAR3_COUNTER_0(i,w) ((w) | iter_s)
    ...
+#endif
```

There is a new straight line path for p3 == 16 as well, and the 64 stored records become 4 plus a wrapped prefix, indexed cyclically:

```
+  const u32 j_step = (p3 <= 16) ? 16 - p3 : 16;
```

Could the description carry those three? It reads differently as a rewrite than as an unrolled loop, and whoever reads the commit later will want the same summary.

For what it is worth on hardware you did not have, it matches master across 180 combinations of the edge suite on AMD and on NVIDIA, and it is faster on all three cards:

```
  RX 7900 XTX     197.6 - 199.4  ->  231.2 - 232.0 kH/s     about +16 percent
  RTX 2080 Ti      45.7 -  47.2  ->   68.5 -  70.2 kH/s     about +49 percent
  RTX 4090        176.9 - 188.0  ->  333.7 - 348.2 kH/s     about +85 percent
```

jsteube:
Good news! The same KDF sits in two other modes. In "m23800-optimized.cl" the iteration core is byte for byte the one this PR rewrites, and "m23700-optimized.cl" carries it twice, once at each of its two entry points:

```
  const u32 p3 = pw_len + salt_len + 3;
  ...
    for (u32 j = 0; j < p3; j++)
  ...
        k += p3;
```

Those three are the only kernels in OpenCL/ that take a per iteration record length from pw_len + salt_len and advance a write offset by it, which is what makes the block alignment cycle and the precomputation possible.

Would 23700 and 23800 come along in this PR? RAR3-p is the format people meet more often than RAR3-hp, and the numbers are reaching only the latter at the moment.

PR #4734, opened 2026-07-29
Partially unroll the SHA3-256 round loop

jeffmcjunkin:
Optimizes the mode 17400 mask kernel by peeling the constant first SHA3-256 round, folding Theta application into the lane updates, and processing the remaining hot loop two rounds at a time.

## Performance

| Mode | Improvement |
|---:|---:|
| 17400 | +16.6% |

## Validation

- Mode 17400 passes startup/self-test on CUDA and NVIDIA OpenCL.
- The change is limited to `m17400_a3-optimized.cl`.

### DIFF
--- a/OpenCL/m17400_a3-optimized.cl
+++ b/OpenCL/m17400_a3-optimized.cl
@@ -80,13 +80,11 @@ DECLSPEC void m17400m (PRIVATE_AS u32 *w0, PRIVATE_AS u32 *w1, PRIVATE_AS u32 *w
       ad = hc_rotl64 (t, r); \
       t = bc0;               \
 
-    #ifdef _unroll
-    #pragma unroll
-    #endif
-    for (int round = 0; round < KECCAK_ROUNDS - 1; round++)
+    // Peeled first round: the initial state has 16 zero lanes plus one
+    // constant lane, so emitting round 0 as straight-line code lets the
+    // compiler fold those constants (Theta parities collapse, D-application
+    // into zero lanes becomes free) without unrolling the whole loop.
     {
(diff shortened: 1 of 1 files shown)


jsteube:
In "m17400_a3-optimized.cl" the new loop unrolls on every backend:

```
+    #pragma unroll 2
+    for (int round = 1; round < KECCAK_ROUNDS - 1; round++)
```

Master guarded that pragma with #ifdef _unroll, and "module_17400.c" has no module_jit_build_options() at all, so the old pragma applied nowhere and the new one applies everywhere.

What that costs depends on the vendor. Three reps of each, master and this PR alternated, -d 1, autotune shape recorded beside every number:

```
  RX 7900 XTX    2381 - 2392  ->  2162 - 2169 MH/s     about -9 percent
  RTX 4090       3962 - 4156  ->  4491 - 4667 MH/s     about +12 percent
  RTX 2080 Ti     913 -  969  ->   985 - 1061 MH/s     about +8 percent
```

The ranges do not overlap on any of the three, so each of those changes is real rather than measurement noise. Your comment reasons about ptxas and LOP3, which is an argument about one vendor, and the pragma lands on all of them.

Could the unrolling be gated so the AMD backends keep the rolled loop? In "module_29452.c" the same define is already gated to NVIDIA alone:

```
  // NVIDIA GPU
  if (device_param->opencl_device_vendor_id == VENDOR_ID_NV)
  {
    hc_asprintf (&jit_build_options, "-D _unroll");
  }
```

Adding that to "module_17400.c" would let the kernel keep its #ifdef _unroll.

PR #4735, opened 2026-07-29
Defer Bitcoin WIF basepoint work until validation passes

jeffmcjunkin:
Optimizes compressed Bitcoin WIF validation by rejecting invalid candidates before elliptic-curve basepoint setup. The mask kernel also hoists the invariant Base58 tail decode and uses a launch shape that better hides latency.

## Performance

| Mode | Improvement |
|---:|---:|
| 28501 | +168.8% |

## Validation

- Mode 28501 completes startup/self-test on CUDA and NVIDIA OpenCL with a valid 52-character candidate.
- The official WIF/password vector cracks correctly on both backends.
- Changes are limited to the three mode 28501 attack kernels; no module behavior is changed.

### DIFF
--- a/OpenCL/m28501_a0-pure.cl
+++ b/OpenCL/m28501_a0-pure.cl
@@ -36,10 +36,6 @@ KERNEL_FQ KERNEL_FA void m28501_mxx (KERN_ATTR_RULES ())
    * base
    */
 
-  secp256k1_t preG; // need to change SECP256K1_TMPS_TYPE above to: PRIVATE_AS
-
-  set_precomputed_basepoint_g (&preG);
-
   COPY_PW (pws[gid]);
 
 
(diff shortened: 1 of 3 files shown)


jsteube:
In "m28501_a3-pure.cl" a vendor macro is redefined for the rest of the file:

```
#if defined IS_CUDA || defined IS_HIP
#undef  KERNEL_FA
#define KERNEL_FA __launch_bounds__ (256, 3)
#endif
```

No other kernel in OpenCL/ redefines KERNEL_FA. For this mode it expands to an empty macro today, because FIXED_LOCAL_SIZE is unset and "module_28501.c" leaves module_kernel_threads_min and max at MODULE_DEFAULT, so the thread count is the autotuner's to pick. After this it is 256 on every CUDA and HIP device, and hashcat reads the bound back through CU_FUNC_ATTRIBUTE_MAX_THREADS_PER_BLOCK, so 256 becomes the ceiling the autotuner reads back rather than one it settled on. The HIP half of that is untested, since the numbers here are from a 3090.

Could the launch shape come out of this PR? If 256 threads and 3 blocks really is the best shape, it belongs in "module_28501.c" as a thread count, where the autotuner can still see it and other devices are not held to a constant picked on one card.

Separately, the validation listed is a self-test with a valid candidate, which walks the path that finds a hit. The tail decode and the prefix reconstruction are new arithmetic over hardcoded limbs of 58^48, and an error there costs candidates that should have matched, which a run that finds its one vector cannot show. There is a "m28501.pm" already, so tools/test_edge.sh -m 28501 would cover the reject paths as well. Could you run that and post the result?

PR #4736, opened 2026-07-29
Hoist invariant SM3 schedule work in the mask kernel

jeffmcjunkin:
Optimizes the mode 31100 mask kernel by hoisting W0-invariant SM3 schedule work, synthesizing sparse schedule deltas, and rejecting nonmatches through the inverse-P0 round-60 state.

## Performance

| Mode | Improvement |
|---:|---:|
| 31100 | +18.5% |

A second isolated-branch confirmation measured +21.0%.

## Validation

- Mode 31100 passes startup/self-test on CUDA and NVIDIA OpenCL.
- The shared SM3 helpers are additive; existing callers are unchanged.

### DIFF
--- a/OpenCL/inc_hash_sm3.cl
+++ b/OpenCL/inc_hash_sm3.cl
@@ -9,6 +9,25 @@
 #include "inc_common.h"
 #include "inc_hash_sm3.h"
 
+// PRMT-offload permutation functions (vector path):
+// P0/P1 each need two rotates of the same value. The second rotate is derived from the
+// first by an extra 8-bit rotate, which we emit as a byte permute (prmt.b32) instead of a
+// second funnel shift (shf). rotl(x,17)=rotl(rotl(x,9),8), rotl(x,23)=rotl(rotl(x,15),8).
+// Byte rotate is bit-exact, so results are identical; the intent is to move one rotate off
+// the saturated SHF pipe onto the PRMT unit. Selector 0x2103 = rotate-left-by-8 (cf. the
+// byteswap selector 0x0123). Temp binds the funnel shift so it is emitted exactly once.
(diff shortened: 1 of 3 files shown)


jsteube:
In "inc_hash_sm3.cl" the vector rotate uses a single selector on both AMD and NVIDIA:

```
+  return x ^ r ^ hc_byte_perm (r, r, 0x2103);        // rotl(x,17) = rotl(r,8) via prmt
```

On an RX 7900 XTX under HIP this builds without a diagnostic and then misses. Mode 31100 gives 27 errors in the edge suite against master's 0, every one of them exhausted rather than cracked, across attack modes 0 and 3 at vector widths 1, 2 and 4. On a single known vector, the empty password against 1ab21d83...82aa2b:

```
master    rc=0   Status: Cracked      Recovered: 1/1 (100.00%)
this PR   rc=1   Status: Exhausted    Recovered: 0/1 (0.00%)
```

Two things sit behind it. The function hc_byte_perm() is defined only under IS_AMD, IS_HIP or IS_AMD_USE_OPENCL and under IS_NV, with no generic path, so a platform that is neither loses the build outright. And 0x2103 is prmt.b32 nibble encoding, while the AMD side is __builtin_amdgcn_perm(), which takes four byte selectors, so the same constant means something else there. In "inc_hash_blake2s.cl" the function blake2s_rot08_S() spells that same rotate as 0x0321 on NV against 0x00030201 on AMD, behind per-vendor guards with a plain rotate fallback.

Could the two selectors and a fallback go in, following that one? The failure is in the constant, not in the schedule rework around it.

Separately, the header reaches mode 35100 as well, which the description does not mention. It compiles the changed macros and executes none of them, since it uses only the scalar path.

PR #4749, opened 2026-08-05
replaced scalar line counting with SIMD-accelerated implementation (AVX2/SSE2/NEON)

matrix:
Replace the scalar newline scan in count_lines() with a SIMD version (AVX2/SSE2/NEON), with a scalar fallback

### DIFF
--- a/src/filehandling.c
+++ b/src/filehandling.c
@@ -16,6 +16,24 @@
 #include <Xz.h>
 #include <XzCrc64.h>
 
+#if defined(__AVX2__)
+#include <immintrin.h>
+#define SIMD_MODE_AVX2
+#elif defined(__SSE2__) || (defined(_MSC_VER) && (defined(_M_X64) || defined(_M_IX86)))
+#include <emmintrin.h>
+#define SIMD_MODE_SSE2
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks for this, and sorry for the delay. I am closing it, but you were right about the gap and it is worth explaining what happened to it.

count_lines() was still scanning one byte at a time when you filed this. The vectorized scanner you added in 86be8dcb2 a year earlier, including the runtime dispatch, was already used by the wordlist feed but had never been connected to count_lines(). So this was not duplicating existing work. It was extending your own implementation to a path that had not reached it yet.

That happened two days later in 93b55d37d. Line reading was reworked to operate on blocks, and count_lines() now uses hc_memchr_get(), so the scan is dispatched at runtime to the AVX2, SSE2 or NEON implementation available on the machine.

I would rather keep that approach than take this patch because of how the implementation is selected. This version uses #if defined(AVX2) at compile time, so a binary built with AVX2 gets that path while one built elsewhere silently does not. That is the same kind of problem we had with -march=native in cross builds. Your 2025 implementation selects the best path at startup, which is the right model for binaries that will run on other machines.

So the speedup you were after is already in, using the runtime dispatch you originally built for it. Sorry it landed two days after you asked for it rather than through this pull request.

PR #4753, opened 2026-08-05
Build for the Raspberry Pi on an ARM64 runner

matrix:
Nothing in CI builds hashcat for ARM, so an aarch64 change is checked only insofar as it did not break x86. This adds two entries on an ARM64 runner, and fixes the Makefile so that what comes out of them runs on the hardware the job is named for. It is one workflow file and three build files, and no hash mode or kernel.

## Reproduction

No job reaches aarch64 today:

```
$ grep -n 'os_name:' .github/workflows/build.yml
85:            os_name: Linux
89:            os_name: Linux
93:            os_name: macOS
97:            os_name: Windows
101:            os_name: Windows
105:            os_name: FreeBSD
```

And the flags the first version of this job would have used produce a binary that faults on a Pi. This is the whole program:

```
$ cat lse.c
#include <stdio.h>
static int v = 0;
int main (void) { int r = __atomic_fetch_add (&v, 1, __ATOMIC_SEQ_CST); printf ("%d\n", r); return 0; }
```

Compiled once per flag set on a Raspberry Pi 4 with Debian clang 19.1.7, counting the LSE instruction and the outline atomics helper in the disassembly, reading the architecture back out of the driver, and then running it:

```
$ clang -O2 $flags -o lsebin lse.c
$ objdump -d lsebin | grep -oE '\bldaddal\b' | wc -l          # the instruction
$ objdump -d lsebin | grep -c '__aarch64_ldadd'               # the helper
$ clang -x c /dev/null -### -c -O2 $flags 2>&1 | tr ' ' '\n' | grep -oE '"\+v8[^"]*"'
$ ./lsebin

flag                                                 ldaddal  helper  feature  run
-march=armv8-a                                       1        4       +v8a     ok
-march=armv8.1-a                                     1        0       +v8.1a   SIGILL
-march=armv8.2-a -mtune=neoverse-n1 -mcpu=cortex-a72 1        0       +v8.2a   SIGILL
-mcpu=cortex-a72                                     1        4       +v8a     ok
-mcpu=cortex-a76                                     1        0       +v8.2a   SIGILL
```

The three that go wrong die the same way:

```
Illegal instruction
$ echo $?
132
```

`ldaddal` is in all five, and the `helper` column is what separates them: at `+v8a` the instruction sits inside the outline atomics helper, which tests for it at run time, and the four references to `__aarch64_ldadd` are that test. From `+v8.1a` up the helper is gone and `ldaddal` is emitted inline with nothing guarding it. The BCM2711 does not have it:

```
$ grep -m1 Features /proc/cpuinfo
Features        : fp asimd evtstrm crc32 cpuid
```

The third row is what `IS_RPI=1` on an ARM64 runner produces, because `-mcpu` sets the tuning and `-march=native` still sets the architecture, and on the runner that is the runner's own core. The fifth row is the same fault from the other direction: `cortex-a76` is the Pi 5 core, so a binary tuned for a Pi 5 does not run on a Pi 4.

## What changes

`MCPU=<core>` names the core to build for, for a build that has to run on a machine other than this one, and it replaces the native flags rather than joining them. On its own `-mcpu` carries the right architecture with it. Read back with `make -n` on the Pi, with the compiler lines trimmed to the flags that differ:

```
$ make -n                        # detection, unchanged from today
gcc -std=gnu99 -march=native -mtune=native -mcpu=cortex-a72

$ make -n MCPU=cortex-a72
gcc -std=gnu99 -mcpu=cortex-a72

$ make -n MCPU=cortex-a72        # the same command without this patch
gcc -std=gnu99 -march=native -mtune=native -mcpu=cortex-a72

$ make MCPU=cortex-a999
src/Makefile:474: *** MCPU=cortex-a999 is not a core gcc builds for with -mcpu.  Stop.
```

`SIMD_NATIVE` and the two Rust `.mk` files follow `MCPU` as well, or the parts of the artifact that do not take their flags from `CFLAGS` would still be built for the machine that ran the build. Those are the three bridges compiled with `SIMD_NATIVE`, `bridge_argon2id_reference`, `bridge_scrypt_jane` and `bridge_scrypt_yescrypt`, and the three crates, `Rust/bridges/dynamic_hash`, `Rust/bridges/generic_hash` and `Rust/feeds/random`. The counts are in the section below.

The Raspberry Pi branch keeps its own `grep Raspberry /proc/cpuinfo` gate, so no other aarch64 machine changes, and no longer hardcodes one core. `Raspberry Pi` is the board name for cores three generations apart, so the core comes from the `CPU part` line instead:

```
$ make -n CPU_PARTS=0xd03
gcc -std=gnu99 -march=native -mtune=native -mcpu=cortex-a53

$ make -n CPU_PARTS=0xd0b
gcc -std=gnu99 -march=native -mtune=native -mcpu=cortex-a76

$ make -n CPU_PARTS=0xdff
src/Makefile:539: Raspberry Pi with CPU part 0xdff, which this Makefile has no -mcpu name for
src/Makefile:540: building without it, which costs scheduling only; add MCPU_PART_0xdff or pass MCPU=
...
gcc -std=gnu99 -march=native -mtune=native
```

`0xd03` is the Pi 2 v1.2, Pi 3, Pi 3+ and Zero 2 W, `0xd08` the Pi 4, Pi 400 and CM4, `0xd0b` the Pi 5, Pi 500 and CM5. A part the file has no name for is a board newer than the file, so it warns and builds: `-march=native` already got the architecture right there and only the scheduling model is missing. The name is taken only when every core reports the same part, and only when the compiler accepts it.

`-mcpu=native` would answer the same question in one flag, and is not used because on a Pi 4 with clang 19 it answers it wrongly:

```
$ clang -x c /dev/null -### -c -mcpu=native      # the target features, filtered
"+aes" "+sha2" "+v8a"
```

## The builds

A Raspberry Pi 4 Model B Rev 1.5, Debian 13, gcc 14.2.0, 4 cores, on this branch merged with master at `46e9958ac`. Four full builds, both link modes against both ways of arriving at the core, each from `make clean`:

```
                                 rc   time   warn   mcpu   march=native   rust          test_package
SHARED=1, detection              0    632s   0      106    109            native        0 failures
SHARED=1, MCPU=cortex-a72        0    565s   0      109    0              cortex-a72    0 failures
SHARED=0, detection              0    557s   0      106    109            native        0 failures
SHARED=0, MCPU=cortex-a72        0    510s   0      109    0              cortex-a72    0 failures
```

Every cell built 596 modules, 6 bridges, 6 feeds and 3 Rust libraries, `bridges/subs/dynamic_hash.so`, `bridges/subs/generic_hash.so` and `feeds/rust_random.so`, and passed `tools/test_package.sh . --no-device` with no failures. The two `SHARED=1` cells also built `libhashcat.so.7`, which is the whole difference between the link modes.

```
$ echo hashcat | ./hashcat -m 0 --quiet -D 1 8743b52063cd84097a65d1633f5c74f5
8743b52063cd84097a65d1633f5c74f5:hashcat
```

## What an unset MCPU changes

Nothing. `MCPU` empty takes the `else` of the new `ifneq`, which holds the block that was there before, so a build that does not pass it is the build it was. On the Pi, `make -n` with the build timestamp and the version tag lines taken out, once without the patch and once with it:

```
$ make -n | grep -v 'COMPTIME\|VERSION_TAG' > before.txt
$ make -n | grep -v 'COMPTIME\|VERSION_TAG' > after.txt
$ diff before.txt after.txt && wc -l < before.txt
721
```

`diff` says nothing and the file is 721 lines, and neither run writes anything to stderr: the probe sits inside the new `ifneq`, so an unset `MCPU` never reaches it. Same on x86, where the pair is 726 lines and stderr is empty on both sides. And counted over the whole dry run, with `cargo` on the path so the crates show their flags:

```
                      -mcpu=cortex-a72   -march=native   target-cpu=native   target-cpu=cortex-a72
make -n      before   105                108             3                   0
make -n      after    105                108             3                   0
make -n MCPU before   105                108             3                   0
make -n MCPU after    108                0               0                   3
```

The first two columns are one lower here than in the build table above, because this filter drops the one line the build table keeps: the final link, which is where `COMPTIME` is passed.

The third row is the point: before the patch, `MCPU=cortex-a72` was a variable the Makefile did not read, so it changed nothing. The fourth is the same command after it. The crates follow:

```
$ make -n MCPU=cortex-a72 | grep -m1 -o 'RUSTFLAGS="[^"]*"'
before: RUSTFLAGS="-C target-cpu=native"
after:  RUSTFLAGS="-C target-cpu=cortex-a72"

$ make -n | grep -m1 -o 'RUSTFLAGS="[^"]*"'      # no MCPU, after the patch
RUSTFLAGS="-C target-cpu=native"
```

The same on x86_64, an i7-12700K with gcc 13.3.0, on the same merge:

```
$ diff before.txt after.txt && wc -l < before.txt
726

$ make -n | grep -m1 -o -- 'gcc -std=gnu99\( -m[^ ]*\)*'
before: gcc -std=gnu99 -march=native -mtune=native
after:  gcc -std=gnu99 -march=native -mtune=native

$ make -n | grep -m1 -o 'RUSTFLAGS="[^"]*"'
before: RUSTFLAGS="-C target-cpu=native"
after:  RUSTFLAGS="-C target-cpu=native"
```

`MCPU` on x86 is refused rather than taken. `-mcpu` is not an x86 flag: gcc there reads it as a deprecated spelling of `-mtune`, says so on stderr, and exits 0 anyway, so a probe on the exit code alone would accept it and the build would lose `-march=native`:

```
$ echo | gcc -mcpu=znver3 -E - > /dev/null ; echo $?
gcc: warning: '-mcpu=' is deprecated; use '-mtune=' or '-march=' instead
0
$ echo | clang -mcpu=znver3 -E - > /dev/null ; echo $?
clang: error: unsupported option '-mcpu=' for target 'x86_64-pc-linux-gnu'
1
```

The probe wants a clean run now, not only a zero exit, so the same command stops, and so does any other core, because what gcc objects to there is the flag and not the name:

```
$ make -n MCPU=znver3
src/Makefile:474: *** MCPU=znver3 is not a core gcc builds for with -mcpu.  Stop.
$ make -n MCPU=cortex-a72
src/Makefile:474: *** MCPU=cortex-a72 is not a core gcc builds for with -mcpu.  Stop.
```

Both of those are the only line the command prints. What the compiler says goes into the probe, which is the whole point of reading it.

On aarch64 the probe is unchanged in effect, because there the flag is native and the compiler says nothing:

```
$ echo | gcc -mcpu=cortex-a72 -E - 2>&1 >/dev/null      # on the Pi, no output
$ make -n MCPU=cortex-a72 | grep -m1 -o -- 'gcc -std=gnu99\( -m[^ ]*\)*'
gcc -std=gnu99 -mcpu=cortex-a72
```

## The job

Two entries on `ubuntu-24.04-arm`, static and shared, the same pair Linux already has, and both run `tools/test_package.sh . --no-device` like every other job, so the plugins have to load and a kernel has to build. They are named for what they build: the artifact name is `hashcat-${{ matrix.os_name_lowercase }}-${{ matrix.shared == 0 && 'static' || 'shared' }}`, so `os_name: Linux-ARM64-RPi4` gives `hashcat-linux-arm64-rpi4-static` and `hashcat-linux-arm64-rpi4-shared` rather than a generic `hashcat-linux-arm64` over a binary tuned for one board. `ubuntu-22.04-arm` was the label in the first version and is not used: actions/runner-images#14254 begins deprecating the Ubuntu 22 images on September 17th 2026, failing jobs on purpose during the brownouts, with retirement on April 17th 2027. There is no `ubuntu-latest-arm`, so the ARM entries name a version where the x86 ones say `ubuntu-latest`.

Python is set up because the image ships 3.12.3 and `src/bridges/bridge_python_generic_hash_sp.mk` sets its own floor at 3.13 and skips itself below it, so without that step the job would build `-m 73000` and silently skip `-m 72000`. `AR=llvm-ar` earns its place on the static entry only, which is the one that calls an archiver:

```
$ make -n ENABLE_LTO=0 SHARED=0 MCPU=cortex-a72 CC=clang CXX=clang++ AR=llvm-ar | grep -c llvm-ar
1
$ make -n ENABLE_LTO=0 SHARED=1 MCPU=cortex-a72 CC=clang CXX=clang++ AR=llvm-ar | grep -c llvm-ar
0
```

`clang` is on the image and neither `lld` nor `llvm` appears in its readme, which is what the `apt-get install` line is for. Whether they are genuinely absent rather than merely unlisted, and which core that runner reports, are the two things only a first run can answer.

## What this leaves out of the earlier version

The branch used to carry four other changes to the same file. The `paths` filter is one pull request now and the mingw Python with the action versions is another. The OpenMP dependencies are dropped: no `pragma omp` and no `-fopenmp` remain under `src/` or `include/`. The separate `build_binaries_arm` job is gone too, because it ran a native `make` on the same runner as the matrix entries and produced the same binary under a second artifact name.


jsteube:
From the description:

```
On a Pi those agree, on a runner whose core is not an A72 they do not.
```

They combine rather than one winning. Clang takes the tuning from "-mcpu" and the architecture from "-march":

```
-mcpu=cortex-a72                              target-cpu cortex-a72  +v8a
-march=armv8.2-a -mtune=neoverse-n1 -mcpu=..  target-cpu cortex-a72  +v8.2a
```

On the runner "-march=native" is the runner's own core, so the artifact comes out built for ARMv8.2-A and tuned for a Cortex-A72. A Pi 4 is ARMv8-A, so that binary can fault on the hardware the job is named for. The Pi cannot show it, because "-march=native" is armv8-a there and the two agree.

So I would drop "IS_RPI=1" before this goes in. That answers both halves: the build then takes the runner's own core, which is what a runner-built artifact should be, and the name it already ships under says the same thing:

```
name: hashcat-${{ matrix.os_name_lowercase }}-...
```

That reads hashcat-linux-arm64, so a Pi tuned binary is under a generic name either way.

matrix:
You are right about the mechanism, and it is worse than the description said. I read the driver's own decision back out of `clang -###` on a Pi 4, clang 19.1.7:

```
flag                                                 target-cpu      feature
-mcpu=cortex-a72                                     cortex-a72      +v8a
-mcpu=cortex-a53                                     cortex-a53      +v8a
-mcpu=cortex-a76                                     cortex-a76      +v8.2a
-march=native -mtune=native                          generic         +v8a
-march=native -mtune=native -mcpu=cortex-a72         cortex-a72      +v8a
-march=armv8.2-a -mtune=neoverse-n1 -mcpu=cortex-a72 cortex-a72      +v8.2a
```

Your line is the last one, and the fault it implies is real. The two middle rows are why the Pi branch cannot keep one hardcoded core either. LSE, the atomics ARMv8.1-A added, is enough to show it because a compiler emits them for ordinary C. This program is one `__atomic_fetch_add`:

```
flag                                                 ldaddal  helper  feature  run
-march=armv8-a                                       1        4       +v8a     ok
-march=armv8.1-a                                     1        0       +v8.1a   SIGILL
-march=armv8.2-a -mtune=neoverse-n1 -mcpu=cortex-a72 1        0       +v8.2a   SIGILL
-mcpu=cortex-a72                                     1        4       +v8a     ok
-mcpu=cortex-a76                                     1        0       +v8.2a   SIGILL
```

`ldaddal` is in all five. The `helper` column is what separates them. At `+v8a` it sits inside the outline atomics helper, which tests for the instruction at run time, and the four references to `__aarch64_ldadd` are that test. From `+v8.1a` up the helper is gone and `ldaddal` is emitted inline with nothing guarding it, so the process dies with 132 on a BCM2711, whose `/proc/cpuinfo` lists `fp asimd evtstrm crc32 cpuid` and no `atomics`.

Your first half is fixed by replacing the flags rather than by dropping them. `IS_RPI=1` added `-mcpu` beside the `-march=native` the Makefile had already set, and that is the combination your third row measures. `MCPU=<core>` sets `-mcpu` and leaves the native flags out, so nothing is inherited from the runner and the architecture comes from the core name, which is the first table's first row: `-mcpu=cortex-a72` is `+v8a`. The artifact cannot come out ARMv8.2-A that way, whatever the runner is.

Your second half is right and the name was wrong. The artifact name is `hashcat-${{ matrix.os_name_lowercase }}-${{ matrix.shared == 0 && 'static' || 'shared' }}`, so with `os_name: Linux-ARM64` those two entries shipped a Pi tuned binary as `hashcat-linux-arm64`. They are `Linux-ARM64-RPi4` now, which reads `hashcat-linux-arm64-rpi4-static` and `hashcat-linux-arm64-rpi4-shared`, and the three steps that test `os_name_lowercase` follow. The name says what is in it.

That is the fix I chose over dropping the core, because a build for the Pi is what the job is for, and `-march=native` on the runner is not the generic arm64 artifact it looks like either: it is a binary for the runner's own core, which is the architecture your comment measures at ARMv8.2-A, and it will not run on a Pi 4. If you would rather have the generic one, that is dropping `MCPU` as well as `IS_RPI`, and I will do that instead.

```
$ make -n                        # on a Pi, detection, unchanged from today
gcc -std=gnu99 -march=native -mtune=native -mcpu=cortex-a72

$ make -n MCPU=cortex-a72
gcc -std=gnu99 -mcpu=cortex-a72
```

The fifth row of the second table is also the reason the Pi branch no longer hardcodes `cortex-a72`. `Raspberry Pi` is the board name for cores three generations apart, so one name tunes two of the three for the wrong machine, and a binary tuned for a Pi 5 does not run on a Pi 4 any more than a runner-built one does. The core now comes from the `CPU part` line, `0xd03` to `cortex-a53`, `0xd08` to `cortex-a72`, `0xd0b` to `cortex-a76`. A part the file has no name for warns and builds without `-mcpu`, and an `MCPU` the compiler does not know is a `$(error)`. The test stays inside the `grep Raspberry /proc/cpuinfo` gate, so no other aarch64 machine changes.

One thing verifying it turned up: the test for whether the compiler takes `-mcpu` was on the exit code alone, and `-mcpu` is not an x86 flag. gcc there reads it as a deprecated spelling of `-mtune`, says so on stderr, and exits 0, so `MCPU=` on an x86 build would have been accepted and would have dropped `-march=native`. clang on x86 refuses it outright. The test wants a clean run now, so `make MCPU=znver3` on x86 stops with `src/Makefile:474: *** MCPU=znver3 is not a core gcc builds for with -mcpu.  Stop.`, and on aarch64 nothing changes because there the compiler says nothing. With `MCPU` unset, `make -n` is identical before and after the patch on both, 726 lines on x86 and 721 on the Pi.

`SIMD_NATIVE` and the two Rust `.mk` files follow `MCPU` as well. Those are the three bridges built with `SIMD_NATIVE`, `bridge_argon2id_reference`, `bridge_scrypt_jane` and `bridge_scrypt_yescrypt`, and the three crates, `Rust/bridges/dynamic_hash`, `Rust/bridges/generic_hash` and `Rust/feeds/random`, and without it they would have been the one part of the artifact still built for the machine that ran the build.

Four full builds on that Pi, both link modes against both ways of arriving at the core:

```
                                 rc   time   warn   mcpu   march=native   rust          test_package
SHARED=1, detection              0    632s   0      106    109            native        0 failures
SHARED=1, MCPU=cortex-a72        0    565s   0      109    0              cortex-a72    0 failures
SHARED=0, detection              0    557s   0      106    109            native        0 failures
SHARED=0, MCPU=cortex-a72        0    510s   0      109    0              cortex-a72    0 failures
```

One thing you did not ask about and I changed anyway: the entries were on `ubuntu-22.04-arm`, and actions/runner-images#14254 has Ubuntu 22 images beginning deprecation on September 17th with jobs failed on purpose during the brownouts before retirement in April. They are on `ubuntu-24.04-arm` now, which is the version `ubuntu-latest` resolves to for the x86 entries beside them.

PR #4756, opened 2026-08-05
fix --segment-size bug

### LINKED ISSUE #4741: Coordinated Vulnerability Disclosure - Integer Underflow via --segment-size=0 in Wordlist Segment Allocation
’m a vulnerability analyst at [VulnCheck](https://www.vulncheck.com/), an exploit intelligence company and research CVE Numbering Authority (CNA), where I'm one of several folks who manage our coordinated vulnerability disclosure (CVD) program.

An external security researcher recently [reported](https://www.vulncheck.com/advisories/report) a potential vulnerability impacting the hashcat and VulnCheck is acting as the intermediary and coordinator.

VulnCheck follows a [120-day disclosure policy](https://www.vulncheck.com/vulnerability-disclosure-policy), meaning we afford vendors/maintainers up to 120 days from the time of receiving the report to address the issue before publishing a CVE record and third-party advisory. For this vulnerability, the 120-day deadline falls on November 28, 2026.

The following CVE ID has been provisionally reserved for reference purposes:

CVE-2026-68769 - Integer Underflow via --segment-size=0 in Wordlist Segment Allocation

Reservation does not constitute publication. The ID will remain in a non-public RESERVED state with no technical details visible in the CVE database until we have aligned on CVE-eligibility, disclosure details, and a timeline, or until the 120-day coordinated disclosure window closes.

They have provided us with a comprehensive technical report, which we have attached below. Let us know if you have any questions for us about the CVD process or for the researcher regarding the reported vulnerability.

________________________________________________________________________________________

Integer Underflow via --segment-size=0 in Wordlist Segment Allocation

Package: hashcat/hashcat Tested Versions: v7.1.2-382-g2d71af371 Affected Files: src/wordlist.c lines 62-69 (underflow + OOB write); src/user_options.c line 2299 (source of 0); src/wordlist.c lines 747-749 (propagation) CWE: CWE-191: Integer Underflow; CWE-787: Out-of-Bounds Write CVSS 3.1: AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H -- 7.8 High

Affected Files
src/user_options.c
  Line 577    user_options->segment_size = hc_strtoul(optarg, NULL, 10);
              -- value taken verbatim; no lower-bound check
  Line 2299   user_options->segment_size *= (1024 * 1024);
              -- 0 * anything = 0; still 0 after scaling

src/wordlist.c
  Lines 747-749  wl_data->buf  = (char *) hcmalloc(user_options->segment_size);  // hcmalloc(0)
                 wl_data->incr = user_options->segment_size;                      // 0

  Lines 62-69    wl_data->cnt = hc_fread(wl_data->buf, 1, wl_data->incr - 1000, fp);
                 -- wl_data->incr - 1000 = 0 - 1000 = 0xFFFFFFFFFFFFFC18 (~16 EiB)
                 -- reads entire wordlist into zero-size buffer
                 wl_data->buf[wl_data->cnt] = 0;
                 -- null write past end of 1-byte allocation: OOB
Root Cause

--segment-size is parsed as an unsigned value with no lower-bound validation. A value of 0 passes the *= 1024*1024 scaling unchanged and propagates into wl_data->incr (a u64). In load_segment(), the read size is computed as wl_data->incr - 1000, which underflows to 0xFFFFFFFFFFFFFC18 (~16 EiB). hcmalloc(0) returns a 1-byte allocation; hc_fread reads the entire wordlist into it; the subsequent buf[cnt] = 0 null-termination write lands outside the allocation. For larger wordlists the following hcrealloc(buf, 0, 0) / NULL path adds use-after-free and NULL dereference behavior.

Chain with Finding 002: --segment-size=0 can be injected remotely via a crafted restore file (Finding 002 — argument injection), elevating this from a local finding to a remote one in that scenario.

PoC
bash
echo '5f4dcc3b5aa765d61d8327deb882cf99' > /tmp/hash.txt   # md5("password")
echo 'password'                          > /tmp/wl.txt

hashcat -a 0 -m 0 --segment-size=0 /tmp/hash.txt /tmp/wl.txt --force --self-test-disable
# -> SIGSEGV (ASAN: heap-buffer-overflow)

ASAN output:

ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7bf0349e0070
WRITE of size 1 at 0x7bf0349e0070 thread T0
    #0 hashcat+0xfaffa
    #1 hashcat+0xfd885
    ...
0x7bf0349e0070 is located 0 bytes inside of 1-byte region
  [0x7bf0349e0070, 0x7bf0349e0071)
  allocated by hcmalloc(segment_size) == hcmalloc(0)
Impact

Heap out-of-bounds write into a 1-byte allocation: hc_fread reads the entire wordlist file into a zero-size buffer (bounded by file size, not the intended segment), and the null-termination write buf[cnt] = 0 goes past the allocation. For larger wordlists, hcrealloc(buf, 0, 0) and NULL dereference behavior are also triggered. When chained with Finding 002 (restore file argument injection), --segment-size=0 can be delivered via a crafted restore file, making this remotely exploitable without direct command-line access.

matrix:
fix #4741

### DIFF
--- a/src/user_options.c
+++ b/src/user_options.c
@@ -1084,6 +1084,16 @@ int user_options_sanity (hashcat_ctx_t *hashcat_ctx)
     return -1;
   }
 
+  if (user_options->segment_size_chgd == true)
+  {
+    if (user_options->segment_size == 0)
+    {
+      event_log_error (hashcat_ctx, "Invalid --segment-size value specified - must be greater than 0.");
+
+      return -1;
+    }
+  }
+
   if (user_options->bitmap_max > 31)
   {
     event_log_error (hashcat_ctx, "Invalid --bitmap-max value specified - must be lower than 32.");


jsteube:
Thanks for this. The bug is real: in v7.1.2, --segment-size=0 reaches wl_data->incr, load_segment underflows incr - 1000 to roughly sixteen exabytes, and the terminating write goes past a one byte allocation. The report is accurate for the released version.

I am closing it because that code no longer exists on master. The wordlist segment reader, wl_data and incr were removed by the feed rework. Wordlists are now mapped directly in src/feeds/feed_wordlist.c, so the arithmetic that underflowed has no remaining path.

segment_size itself survived only as an unused option. It is initialized, set by -c, converted to bytes and logged, but nothing reads it afterwards. Adding a guard for -c 0 would therefore imply that other values still have an effect when they do not.

The real follow up is to decide what happens to --segment-size itself. The help still describes behavior that no longer exists, so the option should either be removed as a backward incompatible change or kept as an explicit no op. That is separate from this patch.

PR #4758, opened 2026-08-06
HIP: select the Hygon DTK ABI at runtime

lql341:
Hygon DTK exposes a legacy HIP ABI through `libgalaxyhip.so` that differs from the ROCm ABI hashcat currently targets: the `hipDeviceAttribute_t` enum values and the 792-byte `hipDeviceProp_t` layout come from `hip_runtime_defines.h` rather than the newer ROCm layout in `include/ext_hip.h`. A default build therefore fails HIP device initialization with `hipDeviceGetAttribute(): invalid argument`.

This PR selects the DTK ABI at runtime instead of with a compile-time switch.

Changes:

- `src/ext_hip.c`: try `libamdhip64.so` (and its sonames) first, then `libgalaxyhip.so`, and record `is_dtk` when the DTK library loads. Keep the HIP DSO loaded on the DTK path because it registers C++ exit handlers that crash after `dlclose`.
- `include/ext_hip.h`: keep the ROCm `hipDeviceAttribute_t` / `hipDeviceProp_t` definitions, add the DTK layout as `hipDevicePropDTK_t`, and map attribute values / properties when `is_dtk` is set.
- `src/ext_hiprtc.c`: inherit `is_dtk` from the HIP runtime and skip `hc_dlclose` on the DTK path.
- `src/backend.c`: drop the unused `regsPerMultiprocessor` workaround (the field is written but never read).

Validation (Hygon BW1000, gfx936, DTK 25.04.2 / HIP 6.3.25422, 80 CUs, 64 GB VRAM):

- HIP device enumeration
- MD5 / NTLM benchmarks
- phpass dictionary attack
- clean process exit after HIP use

Also checked on a Hygon Z100 (gfx906) with DTK 26.04 / HIP 6.3.26113: the attribute enum values and the 792-byte `hipDeviceProp_t` layout match DTK 25.04.2, so the same runtime selection applies.


lql341:
Hi maintainers, just following up on this PR. It adds the opt-in `DTK_HIP=1` path for Hygon DTK 25.04.2 and includes validation on a Hygon BW1000/gfx936 system. There have been no review comments yet, so I wanted to check whether this approach is suitable for hashcat or whether you would prefer changes to the ABI integration, build switch, or test coverage. I am happy to adjust it based on your feedback.

jsteube:
Sorry for the long silence, and thanks for following up. This has not been ignored. There are 3 separate questions here, and I would rather explain them clearly than leave the pull request sitting any longer.

1. We cannot test Hygon hardware

Nobody on the hashcat team has access to a Hygon system. We cannot reproduce your results, verify the ABI values against a real device, or notice later if support breaks. Every other backend in the tree is something one of us can put on a test bench.

That does not mean the patch cannot be accepted. It does mean the implementation has to keep the untested path isolated from the paths we can test. A compile time switch does the opposite.

2. Runtime selection fits better

DTK_HIP=1 places the complete hipDeviceAttribute_t enum and hipDeviceProp_t layout behind #if defined (HC_HIP_DTK). The resulting binary supports either the AMD HIP ABI or the Hygon HIP ABI, depending on how it was built. A distribution would need to ship 2 hashcat binaries, and the additional build would be the one we cannot test. Anyone downloading the normal release would still be unable to use a Hygon device.

hashcat already loads HIP at runtime through hip_init() in src/ext_hip.c. The compression loaders now use the same general model. src/ext_zstd.c, src/ext_lzma.c and src/ext_zlib.c each provide a list of library names, and hc_dynlib_open() tries them until one succeeds. A missing name is not an error by itself.

I think the same design fits here. Try libamdhip64.so first, then libgalaxyhip.so, remember which library was loaded, and select the matching attribute values and property layout. Both ABI definitions can remain in the tree, with the correct one chosen once the runtime is known.

That gives users one binary that works with both implementations. It also keeps the Hygon path separate from AMD at runtime, which makes the change much safer for hardware we can test.

3. The unload behavior needs an explanation

This part needs an answer whichever implementation we choose. In src/ext_hip.c and src/ext_hiprtc.c, the patch skips hc_dlclose() when HC_HIP_DTK is defined. The library handle is deliberately left open, but there is no comment explaining why.

My assumption is that libgalaxyhip crashes or hangs during unload. If that is the reason, please document it next to the condition. That is important behavior and a future reader should not remove it as an apparent leak. If the reason is different, please explain what happens.

4. One part can simply be dropped

I checked the backend.c change expecting a problem and there is none. prop has already been filled by an earlier hc_hipGetDeviceProperties() call in the same function, so reading it there is safe. The current code calls that function 3 times to read 3 fields, and that redundancy is ours rather than yours.

The workaround itself is not needed at all though. device_param->regsPerMultiprocessor is written in 2 places and read nowhere in the tree. Nothing consumes it. The field that matters is device_param->regsPerBlock, which is read at backend.c:18209 and :18234 for the threads per block calculation, and your patch does not touch it.

So that whole #if block can go. Removing it changes nothing on any device, and it takes 1 of the 3 compile time branches out of the patch. The dead field is our mess, not something you introduced.

5. Where this leaves the PR

The branch is now 143 commits behind master and conflicts in src/Makefile, so it needs a rebase in any case. If you are willing to change the implementation to select the ABI at runtime, I will review that version properly.

I also want to be honest about the limitation. We still would not be able to verify the Hygon path ourselves. The support would need to be documented that way, and if it breaks later we would depend on you or another Hygon user to report and test the fix.

Keeping this as a maintained patch against hashcat releases is also a reasonable outcome. I would rather say that clearly now than leave you waiting for a merge that was never going to be straightforward.

lql341:
Thanks for taking the time to explain this so clearly — I really appreciate the detailed guidance. I've reworked the branch accordingly and pushed it.

To respond to each of your points:

1. **Untested hardware and isolating the path.** You're right that a compile-time switch does the opposite of isolating the untested Hygon path. The new version has no compile-time switch at all: the AMD path is unchanged, and the Hygon path is only entered at runtime when `libgalaxyhip.so` is actually loaded.

2. **Runtime selection.** I followed the `hc_dynlib_open`-style model you suggested. `hip_init()` now tries `libamdhip64.so` and its sonames first, then `libgalaxyhip.so`, and records the result in `hip->is_dtk`. The ROCm `hipDeviceAttribute_t` / `hipDeviceProp_t` definitions remain the single compiled-in definitions; the DTK layout lives in a separate `hipDevicePropDTK_t`, and the mapping is applied only on the DTK path, so one binary serves both runtimes.

3. **Unload behavior explanation.** I've added comments at both close paths. The reason is what you suspected: DTK registers C++ exit handlers that call back into the DSO, so `dlclose()` leaves them pointing at unmapped memory and the process crashes at exit. HIPRTC inherits `is_dtk` from the HIP runtime and skips the close for the same reason.

4. **Dropping the `backend.c` workaround.** Agreed and done — the `regsPerMultiprocessor` block is removed. Thanks for pointing out the field is write-only.

5. **Rebase and maintenance expectations.** The branch is now rebased onto current master and the `src/Makefile` conflict is resolved. I also understand and accept that you can't verify the Hygon path yourselves; I'm happy to keep it tested and fixed from the Hygon side when needed.

One extra note on the ABI itself: the same attribute values and 792-byte `hipDeviceProp_t` layout also hold on DTK 26.04, so the runtime selection isn't tied to a single DTK release.

Would you be willing to take another look when you have time? No rush, and I'm happy to adjust anything further.

jsteube:
This is moving in the right direction. I have two remaining points before it can be merged.

1. Try the versioned libgalaxyhip sonames

The ROCm loader immediately above this code does not rely on libamdhip64.so alone. Its comment explains why: the unversioned name is normally a symlink from the development package, while a runtime installation may contain only the versioned soname.

The same may apply to DTK. If Hygon ships libgalaxyhip.so.1 without the unversioned symlink in its runtime package, hashcat will fail to load HIP and the user will only see that the device was not found. Please try the versioned sonames as well, following the existing libamdhip64 loop.

If DTK guarantees that libgalaxyhip.so is always present, document that next to the lookup. I would rather have that guarantee stated than leave the loader depending on an assumption.

2. Explain the hipGetDevicePropertiesDTK function pointer

hipGetDevicePropertiesDTK resolves the same exported symbol as hipGetDeviceProperties but calls it through a different function pointer type. I understand why: the function is the same and only the output structure layout differs between the two ABIs.

Strictly speaking, however, calling a function through an incompatible pointer type is undefined behavior in C. A short comment explaining that the symbol and calling convention are identical and that only the structure layout differs would make the assumption clear to the next person reading this code.

--

One limitation from my earlier review still applies. Nobody on the hashcat team has access to a Hygon device, so this support will be merged as unverified and documented that way. If it stops working, we will depend on you or another DTK user to report and test the problem.

The important improvement is that the DTK ABI is now selected or not at runtime. When libamdhip64 is loaded, the DTK specific path is never entered, so code we cannot test does not alter the AMD path we can test. That isolation is what makes accepting unverified support reasonable.

lql341:
Thanks for the review. Both points are addressed in the new push:

1. The loader now tries `libgalaxyhip.so.N` as well as the unversioned name.
2. The `hipGetDevicePropertiesDTK` function pointer now has a comment explaining that the symbol and calling convention are identical, only the output layout differs.

One additional finding from testing on a Hygon Z100 (gfx906) with DTK 26.04: DTK also exposes `libamdhip64.so` as a symlink to `libgalaxyhip.so`. With the previous `libamdhip64`-first order, hashcat loaded the DTK legacy ABI through the AMD soname, so `is_dtk` was not set and the device failed with `hipDeviceGetAttribute(dev=0, attrib=16): invalid argument`, then crashed. The pushed change loads `libgalaxyhip.so` first and falls back to `libamdhip64.so`. After that change, backend-info and MD5 benchmark both exit 0 on the same node.

This reorder reverses the earlier `libamdhip64`-first suggestion, so I wanted to confirm that shape is acceptable.

jsteube:
Both points are addressed, thanks. And the symlink finding is the useful part of this whole exchange: DTK shipping libamdhip64.so as an alias for libgalaxyhip.so means the soname asserts an ABI compatibility the library does not have.

On the reorder itself. It works, and I understand why you had to make it. What bothers me is that it changes what the code relies on: before, libgalaxyhip.so was only reached after every AMD name had failed, so an AMD machine provably never entered the DTK path. Now the first thing every AMD user's hashcat does is probe for a library that only exists on Hygon. It is harmless in practice, since the dlopen just fails and falls through, but the safety went from structural to circumstantial, and the reason it is safe is now "that file is not on your system" rather than "that code is not reached".

I would rather identify the runtime after opening it than guess from the order of names. That keeps libamdhip64 first, leaves the AMD path exactly as it is today, and cannot be defeated by a symlink. There are 3 ways to do it and I do not know which is available on your hardware, so could you check:

Is libamdhip64.so in the DTK tree a real file or a symlink, and what does each of them resolve to? readlink -f on both would settle it.

Does libgalaxyhip.so export any symbol that ROCm's libamdhip64.so does not? nm -D --defined-only on both and a diff. If there is a unique symbol, a dlsym on the handle we already hold is a definitive test, and it does not care how the file was named or found.

What does hipRuntimeGetVersion return on DTK? hashcat already loads that function, so if the value is distinctive we need no new binding at all, just a comparison after the library opens.

Any one of those is enough. And if DTK always ships that alias, which is what the first question above really decides, the libgalaxyhip search can come out of the patch entirely: DTK hands us libamdhip64.so anyway, so we open the name we already open, ask the library what it is, and pick the ABI from the answer. That removes the reorder, the second soname loop and the whole ordering question together, which makes the change easier to maintain and is a better argument for runtime detection over the compile time flag than anything in my first review.

lql341:
Thanks again — I've pushed a new version that follows this direction exactly: `libamdhip64.so` is tried first again, the `libgalaxyhip` name search and the reorder are gone, and after the library opens the code asks it which runtime it is.

Results of the three checks across DTK 24.04.3 / 25.04.4 / 26.04:

1. `readlink -f` resolves `libamdhip64.so` to `libgalaxyhip.so.5.2.<build>...` in all three releases, and `cmp` says the two files are byte-identical. DTK always hands us the alias.
2. `nm -D --defined-only` across the three releases gives a stable DTK-only export, `hipExtGetNearstCPU` (that spelling, without the second `e`), which is absent from the official ROCm 5.7.3 and 6.3.2 `libamdhip64.so` I checked. The patch uses `dlsym` for that symbol on the already-open handle to set `is_dtk`.
3. `hipRuntimeGetVersion` is not distinctive enough on its own: DTK returns `50724405 / 60325521 / 60326113`, while ROCm 5.7.3 / 6.3.2 return `50731921 / 60342134` — both sides encode a build number in the patch field.

Verified on a Hygon Z100 (gfx906) and a Hygon BW1000 (gfx936), both with DTK 26.04: `./hashcat -I` exits 0 and lists the device, and `./hashcat -b -m 0 -D 2` exits 0 (about 27.4 GH/s on gfx906, 39.5 GH/s on gfx936). Because `is_dtk` is set from a symbol in the loaded DTK runtime rather than from the device ISA, the same branch applies to the other DTK parts that use this runtime.

Your structural-versus-circumstantial point was exactly right, and opening the library and asking it what it is turns out to be a much cleaner solution than the reorder. Thank you for the careful, patient reviews — the guidance has genuinely improved this patch.

lql341:
On the unverified status — understood and accepted. I'll keep this tested from the DTK side; I now have both Hygon Z100 (gfx906) and BW1000 (gfx936) machines available, so I can cover both device generations when something needs checking. And I agree the runtime selection is the exact isolation that makes accepting this reasonable: the DTK path is only entered after the loaded library identifies itself as DTK, so the AMD path remains the only one the team has to test.

jsteube:
The symlink answer is the one that settles the design. DTK reaches its runtime through libamdhip64.so, so the second name search was solving a problem that does not exist, and the loader is back to one list of names with the ABI decided after the library is open. An AMD machine never enters the DTK path now because ROCm does not export the symbol, not because of the order the names happened to be tried in.

hipExtGetNearstCPU is missing a letter, and that misspelling carries more weight than it looks. It is the reason ROCm is unlikely to ever export the same name, so the detection is sound today. It is also the whole of the detection: if Hygon corrects the spelling in a later DTK, is_dtk goes false on DTK hardware and the first symptom is the invalid argument failure you opened this pull request with. Probing both spellings costs a line, and I would like it in before this merges.

Nothing in the patch reports which ABI it selected. Nobody here can run DTK, so a misdetection reaches us as a crash with no sign of which path was taken, and a message when the DTK runtime is identified would close that gap. Emit it only in that case. The AMD library is the ordinary one, so naming it on every AMD machine would suggest something unusual had happened when nothing has.

lql341:
Both points are addressed in the new push: `is_dtk` now probes both `hipExtGetNearstCPU` and the corrected `hipExtGetNearestCPU`, and the loader logs when it selects the DTK ABI (only on DTK). I'll also report the misspelled symbol to the DTK team so the correction can happen upstream.

jsteube:
Both points are in, and probing the corrected spelling next to the current one is the right shape. Reporting the typo to the DTK team is worth doing as well, because the second probe is what keeps hashcat working on the day they fix it.

The message needs a guard before this goes in. event_log_info called from ext_hip.c reaches main_log_info, which sends output to stderr only under --stdout and to stdout the rest of the time, and neither main_log nor the event layer looks at --quiet. hashcat guards informational output at the handler instead, the way main_backend_device_init_pre returns early when quiet is set.

On a Hygon node that means hashcat --keyspace --quiet prints the DTK line above the keyspace number, and a caller reading that number gets the wrong thing. --show and --left have the same shape. Hashtopolis and anything else driving a cluster parses exactly that output, and Hygon boxes in a cluster are the case this pull request exists for. A quiet check around the message settles it, and hip_init takes hashcat_ctx as a void pointer, so user_options has to come off it first.

With that in I will merge it.

lql341:
Done — the DTK ABI message now checks `user_options->quiet` the same way the other informational handlers do, so `--keyspace --quiet`, `--show` and `--left` keep their machine-readable stdout. The branch is updated.

jsteube:
Top, thanks!

PR #4761, opened 2026-08-06
Fix use of rule_buf_out after its scope has ended in slow_candidates

matrix:
Five loops in slow_candidates_seek and slow_candidates_next declared rule_buf_out inside the while body and read it through line_buf after the break, past its scope


matrix:
nothing to fix after commit 24d67b5ee8d0f623259d83f93a11cf7f9aca3c12

PR #4762, opened 2026-08-08
modules: add ZSTD decompression support to the 7-Zip module (mode 11600)

Igloo-Garage:
## Summary
Some 7-Zip archives use the ZSTD codec (e.g. created with mcmilk's
7-Zip-zstd builds, codec id `0x04f71101`). hashcat's 7-Zip module
(mode 11600) only supports LZMA1/LZMA2/DEFLATE, so ZSTD-compressed
archives cannot be cracked. This PR adds ZSTD support.

## Changes
- Define `data_type 8` as ZSTD for the `$7z$` hash format
- Decompress with master's streaming zstd path (`ZSTD_decompressStream`),
  wrapped as `hc_zstd_decompress()` like `hc_inflate_raw()`
- Refuse a `data_type 8` hash at parse time if libzstd is missing
- No GPU/OpenCL kernel changes

## Verification
Verified end-to-end with a real ZSTD-encrypted 7z archive (password
`159357`):
1. Extractor produces `$7z$8$19$0$$16$...` (data_type 8)
2. hashcat `-m 11600 -a 0` with a small dictionary →
   `Status: Cracked`, `Recovered: 1/1`

Decompression uses the libzstd runtime library already loaded by master
(streaming path), not a bundled decoder.


RealEnder:
As collateral bonus, including zstd decompress to hashcat's tree will make it possible to use zstd-compressed dictionaries with a simple additional patch.

jsteube:
Thanks for this, and sorry about the timing. I definitely want the feature, and testing it end to end against a real archive is very helpful. The unlucky part is that master changed only 4 days after you opened this, and compression support moved in pretty much the opposite direction.

Commit 8371d3f76 removed the bundled zlib and LZMA code, around 128k lines in total. Compression libraries are now loaded at runtime with hc_dynlib_open(), so users only need the runtime package and a missing library does not prevent hashcat from starting. That change also added the current ext_zstd.c and ext_zstd.h.

Because of that, the zstd infrastructure in this PR cannot really come in anymore. ext_zstd.c and ext_zstd.h now conflict directly with master, and src/zstddeclib.c would bring another 23k lines of bundled compression code back into the tree, which is exactly what we just removed.

The good news is that the actual feature you added is still completely relevant. Master's zstd support only handles compressed input files. It does not handle zstd data inside a 7-Zip archive, and module 11600 still has no support for codec 0x04f71101. So that part is still yours and I want it.

I think the rebase can actually become very small:

1. Drop src/zstddeclib.c and the ext_zstd files from this PR and use the versions from master.

2. Add ZSTD_decompress() to the existing symbol table and hc_zstd_lib_t. Master currently loads only 6 symbols, the 4 for the streaming path plus ZSTD_isError and ZSTD_getErrorName, so the one shot function you need is simply missing there.

3. Keep the module_11600.c changes.

That should reduce the PR from about 24k lines to maybe 30, which is a pretty nice outcome.

There is one new case to handle because of runtime loading. Hashcat is allowed to run without libzstd now, so if somebody loads a data_type 8 hash without libzstd installed, we should detect that early. Please do not let it fail later inside module_hook23(), since that runs for every candidate. Parse time or the self test would be a much better place.

One detail from your description is also worth keeping in the code. ZSTD_decompress() needs unpack_size rather than data_len because data_len includes the AES-CBC padding. Please leave a comment above the call explaining that. Otherwise I can easily see somebody "cleaning it up" a couple of years from now and accidentally breaking it.

One last thing before you rebase. I already have another unrelated change to module_11600.c queued here. Please hold off for a moment until that lands, and I'll update you here. That way you only need to rebase this once.

Igloo-Garage:
@jsteube Thanks for the review, and sorry for the delay.

Your queued module_11600.c changes look like they have landed on master
(the hook_extra_param signature change and module_hash_hints). If that
was the update you mentioned, I will rebase along the lines you
described:

1. Drop zstddeclib.c and this PR's ext_zstd files; keep master's.
2. Add ZSTD_decompress() to hc_zstd_lib_t and the dynlib symbol table.
3. Keep the module_11600.c feature (data_type 8 / codec 0x04f71101).
4. Fail at parse time if a data_type 8 hash is loaded without libzstd,
   not inside module_hook23().
5. Keep the comment that ZSTD_decompress() must use unpack_size, not
   data_len, because data_len includes AES-CBC padding.

Please confirm it is OK to rebase onto current master.

jsteube:
Yes, please rebase onto current master. The module_11600.c changes are in, and "src/ext_zstd.c" and "include/ext_zstd.h" are there with a symbol table, so points 1, 3 and 4 land where you expect them to.

Point 2 may not be needed. What the table carries is the streaming entry points rather than the one shot call:

```
  HC_DYNLIB_SYM (hc_zstd_lib_t, ZSTD_initDStream,      true),
  HC_DYNLIB_SYM (hc_zstd_lib_t, ZSTD_decompressStream, true),
```

and "src/filehandling.c" already drives them, with the return value marking where a frame ended:

```
    const size_t rc = z->ZSTD_decompressStream (zfp->zds, &outbuf, &zfp->in);

    if (z->ZSTD_isError (rc)) break;

    // Zero is libzstd saying a frame ended exactly here and nothing of it is still owed. What
    // follows is the first byte of the next frame, and a reader handed that offset decodes from it
    // without having read a byte of what came before.
```

That reaches point 5 as well. A stream decode stops at the end of the frame and reports how much of the input it consumed, so the AES-CBC padding sitting after the frame is harmless and unpack_size no longer has to be exact.

Igloo-Garage:
@jsteube 
Rebased onto current master as discussed.
Point 2 was not needed: decompression uses the existing streaming
entry points via a small hc_zstd_decompress() wrapper, the same
shape as hc_inflate_raw(). AES-CBC padding after the frame is left
unread. A data_type 8 hash without libzstd is refused at parse time.

jsteube:
One thing here is hashcat's gap rather than yours, and it needs a small core change before this mode can report the case properly.

```
  if ((data_type == 8) && (hc_zstd_available () == false))
  {
    return (PARSER_HASH_VALUE);
  }
```

Parse time is the only point where the hash has been read far enough for data_type 8 to be visible, and a return code is the only channel a module has there, because module_hash_decode () is handed no context to log with. What is missing is a code that means what this means. PARSER_HASH_VALUE prints "Hash-value exception", so a user on a box without libzstd1 is sent off to look at their hash file for a fault that is not there.

The nearest code that exists is the one mode 17200 added, "Hash contains unsupported compression type for current mode", and it is still not this: every string in that table describes the hash rather than the machine, and the mode does support zstd. Installing the package is what fixes this one, so the string has to name it or a user is no better off.

Would you be up for adding one, or would you rather it landed separately and this waited on it? If you are up for it, -48 and PA_048 are free, and mode 17200 is the shape, three places:

```
  include/types.h   PARSER_PKZIP_CT_UNMATCHED   = -37,

  src/parser.c      static const char *const PA_037 = "Hash contains unsupported compression type for current mode";

  src/parser.c      case PARSER_PKZIP_CT_UNMATCHED:   return PA_037;
```

Either way I would want it in before the mode does, because that message is the whole of what a user gets when the mode will not run for them.

PR #4766, opened 2026-08-10
Fix -m 24420 rejecting the $PEM$2 layout emitted by pem2john.py (bleeding-jumbo)

### LINKED ISSUE #4765: -m 24420 rejects the $PEM$2 layout emitted by pem2john.py (bleeding-jumbo)
### Problem

`-m 24420` rejects the self-describing `$PEM$2` layout that JtR's `pem2john.py` emits when the PBKDF2 PRF is not SHA-1.

Both projects use the prefix `$PEM$2$`, but for different layouts:

| Producer | Layout |
|---|---|
| hashcat `-m 24420` | `$PEM$2$<cid>$<salt>$<iter>$<iv>$<len>$<data>` — the `2` encodes the PRF |
| `pem2john.py` | `$PEM$2$<kdf>$<prf>$<cipher>$<cid>$<salt>$<iter>$<iv>$<len>$<data>` |

So this is not a request to support a foreign format — it is a collision on the same signature, and hashcat currently cannot load the other side of it.

### Self-contained reproduction

No third-party tooling needed; this hash is a throwaway RSA-2048 key, password `hashcat1`:

```
$PEM$2$pbkdf2$sha256$aes256_cbc$4$a37f9280278386b959ea0e752c0a7fcd$2048$346899177d1e400bac501253b0215004$1232$5327114aa4719cea5e6b29f74ae6c016992ea91db86a66e3febbe091cd4902a0c19bd90ba8464e87ee736f80b673b1ea5310eeff6a59f6575dcefd6f25f145393664d992bb5828243fefdaa071df29b6442bcb0f3b1f127f2050085ce4902eb97cc455428f06fdd31f7c7499bae699717d25d497dd826e3b6167d0aaa269b01cf613debe1a1714aeb94e788f95b24819260478661d5a4c5ec57630fc2b0dd2f427a507e2e066a4ef1de25ba0e99fd5a54505fb9bc71448e493fbf03962a441b879eb788fc4eb48534180bcf39b0b1aa6d1fa0dc2c1480a8bf959c0144739ac5bf76b4d62affbf93adb533188c31445281fea4dfb5a697f19d03f7c29c223534f9b74f678627b0879c94c2a2379b3975d6906e741a1b2ccfd8b4c141977d0a170c3ca5d411ba2115e80a50b9d6356b106a73de60910b8b372a7ef67146d05a891ec8b307d4b20e32ed6c78fc941ee1abd2a896f8a7ce0559953946ea8e7bac5652eeabb54d010e1de531a72876b26aaadd05c26a3c11493ca33b57e8de93cf753ae6a2b790b37e3de55092d1f7ea3f2dcc7c7a62db1b2901cdefce49167afa08dff3523ac4f42e4422c4bfdc9823b7681ce97067808b71cf9ff4e315d40565008415919334cd4a29b18995cc05f94fa09c25895a69d645f4540283edaa2a40861cdc107774d95c436d015588150b05cebae0f029da84217640536eb5a75868dc52cfeff0890b19b0d42845bfdc62ae437078c6001725772412257e0f64085ce74d6316d1b68fc019a3e4d454e8c22cf4ef7a5a9a24283aa7e0883226005f501f6e0a30b0c752fbac8ac01bd85188f0472f4f019ac34a3882dd432ead56466ac416ae6475e91f9a27e390d46c8a833d2f19c808c3c8527e22f0a224042133bd4f63ec10146ba9954842ca2800512328b35ec87971aeadc16e71ccf27b974fc504b6bc6aa94d2f9828603867494565fa3a91f2f1154e815f0c67fc67b94750bdd8de9eb3e9516771cea89ca4f89a08364cdb93925105ad151b04df00c64c921a4761ce5f95cee2633c8e973d578f2181740fde8c6f9cacfd82501ed6d4df64a231b4d16a0b4beeececa90afc4ae38ba66a7e54e92c8a128f57b181f224fd01f23fd716b368c3437fb59007ebf046c8c6d1becb8049bc9dc95a53d7c95c66e7909ef4d33ee698f66e7fb9d8b14c719daf80ba581eafdaa819575e4714de0032595852d78914a2e73bb1f038c148aa5024053e74f70415bd9b4a47275769adbde2a7f0332e08712f89dbe7ee2464190c2442c981058a89efb3870ce39cb79a720e470c148042f5cb710ccc39ee479fc21648e6793ef06a80bd406a463936d5a7c6d21cc343b24c60a940d2d292087c908cb040e41f5d60deb50e9bf794a2484e44caaaf3b8a55138edd4dfa4264b212040bc14a3fbb39c1947addb1c4faa1cab0e224079f2a2ceeb1d25cd323a60a2cfca449bbe900abd2155b9c2081edbce3bc1fc10d178b10b5bf23aa5d12a5688d96c2081d06747b8472443534152f34310c57a6169953e3e5fcd169b6e6608669fb2f69ad5354a63139d351804793ea58268a54a40ee2365fc1d347e2cb9a63f44e7714b9a63c6ebb06c44e0be7b76c62ac0e4f9281d35f60a3ab48f84c5d141b257b1d5f24ced41e509fe8fc78d24f9cc3d9c3a059ac34fafd56a70b43fcecc7aed0efe0b3270400ba0549de014127c3a8d1707f83ec297f270f291b0fe88c2a68d6fe
```

```
hashcat -m 24420 verbose.hash wordlist.txt
```

**Actual** (master, `37dfdc9`):

```
Hash parsing error in hashfile: 'verbose.hash' on line 1 ($PEM$2...): Token length exception
* Token length exception: 1/1 hashes
```

**Expected:** loads and recovers `hashcat1`.

The identical key in hashcat's compact layout does load and recover `hashcat1`, which confirms the payload is fine and only the extra descriptor fields are the obstacle:

```
$PEM$2$4$a37f9280278386b959ea0e752c0a7fcd$2048$346899177d1e400bac501253b0215004$1232$5327114aa4719cea5e6b29f74ae6c016992ea91db86a66e3febbe091cd4902a0c19bd90ba8464e87ee736f80b673b1ea5310eeff6a59f6575dcefd6f25f145393664d992bb5828243fefdaa071df29b6442bcb0f3b1f127f2050085ce4902eb97cc455428f06fdd31f7c7499bae699717d25d497dd826e3b6167d0aaa269b01cf613debe1a1714aeb94e788f95b24819260478661d5a4c5ec57630fc2b0dd2f427a507e2e066a4ef1de25ba0e99fd5a54505fb9bc71448e493fbf03962a441b879eb788fc4eb48534180bcf39b0b1aa6d1fa0dc2c1480a8bf959c0144739ac5bf76b4d62affbf93adb533188c31445281fea4dfb5a697f19d03f7c29c223534f9b74f678627b0879c94c2a2379b3975d6906e741a1b2ccfd8b4c141977d0a170c3ca5d411ba2115e80a50b9d6356b106a73de60910b8b372a7ef67146d05a891ec8b307d4b20e32ed6c78fc941ee1abd2a896f8a7ce0559953946ea8e7bac5652eeabb54d010e1de531a72876b26aaadd05c26a3c11493ca33b57e8de93cf753ae6a2b790b37e3de55092d1f7ea3f2dcc7c7a62db1b2901cdefce49167afa08dff3523ac4f42e4422c4bfdc9823b7681ce97067808b71cf9ff4e315d40565008415919334cd4a29b18995cc05f94fa09c25895a69d645f4540283edaa2a40861cdc107774d95c436d015588150b05cebae0f029da84217640536eb5a75868dc52cfeff0890b19b0d42845bfdc62ae437078c6001725772412257e0f64085ce74d6316d1b68fc019a3e4d454e8c22cf4ef7a5a9a24283aa7e0883226005f501f6e0a30b0c752fbac8ac01bd85188f0472f4f019ac34a3882dd432ead56466ac416ae6475e91f9a27e390d46c8a833d2f19c808c3c8527e22f0a224042133bd4f63ec10146ba9954842ca2800512328b35ec87971aeadc16e71ccf27b974fc504b6bc6aa94d2f9828603867494565fa3a91f2f1154e815f0c67fc67b94750bdd8de9eb3e9516771cea89ca4f89a08364cdb93925105ad151b04df00c64c921a4761ce5f95cee2633c8e973d578f2181740fde8c6f9cacfd82501ed6d4df64a231b4d16a0b4beeececa90afc4ae38ba66a7e54e92c8a128f57b181f224fd01f23fd716b368c3437fb59007ebf046c8c6d1becb8049bc9dc95a53d7c95c66e7909ef4d33ee698f66e7fb9d8b14c719daf80ba581eafdaa819575e4714de0032595852d78914a2e73bb1f038c148aa5024053e74f70415bd9b4a47275769adbde2a7f0332e08712f89dbe7ee2464190c2442c981058a89efb3870ce39cb79a720e470c148042f5cb710ccc39ee479fc21648e6793ef06a80bd406a463936d5a7c6d21cc343b24c60a940d2d292087c908cb040e41f5d60deb50e9bf794a2484e44caaaf3b8a55138edd4dfa4264b212040bc14a3fbb39c1947addb1c4faa1cab0e224079f2a2ceeb1d25cd323a60a2cfca449bbe900abd2155b9c2081edbce3bc1fc10d178b10b5bf23aa5d12a5688d96c2081d06747b8472443534152f34310c57a6169953e3e5fcd169b6e6608669fb2f69ad5354a63139d351804793ea58268a54a40ee2365fc1d347e2cb9a63f44e7714b9a63c6ebb06c44e0be7b76c62ac0e4f9281d35f60a3ab48f84c5d141b257b1d5f24ced41e509fe8fc78d24f9cc3d9c3a059ac34fafd56a70b43fcecc7aed0efe0b3270400ba0549de014127c3a8d1707f83ec297f270f291b0fe88c2a68d6fe
```

### Which pem2john.py

This matters for reproducing, and my apologies for omitting it initially:

- **JtR 1.9.0-jumbo** (what most distros package) only ever emits `$PEM$1$` — you will *not* see this with it
- **JtR bleeding-jumbo** (current master) emits the verbose form whenever the PRF is not sha1, at [`pem2john.py` L106-109](https://github.com/openwall/john/blob/bleeding-jumbo/run/pem2john.py):

```python
pem_version = "$PEM$1"
if kdf_params["prf"]["algorithm"] != "sha1":
    pem_version = f'$PEM$2${kdf["algorithm"]}${kdf_params["prf"]["algorithm"]}${cipher}'
```

To generate one:

```
openssl genrsa -out k.pem 2048
openssl pkcs8 -topk8 -v2 aes-256-cbc -v2prf hmacWithSHA256 -in k.pem -out k8.pem
python3 run/pem2john.py k8.pem     # from a bleeding-jumbo checkout
```

### Notes

The three extra fields carry nothing the compact layout lacks: type `2` already implies PBKDF2-HMAC-SHA256, and the cipher id follows them numerically. The two lines describe the same hash, so the verbose one can be reduced to the compact one before tokenizing without touching the existing parse path.

Patch in #4766.

### hashcat version

master, `37dfdc9` (also v7.1.2)

bishwabikash:
Fixes #4765.

### Problem

hashcat and JtR's `pem2john.py` both use the prefix `$PEM$2$`, for different layouts:

| Producer | Layout |
|---|---|
| hashcat `-m 24420` | `$PEM$2$<cid>$<salt>$<iter>$<iv>$<len>$<data>` — the `2` encodes the PRF |
| `pem2john.py` (bleeding-jumbo, PRF != sha1) | `$PEM$2$<kdf>$<prf>$<cipher>$<cid>$<salt>$<iter>$<iv>$<len>$<data>` |

hashcat rejects the second with a token length exception, so a hash in that layout cannot be loaded at all.

To be precise about scope, since I was sloppy about this in the issue at first: this affects JtR **bleeding-jumbo**, not the released 1.9.0-jumbo that distributions package, which only ever emits `$PEM$1$`.

### Approach

The three extra fields carry nothing the compact layout lacks — type `2` already implies PBKDF2-HMAC-SHA256, and the cipher id follows them numerically — so both lines describe the same hash.

Given that, the patch strips the descriptor fields and lets the existing token spec parse the result, rather than threading a second token layout through the decoder. `module_hash_decode()` keeps one parse path. The compact form is untouched: `pem_normalize()` returns early unless the line starts with `$PEM$2$` followed by a non-digit, which only the verbose layout does.

### Testing

Throwaway RSA-2048 key, `aes-256-cbc` / `hmacWithSHA256`, password `hashcat1`.

Before, on master `37dfdc9`:

```
Hash parsing error in hashfile: 'verbose.hash' on line 1 ($PEM$2...): Token length exception
* Token length exception: 1/1 hashes
```

After:

```
$PEM$2$pbkdf2$sha256$aes256_cbc$4$a37f9280278386b959ea0e752c0a7fcd$2048$346899...:hashcat1
```

The same key in the compact layout recovers the same password both before and after this patch, so the two paths agree and the existing one has not regressed. Compiles with no warnings under `-W -Wall -std=gnu99`.

The failing hash is in #4765 if you want to reproduce without generating one.

### Scope

One problem. I have a separate PR (#4767) adding a hash mode; the two do not overlap.


jsteube:
In "module_24420.c" the rewritten line goes into a 4096 byte buffer:

```
  char norm_buf[HCBUFSIZ_TINY];

  const int norm_len = pem_normalize (line_buf, line_len, norm_buf, sizeof (norm_buf));
```

The token spec a few lines below allows 65536 hex characters for the data field, so a line can reach about 65 KB. The example hash in this module is 2533 characters, which is a 2048 bit key and fits. A 4096 bit key is roughly twice that and does not, and pem_normalize() returns -1 on the size check. The verbose line then falls through to the compact parser and is rejected with the same token error this PR sets out to remove, with no sign that the length was the reason. Could you size the buffer from the line rather than from HCBUFSIZ_TINY?

Second, the three descriptor fields are skipped without being read:

```
  while ((skipped < 3) && (rest < (line_buf + line_len)))
  {
    if (*rest == '$') skipped++;

    rest++;
  }
```

Any value passes there, so a line reading "$PEM$2$pbkdf2$sha512$..." is stripped down to "$PEM$2$..." and parsed as PBKDF2-HMAC-SHA256. There is no SHA512 PEM mode in hashcat, so that hash now loads and never cracks, where before it was rejected at load time. A silent wrong parse is worse than the refusal it replaces. Could you compare the kdf and prf fields against pbkdf2 and sha256 and return -1 otherwise?

Sorry for the delay, thus the changes.txt needs a rebase as well

bishwabikash:
Both addressed. Removed the buffer.

The verbose layout has its own token spec now. A bool picks the layout, three descriptor tokens join the spec, and one offset shifts every field from the cipher id onwards, so the line is never copied and there is no size to get wrong. The 4096 bit case you described, a 4879 character line, goes from `Token length exception` to cracking, and the compact and verbose forms of the same key recover the same password.

kdf and prf are read and compared against pbkdf2 and sha256 now, and anything else is refused:

```
$PEM$2$pbkdf2$sha512$aes256_cbc$4$...   Signature unmatched
```

Before, that line loaded and ran to exhaustion against a digest it could never match.

Two things worth flagging rather than leaving for you to find.

The descriptor tokens are bounded at 63 characters, so a kdf or prf name longer than that still reports a token length exception instead of reaching the comparison. That is the one input shape where the old message survives.

`module_hash_encode` still writes the compact layout for a hash read from the verbose one, so `--show`, `--left` and the potfile carry the compact form rather than the line the user supplied. `-m 19700` normalizes its two layouts the same way, so I followed it, but say if you would rather the layout were preserved.

`tools/test_edge.sh -m 24420` reports one `empty test vectors` error on attack type 4. Untouched master reports the same error on attack types 4, 6 and 8 here, so it predates this change.

changes.txt rebased.

jsteube:
> `-m 19700` normalizes its two layouts the same way, so I followed it, but say if you would rather the layout were preserved.

I would keep it as it is. One canonical form is what lets a key read in either layout match the same potfile line, and that is worth more than echoing the input back in --show.

The cipher name is the one descriptor still going unread:

```
  if ((token.len[2] != 6) || (memcmp (token.buf[2], "pbkdf2", 6) != 0)) return (PARSER_SIGNATURE_UNMATCHED);
  if ((token.len[3] != 6) || (memcmp (token.buf[3], "sha256", 6) != 0)) return (PARSER_SIGNATURE_UNMATCHED);
```

A line spelling aes128_cbc in front of a cipher id of 4 loads, and the id wins, which is the same silent disagreement those two lines close. Could the name be checked against the id as well, if pem2john.py spells the four ciphers predictably enough for that? You have its vocabulary in front of you and I do not.

Last, that test_edge note is worth re-running before it stands. On master here the oracle gives vectors for every attack type you named:

```
  mode 24420 attack 0  -> 2 vectors
  mode 24420 attack 4  -> 2 vectors
  mode 24420 attack 6  -> 2 vectors
  mode 24420 attack 8  -> 2 vectors
```

#4863 landed earlier today and closed the attack mode 4 empty word case, so a tree from before it would report exactly what you saw.

bishwabikash:
Cipher name is checked against the id now. `aes128_cbc` in front of an id of 4 gives `Invalid or unsupported cipher`.

The vocabulary is closed. pem2john.py builds the name and the id from one variable and writes no line at all for anything else:

```
tripledes_3key  1
aes128_cbc      2
aes192_cbc      3
aes256_cbc      4
```

All four crack, generated with openssl and run through pem2john.py.

The prf is the opposite case, which is why the sha256 check stays. pem2john passes it through from the key, so sha224, sha384 and sha512 lines all occur.

My test_edge note was wrong, and #4863 was not the reason either: `7015700fc` is in the tree I tested. The cause was `Crypt::DES_EDE3` missing here. `m24420.pm` draws its cipher with `random_number (1, 4)` and dies when it draws 3DES, losing the export, which is why the attack types moved between runs. Installed it, 36 of 36 draws now give 2 vectors:

```
> All tests done in 0d:00h:05m:31s
> Errors detected: 0
```

PR #4767, opened 2026-08-10
Add -m 36800, OpenSSH Private Keys (bcrypt-pbkdf)

bishwabikash:
### Motivation

`ssh-keygen` has written the `openssh-key-v1` container with bcrypt-pbkdf key derivation by default since OpenSSH 7.8 (2018). The existing `-m 22911..22951` handle only legacy PEM keys with the MD5-based KDF, so an SSH private key generated with default options today cannot be attacked with hashcat at all.

The gap is not a missing fast target — it is the opposite. The PEM formats hashcat implements run at ~2 GH/s on this card, while the format `ssh-keygen` actually produces has had no kernel. This adds one.

### Algorithm

Per OpenBSD `bcrypt_pbkdf.c`:

```
sha2pass = SHA512 (password)
for each 32 byte output block:
    sha2salt = SHA512 (salt || BE32 (block))
    out = bcrypt_hash (sha2pass, sha2salt)
    repeat rounds-1 times:
        sha2salt = SHA512 (previous bcrypt_hash output)
        out ^= bcrypt_hash (sha2pass, sha2salt)
    scatter out into the key at stride
```

Two properties of the KDF drive the plugin's shape:

- A `bcrypt_hash()` cannot be split across kernel invocations, so `kernel_loops` is pinned at 1 and the derivation runs inside `_loop`.
- 48 bytes of output (32 byte AES key + 16 byte IV) means `stride = 2`, so **two** independent blocks each run the full round loop. Real work is `2 x rounds x 129` Blowfish key expansions — twice what a naive reading of `bcrypt_pbkdf` suggests.

### Reuse, and one thing that could not be reused

`blowfish_encrypt()`, `c_pbox` and `c_sbox0..3` from `inc_cipher_blowfish.cl` are used unchanged.

`blowfish_set_key_salt()` could not be: it hardcodes a 4 word salt, indexing `salt_buf[(i & 2) + 0]` with S-box loops that reference `salt_buf[0..3]` literally. bcrypt-pbkdf feeds a 64 byte `sha2salt` that the data stream cycles over 16 words at a time. A 16-word `expandstate` is therefore written in `m37500-pure.cl`.

I kept it local to this kernel rather than generalising the shared helper, since changing `inc_cipher_blowfish.cl` would affect `-m 3200` and belongs in its own PR if it is wanted at all.

### Hash format

john's `$sshng$` encoding, so `ssh2john.py` works unmodified:

```
$sshng$<cipher>$16$<salt>$<datalen>$<data>$<rounds>$<ctoffset>
```

Eight tokens against six for the legacy PEM form, so the parsers stay separate rather than `-m 22921` growing a second layout. Cipher id `6` is aes256-ctr and `2` is aes256-cbc; both occur in practice and both are handled.

Verification decrypts the first ciphertext block and compares the two check integers OpenSSH writes equal before encrypting — the same criterion john and `-m 22921` use, at a 2^-32 false positive rate per guess. If a stricter check is preferred I can add a second stage parsing the decrypted key type string; say the word.

### Performance

RTX 4050 Laptop, CUDA 13.3:

```
Speed.#01........:      592 H/s (810.43ms) @ Accel:1 Loops:1 Thr:24 Vec:1
```

Scaling `-m 3200` (36,994 H/s at cost 5, 65 expansions) by the 4128 expansions this mode performs at the default 16 rounds predicts 583 H/s, so the measured figure lands within 1.5% of the cost model.

`Thr:24` is the same occupancy ceiling bcrypt hits: Blowfish needs 4 KB of S-boxes per work item, which caps residency per SM regardless of core count. This mode inherits that exactly, so `OPTS_TYPE_DYNAMIC_SHARED` is handled the way `-m 3200` handles it. The trade-off is inherent to the primitive — there is no arrangement that avoids the 4 KB.

For context, the same key on 16 CPU cores with john runs at 86.5 c/s, so this is roughly a 6.8x speedup. Real and worth having, but bcrypt-pbkdf does not collapse on a GPU and I would not want the mode to imply otherwise.

### Testing

Keys from `ssh-keygen` at `-a 8`, `16` and `64`, covering ed25519, RSA-2048 and ECDSA-256 across both ciphers:

- all recover; a wrong password recovers none
- every cracked line re-encodes byte identical to its input, so the potfile and `--show` round-trip
- self-test passes with the included example hash
- compiles with no warnings under `-W -Wall -std=gnu99`

The derived key was checked against an independent CPU implementation of `bcrypt_pbkdf` before the kernel was trusted, rather than inferring correctness from a successful crack.

### Test module

`tools/test_modules/m37500.pm` implements bcrypt_pbkdf in pure Perl. No CPAN module provides it, and `Crypt::Eksblowfish` cannot substitute: its cost loop runs `ExpandKey(key)` before `ExpandKey(salt)` where bcrypt_pbkdf runs salt before key, and it fixes the salt at 16 bytes where this needs 64. The Blowfish tables in it are generated from `OpenCL/inc_cipher_blowfish.cl` so the two cannot drift apart.

It was cross-checked against the kernel in both directions: hashcat cracks Perl-generated hashes for both ciphers, and the Perl module verifies every key the GPU cracked.

### Scope

One problem, one PR. I have a separate PR fixing an unrelated `-m 24420` parser bug; the two do not overlap. The mode number 37500 is unused on master — happy to renumber if you have a different slot in mind.

### DIFF
--- a/OpenCL/m37500-pure.cl
+++ b/OpenCL/m37500-pure.cl
@@ -1,0 +1,511 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#define NEW_SIMD_CODE
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sha512.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_blowfish.cl)
+#include M2S(INCLUDE_PATH/inc_cipher_aes.cl)
+#endif
+
+#if   VECT_SIZE == 1
+#define uint_to_hex_lower8(i) make_u32x (l_bin2asc[(i)])
+#endif
+
+// Must stay identical to the copy in src/modules/module_37500.c: esalt_bufs[]
+// strides by sizeof(), so a mismatch misreads every hash past the first. Only
+// the fields above data_buf are read here; the blob rides along so the host
+// encoder can reproduce its input line.
+typedef struct sshng_bcrypt
+{
+  u32 cipher;
+  u32 ct_offset;
+  u32 rounds;
+  u32 salt_buf[4];
+  u32 ct_buf[4];
+  u32 data_buf[8192];
+  int data_len;
+
+} sshng_bcrypt_t;
+
+typedef struct sshng_bcrypt_tmp
+{
+  u32 pass_hash[16];  // SHA-512 of the password, as 16 big-endian words
+  u32 dk[12];         // derived 48 bytes: 32 byte AES key + 16 byte IV
+
+} sshng_bcrypt_tmp_t;
+
+// "OxychromaticBlowfishSwatDynamite", read big-endian a word at a time.
+CONSTANT_VK u32 c_bcrypt_magic[8] =
+{
+  0x4f787963, 0x68726f6d, 0x61746963, 0x426c6f77,
+  0x66697368, 0x53776174, 0x44796e61, 0x6d697465
+};
+
+#define SSHNG_CIPHER_AES256_CBC 2
+#define SSHNG_CIPHER_AES256_CTR 6
+
+/**
+ * SHA-512 digest as 16 big-endian u32 words, which is the form
+ * Blowfish_stream2word consumes.
+ */
+DECLSPEC void sha512_digest_to_words (PRIVATE_AS const u64 *h, PRIVATE_AS u32 *w)
+{
+  for (int i = 0; i < 8; i++)
+  {
+    w[(i * 2) + 0] = (u32) (h[i] >> 32);
+    w[(i * 2) + 1] = (u32) (h[i] & 0xffffffff);
+  }
+}
+
+/**
+ * Blowfish_expand0state for a 64-byte (16 word) key.
+ *
+ * Written inline rather than via blowfish_set_key(): that helper re-initialises
+ * P and the S-boxes on entry, but expand0state has to mutate the state left by
+ * the previous call.
+ */
+DECLSPEC void bf_expand0state (PRIVATE_AS u32 *P, LOCAL_AS u32 *S0, LOCAL_AS u32 *S1, LOCAL_AS u32 *S2, LOCAL_AS u32 *S3, PRIVATE_AS const u32 *key)
+{
+  for (u32 i = 0; i < 18; i++)
+  {
+    P[i] ^= key[i % 16];
+  }
+
+  blowfish_encrypt (P, S0, S1, S2, S3);
+}
+
+/**
+ * Blowfish_expandstate with a 64-byte data stream.
+ *
+ * inc_cipher_blowfish's blowfish_set_key_salt() cannot be reused: it hardcodes
+ * a 4-word salt (salt_buf[(i & 2) + 0], and S-box loops referencing
+ * salt_buf[0..3] literally). bcrypt-pbkdf feeds a 64-byte sha2salt, so the data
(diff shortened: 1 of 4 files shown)


jsteube:
In "module_37500.c" the comment and the line four below it disagree:

```
  // rounds -> salt_iter, so the autotuner sees the real cost

  const u32 rounds = hc_strtoul ((const char *) token.buf[6], NULL, 10);

  if (rounds == 0) return (PARSER_SALT_ITERATION);

  // the whole derivation runs in one _loop invocation for now
  salt->salt_iter = 1;
```

With kernel_loops pinned to 1 at both ends, one _loop launch then carries the entire derivation, because the round loop lives inside it:

```
    for (u32 r = 1; r < rounds; r++)
```

The rounds token is 1 to 8 characters wide and only 0 is refused, so a hash file can ask for 99999999 of them. You wrote earlier that the real work is 2 x rounds x 129 Blowfish key expansions, and all of it now lands inside a single launch with no lever on it: the autotuner can still move accel, but the duration of one launch is set by a number in the file. On a card that is also driving a monitor the display driver ends any kernel that runs more than a second or two, so a launch like that is stopped from outside hashcat long before the ceiling is reached. The ssh-keygen default of 16 rounds already puts 4128 key expansions into one launch.

Could you set salt_iter to rounds, the way the first comment describes? Splitting a single bcrypt_hash() across invocations is the part that cannot be done, but one round is one bcrypt_hash(), so the loop divides cleanly at that boundary and kernel_loops can open up again.

bishwabikash:
salt_iter carries the round count now. `_init` seeds each of the two output blocks with `sha2salt = SHA512 (salt || BE32 (block))` and a zeroed accumulator, `_loop` runs LOOP_CNT whole rounds of both blocks, and `_comp` interleaves them into the 48 bytes. One round is one `bcrypt_hash ()` per block, so the loop divides at the boundary you named.

Same tree, only the split differing, on ed25519 keys from ssh-keygen:

```
rounds=8    before  10 H/s (382.07ms) @ Accel:1 Loops:1 Thr:11 Vec:1
rounds=8    after   10 H/s  (47.78ms) @ Accel:1 Loops:1 Thr:11 Vec:1
rounds=64   before  Kernel minimum runtime larger than default TDR
                    * Device #1: skipped, due to kernel autotune failure (-4).
rounds=64   after    0 H/s  (47.77ms) @ Accel:1 Loops:1 Thr:11 Vec:1
rounds=100  before  same autotune failure (-4)
rounds=100  after    0 H/s  (48.02ms) @ Accel:1 Loops:1 Thr:11 Vec:1
```

It was worse than a long launch: `ssh-keygen -a 64` could not be attacked at all, the autotuner refused the device.

You were right about kernel_loops, and I had it pinned for the wrong reason. It is at the module default now, so the autotuner sizes the launch from what it measures, which is what `src/autotune.c` already bounds by target_msec. On this card it settles on Loops:1 either way, so the multi-round path is not reachable here by default; I exercised it with `--force --kernel-loops 2`, `4`, `8` and `16`, and all five test keys recover at each, including a 100 round key where the final launch runs a partial 4 rounds.

One more thing in this push. `tools/test_modules/m37500.pm` had its salt constraint as `[16, 16]`, but test.pl counts characters and the module packs with `pack ("H*")`, so every generated vector carried an 8 byte salt that the parser refused on sight. It reads `[32, 32]` now, the way `m00131.pm` states 8 for a 4 byte salt. With that fixed:

```
> All tests done in 0d:00h:36m:53s
> Errors detected: 0
```

Rebased onto master.

jsteube:
Looks good so far, but could you please move the hash-mode number to 36800?

bishwabikash:
Sure, let me do it in sometime and refresh the PR

bishwabikash:
Renumbered to 36800. Files renamed with `git mv`, entry points `m36800_init`, `_loop` and `_comp`, `KERN_TYPE = 36800`. The test module and the changes.txt line carry no number, so neither changed.

```
> All tests done in 0d:00h:35m:00s
> Errors detected: 0
```

Five ssh-keygen keys at -a 8, 16, 64 and 100 recover under the new number.

PR #4770, opened 2026-08-11
Fix stdin feed blocking on short pipe writes

### LINKED ISSUE #4769: stdin feed delays short pipe writes until 1 MiB or EOF
### Problem

The block-based stdin reader added in `93b55d37d` waits for `fread()` to satisfy a 1 MiB request. `select()` establishes that some data is ready, but the following `fread(buf, 1, STDIN_BLOCK_SIZE, stdin)` can continue waiting until the request is full or the producer closes the pipe.

As a result, a long-running generator can flush many complete candidates without any of them reaching the feed.

### Reproducer

```python
import select
import subprocess

p = subprocess.Popen(
    ["./hashcat", "--stdout", "-w", "1", "-M"],
    stdin=subprocess.PIPE,
    stdout=subprocess.PIPE,
)

raw = b"abc\n" * 250_000  # 1,000,000 bytes: less than the 1 MiB read

p.stdin.write(raw)
p.stdin.flush()

ready = bool(select.select([p.stdout], [], [], 5.0)[0])
print("output before EOF:", ready)

p.stdin.close()
assert p.stdout.read() == raw
assert p.wait() == 0
```

On current master this prints `output before EOF: False`. The completed lines only appear after `p.stdin.close()`.

### Expected behavior

Complete newline-delimited candidates should be made available without requiring the producer to fill the entire 1 MiB request or close stdin.

### Proposed approach

Use a descriptor-level `read()` after `select()` so the reader consumes the bytes currently available. Because descriptor reads may be short, a short newline-free result must be carried into the next read; only a full newline-free block should retain the existing overlong-line behavior. Retry interrupted reads on `EINTR`.

Tested on current master `fc3a57ead` on macOS arm64.

mohammedjhaish:
Fixes #4769.

### Problem

After `select()` reported that stdin was readable, the stdin feed called `fread()` with the entire remaining 1 MiB block as its request. On a pipe, `fread()` could continue waiting for that request to fill or for EOF, even when the producer had already flushed many complete newline-delimited candidates.

### Changes

- Read directly from the stdin descriptor so currently available pipe data is returned without waiting for the full block.
- Retry descriptor reads interrupted by `EINTR`.
- Carry a short newline-free read into the next read instead of treating it as an overlong line.
- Preserve the existing overlong-line behavior only when a newline-free block is actually full.

The Windows path uses `_read(_fileno(stdin), ...)`; POSIX uses `read(fileno(stdin), ...)`.

### Verification

- `make hashcat feeds IS_ARM=1 -j4` on macOS arm64 with `-W -Wall -Wextra`.
- End-to-end `hashcat --stdout -w 1 -M`: 1,000,000 bytes of complete candidates begin appearing while the producer keeps stdin open, and the final output matches byte-for-byte.
- Delayed and fragmented writes: incomplete `abc` produces nothing, adding `\n` immediately produces `abc\n`; the same test passes with an embedded NUL.
- 500 randomized binary records split across 1-47 byte writes matched the expected bytes.
- 10,000 weighted random binary records and 2.5 MiB arbitrary input matched both the length-based model and the wordlist feed.
- 300,000 unique records containing NUL and `0xff`, consumed by 16 threads: no missing, duplicate, or corrupted candidates.
- The concurrent test passed under AddressSanitizer/UndefinedBehaviorSanitizer and ThreadSanitizer.

### Throughput

Five million nine-byte candidates through the feed harness, three interleaved runs:

| Reader | Times (seconds) | Mean |
| --- | --- | --- |
| existing `fread()` | 2.40, 2.50, 2.49 | 2.46 |
| descriptor `read()` | 2.45, 2.46, 2.39 | 2.43 |

This did not show a measurable throughput regression.

### DIFF
--- a/src/feeds/feed_stdin.c
+++ b/src/feeds/feed_stdin.c
@@ -63,6 +63,12 @@
 #include "thread.h"
 #include "feed.h"
 
+#include <errno.h>
+
+#if defined (_WIN)
+#include <io.h>
+#endif
+
 const int GENERIC_PLUGIN_VERSION = FEEDS_INTERFACE_VERSION_CURRENT;
 
 // A pipe carries text lines, so everything hashcat can do to a wordlist line applies here too. The
@@ -149,6 +155,7 @@ typedef struct stdin_global
   int filled_cnt;
 
   bool eof;      // the reader has seen the end of the input
+  bool read_error;
   bool stop;     // the session is over and the reader should wind up
 
   hc_thread_t reader;
@@ -199,6 +206,23 @@ static void stdin_error (generic_thread_ctx_t *thread_ctx, const char *msg)
   thread_ctx->error = true;
 
   snprintf (thread_ctx->error_msg, sizeof (thread_ctx->error_msg), "%s", msg);
+}
+
+static int stdin_read (char *buf, const size_t len)
+{
+  int rc_read;
+
+  do
+  {
+    #if defined (_WIN)
+    rc_read = _read (_fileno (stdin), buf, (unsigned int) len);
+    #else
+    rc_read = (int) read (fileno (stdin), buf, len);
+    #endif
+  }
+  while ((rc_read == -1) && (errno == EINTR));
+
+  return rc_read;
 }
 
 // The reader. It owns the descriptor and nothing else reads from it.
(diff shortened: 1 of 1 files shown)


jsteube:
Thanks, the approach looks right. I have two things I would like fixed before it goes in.

The first is error handling around read(). A return value of 0 means end of input, while minus 1 means the read failed. The current if (rc_read <= 0) treats both as normal EOF, so a real read error silently ends the candidate stream and the session looks complete even though the input was truncated. Your EINTR retry already handles the common harmless failure, which means anything left in that branch is worth reporting as an actual error.

There is one detail in where to report it. stdin_error() takes a generic_thread_ctx_t *, while stdin_reader() only has the stdin_global_t *. Calling it directly from the reader therefore will not compile. I think the smallest solution is a read_error flag in stdin_global_t, set beside eof when read() fails, with thread_next() returning GENERIC_RC_ERROR instead of GENERIC_RC_EOF when that flag is set.

Could you also rebase onto current master before this lands? The branch predates 6b68f1176, which split generic.h into feed.h and feed_ctx.h, so feed_stdin.c as currently pushed no longer compiles by itself.

Issue #4607, opened 2026-01-15
Large hash file does not load under error "Separator Unmatched"


AlexButHimself:
**Describe the bug**
Large hash file fails to load due to "Separator Unmatched" error.

**To Reproduce**
Using a large 750mb zip, use zip2john to create the hash (which should be around 22mil char) for the zip before placing it in hashcat under the command hashcat -a 0 -d 2 -m 13600 hash.txt wordlist.txt

**Expected behavior**
The program to work like normal, load the hash and start going through the wordlist to try to find the password

**Hardware/Compute device (please complete the following information):**
- Compute device name: RTX 3060
- OpenCL/CUDA driver name: OpenCl 3.0 Cuda
- OpenCL/CUDA driver version: 12.9.76
- OpenCL/CUDA driver source: No clue

**Hashcat version (please complete the following information):**
 - OS: Windows
 - Distribution: 22H2 version 19045.6691
 - Version: 7.1.2

**Diagnostic output compute devices:**

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 576.52                 Driver Version: 576.52         CUDA Version: 12.9     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                  Driver-Model | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA GeForce RTX 3060      WDDM  |   00000000:01:00.0  On |                  N/A |
|  0%   39C    P8             15W /  170W |    1570MiB /  12288MiB |     33%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```
```
Post clinfo output
```
Number of platforms:                             2
  Platform Profile:                              FULL_PROFILE
  Platform Version:                              OpenCL 3.0 CUDA 12.9.76
  Platform Name:                                 NVIDIA CUDA
  Platform Vendor:                               NVIDIA Corporation
[...]

Issue #4576, opened 2025-11-26
Keepass v4 (34300) Separator Unmatched


Nygosaki:
**GitHub is for bugs and features - not for support**
For support, please use the hashcat forums https://hashcat.net/forum/

**Check the FAQ**
Some items that might appear to be issues are not issues. Please review the hashcat FAQ https://hashcat.net/wiki/doku.php?id=frequently_asked_questions before submitting a bug report.

**Describe the bug**
When attempting to use the 34300 KeepassV4 hash type, hashcat states "separator unmatched" to an output from keepass2john (after stripping the file name at the start)

**To Reproduce**
Please provide us with all files required to reproduce the bug locally on our development systems. For instance: hash files, wordlists, rule files, ...
Generate a keepass v4 vault on default settings in keepassxc, get the hash using keepass2john, strip the start keeping "$keepass$*4*: and forward, attempt to use hashcat on the hash using hash type 34300.    
For example, when generating a vault with the password "password" (attached) this hash is produced

```
$keepass$*4*34*ef636ddf*67108864*19*4*19aeb99fd0d5f9af0a9568a76e917fd07e491a75246766e4176ad17565e84b70*f342fcee823374fc7e6467c8eba128d69ec9be80c045d4a5ec7dc878eac77f73*03d9a29a67fb4bb500000400021000000031c1f2e6bf714350be5805216afc5aff030400000001000000042000000019aeb99fd0d5f9af0a9568a76e917fd07e491a75246766e4176ad17565e84b700710000000e592d39d23d59a1afaeaf82b21449d2d0b8b00000000014205000000245555494410000000ef636ddf8c29444b91f7a9a403e30a0c05010000004908000000220000000000000005010000004d080000000000000400000000040100000050040000000400000042010000005320000000f342fcee823374fc7e6467c8eba128d69ec9be80c045d4a5ec7dc878eac77f7304010000005604000000130000000000040000000d0a0d0a*cf870dd5ee248743b2f326cba3a128f1f75d1d681408c3a06279b3720bdc8790
```

**Expected behavior**
Hashcat recognizes the hash and starts cracking it

**Hardware/Compute device (please complete the following information):**
- Compute device name: NVIDIA GeForce RTX 3080 Laptop GPU
- OpenCL/CUDA driver name: Nvidia Driver
- OpenCL/CUDA driver version: 581.8 / 13.0
- OpenCL/CUDA driver source: .exe installer

[...]

Issue #4063, opened 2024-08-14
 Separator unmatched No hashes loaded.


MiMaz7707:
**Describe the bug**

I got ": Separator unmatched" error when im trying to crack KRB_AS_REQ TIMESTAMP

```
# hashcat -m 19900 -a 0 $krb5pa$18$user.01$MYLAB.LOCAL$909755c35881e6ceb48473f70e7ba05bc496539586681532a61a8ba381f4dc73653dc3907e091ca3e4897ff117ad72de2c967b2f263f94d4$ ad-dict-pass.txt               
hashcat (v6.2.6) starting

OpenCL API (OpenCL 3.0 PoCL 5.0+debian  Linux, None+Asserts, RELOC, SPIR, LLVM 17.0.6, SLEEF, DISTRO, POCL_DEBUG) - Platform #1 [The pocl project]
==================================================================================================================================================
* Device #1: cpu-haswell-Intel(R) Core(TM) i7-6700HQ CPU @ 2.60GHz, 1425/2914 MB (512 MB allocatable), 4MCU

Minimum password length supported by kernel: 0
Maximum password length supported by kernel: 256

Hash '.01.LOCALc35881e6ceb48473f70e7ba05bc496539586681532a61a8ba381f4dc73653dc3907e091ca3e4897ff117ad72de2c967b2f263f94d4': Separator unmatched
No hashes loaded.

Started: Wed Aug 14 11:32:57 2024
Stopped: Wed Aug 14 11:32:57 2024
```

**Hardware/Compute device (please complete the following information):**
- Compute device name: [e.g. RTX2080Ti]
- OpenCL/CUDA driver name: [e.g. NVIDIA DRIVER]
- OpenCL/CUDA driver version: [e.g. 465.21]
- OpenCL/CUDA driver source: [e.g. runtime installer/.exe installer]

**Hashcat version (please complete the following information):**
 - OS: Linux
 - Distribution: Kali 2024
 - Version: 6.2.6

MiMaz7707:
Fixed after add single quote.

` hashcat -m 19900 -a 0 '$krb5pa$18$user.01$mylab.local$909755c35881e6ceb48473f70e7ba05bc496539586681532a61a8ba381f4dc73653dc3907e091ca3e4897ff117ad72de2c967b2f263f94d4' ad-dict-pass.txt
`

MiMaz7707:
fixed

Issue #3972, opened 2024-03-22
jwt16500Separator unmatched


sexchen:
**Describe the bug**
运行时提示Separator unmatched

**To Reproduce**
 .\hashcat.exe hashcat -a 3 -m 16500 'eyJhbGciOiJIUzI1NiJ9.eyIzNDM2MzQyMCI6NTc2ODc1NDd9.f1nXZ3V_Hrr6ee-AFCTLaHRnrkiKmio2t3JqwL32guY' ?u?l?l?l?l?l?l?l -i --increment-min=6

**Hardware/Compute device (please complete the following information):**
GTX 1660 Ti
Windows 10 1909

**Hashcat version (please complete the following information):**
hashcat-6.2.6

![1711111827961](https://github.com/hashcat/hashcat/assets/17498101/f151eefe-cdd9-4345-ba18-49c1fd210c6a)

I tried using jwtExample, but it prompted "Separate unmatched"

Chick3nman:
Remove the extra "hashcat" at the beginning of your command. This is not a hashcat issue, this is simply a command syntax problem.

sexchen:
> 删除命令开头多余的“hashcat”。这不是 hashcat 问题，这只是命令语法问题。
是的，删除多余的hashcat后成功运行。感谢
.\hashcat.exe -a 3 -m 16500 eyJhbGciOiJIUzI1NiJ9.eyIzNDM2MzQyMCI6NTc2ODc1NDd9.f1nXZ3V_Hrr6ee-AFCTLaHRnrkiKmio2t3JqwL32guY ?u?l?l?l?l?l?l?l -i --increment-min=6
![image](https://github.com/hashcat/hashcat/assets/17498101/dfdc60c4-225e-423a-b26c-ae6c4ef77405)

PR #4772, opened 2026-08-11
Show the offending token when hash parsing fails

thatux:
# Show the offending token when hash parsing fails with "Separator unmatched" / "Token length exception"

_Entirely vibe-coded, I can do a more thorough check if you like the idea._

## Why this is useful

New users hitting a malformed hash today get almost nothing to act on:

```
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Separator unmatched
```

There's no indication of *which* field is missing or *where* in the line things went wrong — just a generic sentence, even when the input is thousands of characters long. This is one of the most common "first five minutes with hashcat" frustrations, and it shows up across totally unrelated hash modes because they all share the same generic tokenizer:

- [#4576](https://github.com/hashcat/hashcat/issues/4576) — KeePass v4 (`-m 34300`) hash from `keepass2john` fails with no clue why
- [#4607](https://github.com/hashcat/hashcat/issues/4607) — ZIP (`-m 13600`) hash file (~22M chars) fails to load, no indication where
- [#3972](https://github.com/hashcat/hashcat/issues/3972) — JWT (`-m 16500`) hash not recognized
- [#3936](https://github.com/hashcat/hashcat/issues/3936) — SSH key (`-m 22951`) hash from hashcat's own examples fails the same way
- [#3949](https://github.com/hashcat/hashcat/issues/3949), [#4063](https://github.com/hashcat/hashcat/issues/4063), [#3752](https://github.com/hashcat/hashcat/issues/3752) — same story, different modes

In every case, the fix that actually helps a user is the same: point at the specific part of the input that didn't parse, instead of making them eyeball the whole (sometimes huge) hash by hand.

### Before / after

```
Before:
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Separator unmatched

After (a trailing field was dropped entirely, so no '*' was left to find):
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Separator unmatched near 'ef636ddf...' (expected '*')
```

```
Before:
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Token length exception

After (field too short):
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Token length exception near 'ef636dd' (8 required but only 7 entered)

After (field too long):
Hash parsing error in hashfile: 'hashes.txt' on line 1 (...): Token length exception near 'ef636ddfa' (only 8 allowed but 9 entered)
```

Snippets are truncated (64 bytes) and non-printable bytes are escaped as `\xHH`, so this also helps with the "invisible character" variant of this bug (stray BOM, whitespace, etc. — see #4576) rather than leaving it completely opaque. The length-mismatch wording adapts to which direction the mismatch went ("required but only entered" vs. "only allowed but entered") instead of using one phrase for both.

## What changed

- `input_tokenizer()` (`src/shared.c`) now captures the offending portion of the input whenever it returns `PARSER_SEPARATOR_UNMATCHED` or `PARSER_TOKEN_LENGTH` — plus, for a missing separator, which byte it was actually scanning for (e.g. `*`), and for a length mismatch, the expected vs. actual length in whichever direction the mismatch went.
- The message is bridged from `module_hash_decode()` back to `hashes.c` through `hashinfo_t` (`include/types.h`), a struct already passed into every `module_hash_decode()` and owned by the main binary. `hashinfo_t` is now allocated unconditionally per hash (previously only under `--username`/`--dynamic-x`/hash-copy/hash-split) so there's always somewhere for a module to report into; `hashes_destroy()` was adjusted to free it unconditionally too, avoiding both a leak and a double-free against the older, still-conditional lookup array.
- `src/hashes.c` prefers this detailed message over the generic one wherever a parsing error is printed.
- Every module's `module_hash_decode()` follows the same shape at the point it calls `input_tokenizer()`:
  ```c
  const int rc_tokenizer = input_tokenizer ((const u8 *) line_buf, line_len, &token);

  if (rc_tokenizer != PARSER_OK) return (rc_tokenizer);
  ```
  That one-line check was mechanically expanded, across all **540 module files** (541 call sites — one module has two), to also bridge the message:
  ```c
  if (rc_tokenizer != PARSER_OK)
  {
    if (hash_info != NULL) hash_info->parser_error_msg = tokenizer_error_dup (rc_tokenizer);

    return (rc_tokenizer);
  }
  ```
  Three modules with unusual control flow around their tokenizer call were left untouched rather than risk misapplying the pattern: `module_09800.c`/`module_09810.c`/`module_09820.c` retry with an alternate token layout before their real final check (which *is* covered), and `module_16801.c` falls back to an alternate PMKID hash-line format on failure without ever hitting this exact one-liner. `PARSER_HASH_LENGTH`/`PARSER_HASH_VALUE`/standalone `PARSER_SIGNATURE_UNMATCHED` checks (bespoke per-module digest/value validation, not the shared tokenizer) are still out of scope.

**Backward compat with un-rebuilt (e.g. third-party) modules:** the new `hashinfo_t` field is appended at the end of the struct, and `hashinfo_t` is only ever allocated by the main binary, never by a module — so a `.so` still compiled against the old layout just never touches the new field and silently falls back to the old, plain message for that mode. Verified by hand: reverted one module's source, compiled it standalone into an old-style `.so`, dropped it in next to the new main binary — no crash, correct fallback text, and rebuilding that one module brings the detailed message right back.

## How to reproduce

```sh
make -j$(nproc)
```

### KeePass v4 (`-m 34300`) — the original #4576 case

```sh
# Separator unmatched: trailing field dropped
printf '$keepass$*4*34*ef636ddf*67108864*19*4*%s*%s*%s' \
  "$(python3 -c "print('a'*64)")" "$(python3 -c "print('b'*64)")" "$(python3 -c "print('c'*506)")" \
  > dropped.hash
./hashcat -m 34300 -a 3 --potfile-disable --quiet dropped.hash '?a'
# -> Separator unmatched near 'cccc...c...' (expected '*')

# Token length exception: UUID field too short (7 hex chars instead of 8)
printf '$keepass$*4*34*ef636dd*67108864*19*4*%s*%s*%s*%s' \
  "$(python3 -c "print('a'*64)")" "$(python3 -c "print('b'*64)")" "$(python3 -c "print('c'*506)")" "$(python3 -c "print('d'*64)")" \
  > baduuid.hash
./hashcat -m 34300 -a 3 --potfile-disable --quiet baduuid.hash '?a'
# -> Token length exception near 'ef636dd' (8 required but only 7 entered)

# Token length exception: UUID field too long (9 hex chars instead of 8)
printf '$keepass$*4*34*ef636ddfa*67108864*19*4*%s*%s*%s*%s' \
  "$(python3 -c "print('a'*64)")" "$(python3 -c "print('b'*64)")" "$(python3 -c "print('c'*506)")" "$(python3 -c "print('d'*64)")" \
  > longuuid.hash
./hashcat -m 34300 -a 3 --potfile-disable --quiet longuuid.hash '?a'
# -> Token length exception near 'ef636ddfa' (only 8 allowed but 9 entered)
```

### ZIP (`-m 13600`) — #4607

```sh
printf '$zip2$*0*1*0*abc*def*trailinggarbagewithoutproperseparator' > badzip.hash
./hashcat -m 13600 -a 3 --potfile-disable --quiet badzip.hash '?a'
# -> Separator unmatched near 'trailinggarbagewithoutproperseparator' (expected '*')
```

### JWT (`-m 16500`) — #3972

```sh
printf 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIn0' > badjwt.hash
./hashcat -m 16500 -a 3 --potfile-disable --quiet badjwt.hash '?a'
# -> Separator unmatched near 'eyJzdWIiOiIxMjM0NTY3ODkwIn0' (expected '.')
```

### md5crypt (`-m 500`)

```sh
printf '$1$abcd' > badmd5crypt.hash
./hashcat -m 500 -a 3 --potfile-disable --quiet badmd5crypt.hash '?a'
# -> Separator unmatched near 'abcd' (expected '$')
```

### BitLocker (`-m 22100`) — also just the mechanical rollout

```sh
printf '$bitlocker$0$16$aaaa$1000$12$bbbb$60$cccc' > badbitlocker.hash
./hashcat -m 22100 -a 3 --potfile-disable --quiet badbitlocker.hash '?a'
# -> Token length exception near 'aaaa$1000$12$bbbb$60$cccc' (32 required but only 25 entered)
```

A valid hash (all fields present, correct lengths) for any of the above should still parse silently — no regression for the success path.


jsteube:
TBH, it is the right thing to do. Pointing directly at the field that failed instead of making someone inspect a line character by character line is definitely the right fix. The module changes are also all identical, so that part looks clean.

There is one correctness issue though. The comment says parsing happens on a single thread before cracking starts, but outfile_check.c calls module_hash_decode from its own thread for the whole run. That means the three static variables can be written there while the main thread reads them through strparser(). Making them thread local should fix that.

The bigger question for me is where this logic should live. If every module has to call tokenizer_error_dup(), then third party modules do not get it automatically, and every new module has to remember adding those lines. It also means hashinfo_t gets allocated for every hash now, which can become expensive on very large hash lists.

I think it would be cleaner to keep this entirely inside input_tokenizer and strparser. They already have the information, and then every module gets the feature automatically. That would also reduce the change from hundreds of files down to just two. The only case I'd check carefully is autodetect, where several modules may parse the same line and the stored message could end up coming from the last one.

thatux:
tnx!

> I think it would be cleaner to keep this entirely inside input_tokenizer and strparser.

Agreed, however, I tried this to start with, but each module gets a private copy of shared.c statically compiled in (AFAI understand now), so that prevents of pushing the error up to hashcat from the module?

thatux:
that should fix
1. hashinfo_t is now only allocated when needed (we use a scratchpad instead)
2. the message-copying logic moved into one centralized tokenizer_error_report() function that frees any prior message before overwriting, fixing a real leak in the split-hash reuse path. 
3. the three transient capture statics in shared.c are now thread-local (HC_THREAD_LOCAL)

jsteube:
The idea behind this PR is justified, but you have to work around an old build system model that forces you to add code to the module, which is exactly what I am trying to avoid. I am using this opportunity to refactor that build model now. Once that is done, integrating your changes as planned should be much easier. I will update you when it is ready.

jsteube:
Merging this now. Thanks for pushing on this, the credit for making me look at it is yours.

What you ran into is a hashcat problem, not a problem with your patch. Every module statically links its own copy of shared.c, so the error the tokenizer captures is not the one the core can read, and the only way through was the call you added to all modules.

The idea is the core is becoming a shared library. Then there is one input_tokenizer() and one captured error. The core reads it directly, and no module has to do anything. The change I'll push soon after this merge replaces some of your patches because hashcat then no longer suffers from that problem.

thatux:
you're the best!

PR #4774, opened 2026-08-13
Add compute sanitizer + asan

thatux:
I was testing the [CUDA Compute Sanitizer](https://developer.nvidia.com/compute-sanitizer), and vibe-coded a test-harness for hashcat. It found a pretty serious bug. I'm not sure why test.sh does not spot this.

These crack on this PR but don't crack on current [master](https://github.com/hashcat/hashcat/commit/77e4ce181d5fcbba19932c31368aaf8448433c62):
```
# MD5  — md5(utf16le($pass).$salt), password "43"
./hashcat -m 30 -a 3 -O --potfile-disable \
  'f08ca4a9c2308460a540a14d302a17a6:2429302669753941929414132' '4?d'

# NTLM — md4(utf16le($pass)), password "hashcathashca42"
./hashcat -m 1000 -a 3 -O --potfile-disable \
  87f0077d7918f2f9c191185112d24576 'hashcathashca?d?d'

# SHA1 — sha1(utf16le($pass).$salt), password "hashcatpw42"
./hashcat -m 130 -a 3 -O --potfile-disable \
  'b629b3a761edde0da043a01406bb31999201a61a:12345678' 'hashcatpw?d?d'

# SHA1 — sha1($salt.utf16le($pass)), password "hashcathashca42"
./hashcat -m 140 -a 3 -O --potfile-disable \
  '01d9857c1ae438c9067bc2b6f4d0b87810b3ba34:12345678' 'hashcathashca?d?d'
```
sometimes I get `cuEventDestroy(): an illegal memory access was encountered` but no cracks.

First I was working on [Valgrind ](https://github.com/thatux/hashcat/tree/valgrind) but that was too slow on .cl files with pocl, I'd still like to try use the combination of Valgrind on the .c files and Compute Sanitizer on the .cl files.

_the remainder of this PR is LLM context, and tools/compute_sanitizer is completely vibe-coded_

# `tools/compute_sanitizer/`: run hashcat's CUDA kernels under NVIDIA Compute Sanitizer

Adds a memory-safety harness that runs hashcat's real CUDA kernels on real GPU
hardware under NVIDIA Compute Sanitizer, and triages each finding down to a
`OpenCL/<file>:<line>` with kernel name and thread/block coordinates.

This is the GPU-kernel counterpart to `DEBUG=2` (hashcat's own ASan mode, host
C only). It replaces an earlier PoCL + Valgrind approach that emulated GPU
kernels on the CPU under Valgrind instrumentation — real hardware is both much
faster and checks the code that actually runs. It is CUDA-only by design and
does not touch the OpenCL/HIP/Metal paths.

## Why it exists

Emulating kernels under Valgrind was too slow to use as a regression tool: a
single GPG mode took 18+ minutes. The same mode under Compute Sanitizer on a
GTX 1650 runs in ~13 seconds. Running on the real backend also removes a whole
class of "does this even reproduce off-device" doubt.

It has already found a real, previously-unreported out-of-bounds read in the
markov candidate generator (`generate_pw()` in `markov_le.cl`/`markov_be.cl`),
affecting common `-a 3` mask attacks. That finding and its fix are handled in a
separate follow-up and are not part of this PR.

## The one source change: `src/backend.c`

`load_kernel()`'s CUDA/NVRTC branch gains two `#if defined (DEBUG)` blocks.
**Release builds are byte-for-byte unchanged** — every added block has an
`#else` (or is absent) that reproduces the previous behaviour exactly.

1. Program name. The 4th argument to `hc_nvrtcCreateProgram()` was the generic
   per-category literal (`"main_kernel"`, `"shared_kernel"`, `"mp_kernel"`,
   `"amp_kernel"`). NVRTC compiles from an in-memory string, so with a generic
   name the embedded line-info table has nothing meaningful to point at and
   Compute Sanitizer reports `main_kernel:527`. In a DEBUG build the argument
   becomes `filename_from_filepath (source_file)` — the real `.cl` name already
   in scope — so the report reads `m17010-pure.cl:527`. In a release build the
   `#else` keeps `kernel_name`, so the call is identical to before:

   ```c
   #if defined (DEBUG)
   const char *nvrtc_program_name = filename_from_filepath (source_file);
   #else
   const char *nvrtc_program_name = kernel_name;
   #endif

   if (hc_nvrtcCreateProgram (hashcat_ctx, &program, kernel_sources[0], nvrtc_program_name, 0, NULL, NULL) == -1) return false;
   ```

2. Line info. A DEBUG build appends `--generate-line-info` to the NVRTC
   options. This keeps kernels optimized but embeds a line table, so
   `--show-backtrace device` resolves to source. Deliberately not `-G` (full
   device debug), which would disable kernel optimization and change the very
   memory-access patterns being tested.

No kernel-cache-key change was needed: the cache key already folds in
`COMPTIME` (`src/Makefile`), which is a fresh timestamp at every link, so a
`make clean && make DEBUG=1` rebuild naturally invalidates any previously
cached kernel binary.

## New tool: `tools/compute_sanitizer/`

Mirrors the structure of the earlier Valgrind tool, adapted to Compute
Sanitizer's simpler, directly-source-resolving log format.

```
tools/compute_sanitizer/
  run.py          # build / check / exec / selftest
  triage.py       # parse compute-sanitizer logs, classify, JSON + terminal
  report.py       # aggregate past runs into a table
  sweep_shim.sh   # BIN indirection target for test.sh/test_edge.sh
  README.md
  modules/        # ground-truth .cu selftest fixtures + expected.json
```

- `run.py build` — `make clean && make DEBUG=1`, copy to a stable
  `./hashcat-sanitizer` sibling, sanity-check it runs.
- `run.py check` — found/missing table for `compute-sanitizer`, `nvcc`,
  hashcat debug info, a visible CUDA device.
- `run.py exec <name> [--tool memcheck|racecheck|synccheck|initcheck] -- <hashcat cmd>`
  — forces CUDA-only (`--backend-ignore-opencl/hip[/metal]`), runs under
  `compute-sanitizer`, writes `command.txt` / `environment.txt` /
  `sanitizer.log` / `summary.{txt,json}` to a fresh timestamped results dir.
- `run.py selftest` — builds the `modules/*.cu` fixtures with `nvcc
  -lineinfo` and asserts each produces the expected finding kind.

Key behaviour: **never passes `--error-exitcode`**, and always
`--check-exit-code no`, so hashcat's own exit code stays authoritative. The
sanitizer verdict comes purely from parsing the log. A real kernel fault
poisons the CUDA context, producing a cascade of `unspecified launch failure`
on every following API call; the parser marks the first real fault `primary`
and those consequential `cudaError*` blocks `secondary`, so the primary
PASS/FAIL count is not drowned out.

## `test.sh` / `test_edge.sh` integration

Both accept `--compute-sanitizer[=<tool>]`, reusing their existing
hash-mode/attack-mode/kernel-type/vector-width test generation:

```
tools/test_edge.sh -m 17010 -a 3 -V 1 --compute-sanitizer
tools/test_edge.sh -m 17010 --compute-sanitizer=initcheck
tools/test.sh --compute-sanitizer
```

Each hashcat invocation is routed through `sweep_shim.sh`, which wraps it in
`run.py exec --sweep`: hashcat's stdout/stderr/exit-code pass through
untouched (so existing pass/fail parsing is unaffected), sanitizer findings
land in a `tools/compute_sanitizer/results/sweep-<ts>/` tree, and a one-line
pointer to `report.py --dir <that tree>` prints at the end. Requires
`run.py build` to have been run first; both scripts check for
`./hashcat-sanitizer` and exit with a clear message if it is missing.

## Verification

- `run.py selftest` — three permanent `modules/*.cu` fixtures (`oob_write`,
  `oob_read`, `uninit_global`) each classified to the expected finding kind,
  including the `initcheck`-specific log shape.
- Injected a one-line OOB read into `OpenCL/m00000_a3-pure.cl`, ran
  `test_edge.sh -m 0 -a 3 -K 0 -t single -V 1 --compute-sanitizer`, and got
  back `InvalidGlobalRead` at `m00000_a3-pure.cl:105` in `m00000_sxx()` with
  thread/block coordinates. Reverted the injection afterwards.
- Exit-code decoupling: a deliberately malformed hash (hashcat rc -1, no
  sanitizer finding) is not reported as a sanitizer failure, and vice versa.
- Timing: GPG mode 17010 example hash — ~13s here vs 18+ min under the earlier
  PoCL+Valgrind path.
- Release build unaffected: `make clean && make` produces a `./hashcat` with
  no `.debug_*` sections and the NVRTC path unchanged.

## Notes / scope

- CUDA-only. On non-NVIDIA setups this tool does nothing; OpenCL/HIP/Metal
  kernels are out of scope.
- Requires the CUDA toolkit's `compute-sanitizer` and `nvcc` (for the selftest
  fixtures) on the machine; not auto-installed.
- The `hashcat-sanitizer` sibling binary, per-run `results/`, and the compiled
  `.cu` fixtures are gitignored; fixture sources are tracked.


jsteube:
This is a good find, and the analysis holds up. I reproduced it on an AMD card through HIP, so this is not a CUDA artifact. It is the buffer sizing.

One thing to fix in the description: it says the generate_pw() fix is in a separate follow up, but it is already part of this pull request in src/backend.c.

What I need changed is the packaging, not the content. There is a four line correctness fix here for a bug mixed with about 1400 lines of new CUDA only tooling. Those need very different reviews, and the bug fix should not wait. Please split this into two pull requests: one with only the SP_PW_MAX restoration and its comment and another with tools/compute_sanitizer/, the two DEBUG blocks in load_kernel() and the test.sh integration.

thatux:
tnx, moved fix to separate PR: https://github.com/hashcat/hashcat/pull/4775
also removed the explanation from this PR.
I'll work on this PR a bit more, so I've converted it to draft.

jsteube:
The four commands in your description crack on current master. Running each of them 10 times on an RTX 2080 Ti and again on an RTX 4090, all four cracked every time, with and without -O. At 77e4ce181, the commit you compared against, all four fail on both cards. The fix is e5a3cac59. The two functions that copy the mask buffers to each device, backend_session_update_mp() and backend_session_update_mp_rl(), run from the main thread, which carries no CUDA context of its own, and they worked only while device enumeration happened to leave one current. AMD was never affected, which is why this reached master at all.

In "tools/asan/parse_harness.c", the mutation block opens with this comment:

> Mutations. Deliberately conservative and structural rather than random: the
> point is reproducibility, not fuzz volume. A real fuzzing campaign belongs
> behind libFuzzer/AFL with this same entry point.

Hashcat has no coverage-guided fuzzing anywhere today, and two functions that sit under every module parser have no direct coverage either. The function input_tokenizer() in "include/parser.h" splits a hash line into fields, and cpu_rule_to_kernel_rule() in "include/rp.h" compiles a rule. A sweep over module_hash_decode() reaches both, but only along the paths a valid example hash and its mutations take.

Would you be up for adding the libFuzzer front end you describe, or would you rather it landed separately on top of this? The function parse_one() already takes a line and a length and frees what it allocates, so it is already the shape of an LLVMFuzzerTestOneInput() body, and those two functions would be two more targets beside it. One thing I would want in the CI job whenever it arrives is the corpus kept between runs, because a coverage-guided run that starts from the seed corpus every time relearns the same paths and never reaches past them.

thatux:
# Coverage guided fuzzing, on top of the sanitizer tooling

This extends #4774 rather than opening a second pull request, which answers the question asked in
https://github.com/hashcat/hashcat/pull/4774#issuecomment-5742417853: the parser target needs the
`hashconfig` builder the ASan harness here already has, and putting them side by side would have
meant writing it twice.

A merge from master, then five commits:

    30fb0027d  Merge branch 'master' into compute_sanitizer
    1e9a1eba4  Add libFuzzer targets for the rule compiler and the token splitter
    4ffa21df3  Run both fuzz targets nightly and keep the corpus between runs
    f9151199d  Add a parser target for one mode at a time
    6b13c8dbe  Document the fuzz targets and what the corpus cache is for
    98091d063  Share the hashconfig builder with the ASan harness

Three targets, the three functions that read text a user did not write:

| target | entry point | what the input is |
| --- | --- | --- |
| `fuzz_parse_<mode>` | `module_hash_decode ()` of that mode | the hash line, byte for byte |
| `fuzz_tokenizer` | `input_tokenizer ()` in src/parser.c | a token spec, then the line |
| `fuzz_rule` | `cpu_rule_to_kernel_rule ()` in src/rp.c | the rule, byte for byte |

## What it found

Nineteen parsers and three shared helpers read past the length they were given, and the four pkzip
parsers leak the line on every error path. The fixes are a pull request of their own, against
master, and it should land first: until it does, the nightly job added here is red.

    printf '$' > /tmp/rule_dollar
    tools/fuzz/build.sh
    ./fuzz_out/fuzz_rule /tmp/rule_dollar

    AddressSanitizer: heap-buffer-overflow, READ of size 1
      rule_utf8_len src/rp.c:292
      cpu_rule_to_kernel_rule src/rp.c:384
    rc=1

That one turned up 16,138 executions into the first run the target ever did, seeded from the rule
files in rules/ and without a dictionary. The others are `rounds_count_length ()` at
src/parser.c:286, which takes an `input_len` and then searches with `strchr ()`, the `%s` in
module_22000.c:848, and the leak in the pkzip parsers.

## Building and running

    tools/fuzz/build.sh
    ./fuzz_out/fuzz_parse_17225 corpus/17225 fuzz_out/work/seeds/parse_17225 -dict=fuzz_out/fuzz_parse.dict

Needs clang, because libFuzzer is a clang feature and the DEBUG=2 build in src/Makefile is not.
The same script serves a build service: it takes `CC`, `CFLAGS`, `LIB_FUZZING_ENGINE`, `OUT` and
`WORK` from the environment wherever they are set, and reads its include paths and its source list
from what the tree actually has.

A target links the code its entry point reaches and nothing else. The file layer, the folder
layout and the generator behind `generate_random_rule ()` are stubs in tools/fuzz/stubs.c rather
than linked, because no target opens a file, and that keeps the compression libraries and the rest
of the tool out of the binary.

## One parser target per mode

Every `src/modules/module_XXXXX.c` defines `module_init`, `module_hash_decode` and the rest of the
API with external linkage, so two modules cannot be linked into one binary. Each parser target
links its own module statically and takes the mode from `-DFUZZ_HASH_MODE`. No `dlopen ()`, no
plugin beside the binary, and one corpus per mode, which is what a per mode format deserves: an
input that reaches deep into the pkzip parser says nothing about sha256crypt.

The `hashconfig` a parser reads is built from the module's own getters the way `interface.c` builds
it, and the buffers are sized from the module's own `dgst_size`, `esalt_size` and `hook_salt_size`,
so an overflow in a target is an overflow of what hashcat would allocate. That is
tools/asan/hashconfig.c now, taken out of parse_harness.c and built into both harnesses, so the
thing a parser sees is defined once.

`FUZZ_MODES` in tools/fuzz/build.sh holds four modes rather than all 599, and the nightly matrix
lists the same four. Two caveats worth reading before agreeing to them:

* The list is written twice, once in build.sh and once in .github/workflows/fuzz.yml. Add a mode to
  `FUZZ_MODES` and the build makes a binary no job runs. The workflow should read the list out of
  build.sh instead, and will if you want it before this lands.
* The four were picked by reasoning and there is now evidence. 17225 and 22000 have both needed a
  memory safety fix in their parsers, 7400 carries a rounds field, and all three have produced
  findings. 29100 was picked because tools/asan/parse_harness.c says its field pump reached that
  mode's overflow by accident, and it has produced nothing in any run. Meanwhile a sweep of 74
  modes found bugs in fourteen others that get no nightly time at all.

## Why this finds what the harness in this branch cannot

`mutate ()` in tools/asan/parse_harness.c is structure aware and better than random mutation at
what it does: truncation at every offset, a length that lies about the buffer, and the charset
preserving pump that binary searches each field to its acceptance boundary. That is the length
overflow class, found by construction rather than by luck.

What it cannot reach, because every family moves one axis of one valid example:

* the other branches of a format with several shapes, where the example hash pins one and the
  mutations stay inside it
* two fields wrong at once
* values rather than lengths, which is where `-use_value_profile=1` and clang's interception of
  `memcmp ()` earn their keep

## Two decisions worth arguing with

**Exact sized buffers.** Each target hands the function an allocation of exactly the length it
passes with it, so a read past that length is a read past the allocation. hashcat's own callers
pass a NUL terminated line buffer, so the same read lands on the terminator there. That makes these
findings correctness findings rather than crashes, and they still have to be fixed: the byte read
is not part of the rule or the line, the function was told how long its input is, and hashes.c
already hands parsers slices of a longer buffer. `-DFUZZ_NUL_TERMINATE` builds the conservative
view of the same run.

**The token spec is part of the tokenizer input.** A sweep over the modules reaches only the specs
the tree declares today; taking the spec from the first bytes reaches the ones the next module will
declare. Two specs are held back, both of which say something about the spec rather than about the
line: a token with no separator and no length, which leaves the next token pointing at nothing and
is dereferenced rather than rejected, and one whose length the tokenizer never measures against
what is left of the line, which walks the next token past the end of a short line. module_34300.c
declares the second and guards it with its own `line_len` check before calling the tokenizer. Both
would stop being reachable if the tokenizer bounded that advance the way it already bounds a fixed
length token, and that is a change to core parsing rather than to tooling.

## The corpus is the point of the CI job

A coverage guided run that starts from the seed corpus every time relearns the same paths and never
reaches past them, so each target keeps its corpus in the actions cache. Three of that cache's
rules decide whether it works:

* an entry cannot be written twice under one key, so the key carries the run id and the restore
  falls back to the newest entry with the same prefix
* an entry is evicted after seven days without a read, which is why this is nightly and not weekly
  like bsd.yml
* the repository holds 10 GB of cache and evicts the least recently used entry when it is full,
  which is why every run merges the corpus down to the inputs that still reach new code before it
  saves

Only the scheduled run saves. A cache written by a pull request lives in that pull request's scope
and is deleted with it, and a fork cannot write one at all, so a pull request fuzzes for two
minutes as a smoke test and nothing outside the default branch can put an input into the corpus. A
crash fails the job and the input is uploaded as an artifact.

Where that is not durable enough, the same targets work against a corpus branch or a small corpus
repository force pushed each night, which has no eviction clock and can be cloned by anyone
reproducing a finding.

## The merge from master

The branch was four weeks and 214 commits behind, which mattered here: a target reports whatever is
in the tree it was built in, and on the old base these were finding things master had already
fixed, such as the class based rule selector at src/rp.c:512.

Three files needed a hand and the merge commit says why. src/backend.c and tools/test_edge.sh were
rewritten on master, so both took master's copy with this branch's changes re-applied: the DEBUG
build still names the .cl file to NVRTC and still asks for `--generate-line-info`, and twenty
command lines go through `HC_BIN` rather than the fourteen this branch last saw. src/Makefile keeps
this branch's `SANITIZE` variable, which defaults to address and so leaves DEBUG=2 doing what
master's line did. `git show --remerge-diff 30fb0027d` is the whole of it, five files.

## OSS-Fuzz, as a suggestion rather than a patch
[oss_fuzz_hashcat.patch](https://github.com/user-attachments/files/32445202/oss_fuzz_hashcat.patch)

Nothing here touches [OSS-Fuzz](https://github.com/google/oss-fuzz), and I would rather ask than
assume. If you want these run continuously rather than nightly, that is where it belongs: it keeps
the corpus, deduplicates crashes, bisects each finding to the commit that introduced it, and costs
the project no hardware. hashcat is not enrolled today, `projects/hashcat/` does not exist there.

The integration is the three files in the
[new project guide](https://google.github.io/oss-fuzz/getting-started/new-project-guide/) in a pull
request against google/oss-fuzz. It is written and tested here already:

* `project.yaml` is where `primary_contact` and `auto_ccs` go, and they are the one part left blank
  rather than guessed at.
* the `Dockerfile` installs exactly one package, `zip`, for the seed corpora. The base builder
  image does not ship it, and the targets need nothing else.
* their `build.sh` is one line calling tools/fuzz/build.sh, so the build stays with the code it
  tests and a mode added to `FUZZ_MODES` never needs a change in the oss-fuzz repository.

Two things would have to be settled first and neither is technical:

* `primary_contact` has to be a hashcat maintainer with a Google account. That address is how
  OSS-Fuzz verifies the integration is wanted and it is who receives the reports, so I cannot enrol
  the project on my own. The criteria are in
  [accepting new projects](https://google.github.io/oss-fuzz/getting-started/accepting-new-projects/).
* OSS-Fuzz publishes a bug 30 days after it is fixed or 90 days after it is reported, whichever
  comes first, under its
  [disclosure guidelines](https://google.github.io/oss-fuzz/getting-started/bug-disclosure-guidelines/).
  For a tool whose users feed it files from strangers that is a policy decision.

Say the word and I will open that pull request; until then the nightly job here is the whole story.

## What the nightly job does not cover

595 of the 599 modes. That is a decision about machine time rather than a limit of the tooling: the
same variable builds every mode there is, and tools/fuzz/README.md says how.

    FUZZ_MODES="$(ls src/modules/module_*.c | sed 's#.*module_##;s#\.c##')" tools/fuzz/build.sh

Measured here, that is about 13 minutes of building and 6 GB of binaries, one per mode at roughly
10 MB, and a minute on each of them is ten hours in a row or about an hour across twelve cores. The
README has the parallel loop. Doing that is also how the mode number came to be read with `10#`
rather than by stripping zeros with sed, which turned 00000 into an empty `-DFUZZ_HASH_MODE=` and
broke the build for mode 0 alone.

So how far the nightly job grows, and how long each target gets, is worth arguing about once it has
run for a week; nothing has to change in the tooling for it.

## Test runs

12 cores, clang 16, AddressSanitizer throughout.

Against this branch, after the merge, twenty seconds on each of the six targets:

    four distinct findings, all of them the ones the parser pull request fixes:
      rule_utf8_len ()        src/rp.c:292          from fuzz_rule
      rounds_count_length ()  src/parser.c:286      from fuzz_tokenizer and fuzz_parse_7400
      the pkzip leak          module_17225.c:222    from fuzz_parse_17225
      the "%s"                module_22000.c:848    from fuzz_parse_22000

    nothing the merge brought in is new

Against a tree carrying those fixes, ten minutes on each of the two entry point targets:

    fuzz_rule        315,599,572 executions   480 edges   corpus 1399   no findings
    fuzz_tokenizer   171,763,073 executions   250 edges   corpus  467   no findings

and a minute on each parser target of every mode the fixes touch, plus three minutes on these two:

    30 targets, 330,819,873 executions, no crashes, no leaks, no artifacts written

jsteube:
The 3 directories this adds under "tools/" are a shape I am happy with, and what they hold is sources and scripts with the build output ignored. Whether they meet the bar is a review rather than a question, and that comes when you mark this ready.

What would help before then is the list of decisions that are mine rather than yours, so you do not build on an answer I would not give. The nightly workflow is one of them: it is recurring Actions time and a share of the cache budget on the project's account, and that is a call to make rather than to find made. Are there others you can already see coming?

thatux:
The policy/resource decisions I can see are:

1. **CI fuzzing**

   * Decision needed: do you want fuzzing in CI?
   * My recommended default: **yes**.
   * On PRs: fuzz only the modes impacted by the PR, with a bounded runtime.
   * Separately: run all modes weekly, but only if there have been new commits since the previous run.

2. **OSS-Fuzz**

   * Decision needed: do you want hashcat enrolled in OSS-Fuzz?
   * My recommended default: **yes, if you are comfortable with the disclosure policy**.
   * Google provides the compute, but findings are subject to its disclosure policy, with publication after at most 90 days if still unresolved.

3. **`test.sh` in CI**

   * Decision needed: do you want `test.sh` to run automatically on pull requests?
   * My recommended default: **yes**.
   * On PRs: run it only for the modes impacted by the PR, with a bounded runtime.
   * Separately: run the broader/full suite weekly when there have been new commits.
   * Before releases: run a larger manually triggered `test.sh` campaign.

4. **Recurring CI resource budget**

   * Decision needed: how much recurring GitHub Actions time and cache/storage budget should this tooling be allowed to consume overall?
   * My recommended default: keep PR CI narrow and change-focused, spend the heavier budget on the weekly all-mode run, and reserve the largest runs for manual pre-release testing.

So my default policy would be:

* PR: impacted-mode `test.sh` + impacted-mode fuzzing.
* Weekly, if there were new commits: all-mode `test.sh` + fuzzing.
* Before release: larger manually triggered `test.sh` + fuzzing campaign.
* OSS-Fuzz: enroll if the disclosure policy is acceptable.

Ultimately I think the project is strengthened by the combination of `test.sh`, coverage-guided fuzzing, and AI-assisted inspection.

For AI inspection, I would keep that as contributor/reviewer work for now rather than add it to CI in this PR. GitHub also has organization-level Copilot code review that could be evaluated separately for the hashcat organization:
https://docs.github.com/en/copilot/how-tos/copilot-on-github/set-up-copilot/configure-code-review alternatively: https://code.claude.com/docs/en/code-review

I don't currently see other policy/resource decisions regarding this.

thatux:
Plz merge #4895 first, that should solve the failing builds here.

jsteube:
The PR fuzzing job in ".github/workflows/fuzz.yml" is the part I would change before this goes in. The header currently says:

    ##   pull request   the parser of every mode the pull request changes, and the
    ##                  rule and tokenizer targets plus a starting set of four
    ##                  parsers when it changes host code they link. 120 seconds a
    ##                  target, at most 20 modes.

Could the PR job replay the existing corpus and stop there, leaving exploration to the weekly and manual runs? A job that generates new inputs can give a different answer when rerun, so I do not think it works well as a gate. That also answers the recurring CI budget question for me. I would keep the PR path to corpus replay and spend the heavier budget on weekly and manual exploration. Up to 20 targets at 120 seconds each ties up concurrency that every other PR in the queue has to share, which matters more on a public repository than the runner minutes themselves.

On OSS Fuzz, no. The 90 day disclosure deadline is not the reason, since "SECURITY.md" already gives a reporter that long here. The issue is that findings become a public feed of memory bugs with reproducers attached, published on a schedule set elsewhere. Which findings become public, and when, is a call I would rather keep with the project.

On "test.sh" in CI, yes, once the move to Python is complete. Installing the Perl modules is awkward enough on a workstation and worse on a runner. The Python suite removes that class of problem instead of working around it, so finishing the migration is what turns that job on.

thatux:
Thanks, all three taken into account.

**Fuzzing on pull requests.** The pull request job now replays the saved corpus and the seeds once each and stops (`-runs=0`, f82e91499). It generates no new inputs, so a rerun gives the same answer, and a replay takes seconds a target instead of 120, which also settles the concurrency cost. The weekly and manual runs keep exploring and keep saving the corpus. Checked on a target that crashes on one input:

    ./fuzz_parse_NNNNN corpus seeds -artifact_prefix=crashes/ -runs=0 -timeout=25
    # crashing input in the corpus: rc=1, reproducer written to crashes/ (the job summary picks it up)
    # without it: rc=0, two runs give identical coverage, and the corpus is left unchanged

**OSS-Fuzz.** Understood. Nothing in the PR sets it up, and I removed the README and build.sh lines that pointed at it (67f31f322).

**test.sh in CI.** I would prefer to keep this one in now rather than wait for the Python move, because it already works:

* `test.sh minimal` passes on the current merge with master: https://github.com/hashcat/hashcat/actions/runs/35985894211/job/107588441502. test.sh itself exits 0 either way, so the job greps its output and fails on any `Kernel-Type ... > Error :` result line and on a mode that gives no result line at all.
* The Perl modules are installed once, by `tools/install_modules.sh`, and `~/perl5` is cached under a key made from that script's hash. A runner pays for the install on the first run and when the module list changes, not on every run.
* It has already paid for itself: the 34100 self-test segfault it hit on this PR became #4913, merged today.

When the Python migration lands, the job switches over and the Perl step goes away. Nothing else in the workflow has to change for that. If the runner budget is the concern, I can hold the pull request path to `-M` only.

thatux:
I plan to do a final review and than mark this as 'ready to review'

thatux:
Ran the final review before marking this ready, at full strength, and it turned up a set of things where a real failure could have slipped through CI or work could have been lost. All fixed and pushed. Head is now e63bc1875.

**CI that could pass on a broken change**

* The test.sh gate only failed on `Kernel-Type ... > Error :` lines. A kernel that fails to compile (hashcat exits -5, which test.sh reports as a Skip), a `Compare Error` line, and test.sh's own container-error lines all went green. The gate now fails on each of those. A Warning (a case that ran into the time limit on a shared runner) is annotated and does not fail. Checked by running the step against a fake test.sh per outcome.
* The ASan sweep logged a module as clean when the harness could not run it at all (a failed dlopen, a missing module_init, or a stale plugin tree), so a stale build read as clean across the board. It now reports those as NOT RUN and counts them, and the harness sets a distinct exit code on a sanitizer finding. The Compute Sanitizer report likewise printed PASS when there was no log to read, and it now prints NO-LOG and the sweep returns nonzero.

**Fuzzing that could lose the corpus or miss a bug**

* A parser target with no example hash (14600 ships ST_HASH NULL) had no seed directory, and libFuzzer exits 1 on a missing directory, so the target was reported as a finding. Every target now gets a seed directory, and 28600's multi-literal example hash is read in full.
* A failed corpus merge ran under `|| true` and then replaced the corpus with the empty merge directory. The corpus is only replaced after a merge that succeeded now.
* A manual run defaulted to 900 s a target, which on the largest shard is over 12 h against a 300 minute job, so it was cancelled and its corpus (saved only on a non-cancelled job) was lost. The seconds a target are sized to the job now, and the step carries its own limit so the save still runs.
* fuzz_report.py did not escape `:` or `,` in an annotation title, so a UBSan finding's title came out garbled. Fixed.

**Smaller**

* The DEBUG build this PR adds for Compute Sanitizer compiles CUDA kernels with `--generate-line-info` and names the NVRTC program after the .cl file, so a finding resolves to a source line. Neither difference was in the kernel cache key. The key starts with COMPTIME, which is normally the build time, so two builds never share a cache entry. But src/Makefile pins COMPTIME to SOURCE_DATE_EPOCH when that is set, and then a `hashcat-sanitizer` built by run.py and a release build in the same tree get the same key. A sanitizer run can then reuse a release kernel without line info (backtraces lose their source lines), or a release run can reuse a DEBUG kernel. Results are never wrong, only the diagnostics. A marker in both CUDA cache keys, present in DEBUG builds only, keeps them apart and leaves the release key unchanged.
* The `--compute-sanitizer` sweep prints the `report.py --dir` pointer at the end now, as the help text promised.
* Moved the cache steps to actions/cache v6 (v4 runs on the deprecated Node 20).
* Removed the OSS-Fuzz-only leftovers (the `.options` files and the `_seed_corpus.zip` packing, unused by CI), read the plugin ABI version and the starting fuzz mode list from their one source instead of a second copy, and split the comments CONTRIBUTING.md flags.

One review point I did not change: `-fork` does not hide timeouts or OOMs. Its defaults (ignore_timeouts=1, ignore_ooms=1) mean the campaign keeps going after one, not that it drops it: every hang and OOM gets a `timeout-` or `oom-` reproducer, and the run exits nonzero at the end (70 after a timeout, 71 after an OOM), which fails the job and is what fuzz_report.py reports. Setting both to 0 would only stop the target at the first one and give a single reproducer instead of all of them. Checked on a target with a hang found at runtime, one already in the corpus, and an allocation over -rss_limit_mb.

CI is green on e63bc1875 under the stricter gate: test.sh minimal https://github.com/hashcat/hashcat/actions/runs/35990773225 and the fuzz shards https://github.com/hashcat/hashcat/actions/runs/35990773204.

PR #4780, opened 2026-08-18
Update expected Python version to 3.13.9

albaropereyra22:


jsteube:
Merging, but a quick note. I think using "3.13" would also work, or "3.14". In that case, pyenv selects the latest. If you think that makes sense, feel free to send a new PR.

PR #4782, opened 2026-08-19
bugfix + add tests: m17030 (GPG, SHA-256 KDF) AES-128 path never crack

thatux:
Generate hash, that never cracks on current master https://github.com/hashcat/hashcat/commit/030b44c160870a92fe0dd6507a8c2197c8702c74
```
sudo apt-get install gnupg1

find kernels -name '*17030*' -delete
export GNUPGHOME=$(mktemp -d); chmod 700 "$GNUPGHOME"
cat > "$GNUPGHOME/params" <<'EOF'
%echo generating test key
Key-Type: RSA
Key-Length: 2048
Name-Real: hashcat test
Name-Email: test@example.com
Expire-Date: 0
Passphrase: hashcat123
%commit
EOF

# m17030 = SHA-256 S2K + AES-128 (cipher_algo 7)
gpg1 --homedir "$GNUPGHOME" --batch \
     --s2k-mode 3 --s2k-digest-algo SHA256 --s2k-cipher-algo AES --s2k-count 65536 \
     --gen-key "$GNUPGHOME/params"

# export the protected secret key and extract the hashcat $gpg$ hash
gpg1 --homedir "$GNUPGHOME" --export-secret-keys > "$GNUPGHOME/secret.gpg"
/home/user/john/run/gpg2john "$GNUPGHOME/secret.gpg" 2>/dev/null | cut -d':' -f2-2 > gpg17030.hash
echo "hashcat123" > pw.txt
./hashcat -m 17030 -a 3 --potfile-disable  gpg17030.hash pw.txt
```

The AES-256 self-test hash (`module_st_hash`) cracks fine before and after —
that is exactly what hides the bug.

hash-mode **17030** ("GPG (AES-128/AES-256 (SHA-256($pass)))") accepts two
OpenPGP cipher algorithms — `7` (AES-128) and `9` (AES-256) — and dispatches
to two device kernels: `m17030_aux1` for AES-128, `m17030_aux2` for AES-256
(`src/modules/module_17030.c:126-133`).

The **AES-128 path (`aux1`) derives the AES key incorrectly**, so *every* valid
GPG secret key that uses the SHA-256 S2K KDF with AES-128 protection fails to
crack (silent **false negative**). The AES-256 path is correct, and the mode's
own self-test hash uses AES-256, so the bug is completely masked by the test
suite and by `--self-test`.

## Root cause

In m17030 the running hash state is 32-bit (SHA-256):

```c
// OpenCL/m17030-pure.cl  (gpg_tmp_t)
u32 h[8];
```

`aux2` (AES-256) reads it correctly as eight u32 words → a 32-byte key:

```c
for (int i = 0; i < 8; i++) aes_key[i] = hc_swap32_S (tmps[gid].h[i]);   // correct
```

But `aux1` (AES-128) was **copy-pasted from the SHA-512 sibling m17020**,
where `h[]` is `u64 h[8]`. It still uses the 64-bit half-word extractors
`h32_from_64` / `l32_from_64` on a **u32** array:

```c
// OpenCL/m17030-pure.cl:390-393  (BUG)
aes_key[0] = hc_swap32_S (h32_from_64 (tmps[gid].h[0]));
aes_key[1] = hc_swap32_S (l32_from_64 (tmps[gid].h[0]));
aes_key[2] = hc_swap32_S (h32_from_64 (tmps[gid].h[1]));
aes_key[3] = hc_swap32_S (l32_from_64 (tmps[gid].h[1]));
```

`h32_from_64(x)` is `(u32)((x) >> 32)`. Applied to a 32-bit `h[0]` this shifts
a 32-bit value right by 32 (→ 0 / UB), so the derived AES-128 key is garbage
(`{0, swap(h0), 0, swap(h1)}`) instead of the first 128 bits of the SHA-256
digest. The decrypt therefore never matches the trailing SHA-1 integrity
check.

The correct extraction is the same as the working sibling **m17010** (SHA-1,
also a `u32 h[]`):

```c
for (int i = 0; i < 4; i++) aes_key[i] = hc_swap32_S (tmps[gid].h[i]);
```

The AES-128 path has never worked since the original commit https://github.com/hashcat/hashcat/commit/f7103f511 (2023-03-02).

## Scope
**m17030 only.** m17010 (SHA-1) uses the correct u32 extraction; m17020
  (SHA-512) correctly uses the 64-bit extractors on its `u64 h[]`; m17040 is
  CAST5-only (`cipher_algo == 3`), no AES key split. m17030 is the unique
  combination of a 32-bit hash state with a dual AES-128/AES-256 dispatch, and
  only its AES-128 arm kept the 64-bit code.

## Fix

```c
-  aes_key[0] = hc_swap32_S (h32_from_64 (tmps[gid].h[0]));
-  aes_key[1] = hc_swap32_S (l32_from_64 (tmps[gid].h[0]));
-  aes_key[2] = hc_swap32_S (h32_from_64 (tmps[gid].h[1]));
-  aes_key[3] = hc_swap32_S (l32_from_64 (tmps[gid].h[1]));
+  for (int i = 0; i < 4; i++) aes_key[i] = hc_swap32_S (tmps[gid].h[i]);
```

## Verification

```
$ ./tools/test.sh -m 17010
[ test_1787165178 ] > Init test for hash type 17010.
[ test_1787165178 ] [ Type 17010, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1787165178 ] [ Type 17010, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped

$ ./tools/test.sh -m 17020
[ test_1787165235 ] > Init test for hash type 17020.
[ test_1787165235 ] [ Type 17020, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1787165235 ] [ Type 17020, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped

$ ./tools/test.sh -m 17030
[ test_1787165307 ] > Init test for hash type 17030.
[ test_1787165307 ] [ Type 17030, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 1 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ test_1787165307 ] [ Type 17030, Attack 0, Mode single, Device-Type Gpu, Kernel-Type Optimized, Vector-Width 4 ] > OK : 0/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
```

---

_completely vibe-coded whilst working on #4774_


thatux:
Two additional ideas I/one could implement to improve more structurally
1. Allow for multiple ST_HASHes on AUX kernels
2. Add support for GPG ` -g    Generate crypto-containers on-the-fly (requires sudo)` in  `./tools/test.sh` i.e. automatically generate all kinds of GPG keys, use gpg2john to convert them into hashes. This'd add an optional dependency to john, but would e.g. also benefit pkzip.

What's your thoughts on this?

jsteube:
1. Yes, but maybe from a broader perspective. Requests for multi test hash support have come up a few times already, but never often enough that I felt there was a real demand for it. Normally I would still implement it, but this is a much larger change than it sounds. I would not do it for now. I may come back to it at some point. It also affects existing third party plugins we know nothing about, so we need to be very careful and surgical here.
2. I do not like that it requires root. We cannot afford that. If you can make it work without sudo, then yes, that would be great.

thatux:
1. ok
2. gnupg1 wouldn't require root, it is required for mounting LUKS2, and that's required because hashcat checks for a filesystem. I'll tried looking for something that can mount a LUKS2 container without root.  but that isn't great either as it requires:
```
sudo tee /etc/polkit-1/rules.d/49-nopasswd-udisks.rules >/dev/null <<'EOF'
polkit.addRule(function(action, subject) {
    if (subject.user !== "user") return; // <--- your username here
    var allow = [
        "org.freedesktop.udisks2.loop-setup",
        "org.freedesktop.udisks2.loop-delete",
        "org.freedesktop.udisks2.encrypted-unlock",   // non-system devices (your loop)
        "org.freedesktop.udisks2.encrypted-lock",
        "org.freedesktop.udisks2.filesystem-mount",
        "org.freedesktop.udisks2.filesystem-unmount",
        "org.freedesktop.udisks2.modify-device"        // needed for Format+take-ownership
    ];
    if (allow.indexOf(action.id) !== -1) return polkit.Result.YES;
});
EOF
```
which isn't great, sudo is required to make that file, and it allows the non-sudo user some additional rights systemwide..

```

PASS='hashcat'                                   # your passphrase
truncate -s 10G container.luks

# format: passphrase via stdin (a passphrase, not a keyfile; no newline)
printf '%s' "$PASS" | cryptsetup luksFormat --type luks2 -q container.luks

# attach loop (authorized by the polkit rule)
LOOP=$(udisksctl loop-setup --no-user-interaction --file "$PWD/container.luks" | awk '{print $NF}' | tr -d '.')
LOOPOBJ="/org/freedesktop/UDisks2/block_devices/$(basename "$LOOP")"

# unlock by passing the PASSPHRASE as a D-Bus string (no keyfile, no prompt)
CLEAROBJ=$(gdbus call --system --dest org.freedesktop.UDisks2 \
  --object-path "$LOOPOBJ" \
  --method org.freedesktop.UDisks2.Encrypted.Unlock "$PASS" '{}' \
  | sed -E "s/.*objectpath '([^']+)'.*/\1/")
CLEAR="/dev/$(basename "$CLEAROBJ" | sed 's/_2d/-/g')"   # dm_2d0 -> /dev/dm-0
echo "loop=$LOOP clear=$CLEAR"

# make + own the filesystem (rootless via udisks)
gdbus call --system --dest org.freedesktop.UDisks2 \
  --object-path "$CLEAROBJ" \
  --method org.freedesktop.UDisks2.Block.Format ext4 "{'take-ownership': <true>}"

udisksctl mount --no-user-interaction -b "$CLEAR"
MNT=$(findmnt -n -o TARGET "$CLEAR")
touch "$MNT/decrypted.txt"

# teardown
udisksctl unmount --no-user-interaction -b "$CLEAR"
udisksctl lock     --no-user-interaction -b "$LOOP"
udisksctl loop-delete --no-user-interaction -b "$LOOP"
```


I like adding `user ALL=(root) NOPASSWD: make-luks-container.sh` to `sudoers`
more than that, as that allows us to use the 'regular' commands.

sudo with LUKS2 seems unavoidable, so accept that, and only error on LUKS2 when using `test.sh -g`. GPG and pkzip can probably be done sudoless.

maybe someone else has an idea?

PR #4786, opened 2026-08-21
test.sh -g: real GPG/PKZIP/RAR containers, new test.pms, and two GPG kernel fixes

thatux:
`test.sh -g` generates **real** containers on the fly and cracks them: GPG,
PKZIP and RAR. Two GPG kernel bugs turned up while wiring the GPG side, and
both are fixed here. Four hash-modes that had no test at all now have one.

Only LUKS2 generation needs `sudo`; GPG, PKZIP and RAR need no privileges.

_Real ground truth to complement the self-contained `test.pl` oracles: the
oracle proves hashcat agrees with our own crypto, `-g` proves it agrees with a
real GnuPG / InfoZip / RAR / John toolchain end to end. As discussed in
https://github.com/hashcat/hashcat/pull/4782#issuecomment-5361193631_

## What is in the branch

| | |
|---|---|
| `test.sh -g`: real GPG and PKZIP containers, folding in `tools/test_pkzip.sh` | `tools/test.sh` |
| `test.sh -g`: real RAR3 and RAR5 archives | `tools/test.sh` |
| `test.sh`: `-g` adds a test instead of replacing one, and no skip stays quiet | `tools/test.sh` |
| `tools/rar_tests/` deleted: 23800 needs `-g`, and a run without it says so | `tools/test.sh` |
| **m17050 kernel fix**: crack RSA and DSA/ElGamal keys, not only ECC/ed25519 | `OpenCL/m17050-pure.cl` |
| **GPG S2K kernel fix**: wrong key for some (count, password length) pairs, all five modes | `OpenCL/m170*-pure.cl` |
| m17050: replace the commented-out range check with why it must not be there | `src/modules/module_17050.c` |
| m17010/17020/17030 oracles exercise both aux kernels (AES-128 and AES-256) | `tools/test_modules/m170*.pm` |
| New oracles for four modes that had none: 22001, 19210, 31100, 99999 | `tools/test_modules/` |

No parser or module behaviour changes. Against master, the only non-test source
changes are the two kernel fixes plus comments in `module_17040.c` and
`module_17050.c`.

## `-g` GPG (`gpg_test`)

- **17010 / 17020 / 17030 / 17040** - `gpg1` (GnuPG 1.4) generates the secret
  key with explicit `--s2k-digest-algo` / `--s2k-cipher-algo`, so each mode's
  S2K variant is produced (SHA1 / SHA512 / SHA256 / CAST5). `gpg2` also produces
[...]

PR #4793, opened 2026-08-23
`test.sh -g`: real 7-Zip, WinZip AES, PDF and OpenSSH artifacts

thatux:
_Builds on #4786, and waits for that to be merged. After these are merged I'd like to refactor `test.sh -g` to make it more modular._

One commit on top of #4786.

Four more formats where a real file can be built headless on Linux, in the same shape as the
generators already there: build it with the tool that owns the format, read the hash back with
John's extractor, hand it to `build_container_cmd`. **None of these needs sudo**, which makes
this the easiest of the `-g` sets to run anywhere, and the first one that could plausibly go in
CI.

| Mode | Built with | Variants |
|---|---|---|
| 11600 7-Zip | `7z a` | header-encrypted and plain (hash types 0 and 2), stored, bzip2 |
| 13600 WinZip AES | `7z a -tzip -mem=` | AES128, AES256 |
| 10400 PDF RC4-40 | `qpdf --encrypt ... 40` | V1/R2 |
| 10500 PDF | `qpdf --encrypt ... 128` | RC4-128 (V2/R3) and AES-128 (V4/R4) |
| 10700 PDF AES-256 | `qpdf --encrypt ... 256` | V5/R6 |
| 22931 OpenSSH keys | `ssh-keygen -m PEM` | RSA, DSA |

## Why 11600 is generated four ways

The mode's work does not end at the KDF. After deriving the key it decrypts a block,
decompresses it and checks a CRC, so the codec is part of what is under test and a single LZMA2
archive only exercises one path through it. Header encryption on and off is a second axis: it
changes the hash type, 0 against 2. Stored and bzip2 cover the other two decompressors the
kernel carries.

## Three boundaries, all found by trying rather than assumed

**qpdf writes neither R5 nor V1-2 with R3.** So 10600 (Adobe's R5 extension) and 10510 have no
generator here and keep the oracles written for them. Worth noting the two do not overlap: the
generator lands on 10400, the oracle on 10510.

**22931 is the only reachable mode of the 229xx family.** `ssh-keygen` without `-m PEM` writes
`openssh-key-v1`, whose bcrypt-pbkdf KDF *no released hashcat mode parses at all*: the hash
comes out as `$sshng$6$...` and every 229xx mode rejects it. That is what open PR #4767
(`-m 37500`) adds. And openssl 3 writes PKCS#8 (`BEGIN ENCRYPTED PRIVATE KEY`) rather than a
DEK-Info header, which `ssh2john.py` will not parse either. So the classic PEM form with
AES-128-CBC is what is left, and that is `$sshng$1$`, which is 22931.

**qpdf 11 refuses RC4** unless told the caller means it, hence `--allow-weak-crypto` on the two
RC4 variants.

## Tools

`7z` and `ssh-keygen` are on nearly every machine already. The rest:

```sh
sudo apt install qpdf ghostscript libcompress-raw-lzma-perl
```

ghostscript is there because qpdf encrypts an existing document rather than making one, so
something has to write the plain PDF first; `gs` writes the smallest valid one. The lzma perl
module is what `7z2john.pl` needs to read a 7-Zip header.

`SEVENZIP_BIN`, `QPDF_BIN`, `SSHKEYGEN_BIN` and the four extractor paths (`SEVENZIP2JOHN`,
`ZIP2JOHN`, `PDF2JOHN`, `SSH2JOHN`) override the defaults. A missing tool is a skip in the
end-of-run summary, not a silent pass:

```
[ test_1787485870 ] > 1 test(s) did not run in full:
[ test_1787485870 ] [ Type 11600 ] > Skip : 7z2john.pl not found (set SEVENZIP2JOHN=/path/to/7z2john.pl)
```

Note the `.py` version of pdf2john is not used: it needs `pyhanko`. The `.pl` has no
dependencies beyond perl.

## Checked

On CPU (`-d 1 -D 1 -V 1`), against this branch:

| Command | Result |
|---|---|
| `-m 11600 -g` | 4 archives, all OK, and the `.pm` oracle still 8/8 alongside |
| `-m 13600 -g` | 2 archives, both OK |
| `-m 10400 -g` | 1 document, OK |
| `-m 10500 -g` | 2 documents, both OK |
| `-m 10700 -g` | 1 document, OK |
| `-m 22931 -g` | 2 keys, both OK, oracle 8/8 alongside |
| `-m 11600`, `-m 10400` without `-g` | unchanged |
| `-m 11600 -g` with a bogus `SEVENZIP2JOHN` | skipped and reported once in the summary |

### DIFF
--- a/tools/README.md
+++ b/tools/README.md
@@ -44,6 +44,10 @@ that names it, for that format only, repeated in a summary at the end.
 | GPG | 17010, 17020, 17030, 17040, 17050 | `gpg2`, `gpg1`, `gpg2john` | No `gpg1`: only what `gpg2` writes by default is covered, which drops the classic S2K combinations and the AES-128 (aux1) path. A note, not a skip. No `gpg2john`: skipped. |
 | PKZIP | 17200, 17210, 17220, 17225, 17230 | `zip`, `zip2john` | Skipped. |
 | RAR | 12500, 13000, 23700, 23800 | `rar` 6.x or older, `rar2john` | Only RAR5 (13000) is built, the RAR3 modes are skipped. 23800 has no `test.pl` oracle at all, so without `-g` it is skipped whatever is installed. |
+| 7-Zip | 11600 | `7z`, `7z2john.pl`, `Compress::Raw::Lzma` | Skipped. |
+| WinZip AES | 13600 | `7z`, `zip2john` | Skipped. |
+| PDF | 10400, 10500, 10700 | `qpdf`, `gs`, `pdf2john.pl` | Skipped. `gs` writes the plain PDF that `qpdf` then encrypts. |
+| OpenSSH key | 22931 | `ssh-keygen`, `ssh2john.py` | Skipped. |
 | LUKS1 | 29511 to 29543 | `cryptsetup`, `sudo` | Skipped. |
 | LUKS2 | 34100 | `cryptsetup`, `sudo` | Skipped. |
 | TrueCrypt | 6211 to 6243, 29311 to 29343 | `tcplay`, `expect`, `sudo` | Skipped. |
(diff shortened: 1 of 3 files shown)


jsteube:
I merged 4792 and 4794, so this one needs a rebase onto current master before it can go in. Sorry, but this is structural rather than anything you did. All three branches started from 4786 and ended up touching the same three files, so whichever went second was always going to conflict. I merged 4794 first only because it applied cleanly on top of 4792 and this one did not.

tools/README.md: keep your 7-Zip, WinZip AES, PDF and OpenSSH rows together with master's LUKS1 row.

tools/install_dependencies.sh: merge your p7zip-full, libcompress-raw-lzma-perl, qpdf, ghostscript and openssh-client additions with master's tcplay and expect.

tools/test.sh: keep both GEN_MODES additions, both new functions, both dispatch branches and merge the -g help text. The help text is the only part that needs actual wording work, combining your 7-Zip, WinZip AES, PDF and OpenSSH notes with LUKS1, TrueCrypt and VeraCrypt

thatux:
Done, tnx

works (applied https://github.com/hashcat/hashcat/pull/4797)

```
./tools/install_dependencies.sh 
./tools/install_modules.sh
./tools/test.sh -g -D1

```

PR #4800, opened 2026-08-25
feeds: add feed_xz and feed_lz4 — seekable xz and lz4 wordlists (follow-up to #4801)

trichimtrich:
# feeds: add feed_xz and feed_lz4 — seekable xz and lz4 wordlists

Follow-up to #4801 (feed_zstd). Adds the two sibling feeds that flank zstd on the ratio/speed curve, using the same seekable design, and extends the `-a 0` magic router to `.xz` and `.lz4`.

> **Stacked on #4801.** This branch sits on top of #4801's branch (it reuses the `generic.c` base-wordlist magic router that #4801 introduces), so the diff currently shows #4801's six commits too. **Only the two top commits are new here** — please review those; once #4801 merges, this reduces to just:
> - `d2c24b6` feeds: add feed_xz — seekable xz wordlists; route .xz under -a 0
> - `edcf907` feeds: add feed_lz4 — seekable lz4 wordlists; route .lz4 under -a 0

## Why

feed_zstd made compressed wordlists seekable for `-a 0`/`-a 8`. zstd is the balanced choice, but two common codecs sit on either side of it and are worth offering:

- **xz** — best ratio (≈6.6× vs zstd 5.2× on our data), i.e. the smallest dictionaries;
- **lz4** — fastest decode (≈1150 MB/s vs zstd ≈460–650), i.e. the least host CPU when feeding fast hashes.

Both are already seekable in principle, so the feed is a small variation on feed_zstd.

## Design (shared with feed_zstd)

Each feed indexes the file's blocks/frames as `{compressed_offset, first_line, uncompressed_size, line_off}`, decodes only the block that holds word N, reassembles lines that cross a block boundary, builds the index in parallel on first use and caches it under the seekdb dir (revalidated by size + content hash), and takes a `threads=` setting. The difference is only where block boundaries come from and how one block is decoded:

- **feed_xz** (`liblzma`): an `.xz` stream carries a **native block Index** in its footer, so there is nothing to reframe — the stock `xz --block-size=16MiB -T0 wordlist` produces a multi-block seekable file directly. The feed reads the index (`lzma_stream_footer_decode` + `lzma_index_buffer_decode` + `lzma_index_iter`) and decodes a block with `lzma_block_decoder`. Single-block `.xz` still works (decodes from the start). Multi-stream `.xz` (manually concatenated) is not handled yet. Links the system `liblzma` via `feed_xz.mk`.
- **feed_lz4** (`liblz4`): the lz4 frame format has no native index, but its blocks are **independent by default** (`-BI`). The feed walks the frame once (reading each block's 4-byte size prefix) to record block offsets, and decodes a block with `LZ4_decompress_safe`. A block-dependent frame (`-BD`) is rejected. Seek granularity is the block max size (up to 4 MiB, `-B7`). Links the system `liblz4` via `feed_lz4.mk`.

`generic.c`: the `-a 0` base-wordlist magic check (added in #4801 for `.zst`) is extended to also route `.xz → feed_xz` and `.lz4 → feed_lz4`. Plaintext and amplifiers are untouched.

## Usage

```sh
# xz: the stock tool makes it seekable; no framing helper needed
xz --block-size=16MiB -T0 -k wordlist
hashcat -a 0 -m 0 hash.txt wordlist.xz -r rules/best64.rule

# lz4: block-independent (the default)
lz4 -9 -BI -B7 wordlist wordlist.lz4
hashcat -a 0 -m 0 hash.txt wordlist.lz4 -r rules/best64.rule

# or invoke a feed explicitly
hashcat -a 8 -m 0 hash.txt feeds/feed_xz.so wordlist.xz threads=8
```

## Verified (measured, on GPU, `-m 0`)

Both, on a 5,000,000-line list with the target at line 4,000,000:

- **feed_xz** — 47-block `.xz`: keyspace 5,000,000; every candidate emitted exactly once (sorted-equal to plaintext); `-a 0` auto-route cracks the target; `-s 3999999 -l 1` cracks (exact seek across blocks) while `-s 4000000` misses; single-block `.xz` keyspace correct; `.xfidx` cache written.
- **feed_lz4** — block-independent `.lz4` (~750 × 64 KiB blocks): keyspace 5,000,000; every candidate once; `-a 0` auto-route cracks; `-s 3999999 -l 1` cracks / `-s 4000000` misses; block-dependent `-BD` rejected; `.lfidx` cache written.

All outputs stay standard files (`unxz` / `lz4 -d` reconstruct them); the index is an external side file.

## Follow-ups

- Vendor `liblzma`/`liblz4` under `deps/` for the Windows/macOS plugin cross-builds (same open item as libzstd in #4801).
- feed_xz: handle multi-stream `.xz` (combine per-stream indexes with `lzma_index_cat`).
- The three feeds share a lot of structure (index/cache/straddle/seek); a future refactor could factor a common "framed seekable feed" core with per-codec hooks.

Codec choice, in one line: **zstd** balanced (default), **xz** for max compression, **lz4** for max feed speed.

### DIFF
--- a/src/feeds/feed_lz4.c
+++ b/src/feeds/feed_lz4.c
@@ -1,0 +1,395 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+// A generic (attack-mode 8) feed that reads an lz4-compressed wordlist and seeks inside it, like
+// feed_zstd / feed_xz. The lz4 frame format has no native index, but its blocks are independent by
+// default (`-BI`), so this feed walks the frame once to record each block's file offset, decodes the
+// blocks in parallel to record the line each opens on (cached), and reaches word N by decoding only
+// its block with LZ4_decompress_safe. lz4's strength is decode speed; its ratio is weak, so this is
+// the "fast" complement to the zstd/xz feeds.
+//
+//   lz4 -9 -BI -B7 wordlist wordlist.lz4     # once, with the stock lz4 tool (block-independent)
+//   hashcat -a 8 -m 0 hash.txt feeds/feed_lz4.so wordlist.lz4 -r rules/best64.rule
+//
+// A block-dependent frame (`-BD`) is not seekable and is rejected. Seek granularity is the block max
+// size (up to 4 MiB with -B7).
+
+#define XXH_INLINE_ALL
+#include "xxhash.h"
+
+#include "common.h"
+#include "types.h"
+#include "memory.h"
+#include "shared.h"
+#include "path.h"
+#include "folder.h"
+#include "memchr.h"
+#include "feed.h"
+
+#include <lz4.h>
+#include <stdlib.h>
+#include <pthread.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <sys/stat.h>
+#include <stdarg.h>
+#include <inttypes.h>
+
+const int GENERIC_PLUGIN_VERSION = FEEDS_INTERFACE_VERSION_CURRENT;
+
+const int GENERIC_PLUGIN_OPTIONS = GENERIC_PLUGIN_OPTIONS_AUTOHEX
+                                 | GENERIC_PLUGIN_OPTIONS_ICONV
+                                 | GENERIC_PLUGIN_OPTIONS_RULES;
+
+#define LFIDX_CACHE_MAGIC  0x315844494C465831ULL
+#define LFIDX_CACHE_VER    1ULL
+#define LFIDX_SAMPLE       65536
+#define LZ4_FRAME_MAGIC    0x184D2204u
+
+typedef struct { u64 comp_off, first_line, uncomp_size, line_off; } lf_frame_t;
+
+typedef struct
+{
+  char   *path;
+
+  u64    *comp_off;    // [nblocks + 1]; block k on-disk size = comp_off[k+1] - comp_off[k]
+  u64    *first_line;  // [nblocks]
+  u64    *uncomp_size; // [nblocks] (actual, from the build decode; informational)
+  u64    *line_off;    // [nblocks]
+
+  u64     nblocks;
+  u64     total_lines;
+  u64     filesize;
+
+  size_t  bms;         // block max size (decode buffer capacity)
+  size_t  max_comp;
+  int     b_csum;      // per-block checksum present
+
+  u64     build_threads;
+} lf_global_t;
+
+typedef struct
+{
+  int             fd;
+  unsigned char  *cbuf;
+  unsigned char  *dbuf;
+  size_t          dlen, doff;
+  u64             block;
+  u64             line;
+  unsigned char  *carry;
+  size_t          carry_cap;
+} lf_thread_t;
+
+static void error_set (generic_global_ctx_t *g, const char *fmt, ...)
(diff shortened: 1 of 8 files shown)


jsteube:
Thanks for this. Same general note as on #4801 since the 2 PRs overlap, plus a couple of things specific to xz and lz4.

Master got zstd support in 8371d3f76 shortly after you opened this. Since then hashcat no longer links compression libraries directly. zlib, liblzma and libzstd are loaded at runtime with hc_dynlib_open(), so building hashcat does not require their development packages.

Right now all 3 new feed makefiles link their libraries directly. Please switch zstd and xz to the loaders already on master. zstd can use src/ext_zstd.c and xz can use src/ext_lzma.c. For lz4 we do not have a loader yet, so that needs a new src/ext_lz4.c following the same pattern. ext_lzma.c already handles the xz container, so everything feed_xz needs should be there.

The other thing to settle is generic_base_feed(). It checks the same compression magic that source_is_compressed() in feed_wordlist.c already checks, just one layer earlier. That means zstd and xz files would suddenly take a different path. I think the routing layer should own that decision, and feed_wordlist should only handle compressed formats that do not have their own seekable feed. Gzip is worth keeping in mind here, since right now it would still use feed_wordlist while zstd and xz would not.

There is also some overlap with #4801. feed_zstd.c, feed_zstd.mk and tools/zsf.c are identical in both PRs, so I'd merge the zstd part there and then rebase this one down to xz and lz4. That also leaves the new lz4 loader as the main additional piece here.

The important part is still very much wanted. Master's compressed seek currently means decoding and discarding when moving forward and restarting when moving backwards, with every thread paying that cost. Proper random access into compressed wordlists is still missing, and that is exactly the gap these feeds solve.

jsteube:
My review notes here never got a reply, and I built both halves out in the meantime. The zstd half is on master in 1172ffb76 and the xz half in df26da13c. Neither became a separate feed. The seek database that feed_wordlist already builds for the line count gained a section recording where each independent piece of a compressed file begins and which line it starts at, and nothing in it names a container, so both codecs fill the same table. #4801 has the longer explanation.

The xz side went differently from this branch, and better. Your feed reads the block index with lzma_stream_footer_decode and lzma_index_buffer_decode. Hashcat was already decoding that index for another reason: xz_read_index_size in filehandling.c built the whole lzma_index just to ask it for the uncompressed size hc_fstat answers with, and then threw it away. The index is kept now and walked with lzma_index_iter, which turned out to be most of the work. Decoding a block on its own is lzma_block_header_decode plus lzma_block_decoder, as you had it, reached through ext_lzma.c rather than by linking liblzma. Multi-stream .xz works too, because lzma_file_info_decoder covers every stream in the file. On a 210 MB wordlist with 15 million lines, 200 random seeks took 27 seconds with a single block, 4.4 seconds with the 9 blocks xz -T0 writes, and 1.6 seconds with the 26 blocks xz -T0 --block-size=8MiB writes.

That leaves lz4, which is the part of this PR still open. I am not convinced it earns its place. It needs a loader of its own and a fourth library in the Windows package, for a format hashcat does not otherwise read, and its ratio is poor on wordlist data. The seeking would be the easy half, because lz4 frame blocks are independent by default and can be indexed by walking the 4 byte size prefixes without decompressing anything. If you have a case where decode speed is what binds and the ratio does not matter, show the numbers and I will look at it again. Otherwise I would close this one.

jsteube:
PR superseded, feel free to reopen

Issue #1644, opened 2018-08-01
Improvement of seeking in wordlist when using --skip


s3inlc:
When using larger wordlists (multiple GBs) and chunking the keyspace, Hashcat needs quite a long time to get to the wordlist position for the later chunks before really start with cracking.

For example when having a ~10GB wordlist, there is a difference of ~50s between the first and one of the last chunks (with chunks of 10 minutes and the same --length value). These 50s are taking place when Hashcat wants to start cracking but first has to get to the right position of the file, so for these chunks we nearly have an overhead of 10% which is quite a lot when considering that this startup time happens with every chunk. And this can even get worse when using larger dictionaries e.g. the weakpass one.

It would be great if at some point this could be improved, especially as Hashcat is creating the dictionary stats anyway, maybe this could also include some information to more quickly jump to certain positions in the wordlist.

davidbolvansky:
I am thinking about two new additions which would help here:

1. --seek N option, which would just do fseek( inputdict, N, SEEK_SET );
2. Store the current (last; after cracking finished) position in the dictionary in status(-json) output. (status seems a better place to store this info than dict stats file)

Thoughts? @jsteube @s3inlc

LywLover:
6 years, Do you have any solutions for this issue? When the wordlist is very large, each chunk spends most of its time reading the file instead of cracking. The larger the value of --skip, the longer it takes to read.

PR #4801, opened 2026-08-25
feeds: add feed_zstd — seekable zstd wordlists for -a 0 / -a 8

trichimtrich:
# feeds: add `feed_zstd` — seekable zstd wordlists for `-a 0` / `-a 8`

Adds a generic (attack-mode 8) feed that reads a **zstd-compressed wordlist and seeks inside it**, so a dictionary attack keeps fast `--skip`/restore **and** a compressed dictionary — which the mmap wordlist feed cannot do. A one-line magic check routes a base `.zst` to it automatically under `-a 0`. A small companion tool (`tools/zsf.c`) prepares/reframes files.

## Problem

The attack-mode-8 wordlist feed (`src/feeds/feed_wordlist.c`) opens the dictionary with `hc_fopen_raw` and `mmap`s the **raw** bytes, then builds the byte-offset seek database over them. That is exactly what makes `--skip` O(1), but it also means **no decompression happens** — the `HCFILE` layer that transparently inflates gzip/zip/xz for other readers is bypassed. So a compressed wordlist is fed to the kernels as *compressed bytes*:

```
$ printf 'a\nb\n...(50000 lines)' | zstd > wl.zst
$ hashcat -a 0 -m 0 hash wl.zst --keyspace
147                # newline count of the *compressed* stream, not 50000
```

`.gz`/`.xz`/`.zst` all behave this way (keyspace = newlines in the compressed data, candidates are `$HEX[...]` garbage). hashcat 6 read wordlists through `hc_fopen` and decompressed them transparently, so this is a silent regression for anyone who fed a compressed dictionary. Tracking the underlying tension: a seekable **plaintext** file (the design in #1644 / the `seekdb`) is fundamentally incompatible with a single compressed stream.

## Idea

A zstd stream is a sequence of **independently decodable frames**. Apply the `seekdb` idea to *frame* offsets instead of raw byte offsets:

- index each frame as `{ compressed_offset, first_line, uncompressed_size }`;
- to reach word **N**, binary-search the index for the frame that holds it, `pread` that one frame, `ZSTD_decompress` it, and walk forward to the line.

A frame boundary is the one place the format *guarantees* a clean decoder slate (window reset, first block re-declares its entropy tables, repeat-offsets reset to `{1,4,8}`), so a checkpoint is just an 8-byte offset and decoding a frame is a public one-shot call. Seek cost is **O(one frame)**, independent of N; the index is ~32 bytes/frame; and the file stays a **standard `.zst`** (`zstd -dc` still reconstructs it — no lock-in).

We deliberately did **not** take the alternative of indexing *into* a single frame with captured decoder state (a zran-style window snapshot). See “Why frames, not window snapshots” below.

## What’s in the PR

Six commits:

1. `feeds: add feed_zstd` — the feed: `global_init/keyspace/term`, `thread_init/next/seek/term`; reads a sidecar index, one-frame decode, binary-search seek.
2. `feed_zstd: auto-build and cache the index; handle straddling lines` — build the index on first use and cache it under the seekdb dir (honouring `--seekdb-path`), keyed by a hash of the file, exactly like the plaintext `seekdb`; reassemble a candidate that crosses a frame boundary so any multi-frame `.zst` works.
3. `feed_zstd: parallel first-run index build` — derive the frame layout from headers and decode/scan frames across a bounded thread pool; serial fallback for unusual inputs.
4. `feed_zstd: settings (frame=/level=/threads=) + on-the-fly reframing` — attack-mode-8 `key=value` settings via the `feed_param` API; `frame=` reframes any zstd (even single-frame) into frames on the fly, cached and reused.
5. `generic: auto-route a base .zst wordlist to feed_zstd (-a 0)` — a base wordlist whose magic is zstd is handed to this feed; plaintext and amplifiers are untouched.
6. `tools: add zsf` — standalone framing/reframing tool.

Files: `src/feeds/feed_zstd.c` (+`.mk`, links system libzstd), `src/generic.c` (+30 lines), `tools/zsf.c`.

## How it works

- **Index (`.zfidx`).** `{magic, version, nframes, total_lines, filesize, ident}` + `{comp_off, first_line, uncomp_size, line_off}` per frame. `first_line[k]..first_line[k+1]` is the line-number range of frame k; `line_off` is where the first whole line begins inside it (nonzero only when a line carried in from the previous frame). Built once, cached, revalidated by size + content hash; shared across a cluster via `--seekdb-path` (built once for all agents).
- **Parallel build.** When frame content sizes are in the headers (`ZSTD_getFrameContentSize`, i.e. one-shot-compressed frames — what `zsf` and per-chunk `zstd -c` emit), the layout is known from `ZSTD_findFrameCompressedSize` alone, so frames decode/scan independently across a pool. Falls back to a serial streaming pass on unknown sizes / skippable frames.
- **Straddle handling.** `line_off` + a carry buffer let the feed reassemble a word that spans frames, so frames need not be line-aligned.
- **Settings (`-a 8`):** `threads=<n>` (cores, 0=all), `frame=<MiB>` (reframe on the fly into that frame size; 0 = use the file’s own frames), `level=<n>` (default 19). Settings are only read in `-a 8`; `-a 0` auto-route uses the file’s own frames and all cores.
- **`tools/zsf`:** `build` (plaintext→frames), `reframe` (existing `.zst`→frames, streaming, no plaintext temp), `skip`, `info`. Lets a site prepare one canonical framed file and replace the original in place (`zsf reframe old.zst new.zst && mv new.zst old.zst`) instead of the feed’s per-file cache copy — matters for very large dicts and read-only run-time mounts.

## Technical challenges

- **Lines that cross frame boundaries.** A frame boundary rarely falls on a newline. The index records where the first *whole* line starts in each frame (`line_off`), and `thread_next`/`thread_seek` follow a straddling line across frames into a carry buffer. This is what lets the feed accept *any* multi-frame `.zst`, not just line-aligned output.
- **Why frames, not window snapshots.** Seeking into a *single* frame would need to resume decoding mid-frame, which requires more than the window: from the decoder source, `ZSTD_DCtx` carries, block to block, the FSE/Huffman tables (a block may be `set_repeat` and carry none of its own — `RETURN_ERROR_IF(dctx->litEntropy==0, …)`), the repeat-offset state `entropy.rep[3]`, and per-frame flags. None of that is exposed by any public save/restore API, so it would mean forking libzstd and shipping version-locked, non-portable snapshots whose per-checkpoint window is MBs (128 MB+ with `--long`). A frame boundary resets all of it, so the checkpoint is 8 bytes and decode is the public one-shot call. (Full write-up in the design note linked below.)
- **Parallel build without holding the file.** The header walk (`ZSTD_findFrameCompressedSize` + `ZSTD_getFrameContentSize`) gives the whole frame layout with no decode; the parallel pass then decodes each frame into its own buffer, bounded so `threads × largest_frame` stays under a memory budget; a serial O(nframes) combine turns per-frame newline counts into `first_line[]`.
- **Standard, portable output.** The framed file is plain concatenated zstd frames (`zstd -dc` reconstructs it byte-for-byte); the index is an external side file, disposable and regenerated on mismatch. No custom container.
- **Non-invasive routing.** `-a 0` auto-route triggers only for a **single base** source whose magic is zstd; multi-source runs, amplifiers (`-a 1/6/7`), and plaintext are left on the wordlist feed.

## Results (measured)

Box: 60-core, 2× RTX 3060, CUDA 13.2. Samples: crackstation slices.

**Framing cost (512 MiB slice, single-stream vs framed):**

| layout | ratio | index | note |
| --- | --- | --- | --- |
| zstd `-19 --long`, single frame (not seekable) | 5.20× | — | baseline |
| 16 MiB frames (seekable) | ~5.13× (**+1.4 %**) | ~1 KB | ~25 ms forward-decode/seek |
| 4 MiB frames | 5.09× (**+2.4 %**) | ~3 KB | finer seek |
| 64 MiB frames | 5.16× (**+0.8 %**) | tiny | coarser seek |

**First-run index build + cache (2 GiB slice, 183 M lines, 128 frames):**

| operation | time |
| --- | --- |
| serial build ≈ whole-file 1-thread decode | 2.67 s |
| **parallel build** + keyspace (cache cleared) | **0.27 s** (~10×, incl. startup) |
| cached keyspace (2nd run, reads the 4 KB index) | 0.06 s |

**Seek (2 GiB slice):** seek to word **100,000,000** matches `sed -n` exactly; decodes one 16 MiB frame. **Reframe** of a 2 GiB single-frame `.zst` → 128 frames in **~5 s** (parallel), output byte-identical and `zstd -dc`-readable.

**End-to-end on GPU (`-m 0`), 5 M-line / 47-frame `.zst`, target at line 4,000,000:** keyspace 5,000,000; every candidate emitted exactly once (sorted-equal to plaintext); full run cracks; `-s 3999999 -l 1` cracks (seek lands on the exact word) while `-s 4000000` misses; byte-split (mid-line frames) `--stdout` is byte-identical to the plaintext; `-a 0` auto-route and `frame=` reframe both crack; unknown settings are rejected with the valid list.

## Usage

```sh
# -a 0: a base .zst is auto-routed to feed_zstd (index built + cached on first use)
hashcat -a 0 -m 0 hash.txt wordlist.zst -r rules/best64.rule

# -a 8 with settings: reframe any zstd (even single-frame) into 16 MiB frames on 8 cores
hashcat -a 8 -m 0 hash.txt feeds/feed_zstd.so wordlist.zst frame=16 level=19 threads=8

# prepare a canonical framed file once (replaces the original, single file)
cc -O2 tools/zsf.c -lzstd -lpthread -o zsf
./zsf reframe wordlist.zst wordlist.new.zst && mv wordlist.new.zst wordlist.zst
```

## Build / compatibility

`feed_zstd` links the system `libzstd` (via `src/feeds/feed_zstd.mk`), like the other feeds link their deps. Needs `libzstd` available at build time on the target platform. **Vendoring libzstd under `deps/` for the Windows/macOS cross-builds is a follow-up** (this branch was developed and tested on Linux with the distro `libzstd`).

## Follow-ups / open questions

- Vendor `libzstd` under `deps/` so the Windows/macOS plugin cross-builds don’t need a system copy.
- Should the feed’s on-the-fly `frame=` reframe be gated (size threshold) the way `--wordlist-index-cache` is in #4703, to avoid a surprise recompress on small inputs? Currently it is opt-in per run.
- `-a 0` auto-route currently covers `.zst` only; `.gz`/`.xz` remain on the mmap feed (still affected by the regression). A parallel `feed_gzip`/`feed_xz`, or restoring decompression to the wordlist feed for those, is out of scope here.
- Wire `tools/zsf` into the build (there is currently no C tool built by the Makefile), or leave it standalone as here.

Design rationale (frames vs window snapshots, the zstd-internals argument) and the full benchmark set are in the companion write-up.

### DIFF
--- a/src/feeds/feed_zstd.c
+++ b/src/feeds/feed_zstd.c
@@ -3,22 +3,33 @@
  * License.....: MIT
  */
 
-// A generic (attack-mode 8) feed that reads a SEEKABLE zstd wordlist: a file of independent,
-// line-aligned zstd frames plus a sidecar index (<path>.idx) mapping each frame to its compressed
-// offset and the line number it starts at. Because every frame decodes on its own, reaching word N
-// decompresses exactly one frame rather than the whole file, so --skip stays O(1) and the wordlist
-// stays compressed. This is the seekdb idea (a checkpoint -> position map) applied to frames instead
-// of raw byte offsets.
+// A generic (attack-mode 8) feed that reads a zstd-compressed wordlist and seeks within it without
+// decompressing the whole file. A zstd stream is a sequence of independently decodable frames, so a
+// small index -- frame k starts at compressed offset X and at line L -- turns "reach word N" into
+// "decode the one frame that holds it". That is the seekdb idea (a checkpoint -> position map) moved
+// from raw byte offsets, which need a plaintext file, onto frame offsets, which do not. The stock
+// mmap wordlist feed cannot seek a compressed file at all.
 //
-// The .zsf / .zsf.idx pair is produced offline by the `zsf` tool (hashcat-seekable repo). A plain
-// single-frame .zst also works, but is one frame, i.e. seeking still decodes from the start.
+// The index is built on first use by one streaming pass over the file and then cached, named by a
+// hash of the file the same way the plaintext seek database is, so a second run and every other
+// machine sharing the file (see --seekdb-path) reuse it. A sidecar written by the `zsf` tool
+// (<path>.idx) is used directly when present.
 //
-//   hashcat -a 8 -m 0 hash.txt feeds/feed_zstd.so wordlist.zsf -r rules/best64.rule
+// A file compressed as a single frame still works but is one frame, so a seek decodes from the
+// start: to get fast --skip the wordlist must be compressed in frames (the `zsf` tool, or zstd's
+// seekable format). Lines that cross a frame boundary are handled, so any multi-frame zstd works.
+//
+//   hashcat -a 8 -m 0 hash.txt feeds/feed_zstd.so wordlist.zst -r rules/best64.rule
+
+#define XXH_INLINE_ALL
+#include "xxhash.h"
 
 #include "common.h"
 #include "types.h"
 #include "memory.h"
 #include "shared.h"
+#include "path.h"
+#include "folder.h"
 #include "memchr.h"
 #include "feed.h"
 
@@ -35,32 +46,41 @@ const int GENERIC_PLUGIN_OPTIONS = GENERIC_PLUGIN_OPTIONS_AUTOHEX
                                  | GENERIC_PLUGIN_OPTIONS_ICONV
                                  | GENERIC_PLUGIN_OPTIONS_RULES;
 
-// index sidecar format, identical to prototype/zsf.c
-#define ZSF_IDX_MAGIC 0x315A53464B434821ULL
+#define ZFIDX_CACHE_MAGIC  0x315844495A465A46ULL // cache file this feed writes
+#define ZFIDX_CACHE_VER    1ULL
+#define ZSF_SIDECAR_MAGIC  0x315A53464B434821ULL // sidecar written by prototype/zsf.c
 
+#define ZFIDX_SAMPLE       65536
+#define BUILD_OUTBUF       (4 * 1024 * 1024)
+
+// One frame: where it starts in the compressed file, the line number of the first line that begins
+// inside it, how many bytes it decodes to, and where in those bytes that first line begins (nonzero
+// only when the frame opens in the middle of a line carried over from the frame before).
 typedef struct
 {
   u64 comp_off;
   u64 first_line;
   u64 uncomp_size;
-} zsf_entry_t;
+  u64 line_off;
+} zf_frame_t;
 
 typedef struct
 {
-  char   *path;
+  char       *path;
 
-  u64    *comp_off;    // [nframes + 1], comp_off[nframes] = filesize
-  u64    *first_line;  // [nframes]
-  u64    *uncomp_size; // [nframes]
+  u64        *comp_off;    // [nframes + 1], comp_off[nframes] = filesize
+  u64        *first_line;  // [nframes]
+  u64        *uncomp_size; // [nframes]
+  u64        *line_off;    // [nframes]
 
-  u64     nframes;
-  u64     total_lines;
-  u64     filesize;
+  u64         nframes;
+  u64         total_lines;
+  u64         filesize;
 
-  size_t  max_comp;    // largest compressed frame, sizes the thread read buffer
-  size_t  max_uncomp;  // largest decompressed frame, sizes the thread frame buffer
+  size_t      max_comp;    // largest compressed frame  -> thread read buffer
+  size_t      max_uncomp;  // largest decompressed frame -> thread frame buffer
 
-} zsf_global_t;
+} zf_global_t;
 
 typedef struct
 {
@@ -75,7 +95,10 @@ typedef struct
   u64             frame;  // frame currently in dbuf, or (u64) -1 when none is loaded
   u64             line;   // global line number of the next candidate
 
-} zsf_thread_t;
+  unsigned char  *carry;  // a line that spans frames is assembled here
+  size_t          carry_cap;
+
+} zf_thread_t;
 
 static void error_set (generic_global_ctx_t *global_ctx, const char *fmt, ...)
 {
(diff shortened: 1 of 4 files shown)


jsteube:
Thanks for this, and sorry about the timing. 4 days after you opened the PR, master got zstd support in 8371d3f76, so some of the plumbing changed underneath you. The idea of this PR is still useful, we just need to hook it into what master does now.

The main change is that hashcat no longer links compression libraries directly. zlib, liblzma and libzstd are loaded at runtime with hc_dynlib_open(), and not having one installed is fine. This avoids requiring development packages just to build hashcat.

For zstd, src/ext_zstd.c and include/ext_zstd.h already wrap the 6 functions we use. feed_wordlist also already detects zstd by magic bytes and decompresses it transparently, together with gzip and xz.

Because of that, there are a few things I'd change here.

1. feed_zstd.mk currently adds -lzstd, and feed_zstd.c includes zstd.h directly. That brings back the build dependency we just removed. Please use hc_zstd_lib_t from ext_zstd.c instead. It should mostly just be replacing the direct zstd calls.

2. The zstd magic check now exists twice. generic_base_feed() in src/generic.c detects it to select a feed, while feed_wordlist.c detects the same magic to select its decompression path. #4800 adds the same check again for xz and lz4 on top of that. Since the routing check happens first, zstd files would stop using the path master already has. I'd prefer the routing layer to own this and let feed_wordlist handle only compressed formats that do not have their own seekable feed.

3. Gzip is worth thinking about too. Right now a .zst file would get a dedicated feed while a .gz file still goes through feed_wordlist, even though that difference is invisible to the user. We should probably make a deliberate decision there instead of ending up with it by accident.

The important part is that none of this changes what your PR is actually solving. Master's compressed seek is still pretty basic. Going forward means decoding and discarding data, and going backwards means restarting from the beginning. Every thread does that independently. Proper random access into a zstd stream is still missing, and tools/zsf.c for creating seekable files fits nicely with that. That's the part I definitely want to keep.

About #4800, feed_zstd.c, feed_zstd.mk and tools/zsf.c are identical in both PRs, so we can only merge them once. I'd put the zstd feed here, since it does not require another loader, then rebase #4800 so it only contains feed_xz and feed_lz4. feed_xz can use the existing ext_lzma.c. For lz4 we'd need a new ext_lz4.c following the same pattern as the other compression loaders.

jsteube:
Compressed wordlists could only be seeked by decoding from the beginning, once per device, and this PR went after exactly the right gap. My review notes here never got a reply, and I built it out in the meantime, so it landed on master in 1172ffb76 and this PR is superseded rather than merged.

The shape came out different because master already routes -a 0 base words through feed_wordlist. Putting the frame index there and in the seek database covers -a 0, -a 1, -a 6, -a 7 and -a 8 at once, with no routing change in generic.c and no new plugin. The concern raised in review, that generic_base_feed checks the same magic source_is_compressed already checks, disappears with it. Both are now one function, hc_path_is_compressed, exported from the file layer.

The seek database already exists and already costs one full pass over the file, so the frame index went into it rather than into a .zfidx beside the wordlist. pzstd already writes multi-frame .zst as well, so no equivalent of zsf was needed. Its frame size follows the compression level rather than the thread count, so a wordlist of any size ends up with frames every few megabytes, and concatenating .zst files works too.

One thing turned up that your branch would have hit as well. Files written by pzstd open with a skippable frame, magic 0x184D2A50 through 0x184D2A5F, rather than 0x28B52FFD, and hashcat matched only the second. The keyspace of a pzstd wordlist was therefore the line count of its compressed bytes, and its candidates were those bytes. I fixed that in the same commit, and it applies to hash lists and rule files as well.

On a 210 MB wordlist with 15 million lines, 200 random seeks took 11 seconds with a single frame and 0.9 seconds with the 26 frames pzstd writes. The xz half of #4800 has landed as well, in df26da13c.

jsteube:
PR superseded, feel free to reopen

PR #4803, opened 2026-08-25
Add AI coding instructions: CONTRIBUTING.md

thatux:
# Add CONTRIBUTING.md, and point README.md and the agent files at it

`README.md` carried the whole contributing policy inline, and nothing in the
tree told an AI coding agent anything at all. This moves the first into a
`CONTRIBUTING.md` and adds the second to it.

## Where an agent picks this up

`AGENTS.md`, `CLAUDE.md` and `.github/copilot-instructions.md` are symlinks to
`CONTRIBUTING.md`, mode `120000`, so an agent that looks for any of the three by
convention finds part 2 without the contributor having to point it there.

Three and no more, deliberately. `AGENTS.md` is the convention the ecosystem is
converging on, `CLAUDE.md` and `.github/copilot-instructions.md` cover the two
agents most likely to touch this repository. A directory of `.cursor/rules`,
`.windsurfrules`, `.clinerules`, `GEMINI.md` and the rest would be vendor sprawl
in a C codebase for very little more coverage.

The one caveat is Windows. Without developer mode or `core.symlinks=true`, git
checks a symlink out as a small text file containing the target path, so a
Windows contributor gets a one line `AGENTS.md` reading `CONTRIBUTING.md`. That
is harmless and still points at the right place, but if you would rather not
rely on it, say so and I will make them stub files instead.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+../CONTRIBUTING.md
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,1 @@
+CONTRIBUTING.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+CONTRIBUTING.md
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,245 @@
+# Contributing to hashcat
+
+This file has two parts, and they are addressed to different readers.
+
+**Part 1 is for human contributors.** It is the code style and pull request
+policy that used to live in `README.md`, plus the conventions that keep coming
+up in review.
+
+**Part 2 is for AI coding agents.** It starts at
+[Part 2](#part-2-instructions-for-ai-coding-agents), and it is additional and
+stricter. If you are an agent, all of part 1 binds you too. Part 2 in short,
+with the reasoning for each in the section itself:
+
+1. Do not sign commits as an AI. Commit as the human you are working for.
+2. If you add a hash mode, add its `tools/test_modules/mXXXXX.pm` in the same
+   pull request, and run it.
+3. Never present output you did not actually run and read back. If you could
+   not run something, say which thing and why.
+4. Verify instead of asserting. Run both sides of the change and read the exit
+   code before you call anything fixed.
+5. Stay inside the change you were asked for. One problem per pull request.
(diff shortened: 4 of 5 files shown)


thatux:
Additionally I can add extra CI checks to  .github/workflows/

A small workflow fits the existing shape. Three checks, all deterministic and cheap, all matching rules the file already states:

1. added lines containing non-ASCII — literally the git diff -U0 | grep -nP '[^\x00-\x7F]' the file tells you to run
2. any commit on the PR carrying a Co-Authored-By: trailer naming a model
3. a new src/modules/module_NNNNN.c arriving without tools/test_modules/mNNNNN.pm

This inverts the problem. Instead of hoping the agent read the instructions, the agent gets a red X with a message naming the rule — and agents are very good at iterating against failing CI without being asked.

jsteube:
Thanks for this. I am closing the pull request, but not because the direction was wrong or the input was unwelcome. The README had been doing too many jobs for a while, and the tree had no guidance for AI coding agents at all. Your patch put both problems in front of me at the same time, and that is what finally pushed me to fix them.

The reason I am not merging this version is that a contribution policy speaks for the project. It defines what we expect from everyone sending patches from now on, so the wording has to be something the maintainers write and own ourselves. I wrote the policy from scratch and it is now on master in 3e6130fe7. The structure you proposed is what shipped: CONTRIBUTING.md contains the policy, AGENTS.md contains the additional guidance for coding agents, and README.md is a project front page again. The commit message credits this pull request for starting that work.

I took a few different decisions along the way. AGENTS.md is a normal file and the only agent entry point, with no CLAUDE.md, no .github/copilot-instructions.md and no symlink. AI attribution is also allowed rather than discouraged. Whether to credit a model is up to the contributor, but if they do, we ask that the Co-Authored-By trailer appears once per pull request rather than on every commit.

The rewrite also picked up a few things outside your original diff, including outdated hash mode counts, dead Travis and Coverity references, missing BSD and Rust workflow badges, and remaining non ASCII punctuation in BUILD.md and several other files.

One part of your follow up about CI does interest me. Checking for a new src/modules/module_NNNNN.c without a matching tools/test_modules/mNNNNN.pm would be useful. There are already 95 modes on master without a test module, so it would have to warn only for newly added modules and must not gate the existing tree. If you want to open a separate pull request for that, please do.

Thanks again for pushing this forward. The final policy is different in wording, but the split itself came directly from this pull request.

PR #4804, opened 2026-08-27
OpenCL: work around Mesa rusticl compiler crashes on gfx1013

Naster17:
## Problem

Two constructs in shared OpenCL kernel code (`inc_common.cl`, `inc_rp.cl`) cause the AMDGPU backend of Mesa rusticl (LLVM 21.1.2) to crash with SIGSEGV during kernel compilation.  The compiler process dies inside `libLLVM`, preventing **all** OpenCL kernels from building on affected hardware.

Tested on an **AMD BC-250** (Cyan Skillfish APU, gfx1013, Alpine Linux 3.23, Mesa 25.2.7, LLVM 21.1.2).  The crash is deterministic and affects every kernel build.

## Root cause

The LLVM AMDGPU backend crashes when it encounters three or more variable-index `u8` loads from private (stack) memory through a casted pointer in the same function.  Both hit sites follow the same pattern:

```c
u8 *ptr = (u8 *) u32_private_array;
u8 val = ptr[variable_index];  // ≥3 such loads → SIGSEGV in LLVM
```

### Hit site 1 — `hc_enc_next()` in `OpenCL/inc_common.cl`

The UTF-8 decode switch reads bytes one at a time via `src_ptr[src_pos++]`.  Case 3 has four such loads; case 2 has three.  Together they exceed the threshold and crash the compiler.

### Hit site 2 — `exchange_byte()` in `OpenCL/inc_rp.cl`

Swaps two bytes via `u8 *ptr = (u8 *) buf; ... ptr[off_src]`.  This function is inlined by ~80 rule-mangle functions, accumulating enough u8 loads across the call graph to trigger the same crash during `apply_rule` compilation.

## Fix

Both fixes avoid the crashing pattern by reading through `u32` word loads instead of `u8` pointer casts.  The change is semantically identical on all little-endian targets and measures the same after optimization on other OpenCL platforms.

**`inc_common.cl`** — new `hc_enc_src_byte()` helper:
```c
DECLSPEC u32 hc_enc_src_byte (PRIVATE_AS const u32 *src_buf, const int pos)
{
  return (src_buf[pos >> 2] >> ((pos & 3) << 3)) & 0xff;
}
```
All `src_ptr[x]` calls in `hc_enc_next()` replaced with `hc_enc_src_byte(src_buf, x)`.  `hc_enc_next_global()` is untouched (doesn't crash).

**`inc_rp.cl`** — `exchange_byte()` rewritten using word-based access:
```c
const int sd  = off_src / 4;
const int sm  = off_src & 3;
const int dd  = off_dst / 4;
const int dm  = off_dst & 3;
// ... extract, swap, XOR back via u32 operations
```
This was already present in the file as a commented-out variant.

## Verification

| Platform | Self-test | WPA1 | WPA2 | WPA3 | PMKID | Rules |
|---|---|---|---|---|---|---|
| BC-250 (rusticl/LLVM 21, gfx1013) | pass | crack | crack | crack | crack | crack |
| AMD dev machine (AMDGPU driver, Radeon 760M) | pass | — | — | — | — | pass |

No change in benchmark numbers on either platform.

## Bugs fixed

- Fixes Rusticl crash on BC-250 / Cyan Skillfish (gfx1013)
- Related: https://gitlab.freedesktop.org/mesa/mesa/-/issues/12269 (same LLVM backend crash pattern)

## Files changed

- `OpenCL/inc_common.cl` — `hc_enc_src_byte()` helper + 12 call sites in `hc_enc_next()`
- `OpenCL/inc_rp.cl` — `exchange_byte()` rewritten to use u32 word access

### DIFF
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -2127,9 +2127,20 @@ DECLSPEC int hc_enc_validate_utf8_global (GLOBAL_AS const u32 *src_buf, const in
 // Input buffer and Output buffer size has to be multiple of 4 and at least of size 4.
 // The output buffer is not zero padded, so entire buffer has to be set all zero before entering this function or truncated afterwards.
 
+// Read one byte of src_buf through a u32 load instead of a u8 load. Three or more
+// u8 loads from private (stack) memory through a casted pointer with a variable
+// index crash the AMDGPU back end of Mesa rusticl (LLVM 21) while compiling
+// gpu_utf8_to_utf16 - the process dies with SIGSEGV inside libLLVM. Extracting
+// the byte from a word load is semantically identical on little-endian targets
+// and measures the same after optimization on other OpenCL platforms.
+
+DECLSPEC u32 hc_enc_src_byte (PRIVATE_AS const u32 *src_buf, const int pos)
+{
+  return (src_buf[pos >> 2] >> ((pos & 3) << 3)) & 0xff;
+}
+
 DECLSPEC int hc_enc_next (PRIVATE_AS hc_enc_t *hc_enc, PRIVATE_AS const u32 *src_buf, const int src_len, const int src_sz, PRIVATE_AS u32 *dst_buf, const int dst_sz)
 {
-  PRIVATE_AS const u8 *src_ptr = (PRIVATE_AS const u8 *) src_buf;
   PRIVATE_AS       u8 *dst_ptr = (PRIVATE_AS       u8 *) dst_buf;
 
   int src_pos = hc_enc->pos;
@@ -2142,7 +2153,7 @@ DECLSPEC int hc_enc_next (PRIVATE_AS hc_enc_t *hc_enc, PRIVATE_AS const u32 *src
 
   while ((src_pos < src_len) && (dst_pos < dst_sz))
   {
-    const u8 c = src_ptr[src_pos];
+    const u8 c = hc_enc_src_byte (src_buf, src_pos);
 
     int extraBytesToRead = -1;
 
@@ -2213,25 +2224,25 @@ DECLSPEC int hc_enc_next (PRIVATE_AS hc_enc_t *hc_enc, PRIVATE_AS const u32 *src
         break;
       */
       case 3:
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++];
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++;
         ch -= offsetsFromUTF8_3;
         break;
       case 2:
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++];
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++;
         ch -= offsetsFromUTF8_2;
         break;
       case 1:
-        ch += src_ptr[src_pos++]; ch <<= 6;
-        ch += src_ptr[src_pos++];
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++; ch <<= 6;
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++;
         ch -= offsetsFromUTF8_1;
         break;
       case 0:
-        ch += src_ptr[src_pos++];
+        ch += hc_enc_src_byte (src_buf, src_pos); src_pos++;
         ch -= offsetsFromUTF8_0;
         break;
     }
--- a/OpenCL/inc_common.cl
+++ b/OpenCL/inc_common.cl
@@ -2197,115 +2197,51 @@ DECLSPEC int hc_enc_has_next (PRIVATE_AS hc_enc_t *hc_enc, const int sz)
   return 0;
 }
 
-DECLSPEC int hc_enc_validate_utf8 (PRIVATE_AS const u32 *src_buf, const int src_pos, const int extraBytesToRead)
+DECLSPEC u32 hc_enc_seq_byte (const u32 w0, const u32 w1, const int p, const int j)
 {
-  PRIVATE_AS const u8 *src_ptr = (PRIVATE_AS const u8 *) src_buf;
+  const int k = p + j;
 
-  if (extraBytesToRead == 0)
-  {
-    const u8 c0 = src_ptr[src_pos + 0]; if (c0 >= 0x80) return 0;
-  }
-  else if (extraBytesToRead == 1)
-  {
-    const u8 c0 = src_ptr[src_pos + 0]; if ((c0 < 0xc2) || (c0 > 0xdf)) return 0;
-    const u8 c1 = src_ptr[src_pos + 1]; if ((c1 < 0x80) || (c1 > 0xbf)) return 0;
-  }
-  else if (extraBytesToRead == 2)
-  {
-    const u8 c0 = src_ptr[src_pos + 0];
+  const u32 w = (k < 4) ? w0 : w1;
 
-    if ((c0 >= 0xe0) && (c0 <= 0xe0))
-    {
-      const u8 c1 = src_ptr[src_pos + 1]; if ((c1 < 0xa0) || (c1 > 0xbf)) return 0;
-      const u8 c2 = src_ptr[src_pos + 2]; if ((c2 < 0x80) || (c2 > 0xbf)) return 0;
-    }
-    else if ((c0 >= 0xe1) && (c0 <= 0xec))
-    {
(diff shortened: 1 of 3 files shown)


jsteube:
Thanks for tracking this down. I do not want to merge the patch in its current form, mainly for two reasons. First, the performance cost is measurable even on hardware that cannot hit the reported Mesa issue. On the UTF8 path I measured regressions from 2.3 to 9.0 percent across RTX A400, RTX 4090, RX 7900 XTX and Radeon Pro W7800, using both CUDA and HIP. The test was NTLM with the pure kernel under -a 3, using valid 2 byte UTF8 so hc_enc_next() runs for every candidate, with an ASCII mask of the same shape as the control.

The bigger concern is that the aliasing pattern blamed by the commit message is still present after the patch. hc_enc_next() keeps the write side alias, hc_enc_validate_utf8() creates the same PRIVATE_AS u8 * alias and is called on the same buffer in the same loop iteration, and the global variants repeat the same pattern. amp_a0.cl also reaches the same kind of aliasing through apply_rule(). If this is what LLVM 21 fails on, the patched build still contains it. If it is not, then we are paying the performance cost without addressing the actual cause.

I think the better direction is to remove the pattern completely. A UTF8 sequence is at most 4 bytes and spans at most 2 consecutive u32 words, so those words can be loaded once into a u32 window and passed to the validator by value. That removes the u8 alias from all 4 functions and should also reduce repeated reads, since the current code reads the same sequence bytes during lead classification, validation and decoding. For output I would accumulate complete words before storing them. I would avoid a masked insert that reads dst_buf back, since that version measured about 45 percent slower than master.

There are two details to keep exactly as they are today. The second word load must be conditional because src_buf[idx + 1] can otherwise read word 64 of a u32[64], and the lead byte classification and src_sz checks should stay unchanged so this remains a refactor. I could not reproduce the original crash locally, even with AMD_DEBUG=usellvm and LLVM 21.1.8 on gfx1100, so whichever version we land should be confirmed on your gfx1013 system before we call it fixed.

Naster17:
Thanks for the detailed review - I reworked the patch along the lines you suggested.

**What changed**

- The u8 alias is gone from all four functions. Each sequence is now decoded from a window of two consecutive u32 words (w0/w1) that is loaded once per sequence. The same window is passed to the validator by value, so hc_enc_validate_utf8() and hc_enc_validate_utf8_global() collapsed into a single address-space-agnostic hc_enc_validate_utf8(w0, w1, src_pos, extraBytesToRead). Bytes are extracted from the window registers, not re-read from the buffer, so the lead classification, validation and decoding no longer touch memory three times for the same bytes.
- Output is accumulated into full u32 words before storing: BMP writes go to a pending accumulator when the position is word-aligned and are flushed as one store when the word completes; surrogate pairs store one or two full words. There is no masked insert and no read of dst_buf anywhere.
- The second word load is conditional on the sequence actually crossing the word boundary (p + extraBytesToRead > 3), so src_buf[idx + 1] never reads word 64 of a u32[64]. The lead byte classification and the src_sz check are untouched.
- exchange_byte() in inc_rp.cl keeps the word-based body from the previous version (apply_rule() path).

**Verification**

- Byte-for-byte equivalence: a differential harness ran 652,000 cases comparing the old implementation against the new one (every 1-4 byte sequence shape at every byte offset across a u32[64] buffer, invalid leads, truncated sequences, all dst_sz values 4..256 including the surrogate carry path, plus 300k random fuzz cases, streaming across multiple hc_enc_next() calls like the real callers do). Zero mismatches, and the source buffer was placed against a guard page so any unconditional second word load would fault. Clean under UBSan.
- On the gfx1013 system where the original crash was found (Mesa rusticl, LLVM 21.1.2): building the shared module with upstream master dies with SIGSEGV inside libLLVM, reproducing the original report. With this patch the same build compiles in 0.5 s and hashcat runs end to end (self-test passes).
- Functional check on the same device: NTLM pure kernel cracks 7/7 passwords covering ASCII, 2-byte, 3-byte and 4-byte (surrogate pair) UTF-8 input.
- Performance on that device, NTLM pure -a 3, mask of 8 valid 2-byte UTF-8 characters so hc_enc_next() runs for every candidate, ASCII mask of the same shape as control (two rounds each):

  | | UTF-8 mask | ASCII mask |
  |---|---|---|
  | previous patch | 2816 / 2820 MH/s | 9343 / 9380 MH/s |
  | this rework | 3357 / 3474 MH/s | 8843 / 9390 MH/s |

  On the UTF-8 path the rework is about 21% faster than the previous version of the patch, and the ASCII control shows no regression. For reference, upstream master cannot run on this device at all, since it segfaults while compiling the shared module.

The branch is rebased onto current master (it had also picked up the lead-classification range fix from 5c840aaec that the old base predated).

jsteube:
Looks good so far. Most of the hardware I have benefit from it. There is one file I would leave out, which I come to at the end.

I tested 9 device and backend combinations against master with your commit cherry picked on top, so this change is the only difference between the two sides. Every figure is the median of 3 runs per side with the order alternated, 25 seconds each, NTLM with the pure kernel. The UTF8 case is a mask of 8 valid 2 byte sequences, so hc_enc_scan fails and hc_enc_next runs for every candidate:

```
./hashcat -m 1000 -a 3 --hex-charset -1 c3 -2 808182838485868788898a8b8c8d8e8f909192939495969798999a9b9c9d9e9fa0a1a2a3a4a5a6a7a8a9aaabacadaeafb0b1b2b3b4b5b6b7b8b9babbbcbdbebf 00000000000000000000000000000001 ?1?2?1?2?1?2?1?2?1?2?1?2?1?2?1?2 -w 3 --runtime 25
```

The ASCII control is the same command with -1 41 and -2 set to 40 through 7f, same shape and keyspace, on the fast path. The 3 byte case uses lead bytes e1 through ec, the only one where a sequence crosses the u32 window boundary and the second word load fires. The 4 byte case uses f1 through f3, so every character is a surrogate pair.

UTF8 path

```
    device                        2 byte    3 byte    4 byte    ascii control
    Arc A770      OpenCL          +45.8%    +26.5%     +1.8%      +0.0%
    Radeon Pro W7800  HIP         +33.6%    +39.0%    +34.8%      -0.8%
    RTX 4090      CUDA            +28.8%    +16.7%    +17.6%      -1.2%
    RX 7900 XTX   HIP             +23.7%    +25.7%    +22.7%      -3.6%
    Radeon Pro W5700X OpenCL      +20.3%    +16.6%    +27.1%      +0.1%
    RX 9070 XT    HIP              -1.3%     +2.2%     +0.2%     -11.9%
    RTX A400      CUDA             -9.1%     -7.4%     -8.4%      -1.3%
    Radeon Pro W5700X Metal       -35.7%    -35.7%    -34.1%     -32.0%
    Core i5-9600K OpenCL           +0.1%     +0.3%     +0.2%      +0.0%
```

A -m 0 noise floor stayed inside 0.3 percent on all 9, so the movement above is real. The i5 is the exception: every mask case there lands between 541 and 569 MH/s whatever the input, so the host is the bottleneck and the row says nothing about the kernel. Its rules figure below is unaffected, because -a 0 does far more work per candidate.

Mode 1000 -a 3 pure on gfx1201, from a clean kernel cache, agrees with the fast rows:

```
master     162 scratch ops   336 B private   81 vgpr   0 spills
this PR     68 scratch ops   288 B private   82 vgpr   0 spills
```

58 percent less scratch traffic and 48 bytes less private memory, for 1 extra vgpr and 3 percent more ALU. That is the u8 alias being gone.

The RX 9070 XT loses 12 percent on the ASCII control, which never executes the changed code, and that ISA is from the same card. It is autotune: Thr 384 for the old kernel, Thr 768 for the new one, and the second choice is worse. That is a hashcat problem your patch exposes, not a defect in it.

Metal is the bad one, and the same card answers it: the W5700X loses about a third on all four mask cases under Metal and gains 20 percent under OpenCL, same machine. It is not inlining, since HC_NOINLINE_ALWAYS on hc_enc_next and hc_enc_next_global changes nothing and neither does HASHCAT_FORCE_NO_INLINE=1. That leaves Apple's Metal compiler, and I will not hold the patch for it.

The RTX A400 I cannot explain: it is the only device that loses on the UTF8 path while its ASCII control stays flat, and the RTX 4090 gains 29 percent on the same toolchain.

Please take inc_rp.cl back out

The exchange_byte change is a separate change from the crash this pull request fixes, it is not needed for that fix, and on most of this hardware it is expensive.

```
    device                        master        this PR      delta
    RTX 4090                  735.0 MH/s     918.4 MH/s     +24.9%
    RTX A400                  104.1 MH/s     126.6 MH/s     +21.6%
    Radeon Pro W5700X Metal    42.4 MH/s      41.4 MH/s      -2.4%
    Radeon Pro W5700X OpenCL  194.1 MH/s     184.7 MH/s      -4.8%
    RX 7900 XTX               454.1 MH/s     345.2 MH/s     -24.0%
    Arc A770                   82.1 MH/s      59.9 MH/s     -27.1%
    RX 9070 XT                430.6 MH/s     284.2 MH/s     -34.0%
    Radeon Pro W7800          380.0 MH/s     207.8 MH/s     -45.3%
    Core i5-9600K              30.2 MH/s      16.0 MH/s     -47.2%
```

That is -a 0 with rules, so it is every hash mode and one of the most common ways hashcat gets used. The workload is 20000 words of length 16 against 512 rules of 10 operations each, drawn from the two rotates and reverse, so exchange_byte dominates:

```
    python3 -c "
    import random
    random.seed(11)
    ops=['{','}','{','}','r']
    print('\n'.join(' '.join(random.choice(ops) for _ in range(10)) for _ in range(512)))
    " > rot.rule
```

inc_rp.cl alone on master reproduces the full patch on 3 cards, -23.7 against -24.0 on the RX 7900 XTX and -36.0 against -34.0 on the RX 9070 XT, so the two halves are independent. On the RTX 4090 master has two performance states here, and 8 runs per side of the isolated change put it at +36.8 rather than +24.9.

The word based version also does not remove the access it was meant to avoid. It still indexes a private array through a variable subscript, so the scratch access stays and the new code only adds arithmetic around it: on gfx1201, 1.5 percent more instructions and slightly more scratch, not less.

If it is worth having on NVIDIA it wants its own pull request and probably a vendor gate. Plus 37 percent on one vendor against minus 45 on another should not ride along inside an unrelated fix.

PR #4809, opened 2026-08-29
Add static asserts to the SHA256 crypto functions

### LINKED ISSUE #4773: `sha256_hmac_init`: if `len < 64`, potentially uninitialised data is read
I could not find any Github issues about this, or any mention of this in the FAQ.

**Describe the bug**
When writing a custom kernel that uses `sha256_hmac_init`, it's very easy to get wrong results. This happens because even though `sha256_hmac_init` takes the length of the passed buffer as argument (`len`). The implementation of `sha256_hmac_init` assumes that this `len` is at least 64 bytes long.

**To Reproduce**
1. Write a custom kernel that at some point contains the snippet:

```c
u32 buf[8] = { 0 };
// omitted: fill buf with values

sha256_hmac_ctx_t ctx;
sha256_hmac_init(&ctx, buf, sizeof(buf));
```

2. Compile your kernel, and run it with a test hash.
3. Observe that the test hash doesn't crack.

**Expected behavior**
The test hash cracks.

**Hashcat version:**
 - OS: Linux
 - Distribution: Ubuntu 24.04
 - Version: 5d5990ce38311260724aa914955c532e1f5188e9

**Additional context**
Looking at the implementation of `sha256_hmac_init`:

https://github.com/hashcat/hashcat/blob/5d5990ce38311260724aa914955c532e1f5188e9/OpenCL/inc_hash_sha256.cl#L1089-L1144

Notice that `w` is accessed as if it's a 16-element `u32` array if `len <= 64`. In the example above, where the passed buffer was only 8 elements long, this caused `sha256_hmac_init` to read uninitialised memory, which leads to unpredictable results, and bugs that are very hard to track down.

I see a few options on how to resolve this, all of them seem to have their downsides. I'll order them based on my personal preference (from best to worst), as someone who occasionally builds a hashcat kernel, but not often enough to remember this footgun.

1. Add a compile-time assert that `len == 64` in the `else` case. I'm not sure if the way kernels are compiled currently allows for compiler-time asserts. I couldn't find any in the code. However, this would be a nice option, since this would fail at compile-time, and it should show the kernel developer exactly what's wrong. Another upside to this option, is that it would not change the code that the compiler produces for `sha256_hmac_init`, so it has no performance impact.
2. Add a separate branch for `len < 64`, where the passed buffer is copied only up to `len` and the rest of the buffer is padded with null bytes. This will always work, but it might be slower. It seems like just giving 64-bytes to `sha256_hmac_init` is always faster. In this option, it will work, but it will work silently, not letting the kernel developer know there's room for improvement. The `len == 64` case might be constant-folded by the compiler, so it seems plausible that there is no performance impact if the passed length is 64.
3. Add documentation to `sha256_hmac_init` to never pass in a buffer that is shorter than 64 bytes. This is probably the easiest patch, but a comment in the header file might be easily overlooked by kernel developers.

I wonder what the hashcat maintainers think about this. Is this an issue worth fixing? If so, would any of the methods I mentioned be a good idea?

LukeSerne:
Fixes #4773

This catches some common error cases, where a kernel does not use correctly sized buffers. Passing buffers that are too short will lead to unexpected behaviour that's hard to debug. While the docs mention that the buffers should be large enough, adding an explicit check at compile time will save kernel developers some time debugging this issue.

This check is not perfect. It also allows buffers that happen to be exactly the same size as a pointer, to avoid errors when these functions are passed pointers instead of buffers.

If I modify the `HC_SHA256_BLOCK_SIZE` to `7` (just to make the assert fail) and try to crack a hash using sha256, the kernel fails to compile with a clear error message:

```
~/git/hashcat » ./hashcat -m 1400 0000000000000000000000000000000000000000000000000000000000000000 --runtime=1 --potfile-disable
hashcat (v7.1.2) starting

nvmlDeviceGetFanSpeed(): Not Supported

CUDA API (CUDA 12.6)
====================
* Device #01: NVIDIA RTX 2000 Ada Generation Laptop GPU, 7725/7838 MB, 24MCU

OpenCL API (OpenCL 3.0 CUDA 12.6.65) - Platform #1 [NVIDIA Corporation]
* Device #02: NVIDIA RTX 2000 Ada Generation Laptop GPU, skipped

Bitmaps: 13 bits, 8192 entries, 0x00001fff mask, 32768 bytes, 5/13 rotates
Rules: 1

Optimizers applied:
* Zero-Byte
* Early-Skip
* Not-Salted
* Not-Iterated
* Single-Hash
* Single-Salt
* Raw-Hash

Pure kernels can crack longer passwords, but drastically reduce performance.
If you want to switch to optimized kernels, append -O to your commandline.

main_kernel(50): error: static assertion failed with "Size of buffer is not a multiple of SHA256 block size (64)!"
      sha256_update_swap (&ctx, tmp.i, tmp.pw_len);
      ^

main_kernel(106): error: static assertion failed with "Size of buffer is not a multiple of SHA256 block size (64)!"
      sha256_update_swap (&ctx, tmp.i, tmp.pw_len);
      ^

2 errors detected in the compilation of "main_kernel".

* Device #1: Kernel ~/git/hashcat/OpenCL/m01400_a0-pure.cl build failed.

* Device #1: Kernel ~/git/hashcat/OpenCL/m01400_a0-pure.cl build failed.

Started: Sat Aug 29 23:08:07 2026
Stopped: Sat Aug 29 23:08:09 2026
```

Leaving the code as it is in this PR, the static assert never triggers.

There is definitely room for improvement (maybe the assert message can be changed to reflect the computed `sizeof(buffer)`, maybe this check can be extended to other parts of the crypto library, etc.), but I decided against including that in this PR because either my OpenCL is not good enough, or I wanted to avoid this PR becoming too big to review.

I'd be interested to hear what the hashcat maintainers think about this.

### DIFF
--- a/OpenCL/inc_hash_sha256.cl
+++ b/OpenCL/inc_hash_sha256.cl
@@ -345,7 +345,7 @@ DECLSPEC void sha256_update_64 (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS u32 *w0
   }
 }
 
-DECLSPEC void sha256_update (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
+DECLSPEC void sha256_update_impl (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
 {
   u32 w0[4];
   u32 w1[4];
@@ -397,7 +397,7 @@ DECLSPEC void sha256_update (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS const u32 
   sha256_update_64 (ctx, w0, w1, w2, w3, len - pos1);
 }
 
-DECLSPEC void sha256_update_swap (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
+DECLSPEC void sha256_update_swap_impl (PRIVATE_AS sha256_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
 {
   u32 w0[4];
   u32 w1[4];
@@ -1086,7 +1086,7 @@ DECLSPEC void sha256_hmac_init_64 (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS
   sha256_update_64 (&ctx->opad, b0, b1, b2, b3, 64);
 }
 
-DECLSPEC void sha256_hmac_init (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
+DECLSPEC void sha256_hmac_init_impl (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
 {
   u32 w0[4];
   u32 w1[4];
@@ -1143,7 +1143,7 @@ DECLSPEC void sha256_hmac_init (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS co
   sha256_hmac_init_64 (ctx, w0, w1, w2, w3);
 }
 
-DECLSPEC void sha256_hmac_init_swap (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
+DECLSPEC void sha256_hmac_init_swap_impl (PRIVATE_AS sha256_hmac_ctx_t *ctx, PRIVATE_AS const u32 *w, const int len)
 {
   u32 w0[4];
   u32 w1[4];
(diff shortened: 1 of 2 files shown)


jsteube:
Thanks for this, and for opening #4773 before it. I am closing the pull request, but not because the idea was rejected. It led directly to the fix that landed, and I want to explain how we got from your patch to that result.

What did not convince me about the approach:

The plugin guide described two requirements for these buffers. The buffer had to be a multiple of the block size, and every byte after the stated data length had to be zero. Your assertion checks the first requirement, but a static assertion cannot check the second because zero padding is a property of the data at runtime, not of the buffer type.

That distinction matters. A buffer that is too small may cause an out of bounds read, which a sanitizer or crash can expose. A buffer with uninitialized data after the stated length can silently mix those bytes into the hash block and return the wrong digest. An assertion covering only the size could therefore give a developer false confidence while leaving the more difficult failure untouched. For example, u32 buf[16]; can satisfy the assertion, contain only five initialized bytes, and still produce a wrong result.

The patch also covered only six SHA256 entry points. The UTF16 and vector variants were still exposed, and SHA256 is only one of 21 hash families using the same pattern. HC_STATIC_ASSERT may also compile to nothing on the OpenCL paths where **STDC_VERSION** is defined, while the example was demonstrated through NVRTC, which compiles as C++. The macro itself ended in while (0); with the semicolon inside it, which would break an else following an unbraced use. Nothing in the tree currently uses it that way, but hashcat commonly omits braces around single statements.

None of those points is the main reason I am closing it.

What the discussion turned into:

Both requirements in the guide came from the same behavior. sha256_update read all 16 words of the final block regardless of the length passed by the caller. That caused both problems. A short buffer could be read past its end, and any nonzero data after the valid bytes could be included in the hash.

An that's what causes the confusion for new developers, because it's not intuitive. A normal host API such as SHA256_Update reads exactly the number of bytes it is given. Its callers do not need to allocate a complete block or clear bytes outside the stated input. There was no good reason for the kernel API to require anything different. The reason why it was there is performance, and the new code base solved that through cmov.

What landed:

Commit 5243999689 changes every affected helper to read only the words covered by len, using (len + 3) / 4, and masks the final partial word. A word outside the stated length is never read. This includes the UTF16 variants and the HMAC key readers, which previously loaded a complete 64 or 128 byte block regardless of the key length.

The plugin guide no longer requires a buffer whose size is a multiple of the block size, and it no longer requires zero padding after the data. A caller now only has to provide the words covered by the stated length. Your original example from #4773, u32 buf[8] passed to sha256_hmac_init (&ctx, buf, sizeof (buf)), is valid as written.

No kernel in the tree needed to change, and no measurable performance cost showed up. Full benchmarks using the same binary with only the OpenCL headers exchanged stayed within normal variation across CUDA, NVIDIA OpenCL, HIP, AMD OpenCL, rusticl and Intel OpenCL.

So the underlying trap is gone rather than guarded, which leaves nothing for the assertion to check. That is why I am closing this pull request. Thank you for pushing on it. The same commit also fixes #4773.

PR #4810, opened 2026-08-31
Crack non-ASCII passwords with -a 3, and make test.sh find those bugs

thatux:
Quite some kernels don't crack non-ASCII password. Optimized kernels ship with a message that they don't (e.g. -m1000 `Module 1000 advice notice: When using the optimized-kernel: some byte encoding is simplified that would make some non-ASCII passwords never be cracked, even if they are under the optimized byte limit. The pure-kernel implements true UTF16 handling so doesn't have this problem and should crack all passwords.`). However, currently also the pure kernel didn't work.

I'm not 100% sure how to fix this, but tried. This continues 5c840aaec, "Crack non ASCII passwords again in 41 hash modes, and repair 9 more" (Jens Steube, 2026-08-28). That commit widened the lead byte ranges in
`hc_enc_next()`, which repaired the scalar UTF-8 to UTF-16 path. `-a 3` does not use it: 7
modes take a widening branch above vector width 1, and 16 more have no scalar branch at all,
so 22 modes still cannot crack a euro sign, Japanese kana or a CJK character. That is fixed
here.

The reason it could go unnoticed is that `tools/test.pl` builds every password out of the
digits `0` to `9`, so a non-ASCII password has never reached a kernel and the whole path is
untested. This branch generates them, and fixes the test oracles that turn out to have been
wrong about them.

## Reproduction

Measured against pristine upstream/master 1a7af0976. Run `make` on each side: the kernel
cache is keyed on the binary's compile time, so a rebuild invalidates it on its own and no
`rm -rf kernels/` is needed. `Vec:` in the status output says which binary you are on, since
the fix pins these modes to vector width 1, so `Vec:4` under `--backend-vector-width 4` means
the fix is not in the binary you are running.

```
make
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 '7c23a6931f4a44e3f031202f065b5c4e' 'hashcat€'
./hashcat -m 130 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455' '1234567?d€'
```

```
before: m1000 Status Exhausted, Recovered 0/1, rc=1   at any vector width above 1,
                                                     Cracked rc=0 at width 1
        m0130 Status Exhausted, Recovered 0/1, rc=1   at every vector width, 1 included
after:  both  Status Cracked,   Recovered 1/1, rc=0
```

After the fix the first command reports `Vec:1` although it asked for 4. That is the pin, not
a bug.

`7c23a6931f4a44e3f031202f065b5c4e` is NTLM of `hashcat` followed by U+20AC;
`bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455` is `sha1(utf16le('12345678' + U+20AC) .
'2455')`.

An ASCII password is unaffected either way, which is what says the fault is in the UTF-8
handling and not in the mask:

```
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 'b4b9b02e6f09a9bd760f388b67351e2b' 'hashca?l'
./hashcat -m 130 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'a6d8e5ca22a3f603dad986f5ada47728a6d91734:245755154711915074179' '?d?d?d?d?d'

before and after: Cracked, rc=0
```

And the suite cannot see any of it. Revert only the `hc_enc_next()` hunk of 5c840aaec, the
commit this one continues, and every test still passes:

```
git show 5c840aaec -- OpenCL/inc_common.cl | git apply -R -
rm -rf kernels/
./tools/test.sh -m 1000 -a all -t all -D 1 -P -f -V 1
```

```
before this branch: 16 OK, 0 Error, 1 Skip   the suite does not notice
after  this branch:  3 OK, 13 Error, 1 Skip  7/8 not found under -a 0 alone
```

The one Skip is `Attack 4, Mode single`, which master skips either way.

## What was broken

Two faults, both `-a 3` only. Every other attack mode was already safe: `-a 0`, `-a 1`,
`-a 4`, `-a 6`, `-a 7`, `-a 8` and `-a 9` use the scalar conversions, and `-a 12` is pinned to
vector width 1 by the association-mode block in `src/backend.c`.

### There is no vector form of hc_enc_next()

Of the 121 UTF-16 conversion helpers in `OpenCL/inc_hash_*.cl`, 47 decode the UTF-8 through
`hc_enc_next()` and 74 widen the bytes instead. Widening turns a euro sign into three
characters, so it never matches. `hc_enc_next()` works on one candidate at a time and the
UTF-16 length of a candidate is not known before the conversion, so there is no vector form of
it and there cannot be one while a `*_ctx_vector_t` carries a single scalar `len` for all its
lanes.

`m00030`, `m00040`, `m00070`, `m01000`, `m01100`, `m05500` and `m31300` pick the helper with
an `#if VECT_SIZE == 1`, so they are right at width 1 and widen above it.

`src/backend.c` now pins UTF-16 modes to vector width 1, next to the two cases that already do
the same, one of them for exactly this reason:

> // We can't have SIMD in kernels where we have an unknown final password length

`hashconfig->opts_type` cannot be read for this. `src/interface.c:442` clears
`OPTS_TYPE_PT_UTF16LE` and `OPTS_TYPE_PT_UTF16BE` whenever the optimized kernel is not used,
and the pure kernels are exactly the ones that convert in the kernel, so the module is asked
directly.

The pin is on the pure kernels only, which is the same condition as the block above it. An
optimized kernel widens the bytes at any vector width, so it cannot crack a multi byte
password either way and there is nothing to buy by giving up its SIMD.

CUDA and HIP already ran the pure kernels at width 1, because the block above pins every pure
kernel on a GPU to 1, so nothing changes there. The cost lands on CPU and on non GPU OpenCL.
Measured on PoCL, `-m 1000 -a 3`, a 20 second run of `?d` x 10:

```
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --runtime 20 'ffffffffffffffffffffffffffffffff' '?d?d?d?d?d?d?d?d?d?d'

master, pure kernel:  467.0 MH/s   Vec:8
branch, pure kernel:   81.9 MH/s   Vec:1     5.7x slower
branch, -O:           350.3 MH/s   Vec:8     unchanged
```

That is the price of the blanket pin, and it is worth saying plainly: someone running a
UTF-16 mode on CPU without `-O` gets a fifth of the throughput for a property they may not
need on that particular run.

**A narrower pin is possible and I have not built it.** In `-a 3` the mask is known on the
host before the kernel runs. If no position in it can emit a byte above 0x7f, every `?d`,
`?l`, `?u`, `?s` and every ASCII literal, then no candidate can be multi byte and the vector
path is provably safe for that run. Pinning only when the mask actually carries a high byte
would keep full speed for the common case and drop to width 1 only where the wide path would
otherwise be wrong. It needs the mask, which `mpsp.c` parses, to be reachable from
`backend_session_begin()`, and `--increment` and hcmask files each want thinking about, so it
is a design call rather than a mechanical change. Happy to build it instead of the blanket
pin if that is preferred.

### Sixteen kernels had no scalar branch at all

`m00130`, `m00140`, `m00170`, `m01430`, `m01440`, `m01470`, `m01730`, `m01740`, `m01770`,
`m10830`, `m10840`, `m10870`, `m13500`, `m13800`, `m29000` and `m29200` called a widening
helper unconditionally, so they were broken at width 1 as well. They get a scalar branch.

These are `OPTS_TYPE_PT_GENERATE_BE` modes and `markov_be.cl` hands their brute force
candidates over in big endian word order. `hc_enc_next()` reads the UTF-8 byte by byte and
cannot make sense of that, so the scalar branch puts the words back in native order first.
Without that step the digest comes out wrong for an ASCII password too, which the self-test
catches immediately.

### Nothing told the user

`docs/*.md` and `docs/*.txt` say nothing about which modes handle a non-ASCII password. The
only statement in the tree is `module_advice_notice()` in `src/modules/module_01000.c:58`,
which is 1 module out of 593 and prints only for `-m 1000`:

> "When using the optimized-kernel: some byte encoding is simplified that would make some
> non-ASCII passwords never be cracked, even if they are under the optimized byte limit. The
> pure-kernel implements true UTF16 handling so doesn't have this problem and should crack all
> passwords."

The last sentence was never true for `-a 3`, before or after 5c840aaec. It is now.

Two modes are still outside it and are left that way on purpose. **m15500** (JKS Java Key
Store) and **m35200** (AS/400 SSHA1) are the genuine UTF-16BE modes: their `a0-pure` kernel
uses `sha1_update_utf16be_swap()`, and no UTF-16BE helper anywhere calls `hc_enc_next()`, so
hashcat has no UTF-8 path for them in any attack mode. Giving them one is a separate change.
This branch keeps their oracles and their generated passwords on ASCII rather than reporting
them red.

### 41 test oracles hashed the password as Latin-1

Ground truth against a real archive:

```
7z a -mhe=on -p'pass€1' /tmp/t.7z /etc/hostname
perl ~/john/run/7z2john.pl /tmp/t.7z | grep -oE '\$7z\$[^:]*' > /tmp/t.hash
./hashcat -m 11600 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 "$(cat /tmp/t.hash)" 'pass€?d'

Status Cracked, Recovered 1/1, rc=0
```

hashcat and 7-Zip agree, so the kernel is right. `tools/test_modules/m11600.pm` called
`encode("UTF-16LE", $word)` on the raw bytes, which widens each byte as if it were Latin-1.
39 modules had the same bug and now decode first. The modules whose kernel widens too are
left alone, because there the oracle was right.

m05500 and m05600 took the NT hash from `Authen::Passphrase::NTHash`, which is wrong for a
character outside the BMP:

```
perl -MAuthen::Passphrase::NTHash -MEncode -MDigest::MD4=md4_hex -e \
  '$c = decode("utf-8", "12345678\xf0\x9f\x98\x80");
   print Authen::Passphrase::NTHash->new(passphrase => $c)->hash_hex, "\n",
         md4_hex(encode("UTF-16LE", $c)), "\n"'

36a5dbdc75876d5364ef604c79180078
1a160fb7177bbdf8e8edbe681acd5380
```

hashcat agrees with the second, so both modules compute it directly now.

### Why the suite never saw any of this

5c840aaec repaired `hc_enc_next()` for 41 hash modes, and the suite did not notice it going
away again, as the revert at the top of this description shows. Same build, same kernels,
three passwords hashcat is supposed to crack:

```
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 '7c23a6931f4a44e3f031202f065b5c4e' 'hashcat€'
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'a321b58ec81252d199011ef6ed4f09d3' 'かたかな123'
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'a6ea5bd6eb61f1c5e735a08f621b3b9a' '中文密码'
```

```
5c840aaec reverted: Status Exhausted, Recovered 0/1, rc=1   (all three)
5c840aaec present:  Status Cracked,   Recovered 1/1, rc=0   (all three)
```

Euro U+20AC is lead byte `0xe2`, kana `0xe3`, CJK `0xe4`; none was in the old
`{0xe0, 0xec, 0xed, 0xef}` whitelist.

### tools/test.pl

`random_non_ascii_string()` joins `random_numeric_string()` and substitutes multi byte
characters into the digits it returns. The substitution is in place, so the byte length does
not change and the password still fits the mode's `Pwd.Len.Max`. Eleven characters, covering
the lead bytes `hc_enc_next()` has to get right, including one 4 byte character: euro,
hiragana, katakana, four CJK, hangul, devanagari, fullwidth latin, and U+1F600. It comes back
as plain digits for a mode that cannot take one, so a caller never has to ask which.

Only the two callers that generate a password use it. Salts, site keys and challenge
characters keep calling `random_numeric_string()`, which is untouched, so a test module added
later cannot pick up multi byte salts by accident and `tools/test_modules/README.md`, which
points modules at `random_numeric_string` for exactly that job, stays correct as written.

Which modes get them is read out of the kernels, because there is no flag for it.
`OPTS_TYPE_PT_ALWAYS_ASCII` is about hexifying output, not about what a mode can crack, and
only modules 3000 and 25400 set it. So `inc_hash_*.cl` is split into the helpers that call
`hc_enc_next()` and the ones that widen, and a mode gets multi byte passwords when the pure
kernel that will run uses a decoding one. Modes that case fold the plaintext, spell it in hex
or base58, or build it in `module_get_random_password()` stay on ASCII, as do `-O` runs of
modes whose optimized kernel widens.

Injection starts at the ninth byte, so the multi hash tests, whose passwords are 2 to 8 bytes
and which share one mask per length, stay on digits.

`NO_NON_ASCII=1` turns it off.

### tools/test.sh, so the attacks can spell such a password

- `export LC_ALL=C`, so `${#pass}` and `cut -c` count bytes the way hashcat does.
- `mask_literalize()`, because no `?d` produces byte `0xe2`. A mask position over a non-digit
  byte becomes that byte, which leaves the keyspace unchanged. Used by `-a 3`, `-a 6`, `-a 7`
  and the four `-a 12` shapes.
- `utf8_split_point()`, which moves a dict/mask split back to a character boundary. The
  combinator kernels convert the word and the amplifier separately
  (`OpenCL/m01000_a1-pure.cl:65-66`, one call per buffer), so each half has to be valid UTF-8
  on its own:

    ```
    printf '12345\xe7\n' > /tmp/d1 ; printf '12345\n' > /tmp/d2
    H=ac25eecbd4dd9f58b674d3bebe463f9e     # NTLM of 12345<U+7801>678
    ./hashcat -m 1000 -a 6 -D 1 --force --potfile-disable --backend-vector-width 1 "$H" /tmp/d2 "$(printf '\xe7\xa0\x81?d?d?d')"
    ./hashcat -m 1000 -a 6 -D 1 --force --potfile-disable --backend-vector-width 1 "$H" /tmp/d1 "$(printf '\xa0\x81?d?d?d')"

    split on a character boundary: Status Cracked,   rc=0
    split inside the character:    Status Exhausted, rc=1
    ```

    Both spell the same 11 bytes. `-m 0` cracks either. This one is by design, not a bug.

## Testing

PoCL CPU only, `-D 1 --force`, so `-a all -t all` on every mode is not practical here.

All 22 repaired modes, `./tools/test.sh -m <mode> -a 3 -t single -D 1 -P -f -V 4`, every one
`OK 0/7`, against `Error 6/7` or `Error 7/7` before, depending on how many of the eight
generated passwords happened to carry a multi byte character:

```
30 40 70 1000 1100 5500 31300
130 140 170 1430 1440 1470 1730 1740 1770 10830 10840 10870 13500 13800 29000
```

m29200 got the same kernel change but the suite generates no `-a 3` case for it.

Green with multi byte passwords at `-a all -t all`, which on master 1a7af0976 now also covers
attack types 4, 8 and 9: m1000 at `-V 1 -P`, every combination OK except `Attack 4, Mode
single`, which pristine master skips the same way. m130 the same at `-V 4 -P`. Green again
with `NO_NON_ASCII=1`, so ASCII did not regress.

Green with multi byte passwords at `-V 4 -P -a 0`: all 41 modules whose oracle changed, re-run
after the kernel and backend change rather than before it.

Green with multi byte passwords at `-V 1 -P`: m5500 and m5600 at `-a all -t all`; m0, m10,
m400, m11600, m22000 and a 24 mode spread: 20, 100, 110, 300, 500, 900, 1400, 1410, 1700, 1710,
2611, 2711, 2811, 3710, 4400, 4500, 4700, 6000, 6100, 6900, 11500, 17300, 21100, 24900.

Optimized kernels, `-a all -t single -V 4`: m0, m130 and m1000. The width pin does not reach a
mode that does no UTF-16 conversion:

```
./hashcat -m 0 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 '5f4dcc3b5aa765d61d8327deb882cf99' 'passwor?d'
Speed.#01: ... Vec:4
```

6 of 8 single hash passwords carry multi byte characters; every multi hash password is ASCII,
as designed.

## Not covered

- **No GPU.** PoCL CPU only, so the 17 changed kernels have never been compiled by CUDA, HIP
  or AMD OpenCL. This is the gap worth closing before merging.
- `-a all -t all` was run for m130 and m1000 rather than for all 22 repaired modes.
- The narrower, mask aware version of the vector width pin described above is not built.
- **m24800 produces 0 test vectors.** Pre-existing, identical with `NO_NON_ASCII=1`, untouched
  here.

## Commits

1. `test modules: decode the UTF-8 password before encoding it as UTF-16` (41 files, no
   observable change while the passwords are digits)
2. `test.sh: generate passwords that are not 7 bit ASCII`
3. `Crack non ASCII passwords under -a 3 in 22 hash modes` (17 kernels, `src/backend.c`)

### DIFF
--- a/OpenCL/m00130_a3-pure.cl
+++ b/OpenCL/m00130_a3-pure.cl
@@ -61,6 +61,31 @@ KERNEL_FQ KERNEL_FA void m00130_mxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_update (&ctx, s, salt_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -70,6 +95,8 @@ KERNEL_FQ KERNEL_FA void m00130_mxx (KERN_ATTR_VECTOR ())
     sha1_update_vector (&ctx, s, salt_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
@@ -139,6 +166,31 @@ KERNEL_FQ KERNEL_FA void m00130_sxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_update (&ctx, s, salt_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -148,6 +200,8 @@ KERNEL_FQ KERNEL_FA void m00130_sxx (KERN_ATTR_VECTOR ())
     sha1_update_vector (&ctx, s, salt_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
--- a/OpenCL/m00140_a3-pure.cl
+++ b/OpenCL/m00140_a3-pure.cl
@@ -58,6 +58,27 @@ KERNEL_FQ KERNEL_FA void m00140_mxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx = ctx0;
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector_from_scalar (&ctx, &ctx0);
@@ -65,6 +86,8 @@ KERNEL_FQ KERNEL_FA void m00140_mxx (KERN_ATTR_VECTOR ())
     sha1_update_vector_utf16beN (&ctx, w, pw_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
@@ -131,6 +154,27 @@ KERNEL_FQ KERNEL_FA void m00140_sxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx = ctx0;
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector_from_scalar (&ctx, &ctx0);
@@ -138,6 +182,8 @@ KERNEL_FQ KERNEL_FA void m00140_sxx (KERN_ATTR_VECTOR ())
     sha1_update_vector_utf16beN (&ctx, w, pw_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
--- a/OpenCL/m00170_a3-pure.cl
+++ b/OpenCL/m00170_a3-pure.cl
@@ -52,6 +52,29 @@ KERNEL_FQ KERNEL_FA void m00170_mxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -59,6 +82,8 @@ KERNEL_FQ KERNEL_FA void m00170_mxx (KERN_ATTR_VECTOR ())
     sha1_update_vector_utf16beN (&ctx, w, pw_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
@@ -119,6 +144,29 @@ KERNEL_FQ KERNEL_FA void m00170_sxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -126,6 +174,8 @@ KERNEL_FQ KERNEL_FA void m00170_sxx (KERN_ATTR_VECTOR ())
     sha1_update_vector_utf16beN (&ctx, w, pw_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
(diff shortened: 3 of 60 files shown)


thatux:
m11600 is broken. I'll fix

jsteube:
Thanks for digging into this. I think the diagnosis is stronger than the proposed fix, and there are 2 separate problems mixed together here.

What the mask engine does today using the same binary and attack mode, these 3 commands produce very different results:

```
$ ./hashcat --stdout -a 3 'X€'
X€                                          <- E2 82 AC, correct

$ ./hashcat --stdout -a 3 --hex-charset -1 e2 -2 82 -3 ac '?1?2?3'
€                                           <- E2 82 AC, correct

$ ./hashcat --stdout -a 3 -1 'a€' '?1'
a
<0xe2>
<0x82>
<0xac>                                      <- 4 candidates, no Euro anywhere
```

Literal multibyte characters in the mask itself work and always have. Multibyte characters inside a custom charset do not. mp_add_cs_buf() in src/mpsp.c treats every input byte as a separate charset element:

```
u32 u = in_buf[i] & 0xff;
```

So -1 'a€' does not add 2 choices, a and €. It adds 4 choices: a, 0xE2, 0x82 and 0xAC. The mask then selects one byte at each position. There is no warning, and the generated candidates are not what the command appears to request.

That is the first problem. The current behavior is inconsistent, and the inconsistency is visible only when someone checks the output with --stdout.

Your patch addresses what happens after multibyte input reaches a kernel that converts UTF-8 to UTF-16. That is a valid case. For example, -a 3 'X€?d?d' against a UTF-16 mode sends literal multibyte input through the conversion path, and a defect there should be fixed.

It does not address the case most people will associate with a title such as "Crack non-ASCII passwords with -a 3". Someone will try:

```
-m 0 -a 3 -1 '?l€' '?1?1?1'
```

That still cannot produce the requested candidates because the charset was split into bytes before any kernel sees it. This limitation exists above every hash mode and every kernel, and the branch does not change it.

There is another path that may not reach this code at all. For fast UTF-16 hash modes under -a 3, mp_css_utf16le_expand() performs the conversion on the host by expanding every mask position and inserting a zero byte. hc_enc_next() is not involved there. The plugin guide has always described that path as ASCII only.

Which modes did you confirm as failing? If any are fast UTF-16 modes under -a 3, then the change is being made in the wrong layer. The description also says 22 modes in one place and then 7 plus 16 in another, so I am not sure which total is intended.

How I would split this

1. The test coverage should land separately

tools/test.pl currently builds passwords only from the digits 0 through 9, which means this entire area has gone untested. That is a more important finding than the exact number of affected modes. Please split the tests into their own pull request. I would like those in the tree before changing the mask engine, because they are what the eventual fix needs to be measured against.

2. The kernel conversion fix can also stand on its own

It needs a narrower title and description that name the UTF-8 to UTF-16 conversion path directly. That would describe exactly what it repairs without suggesting that custom charsets now support arbitrary Unicode characters. A narrow claim that remains true is better than a broad one that covers only half of the path.

3. I would like to handle the charset side myself

Making a charset contain complete characters instead of bytes means a single position no longer contributes a fixed number of bytes, so candidates generated from one mask can have different lengths. The current -a 3 design depends heavily on that not happening. Kernels are compiled for fixed length classes, autotune assumes a fixed shape, and much of the performance model follows from it.

That makes native UTF-8 charsets a change to the center of the attack mode rather than a small parser fix. I have wanted proper UTF-8 masks for a long time, but I have not yet found a design that preserves the speed of the current engine. It needs to be solved as one complete design, which is why I do not want a partial implementation to land first.

thatux:
Cool, tnx, I'll split this into 1. and 2. and leave 3. to you, right?


> Which modes did you confirm as failing?

I focused on pure kernels, not optimized as I know -m1000 -O has the 'advice notice' I started the PR with. To me it makes the most sense to have all impacted -O kernels have that notice. I guess that are typically UTF-16 kernels.

PR #4812, opened 2026-08-31
Crack non ASCII passwords under -a 3 in 22 hash modes

thatux:
`-a 3` cannot crack a password containing a euro sign, Japanese kana or a CJK character in 22
modes. `-a 0`, `-a 1`, `-a 4`, `-a 6`, `-a 7`, `-a 8` and `-a 9` already use the scalar
conversions and were never affected, and `-a 12` is pinned to vector width 1 by the
association-mode block at [backend.c:15651](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/src/backend.c#L15651).

This .2 of the split discussed in https://github.com/hashcat/hashcat/pull/4810#issuecomment-5483004665. It is the kernel conversion path only. **It does not touch the mask processor or custom charsets**, so the `-1 'a€'` behaviour you described, where a multi byte charset is split into single bytes, is untouched and still yours. The test coverage is part 1, in a separate PR, and this branch is what that coverage catches.

This continues 5c840aaec, "Crack non ASCII passwords again in 41 hash modes, and repair 9
more". That commit widened the lead byte ranges in `hc_enc_next()`, which repaired the scalar
path. `-a 3` does not use it.

## Reproduction

Against master [`1a7af0976`](https://github.com/hashcat/hashcat/commit/1a7af0976474eef81d1bb0bc19030d2651383a95). Run `make` on each side: the kernel cache is keyed on the binary's compile
time, so a rebuild invalidates it on its own. `Vec:` in the status output says which binary you
are on, since the fix pins these modes to vector width 1.

```
make
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 '7c23a6931f4a44e3f031202f065b5c4e' 'hashcat€'
./hashcat -m 130  -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455' '1234567?d€'
```

```
before: m1000 Status Exhausted, Recovered 0/1, rc=1   at any vector width above 1,
                                                     Cracked rc=0 at width 1
        m130  Status Exhausted, Recovered 0/1, rc=1   at every vector width, 1 included
after:  both  Status Cracked,   Recovered 1/1, rc=0
```

`7c23a6931f4a44e3f031202f065b5c4e` is NTLM of `hashcat` followed by U+20AC.
`bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455` is `sha1(utf16le('12345678' + U+20AC) .
'2455')`.

An ASCII password is unaffected either way, which is what says the fault is in the UTF-8
handling and not in the mask:

```
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 'b4b9b02e6f09a9bd760f388b67351e2b' 'hashca?l'
./hashcat -m 130  -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 'a6d8e5ca22a3f603dad986f5ada47728a6d91734:245755154711915074179' '?d?d?d?d?d'

before and after: Cracked, rc=0
```

After the fix the first command reports `Vec:1` although it asked for 4. That is the pin, not a
bug.

### There is no vector form of hc_enc_next()

Of the 121 UTF-16 conversion helpers in `OpenCL/inc_hash_*.cl`, 47 decode the UTF-8 through
`hc_enc_next()` and 74 widen the bytes instead. Widening turns a euro sign into three
characters, so it never matches. `hc_enc_next()` works on one candidate at a time and the
UTF-16 length of a candidate is not known before the conversion, so there is no vector form of
it and there cannot be one while a `*_ctx_vector_t` carries a single scalar `len` for all its
lanes.

`m00030`, `m00040`, `m00070`, `m01000`, `m01100`, `m05500` and `m31300` pick the helper with an
`#if VECT_SIZE == 1`, [m01000_a3-pure.cl:55](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/OpenCL/m01000_a3-pure.cl#L55) for one, so they
are right at width 1 and widen above it.

This branch pins UTF-16 modes to vector width 1, immediately before
[backend.c:15659](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/src/backend.c#L15659) where the width is assigned, and next to the two
cases that already do the same on master: [backend.c:15637](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/src/backend.c#L15637) for an
unknown final password length, and [backend.c:15651](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/src/backend.c#L15651) for association
mode. The first of those says it for this reason too:

> // We can't have SIMD in kernels where we have an unknown final password length

The pin is on the pure kernels only, which is the same condition as the block above it. An
optimized kernel widens the bytes at any vector width, so it cannot crack a multi byte password
either way and there is nothing to buy by giving up its SIMD.

`hashconfig->opts_type` cannot answer this. [interface.c:442](https://github.com/hashcat/hashcat/blob/1a7af0976474eef81d1bb0bc19030d2651383a95/src/interface.c#L442) clears
`OPTS_TYPE_PT_UTF16LE` and `OPTS_TYPE_PT_UTF16BE` whenever the optimized kernel is not used,
and the pure kernels are exactly the ones that convert in the kernel, so the module is asked
directly.

### Sixteen kernels had no scalar branch at all

`m00130`, `m00140`, `m00170`, `m01430`, `m01440`, `m01470`, `m01730`, `m01740`, `m01770`,
`m10830`, `m10840`, `m10870`, `m13500`, `m13800`, `m29000` and `m29200` called a widening
helper unconditionally, so they were broken at width 1 as well. They get a scalar branch.

These are `OPTS_TYPE_PT_GENERATE_BE` modes and `markov_be.cl` hands their brute force
candidates over in big endian word order. `hc_enc_next()` reads the UTF-8 byte by byte and
cannot make sense of that, so the scalar branch puts the words back in native order first.
Without that step the digest comes out wrong for an ASCII password too, which the self test
catches immediately.

## Cost

CUDA and HIP already ran the pure kernels at width 1, because the block above pins every pure
kernel on a GPU to 1, so nothing changes there. The cost lands on CPU and on non GPU OpenCL.
Measured on PoCL, a 20 second run:

```
./hashcat -m 1000 -a 3 -D 1 --force --potfile-disable --runtime 20 'ffffffffffffffffffffffffffffffff' '?d?d?d?d?d?d?d?d?d?d'

master, pure kernel:  467.0 MH/s   Vec:8
branch, pure kernel:   81.9 MH/s   Vec:1     5.7x slower
branch, -O:           350.3 MH/s   Vec:8     unchanged
```

**A narrower pin is possible and I have not built it.** In `-a 3` the mask is known on the host
before the kernel runs. If no position in it can emit a byte above 0x7f, every `?d`, `?l`, `?u`,
`?s` and every ASCII literal, then no candidate can be multi byte and the vector path is
provably safe for that run. Pinning only when the mask actually carries a high byte would keep
full speed for the common case. It needs the mask, which `mpsp.c` parses, to be reachable from
`backend_session_begin()`, and `--increment` and hcmask files each want thinking about, so it
is a design call rather than a mechanical change. Happy to build it instead if you prefer.

## Two modes deliberately left alone

**m15500** (JKS Java Key Store) and **m35200** (AS/400 SSHA1) are the genuine UTF-16BE modes.
Their `a0-pure` kernel uses `sha1_update_utf16be_swap()`, and no UTF-16BE helper anywhere calls
`hc_enc_next()`, so hashcat has no UTF-8 path for them in any attack mode. Giving them one is a
separate change, not a widening of this one.

## Testing

PoCL CPU only, `-D 1 --force`, with the test coverage from part 1 applied on top.

All 22 modes, `./tools/test.sh -m <mode> -a 3 -t single -D 1 -P -f -V 4`, every one `OK 0/7`,
against `Error 6/7` or `Error 7/7` before, depending on how many of the eight generated
passwords happened to carry a multi byte character:

```
30 40 70 1000 1100 5500 31300
130 140 170 1430 1440 1470 1730 1740 1770 10830 10840 10870 13500 13800 29000
```

m29200 got the same kernel change but the suite generates no `-a 3` case for it.

Green at `-a all -t all`: m1000, m130, m5500. Green again with `NO_NON_ASCII=1`, the switch
that turns the multi byte passwords off, so ASCII did not regress. That switch lives in part 1,
not here: this branch changes no test code at all, so on its own it is measured with whatever
passwords master's `tools/test.pl` generates, which are digits. The width pin does not reach a mode that does no UTF-16 conversion:

```
./hashcat -m 0 -a 3 -D 1 --force --potfile-disable --backend-vector-width 4 '5f4dcc3b5aa765d61d8327deb882cf99' 'passwor?d'
Speed.#01: ... Vec:4
```

## Not covered

- **No GPU.** PoCL CPU only, so the 17 changed kernels have never been compiled by CUDA, HIP or
  AMD OpenCL. This is the gap worth closing before merging.
- The narrower, mask aware version of the vector width pin described above is not built.

### DIFF
--- a/OpenCL/m00130_a3-pure.cl
+++ b/OpenCL/m00130_a3-pure.cl
@@ -53,38 +53,44 @@ KERNEL_FQ KERNEL_FA void m00130_mxx (KERN_ATTR_VECTOR ())
 
   u32x w0l = w[0];
 
+  #if VECT_SIZE == 1
+
+  // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+  // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+  // by byte. Put the words back in native order for it. Only w[0] changes from one
+  // candidate to the next, so everything above it is swapped once here and the loop is
+  // left with the single word it rewrites.
+
+  for (u32 i = 4, idx = 1; i < pw_len; i += 4, idx += 1)
+  {
+    w[idx] = hc_swap32_S (w[idx]);
+  }
+
+  #endif
+
   for (u32 il_pos = 0; il_pos < IL_CNT; il_pos += VECT_SIZE)
   {
     const u32x w0r = words_buf_r[il_pos / VECT_SIZE];
 
     const u32x w0 = w0l | w0r;
 
-    w[0] = w0;
-
     #if VECT_SIZE == 1
 
-    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
-    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
-    // by byte. Put the words back in native order for it.
-
-    u32 c[64] = { 0 };
-
-    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
-    {
-      c[idx] = hc_swap32_S (w[idx]);
-    }
+    w[0] = hc_swap32_S (w0);
 
     sha1_ctx_t ctx;
 
     sha1_init (&ctx);
 
-    sha1_update_utf16le_swap (&ctx, c, pw_len);
+    sha1_update_utf16le_swap (&ctx, w, pw_len);
 
     sha1_update (&ctx, s, salt_len);
 
     sha1_final (&ctx);
 
     #else
+
+    w[0] = w0;
 
     sha1_ctx_vector_t ctx;
 
@@ -158,38 +164,44 @@ KERNEL_FQ KERNEL_FA void m00130_sxx (KERN_ATTR_VECTOR ())
 
   u32x w0l = w[0];
 
+  #if VECT_SIZE == 1
+
+  // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+  // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+  // by byte. Put the words back in native order for it. Only w[0] changes from one
+  // candidate to the next, so everything above it is swapped once here and the loop is
+  // left with the single word it rewrites.
+
+  for (u32 i = 4, idx = 1; i < pw_len; i += 4, idx += 1)
+  {
+    w[idx] = hc_swap32_S (w[idx]);
+  }
+
+  #endif
+
   for (u32 il_pos = 0; il_pos < IL_CNT; il_pos += VECT_SIZE)
   {
     const u32x w0r = words_buf_r[il_pos / VECT_SIZE];
 
     const u32x w0 = w0l | w0r;
 
-    w[0] = w0;
-
     #if VECT_SIZE == 1
 
-    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
-    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
-    // by byte. Put the words back in native order for it.
-
-    u32 c[64] = { 0 };
-
-    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
-    {
-      c[idx] = hc_swap32_S (w[idx]);
-    }
+    w[0] = hc_swap32_S (w0);
 
     sha1_ctx_t ctx;
 
     sha1_init (&ctx);
 
-    sha1_update_utf16le_swap (&ctx, c, pw_len);
+    sha1_update_utf16le_swap (&ctx, w, pw_len);
 
     sha1_update (&ctx, s, salt_len);
 
     sha1_final (&ctx);
 
     #else
+
+    w[0] = w0;
 
     sha1_ctx_vector_t ctx;
 
--- a/OpenCL/m00130_a3-pure.cl
+++ b/OpenCL/m00130_a3-pure.cl
@@ -61,6 +61,31 @@ KERNEL_FQ KERNEL_FA void m00130_mxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_update (&ctx, s, salt_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -70,6 +95,8 @@ KERNEL_FQ KERNEL_FA void m00130_mxx (KERN_ATTR_VECTOR ())
     sha1_update_vector (&ctx, s, salt_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
@@ -139,6 +166,31 @@ KERNEL_FQ KERNEL_FA void m00130_sxx (KERN_ATTR_VECTOR ())
 
     w[0] = w0;
 
+    #if VECT_SIZE == 1
+
+    // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+    // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+    // by byte. Put the words back in native order for it.
+
+    u32 c[64] = { 0 };
+
+    for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+    {
+      c[idx] = hc_swap32_S (w[idx]);
+    }
+
+    sha1_ctx_t ctx;
+
+    sha1_init (&ctx);
+
+    sha1_update_utf16le_swap (&ctx, c, pw_len);
+
+    sha1_update (&ctx, s, salt_len);
+
+    sha1_final (&ctx);
+
+    #else
+
     sha1_ctx_vector_t ctx;
 
     sha1_init_vector (&ctx);
@@ -148,6 +200,8 @@ KERNEL_FQ KERNEL_FA void m00130_sxx (KERN_ATTR_VECTOR ())
     sha1_update_vector (&ctx, s, salt_len);
 
     sha1_final_vector (&ctx);
+
+    #endif
 
     const u32x r0 = ctx.h[DGST_R0];
     const u32x r1 = ctx.h[DGST_R1];
--- a/OpenCL/m00140_a3-pure.cl
+++ b/OpenCL/m00140_a3-pure.cl
@@ -50,34 +50,40 @@ KERNEL_FQ KERNEL_FA void m00140_mxx (KERN_ATTR_VECTOR ())
 
   u32x w0l = w[0];
 
+  #if VECT_SIZE == 1
+
+  // The mask processor hands a OPTS_TYPE_PT_GENERATE_BE mode its candidates in big endian
+  // word order (markov_be.cl), which hc_enc_next() cannot read: it decodes the UTF-8 byte
+  // by byte. Put the words back in native order for it. Only w[0] changes from one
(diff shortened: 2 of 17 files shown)


jsteube:
Swapping the words back to native before hc_enc_next() sees them is genuinely needed rather than defensive, markov_be.cl really does hand GENERATE_BE modes their candidates in big endian word order, and putting the new code under #if VECT_SIZE == 1 with the vector path untouched in the #else is about as contained as it could be. The comment explaining the swap is the kind I wish more kernels carried.

This is getting complicated now but there's no way around. The array and the swap are both inside the il_pos loop:

```
u32 c[64] = { 0 };

for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
{
  c[idx] = hc_swap32_S (w[idx]);
}
```

The kernel already has u32x w[64] = { 0 } at outer scope, so this is a second 256 byte array, rebuilt per candidate. But only w[0] changes in that loop. Everything from w[1] up is set once before it from pws[gid].i[idx] and never touched again, so the swap of those words is being redone for every candidate and carried in a whole second array to hold a value that is constant except for its first word.

Swapping w[1] and up once before the loop, then writing w[0] = hc_swap32_S (w0) inside it, gets rid of the second array completely and leaves 1 swap per candidate. In place on w is safe done that way, because w[0] is overwritten each iteration anyway.

That matters more than it looks. These are pure kernels, which already run scalar on GPU and are already the register constrained ones, the comment in backend.c says as much. An extra 64 words of private array per work item in the inner loop is exactly the wrong thing to add there, and as far as I can see none of it has to be there at all.

Now the backend.c block, which I think is a separate question from the kernel work.

Master already has this a few lines above where you added yours:

```
// We can't have SIMD in kernels where we have an unknown final password length
// It also turns out that pure kernels (that have a higher register pressure)
// actually run faster on scalar GPU (like 1080) without SIMD

if ((hashconfig->opti_type & OPTI_TYPE_OPTIMIZED_KERNEL) == 0)
{
  if (device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
  {
    vector_width = 1;
  }
}
```

Yours is the same outer condition with the GPU test removed, so on GPU it changes nothing, the width was already 1. Its entire effect is on non GPU devices, in practice CPU OpenCL, where the width comes from CL_DEVICE_NATIVE_VECTOR_WIDTH_INT and is normally 4, 8 or 16.

So the kernel change is what GPUs actually run, and this block only ever costs CPU. hashcat is tuned for GPU first, and I am not keen on taking a 4x to 16x vector width cut on CPU to buy correctness there when the mechanism for saying "this kernel cannot do that" already exists. Dropping the block and letting CPU pure kernels stay on the vector path is a defensible answer, as long as it is said out loud rather than left silent.

If you would rather keep it, I want a number against it, measured on CPU OpenCL for a few of the affected modes.

Either way it needs narrowing, because right now it catches more than it fixes. 27 modes declare OPTS_TYPE_PT_UTF16LE or PT_UTF16BE and have an _a3-pure kernel, so all 27 hit the block. 16 get a scalar branch from you, 7 already had one, 00030 00040 00070 01000 01100 05500 and 31300. That leaves 4 with neither:

```
  m05600   md4_update_utf16le,          no VECT_SIZE == 1 branch
  m15500   sha1_update_vector_utf16le,  no VECT_SIZE == 1 branch
  m24800   no utf16 call at all,        no VECT_SIZE == 1 branch
  m35200   sha1_update_vector_utf16le,  no VECT_SIZE == 1 branch
```

On CPU those 4 give up their vector width and still widen bytes, so they are slower and just as unable to crack a multi byte password. They want either a scalar branch of their own or an exclusion.

Also 16 plus 7 is 23 rather than the 22 in the title, so I would like to know which set you actually measured.

One thing that is not really about this PR, but I do not want it to get lost. Your block leaves optimized kernels alone, which is right and matches what module_01000.c:58 documents. After this lands though, pure and optimized visibly disagree on 27 modes and only module_01000 carries the advice notice. Somebody running -m 5500 -O gets silent false negatives while -P works, with nothing telling them why. Together with #4813 keeping non ASCII passwords away from -O, the optimized side ends up both broken and untested. 

This PR and the #4813 has the potential to get out of control quickly and sooner or later will spread across all hashcat kernels, even non-utf16 ones, but that's what it really takes. Encoding is a nightmare as it always was. I think a full rework on multibyte is what it really takes, that would then also justify a speed drop.

thatux:
For now I've implemented as you suggested (I think).


### The second array is gone

Done exactly as described, in all 16 kernels:

```c
  #if VECT_SIZE == 1

  // ... Only w[0] changes from one candidate to the next, so everything above it
  // is swapped once here and the loop is left with the single word it rewrites.

  for (u32 i = 4, idx = 1; i < pw_len; i += 4, idx += 1)
  {
    w[idx] = hc_swap32_S (w[idx]);
  }

  #endif

  for (u32 il_pos = 0; il_pos < IL_CNT; il_pos += VECT_SIZE)
  {
    const u32x w0 = w0l | w0r;

    #if VECT_SIZE == 1

    w[0] = hc_swap32_S (w0);

    sha1_ctx_t ctx;
    ...
```

The `w[0] = w0` assignment moved inside both arms of the existing `#if`, so there
is one preprocessor branch rather than two, no `u32 c[64]`, and one swap per
candidate instead of a whole password. In place on `w` is safe the way you said:
the loop overwrites `w[0]` every iteration, and `w0l` is read before the pre-swap
so it is unaffected.

    ./hashcat -m 130 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 \
      'bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455' '1234567?d€'

    before and after: Status Cracked, Recovered 1/1, rc=0

### The backend.c block is dropped

You are right that it only ever reached non GPU devices, and here is the number
you asked for. CPU OpenCL, PoCL 5.0, `-a 3`, `--runtime 12`, native width against
the pinned width:

    m130      60.0 MH/s Vec:1     239.3 MH/s Vec:8     4.0x
    m1000    118.5 MH/s Vec:1     579.1 MH/s Vec:8     4.9x
    m1430     37.2 MH/s Vec:1     110.9 MH/s Vec:8     3.0x
    m13500    70.0 MH/s Vec:1     122.2 MH/s Vec:8     1.7x
    m5600      9.8 MH/s Vec:1       9.2 MH/s Vec:8     0.9x

3x to 5x for correctness on the one device class hashcat is not tuned for is not
a trade I want to argue for, so the block is gone and `src/backend.c` is back to
master. It is its own commit, so `git revert` puts it back if you decide the
other way.

Said out loud, as you asked: on CPU and other non GPU OpenCL devices, `-a 3`
still cannot crack a password outside ASCII in these modes unless the width is 1.

    ./hashcat -m 130 -a 3 -D 1 --force --potfile-disable --backend-vector-width 8 \
      'bc93548344c1d91dccb23ffb78440cdd7d4f40b4:2455' '1234567?d€'
    Status Exhausted, rc=1

    the same command with --backend-vector-width 1
    Status Cracked, rc=0

One consequence worth naming before it surprises anyone: #4813 will therefore
still be red on a CPU only test box after this merges, for the same 22 modes,
unless it is run with `-V 1`. On GPU both are green, because the block above mine
already pinned pure kernels there.

The mask aware version, pinning only when the mask can emit a byte above 0x7f,
would keep the speed and the correctness. `backend_session_begin()` does get the
whole `hashcat_ctx`, so `mask_ctx->masks[]` is reachable and it is buildable. It
needs a rule for hcmask lines that carry their own charsets, and by your own last
paragraph it belongs with the wider multibyte work rather than bolted on here.

### The four modes, of which two are not what they look like

I checked each helper for an `hc_enc_next()` call rather than trusting the name:

| mode | what it actually does | |
| --- | --- | --- |
| m05600 | `md4_update_utf16le`, which **decodes** | scalar kernel, `u32 w[64]` |
| m24800 | calls `hc_enc_next()` **directly**, no helper at all | scalar kernel |
| m15500 | `sha1_update_vector_utf16leN`, widens | genuine UTF-16BE, no decoder exists |
| m35200 | `sha1_update_global_utf16be_swap`, widens | same |

m05600 and m24800 look unbranched to a grep because they never needed a branch:
both are written scalar throughout. The measurement above says the same thing
from the other side, m5600 runs at the same speed at Vec:1 and Vec:8 because its
kernel takes nothing from the vector width, so pinning it cost nothing and bought
nothing.

That leaves m15500 and m35200, which are the two the PR body already lists as
deliberately untouched. Dropping the block moots the exclusion either way.

### 23, not 22

You are right. 16 kernels changed here plus the 7 that already had a branch is
23 modes. The title says 22 because the suite generates no `-a 3` vector for
m29200, so 22 is the number I could measure, not the number I changed. That
distinction is a footnote in the body and should be in the title instead.

### The optimized side

Agreed, and #4813 now does something about it rather than routing around it. The
oracles branch on `IS_OPTIMIZED`, so the optimized kernels get non-ASCII
passwords too and are checked against the widening they are documented to do.
The advice notice in module_01000.c stops being folklore and becomes something
the suite asserts. It is still only that one module carrying the notice, which is
a real gap, but it is now a gap with a test behind it.

PR #4813, opened 2026-08-31
test.sh: test passwords that are not 7 bit ASCII, and build the -g containers with one

thatux:
`tools/test.pl` builds every password out of the digits `0` to `9`, and `tools/test.sh` builds
every `-g` container with the literal `hashcat` and searches for it with a literal mask.
Neither can fail for the right reason. This branch changes both.

This is .1 of the split discussed in https://github.com/hashcat/hashcat/pull/4810#issuecomment-5483004665: only improve test coverage.

This depends on the fix in https://github.com/hashcat/hashcat/pull/4812 hence it is branched as such (_I'll remove the draft label on this PR once that PR lands in master_). Without it this branch is red for 22 modes under
`-a 3`, which is the point of it: `./tools/test.sh -m 130 -a all -t all -D 1 -P -f -V 1` gives
`Error 7/7` for `-a 3` on master and `OK 16, Error 0` with the fix applied. Merge the fix first,
or merge them together.

## Reproduction

The suite cannot see a UTF-8 bug today. Revert only the `hc_enc_next()` hunk of 5c840aaec,
"Crack non ASCII passwords again in 41 hash modes", and every test still passes:

```
git show 5c840aaec -- OpenCL/inc_common.cl | git apply -R -
rm -rf kernels/
./tools/test.sh -m 1000 -a all -t all -D 1 -P -f -V 1
```

```
before this branch: 16 OK,  0 Error, 1 Skip   the suite does not notice
after  this branch:  3 OK, 13 Error, 1 Skip   7/8 not found under -a 0 alone
```

The one Skip is `Attack 4, Mode single`, which master skips either way.

Same for the containers. `-g` built them with `hashcat` and searched `hashca?l`, both
hardcoded, so a pass proved only that the literal matched itself:

```
./tools/test.sh -m 22931 -g -a all -t single -D 1 -f -V 1

before: every run builds the key with 'hashcat' and searches 'hashca?l'
after:  one run built '7<U+0939>14551048', the next a different one, both Cracked, rc=0
```

## What the passwords look like now

```
7<U+0939>60778768        2886128986184<U+4E2D>0542
0489<U+AC00>5            ह4€0龍ह9文44
```

Eleven characters, covering the lead bytes `hc_enc_next()` has to get right, including one
4 byte character: euro, hiragana, katakana, four CJK, hangul, devanagari, fullwidth latin, and
U+1F600. A character can land anywhere, including the first byte.

`random_non_ascii_string()` joins `random_numeric_string()` and substitutes them into the digits
it returns. The substitution is in place, so the byte length does not change and the password
still fits the mode's `Pwd.Len.Max`. It comes back as plain digits for a mode that cannot take
one, so a caller never has to ask which. `random_numeric_string()` is untouched, so salts, site
keys and challenge characters keep getting what they always got, a test module added later
cannot pick up multi byte salts by accident, and `tools/test_modules/README.md`, which points
modules at it for exactly that job, stays correct as written.

Which modes get them is read out of the kernels, because there is no flag for it.
`OPTS_TYPE_PT_ALWAYS_ASCII` is about hexifying output, not about what a mode can crack, and only
modules 3000 and 25400 set it. So `inc_hash_*.cl` is split into the helpers that call
`hc_enc_next()` and the ones that widen, and a mode gets multi byte passwords when the pure
kernel that will run uses a decoding one. Modes that case fold the plaintext, spell it in hex or
base58, or build it in `module_get_random_password()` stay on ASCII, as does any UTF-16 mode
under `-O`.

### Turning it off

`tools/test.pl` honours one environment variable, `NO_NON_ASCII`. Set it to anything and every
generated password comes back as plain digits:

```
tools/test.pl:1070    return 0 if exists $ENV{"NO_NON_ASCII"};
```

It is one line, nothing depends on it, and it is deliberately not in `test.sh -h`: it is for
whoever is working on the suite, not for whoever is running it. What it buys is the answer to
the first question a red mode raises, is this the encoding or was it already broken, in one run
instead of an edit to `test.pl`. It earned its place three times while this branch was written:

```
m11600   non-ASCII on -> Error 5/8    NO_NON_ASCII=1 -> OK 0/8     the -O gate, not the oracle
m22931   -a 1/6/7 red                 NO_NON_ASCII=1 -> green      the split, not the container work
m130     -a 3 Error 7/7               NO_NON_ASCII=1 -> OK 0/7     the kernel, not the mask
```

It is also how the before lines in this description were measured against the same passwords.

## What had to change so the attacks can spell one

- `export LC_ALL=C`, so `${#pass}` and `cut -c` count bytes the way hashcat does.
- `mask_literalize()`, because no `?d` produces byte `0xe2`. A mask position over a non-digit
  byte becomes that byte, which leaves the keyspace unchanged.
- `utf8_split_point()`, which moves a dict/mask split to a character boundary. The combinator
  kernels convert the word and the amplifier separately, so each half has to be valid UTF-8 on
  its own:

    ```
    printf '12345\xe7\n' > /tmp/d1 ; printf '12345\n' > /tmp/d2
    H=ac25eecbd4dd9f58b674d3bebe463f9e     # NTLM of 12345<U+7801>678
    ./hashcat -m 1000 -a 6 -D 1 --force --potfile-disable --backend-vector-width 1 "$H" /tmp/d2 "$(printf '\xe7\xa0\x81?d?d?d')"
    ./hashcat -m 1000 -a 6 -D 1 --force --potfile-disable --backend-vector-width 1 "$H" /tmp/d1 "$(printf '\xa0\x81?d?d?d')"

    split on a character boundary: Status Cracked,   rc=0
    split inside the character:    Status Exhausted, rc=1
    ```

    Both spell the same 11 bytes. `-m 0` cracks either. This one is by design, not a bug.
- A character is only written where the password keeps at least one byte outside it. A password
  that is nothing but one character cannot be cut in two on a boundary, and `-a 7` came back
  with an empty dictionary. That costs nothing above 4 bytes.

### Multi hash

The eight passwords of a length share one mask, so `tools/test.pl` draws the layout, which
character and where, from a generator seeded by the password length rather than from `rand()`.
All eight come out with their characters in the same places and one mask can spell them; the
digits stay random per password.

```
len 8   2327<U+AC00>1   0629<U+AC00>5   0489<U+AC00>5
```

`-a 6`, `-a 7` and `-a 12` multi build their mask from what the dictionary holds rather than
from `mask_6[]` and `mask_7[]`, which were sized for a split on a byte.

`-a 3` multi is one `--increment` run over passwords of many lengths at once, so no single mask
spells all of them however the layout is drawn. It already had a second path for the cases where
a mask will not do, handing hashcat an hcmask file, which is one mask per line. That path is
taken whenever a character is in play, with a line per password, so the run keeps its shape.

### 41 test oracles hashed the password as Latin-1

Ground truth against a real archive:

```
7z a -mhe=on -p'pass€1' /tmp/t.7z /etc/hostname
perl ~/john/run/7z2john.pl /tmp/t.7z | grep -oE '\$7z\$[^:]*' > /tmp/t.hash
./hashcat -m 11600 -a 3 -D 1 --force --potfile-disable --backend-vector-width 1 "$(cat /tmp/t.hash)" 'pass€?d'

Status Cracked, Recovered 1/1, rc=0
```

hashcat and 7-Zip agree, so the kernel is right. `tools/test_modules/m11600.pm` called
`encode("UTF-16LE", $word)` on the raw bytes, which widens each byte as if it were Latin-1. 39
modules had the same bug and now decode first. The modules whose kernel widens too are left
alone, because there the oracle was right.

m05500 and m05600 took the NT hash from `Authen::Passphrase::NTHash`, which is wrong for a
character outside the BMP:

```
perl -MAuthen::Passphrase::NTHash -MEncode -MDigest::MD4=md4_hex -e \
  '$c = decode("utf-8", "12345678\xf0\x9f\x98\x80");
   print Authen::Passphrase::NTHash->new(passphrase => $c)->hash_hex, "\n",
         md4_hex(encode("UTF-16LE", $c)), "\n"'

36a5dbdc75876d5364ef604c79180078
1a160fb7177bbdf8e8edbe681acd5380
```

hashcat agrees with the second, so both modules compute it directly now.

## The -g containers

Making the container password random turned two container tests red, and both were red for
reasons that had nothing to do with the password.

**The LUKS `-a 3` mask had been collapsing to `?l`.** `luks_test()` reads `LUKS_PASSWORD`, which
is only ever assigned inside `luks_legacy_test()`. For mode 29511 that function has not run, so
the variable is empty, `${#LUKS_PASSWORD} - 1` is `-1`, and the mask degenerates. Not new:

```
git checkout master -- tools/test.sh
./tools/test.sh -m 29511 -g -a 3 -t single -D 1 -f -V 1

Error 1/1 six times, mask '?l', the same as before the change
```

It stayed invisible because the mask never had to match anything until `-g` started building the
container: against a container that is not there, a failed crack and a skipped test look alike.

**`-g` never built a LUKS2 container**, while claiming it could. 34100 was in `GEN_MODES` and in
`GEN_SUDO_MODES`, so it asked for root and reported `OK`, but there was no `luks2_generate()`,
`luks2_test()` walked `${TDIR}/luks2_tests` which `-g` never fills, the dispatch called it only
when `GENERATE_CONTAINERS` is 0, and the guard read a global only the download path sets. It
reported `OK` on the strength of its `test.pl` oracle and looked like it had covered the
container. There is a generator now, six containers, every combination the mode accepts.
`cryptsetup luksDump` on one reports `Version: 2`, `PBKDF: argon2id`, `Time cost: 4`,
`Memory: 16384`, `Threads: 1`, and the `OK` is not vacuous:

```
./hashcat -a 3 -m 34100 <hash> 'zzzzzz?l'   Exhausted, 0/1
./hashcat -a 3 -m 34100 <hash> 'lbvejc?l'   Cracked,   1/1
```

**The generators needed a UTF-8 locale.** `LC_ALL=C` is what makes `${#pass}` count bytes, and
veracrypt reads its password through wxWidgets, which decodes argv with that same variable.
Under `LC_ALL=C` a UTF-8 password is read as Latin-1 and re-encoded, so the volume answers to
bytes nobody chose:

```
veracrypt --text --create /tmp/v.vc ... --password="$(printf '7\xe0\xa4\xb960778768')"
printf '7\xe0\xa4\xb960778768\n' > /tmp/v.txt
./hashcat -a 0 -m 13721 /tmp/v.vc /tmp/v.txt

created under LC_ALL=C      rc=1, not cracked
created under LC_ALL=C.utf8 rc=0, cracked
```

hashcat is not involved, it gets the bytes it was given. Confirmed by cracking the volume with
the mojibake and only with it: `utf-8` rc=1, `latin1 -> utf8` rc=0, `utf-16le/be` rc=1. tcplay
is driven through expect, whose Tcl picks its encoding the same way. So the generators that take
a password on the command line run under a UTF-8 locale, picked from whichever of `C.UTF-8`,
`C.utf8`, `en_US.UTF-8` or `en_US.utf8` the machine has, while the shell keeps `LC_ALL=C`.

**The LUKS oracles left every container they built in `/tmp`.** They build a real 20 MiB
container to get a hash out of it, hardcoded `/tmp/out` and `/tmp/mnt`, and deleted none of it
except on their two failure paths. On the machine this was written on `/tmp/out` had reached
4.2 GB across 416 files and `/tmp/mnt` held 280 empty directories.

```
./tools/test_modules/m34100.sh 'hashcat' >/dev/null

before: /tmp/out 4545M -> 4566M, /tmp/mnt 280 -> 281 entries, and they stay
after:  both unchanged, 1 MiB of hash and log in the scratch directory the caller names
```

The container goes once the hash is out of it, except for mode 14600 which is handed the
container itself; the mount point goes with it; and `OUTPUT_DIR`, `MOUNT_DIR` and the log honour
`HCTEST_SCRATCH_DIR` and `HCTEST_MOUNT_DIR`, which `tools/test.sh` points at the run's own
directory. Run by hand the scripts keep the paths they always had.

## Testing

PoCL CPU only, `-D 1 --force`, with the kernel fix from the other PR applied.

- `-m 1000` at `-a all -t all -V 1 -P`: OK 16, Error 0, Skip 1, with characters in both single
  and multi hash mode. Green again with `NO_NON_ASCII=1`, so the digits path did not regress.
- all 41 modules whose oracle changed, `-a 0 -t single -V 4 -P`, re-run after the kernel change
  rather than before it
- m130, m5500, m5600, m11600, m17200, m22931 at `-a all`
- m0, m10, m400, m22000 and a 24 mode spread: 20, 100, 110, 300, 500, 900, 1400, 1410, 1700,
  1710, 2611, 2711, 2811, 3710, 4400, 4500, 4700, 6000, 6100, 6900, 11500, 17300, 21100, 24900
- containers as root, `-g -a all -t single`, each run building its own password:

```
m6211  TrueCrypt   OK 3  Error 0
m13721 VeraCrypt   OK 5  Error 0
m29511 LUKS1       OK 9  Error 0  Skip 1
m34100 LUKS2       OK 7  Error 0
m22931 SSH         OK 10 Error 0  Skip 1
m17200 PKZIP       OK 8  Error 0  Skip 1
m13711 VeraCrypt   Skip: veracrypt 1.26 dropped RIPEMD-160 volumes
```

The Skips are `Attack 4, Mode single`, which master skips the same way.

## Not covered

- **No GPU.** PoCL CPU only.
- `test.sh` is not safe to run concurrently with another instance: two runs share the device,
  and with `-g` they also share sudo, loop devices and device-mapper names. The failure mode is
  silent, plausible-looking `Error : N/N` lines rather than a clash. Pre-existing, not addressed
  here.
- **m24800 produces 0 test vectors.** Pre-existing, identical with `NO_NON_ASCII=1`.

### DIFF
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -1059,7 +1059,6 @@ sub non_ascii_supported
 
   my $any_utf16    = 0;
   my $pure_decodes = 0;
-  my $opt_widens   = 0;
 
   for my $kernel (glob (sprintf ("%s/../OpenCL/m%05d*.cl", $FindBin::Bin, $mode)))
   {
@@ -1081,8 +1080,7 @@ sub non_ascii_supported
       if ($decoding->{$1}) { $decodes = 1; } else { $widens = 1; }
     }
 
-    if ($kernel =~ /-pure\.cl$/) { $pure_decodes ||= $decodes; }
-    else                        { $opt_widens   ||= $widens;   }
+    $pure_decodes ||= $decodes if $kernel =~ /-pure\.cl$/;
   }
 
   if ($any_utf16)
@@ -1091,9 +1089,13 @@ sub non_ascii_supported
 
     return 0 if $pure_decodes == 0;
 
-    # the optimized kernel is the one that would run, and it widens
+    # -O changes the plaintext path even for a mode that has no optimized kernel of its own.
+    # m11600 is one: the same generated vector cracks under -P and comes back not found under
+    # -O, which is the default the suite runs with. So a mode that converts UTF-16 at all keeps
+    # its passwords ASCII whenever -O is in play, rather than only when an optimized kernel of
+    # its own would have widened them.
 
-    return 0 if $IS_OPTIMIZED == 1 && $opt_widens == 1;
+    return 0 if $IS_OPTIMIZED == 1;
   }
 
   return 1;
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -571,7 +571,16 @@ sub password
 
   return unless is_count ($count);
 
-  my $string = random_non_ascii_string ($count) // "";
+  # A real archive or volume carries whatever encoding the application wrote, and the
+  # optimized path cannot match a multi byte one: a genuine 7-Zip archive built with a euro
+  # sign in its password cracks under -P and comes back not found under -O. So a -O run has
+  # to build its containers out of ASCII, and test.sh sets NO_NON_ASCII to ask for that.
+
+  my $string = (exists $ENV{"NO_NON_ASCII"})
+             ? random_numeric_string ($count)
+             : random_non_ascii_string ($count);
+
+  $string //= "";
 
   print "$string\n";
 }
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -12,7 +12,7 @@ use File::Basename;
 use FindBin;
 use List::Util 'shuffle';
 
-my $TYPES = [ 'edge', 'single', 'passthrough', 'potthrough', 'verify' ];
+my $TYPES = [ 'edge', 'single', 'password', 'passthrough', 'potthrough', 'verify' ];
 
 my $TYPE = shift @ARGV;
 my $MODE = shift @ARGV;
@@ -71,12 +71,16 @@ my @NON_ASCII_CHARS =
   "\xf0\x9f\x98\x80",  # U+1F600 grinning face
 );
 
-# The first bytes of a password are left as digits. tools/test.sh builds the -a 3, -a 6 and
-# -a 7 masks out of '?d' groups, and the multi hash tests share one such mask across every
-# password of a given length, which only works while those bytes are digits. Multi hash
-# passwords are 2 to 8 bytes long, so a password of 8 bytes or less stays pure ASCII.
+# Where the characters may land. Anywhere, including the first byte: tools/test.sh rewrites
+# the '?d' at a position whose byte is not a digit into that byte, so a mask can spell one
+# wherever it turns up.
+#
+# The multi hash tests are the reason the layout is not drawn from rand(). They share one mask
+# across every password of a given length, so the characters have to sit in the same places in
+# all of them. The generator below is seeded from the length alone, which makes the shape of a
+# password a function of its length while the digits around it stay random per password.
 
-my $NON_ASCII_SKIP_BYTES = 8;
+my $NON_ASCII_SKIP_BYTES = 0;
 
 # Roughly how often an eligible position is turned into a multi byte character. Low enough
 # that a generated set still holds plain ASCII passwords, high enough that a set of 8 almost
@@ -95,6 +99,12 @@ if ($TYPE eq 'edge')
 elsif ($TYPE eq 'single')
 {
   single (@ARGV);
+}
+elsif ($TYPE eq 'password')
+{
+  usage_exit () if scalar @ARGV > 1;
+
+  password (@ARGV);
 }
 elsif ($TYPE eq 'passthrough')
 {
@@ -543,6 +553,21 @@ sub single
   }
 }
 
+sub password
+{
+  # One password for this mode, on stdout, nothing else. tools/test.sh builds its -g containers
+  # with it, so a container gets the same multi byte characters the oracle passwords get, and
+  # the same per mode gate decides whether it gets any.
+
+  my $count = shift // 12;
+
+  return unless is_count ($count);
+
+  my $string = random_non_ascii_string ($count) // "";
+
+  print "$string\n";
+}
+
 sub passthrough
 {
   my $option = shift || '';
@@ -989,15 +1014,28 @@ sub sprinkle_non_ascii
 
   my $len = length $string;
 
+  # Seeded from the length, so every password of a given length comes out with its characters
+  # in the same places. A plain rand() here would give each of the eight multi hash passwords
+  # of a length a different shape, and no single mask could spell all eight.
+
+  my $seed = ($len * 2654435761) % 4294967291;
+
+  my $rand = sub
+  {
+    $seed = ($seed * 1103515245 + 12345) % 2147483648;
+
+    return $seed / 2147483648;
+  };
+
   my $pos = $NON_ASCII_SKIP_BYTES;
 
   while ($pos < $len)
   {
-    my $char = $NON_ASCII_CHARS[rand @NON_ASCII_CHARS];
+    my $char = $NON_ASCII_CHARS[int ($rand->() * scalar @NON_ASCII_CHARS)];
 
     my $char_len = length $char;
 
-    if ((($pos + $char_len) <= $len) && (rand () < $NON_ASCII_RATE))
+    if ((($pos + $char_len) <= $len) && ($rand->() < $NON_ASCII_RATE))
     {
       substr ($string, $pos, $char_len) = $char;
 
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -82,6 +82,14 @@ my @NON_ASCII_CHARS =
 
 my $NON_ASCII_SKIP_BYTES = 0;
 
+# A character is only written where the password keeps at least one byte outside it. -a 1, -a 6,
+# -a 7 and -a 12 cut the password in two and both halves have to land on a character boundary,
+# so a password that is nothing but one character cannot be cut at all and leaves one of the two
+# dictionaries empty. This is the smallest rule that avoids it: it costs nothing above 4 bytes,
+# where the old flat minimum of 6 cost every password of 4 and 5 bytes as well.
+
+my $NON_ASCII_MIN_SPARE = 1;
+
 # Roughly how often an eligible position is turned into a multi byte character. Low enough
 # that a generated set still holds plain ASCII passwords, high enough that a set of 8 almost
 # always holds at least one that is not.
@@ -1035,7 +1043,7 @@ sub sprinkle_non_ascii
 
     my $char_len = length $char;
 
-    if ((($pos + $char_len) <= $len) && ($rand->() < $NON_ASCII_RATE))
+    if ((($pos + $char_len) <= ($len - $NON_ASCII_MIN_SPARE + ($pos > 0 ? 1 : 0))) && ($rand->() < $NON_ASCII_RATE))
     {
       substr ($string, $pos, $char_len) = $char;
 
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -1135,13 +1135,11 @@ sub non_ascii_supported
 
     return 0 if $pure_decodes == 0;
 
-    # -O changes the plaintext path even for a mode that has no optimized kernel of its own.
-    # m11600 is one: the same generated vector cracks under -P and comes back not found under
-    # -O, which is the default the suite runs with. So a mode that converts UTF-16 at all keeps
-    # its passwords ASCII whenever -O is in play, rather than only when an optimized kernel of
-    # its own would have widened them.
-
-    return 0 if $IS_OPTIMIZED == 1;
+    # -O leaves OPTS_TYPE_PT_UTF16LE set, so interface.c has the host widen the candidate
+    # before the kernel ever sees it, whatever the kernel itself would have done. That is why
+    # a mode with no optimized kernel of its own, m11600 for one, still behaves the optimized
+    # way under -O. The oracles follow the same split through $PW_CHARSET, so both families
+    # get non-ASCII passwords and each is checked against what it actually does.
   }
 
   return 1;
--- a/tools/test.pl
+++ b/tools/test.pl
@@ -50,6 +50,42 @@ my $single_outputs = 8;
 
 my $constraints = get_module_constraints ();
 
+# Multi byte UTF-8 characters the generated passwords are seeded with. Every entry is a raw
+# byte string, and together they cover the lead bytes a kernel has to get right: a euro sign,
+# hiragana, katakana, CJK, hangul, devanagari, fullwidth latin and one 4 byte character.
+# Passwords made only of digits never leave the ASCII path, so a mode that mangles anything
+# above 0x7f used to pass the suite unnoticed.
+
+my @NON_ASCII_CHARS =
+(
+  "\xe0\xa4\xb9",      # U+0939  devanagari letter ha
+  "\xe2\x82\xac",      # U+20AC  euro sign
+  "\xe3\x81\x8b",      # U+304B  hiragana letter ka
+  "\xe3\x82\xab",      # U+30AB  katakana letter ka
+  "\xe4\xb8\xad",      # U+4E2D  cjk ideograph 'middle'
+  "\xe6\x96\x87",      # U+6587  cjk ideograph 'script'
+  "\xe7\xa0\x81",      # U+7801  cjk ideograph 'code'
+  "\xe9\xbe\x8d",      # U+9F8D  cjk ideograph 'dragon'
+  "\xea\xb0\x80",      # U+AC00  hangul syllable ga
+  "\xef\xbc\xa1",      # U+FF21  fullwidth latin capital a
+  "\xf0\x9f\x98\x80",  # U+1F600 grinning face
+);
+
+# The first bytes of a password are left as digits. tools/test.sh builds the -a 3, -a 6 and
+# -a 7 masks out of '?d' groups, and the multi hash tests share one such mask across every
+# password of a given length, which only works while those bytes are digits. Multi hash
+# passwords are 2 to 8 bytes long, so a password of 8 bytes or less stays pure ASCII.
+
+my $NON_ASCII_SKIP_BYTES = 8;
+
+# Roughly how often an eligible position is turned into a multi byte character. Low enough
+# that a generated set still holds plain ASCII passwords, high enough that a set of 8 almost
+# always holds at least one that is not.
+
+my $NON_ASCII_RATE = 0.34;
+
+my $NON_ASCII_OK = non_ascii_supported ($MODE);
+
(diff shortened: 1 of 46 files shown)


jsteube:
That was fast and I see it's still a draft, but there's some correction needed early, thus the comment. The split did not quite happen. This branch still carries everything from #4812, byte for byte, the 28 line vector_width block in src/backend.c and all 16 of the m*_a3-pure.cl kernels. So merging it merges the fix too.

That is the bit I was trying to avoid. What I want is to put these tests on current master and watch them go red across the 22 modes, because that failing run is the argument for #4812. Once both arrive in the same commit the suite is green on the first try and neither of us learns anything. So drop src/backend.c and the kernels from here, leave them in #4812, and we land this one red and that one green.

Now the interesting part, which is the oracle.

I read the .pm changes as a straightforward fix at first, encode ('UTF-16BE', $word) missing its decode, and it is more than that. Widening each byte on its own is not a mistake in isolation, it is what the optimized kernels actually do. module_01000.c:58 spells it out:

```
When using the optimized-kernel: some byte encoding is simplified that would make some
non-ASCII passwords never be cracked, even if they are under the optimized byte limit.
The pure-kernel implements true UTF16 handling so doesn't have this problem and should
crack all passwords.
```

So the 2 kernel families are supposed to disagree here. With passwords made of digits they do not, because widening and real UTF-16 give the same bytes, which is how one oracle managed to serve both for years without anyone noticing. Feed it a euro sign and that stops being true, and no single oracle can be right for both any more. decode ('utf-8', $word) picks the pure kernel and quietly abandons the optimized one.

I can see how you dealt with that:

```
return 0 if $IS_OPTIMIZED == 1;
```

which works, but test.sh runs -O by default, so in practice the default run now generates no non-ASCII passwords at all and the optimized kernels stay untested in exactly the area this is about.

I think branching the oracle beats skipping the test, and you are already holding the variable that does it. test.sh exports IS_OPTIMIZED at line 6034 from the same value it uses 2 lines further down to add -O, so they cannot drift, and since the .pm gets required into the same perl process you can just read it. m32300.pm has been doing this on its own for a while:

```
my $IS_OPTIMIZED = 1;

if (exists $ENV{"IS_OPTIMIZED"} && defined $ENV{"IS_OPTIMIZED"})
{
  $IS_OPTIMIZED = $ENV{"IS_OPTIMIZED"};
}

my $salt2_max_len = $IS_OPTIMIZED == 1 ? 33 : 238;
```

Same shape on the UTF-16 oracles gives each mode whichever encoding its kernel really uses, widen when optimized, decode when pure. Then the default run keeps its coverage, both families get tested against what they are documented to do, and that advice notice becomes something the suite checks instead of something we route around. It is also the better outcome for #4812: if optimized kernels are meant to fail on multi byte passwords, I would rather see the suite prove they fail the documented way than never ask them.

The other thing I keep coming back to is the 43 modules. Byte widening is genuinely correct for some formats, plenty of applications widen rather than convert, so decode is not automatically the right answer everywhere. Which ones did you actually check against the application or a real artifact, and which came along because they matched the pattern? Not asking you to have done all 43 by hand, I just want to know which ones to look at twice. These files end up being the reference for every kernel change after this, and a wrong oracle costs far more than a wrong kernel.

Two smaller things. test.sh is +663/-192 which is a lot of movement in a harness that has hidden things from us before, so I will go through it properly once the kernel changes are out and the diff is only tests. And the HCTEST_SCRATCH_DIR, mkfs and rmdir bits look like they belong to something else entirely, no objection to them, they would just be easier to review on their own.

thatux:
Yeah, I originally built this branch on #4812, but rebased on master now.
Not sure I understand you completely, I guess all these UTF-16 `-O` kernels should have that `advice notice` i.e. `This attack with -O can only crack ASCII, run without -O if non-ASCII characters are in dictionary`

I think I'd prefer test.sh not to break on `-O` by default. I think the default test, should be regression, and for `-O` I guess that should be ASCII-only.

### The split, this time for real

    git rebase --onto upstream/master ac0c75b80 testsh_non_ascii

11 commits, `tools/` only, no `src/backend.c` and no kernels. It lands red where
you wanted it red:

    ./tools/test.sh -m 130 -a 3 -t single -D 1 -f -V 1 -P
    Error : 7/7 not found

    ./tools/test.sh -m 130 -a 3 -t single -D 1 -f -V 1
    OK : 0/6 not found

which is the failing run that argues for #4812, on master, with no kernel change
in sight.

### The oracle branches instead of skipping

You were right that skipping was the wrong answer, and the variable was already
there. Every UTF-16 oracle now carries:

```perl
my $PW_CHARSET = "latin1";

if (exists $ENV{"IS_OPTIMIZED"} && defined $ENV{"IS_OPTIMIZED"} && $ENV{"IS_OPTIMIZED"} == 0)
{
  $PW_CHARSET = "utf-8";
}
```

with `decode ("utf-8", $word)` becoming `decode ($PW_CHARSET, $word)`.
`decode("latin1")` reproduces byte widening exactly, so one variable covers both
families and `return 0 if $IS_OPTIMIZED == 1` is gone.

    ./tools/test.sh -m 1000  -a 3 -t single -D 1 -f -V 1        OK : 0/7
    ./tools/test.sh -m 1000  -a 3 -t single -D 1 -f -V 1 -P     OK : 0/7
    ./tools/test.sh -m 5500  -a 3 -t single -D 1 -f -V 1        OK : 0/7
    ./tools/test.sh -m 5500  -a 3 -t single -D 1 -f -V 1 -P     OK : 0/6
    ./tools/test.sh -m 11600 -a 0 -t single -D 1 -f -V 1        OK : 0/8
    ./tools/test.sh -m 11600 -a 0 -t single -D 1 -f -V 1 -P     OK : 0/8

all on non-ASCII passwords, where before the default run generated none at all.

m11600 is worth calling out because it was the reason for the blanket skip in the
first place, and my old comment had the cause wrong. `-O` leaves
`OPTS_TYPE_PT_UTF16LE` set, so `interface.c` has the **host** widen the candidate
before the kernel sees it, whatever the kernel itself would have done. That is
why a mode with no optimized kernel of its own still behaves the optimized way
under `-O`, and why following `IS_OPTIMIZED` rather than which kernel exists is
the right split.

One file in that commit is not one this branch had touched. `m01000.pm` on master
already reads

```perl
my $utf16le = encode("UTF-16LE", decode("utf-8", $word));
```

so the tree has been carrying the pure kernel's answer for NTLM all along. It
never showed, because master generates digits and widening and decoding agree
there. It needed the same branch as the rest.

### Which of them I actually checked

41, not 43. The other two files in that diff are `luks1.sh` and `m34100.sh`,
which are container scripts rather than oracles.

**Checked against something outside hashcat.** m01000, on the definition of the
NT hash:

    perl -MDigest::MD4 -MEncode -e 'print md4_hex (encode ("UTF-16LE", "hashcat\x{20ac}"))'
    7c23a6931f4a44e3f031202f065b5c4e

which is the vector in #4812. m11600, against an archive the run builds with the
real `7z` binary and its own non-ASCII password, so the whole path from the
application to the kernel is exercised:

    ./tools/test.sh -m 11600 -g -a 0 -t single -D 1 -f -V 1 -P     OK    : 0/1 not found
    ./tools/test.sh -m 11600 -g -a 0 -t single -D 1 -f -V 1        Error : 1/1 not found

That is the strongest evidence in either PR and it cuts both ways. 7-Zip converts
properly, the pure kernel matches it, and the optimized path cannot: the second
line is the module_01000.c:58 limitation demonstrated against an application
instead of asserted.

It also has a consequence I had not thought through. A container test cannot
branch the way an oracle can, because the artifact carries whatever the
application wrote and nobody gets to choose. So an -O run now builds its
containers out of ASCII, and both lines above read OK. Only the artifacts step
back, and only under -O; the oracle passwords keep their characters in both
families.

**Right by the name of the mode.** m00030, m00040, m00070, m00130, m00140,
m00170, m01430, m01440, m01470, m01730, m01740, m01770, m10830, m10840, m10870.
These are hashcat's own constructions and the mode name says `utf16le($pass)`, so
UTF-16LE of the password is the definition, not an inference.

**Windows and Kerberos formats that build on the NT hash or on a UTF-16 string by
construction.** m01100, m02100, m05500, m05600, m07500, m09400, m09500, m09600,
m12800, m13100, m13800, m15300, m15310, m15900, m15910, m18200, m22100, m25300,
m31300. I did not open a real artifact for any of these. I am confident in the
class, not in each member.

**The ones to look at twice.** m13500 PeopleSoft, m24800 Umbraco, m29000, m29200
Radmin3, and in particular m22700 and m27700, the two MultiBit modes, which are
the only UTF-16**BE** oracles in the set. These came along because they matched
the pattern. If any of them is a format that genuinely widens rather than
converts, its oracle is now wrong in the pure direction and I would rather hear
it from you than find out later.

Worth being precise about one thing in the #4812 body while we are here: "the BE
modes are left alone" is true of the kernels, m15500 and m35200, and not of these
two oracles.

For all of the rest, the evidence is that the oracle and the kernel agree, which
proves they agree and nothing more. You put that better than I would have: a
wrong oracle costs far more than a wrong kernel.

### The two smaller things

`tools/test.sh` is the whole diff now that the kernels are out, so the review you
wanted is possible.

The scratch directory work is already its own commit, `test.sh: stop the LUKS
oracles leaving their containers in /tmp`, and nothing else depends on it. Say
the word and it goes out as a separate PR; it is a cherry-pick either way.

One tidy up I have not done: `test.sh: keep a UTF-16 mode on ASCII passwords
whenever -O is in play` is the commit that added the skip, and the oracle commit
above reverses it. The history is honest as it stands but reads oddly. I can
squash the pair out if you would rather the branch not carry a decision and its
reversal.

jsteube:
Merging this. The split is right, the oracle branching is right, and it unblocks the test.py work. That is more useful than spending more time polishing behavior that will disappear with test.pl.

There is one design question to settle in test.py rather than here. -O does not guarantee that an optimized kernel actually runs. When a mode has no optimized kernel, interface.c falls back to the pure kernel and clears OPTS_TYPE_PT_UTF16LE. That applies to 15 of the 42 .pm files in this change, including m02100, so a default -O run silently tests the pure kernel for those modes and reports OK without saying which kernel family it exercised. test.pl does this deliberately. get_module_constraints() copies the constraints in both directions and sets IS_OPTIMIZED to match the kernel that will really run.

I am not convinced that substitution is what the new test suite should do. Someone asking for -O probably wants to know whether the optimized kernels are healthy. A mode with no optimized kernel could instead be reported as not applicable, with -P covering the pure kernel. That would also simplify the oracle selection because only one kernel family could be active, and test.py could drop the constraint copying instead of carrying it forward. This is worth deciding now, while test.py is still small enough to change cheaply.

Also, please do not carry over the comment near test.pl:1149. It says that -O leaves OPTS_TYPE_PT_UTF16LE set so the host performs the widening, and names m11600 as a mode without an optimized kernel. interface.c does the opposite, and m11600 has an optimized kernel and never sets OPTS_TYPE_PT_UTF16LE.

You are right about the advice notice as well. That belongs in a separate change.

PR #4816, opened 2026-09-02
regression in test_edge.sh due to test.sh non-ASCII

thatux:
# test_edge.sh: build masks that can spell a non-ASCII password

#4813 taught `tools/test.sh` about multi-byte passwords and never touched
`tools/test_edge.sh`, so every mask attack in the edge suite still assumes a password is digits
from end to end. It is not one mode: on master the suite fails on `-a 3`, `-a 6`, `-a 7` and
`-a 12` for every hash mode whose oracle produces a non-ASCII password, which is most of them.

## Reproduction

Mode 0, every attack type, both kernel families, single and multi:

    ./tools/test_edge.sh -m 0 -V 1 -D 1 -f --skip-clean-cache

    before: Errors detected: 10      failing attack types 3, 6, 7 and 12
    after:  Errors detected: 0

One of the ten, verbatim:

    !> error (1) detected with CMD: ./hashcat ... -O --backend-vector-width 1
       -m 0 '4d129e1dab3361502...' -a 3 €文65中2가33Ａ0912771973😀90가ह262か110가9
    !> Hash-Type 0, Attack-Type 3, Kernel-Type optimized, Vector-Width 1, Test ID 2,
       Word len 55, Word '€文65中2가33Ａ0912771973😀90가ह262か110가9'

Distilled to hashcat, no suite involved. The password is 1234가0 and its MD5 is
82bf84a88f5aab445c56ebb28a227ad6:

    ./hashcat -m 0 -a 3 -D 1 --force --potfile-disable '82bf84a88f5aab445c56ebb28a227ad6' '123?d?d?d'
    Status Exhausted, rc=1

    ./hashcat -m 0 -a 3 -D 1 --force --potfile-disable '82bf84a88f5aab445c56ebb28a227ad6' '1234가?d'
    Status Cracked, rc=0

The first mask is what `test_edge.sh` builds. The second is what it builds now.

## What is wrong

Two independent faults in the same few lines.

**A `?d` cannot produce a byte above 0x7f.** `-a 3` does
`word_1="${word%???}"` with `mask_1="?d?d?d"`
([test_edge.sh:1009](tools/test_edge.sh#L1009)). One of the three characters it strips is
usually not a digit, so no candidate the mask emits can equal the password. `test.sh` solved this
with `mask_literalize()`, which rewrites the positions that are not digits as literals.

**The split counts bytes, not characters.** `test_edge.sh` exports `LANG=C`
([test_edge.sh:181](tools/test_edge.sh#L181)), under which bash's `${word%???}` and `cut -c`
strip three *bytes*, so the cut lands inside a multi-byte character:

    LC_ALL=C.UTF-8  mask = 123?d?d?d       valid
    LC_ALL=C        mask = 1234M-j?d?d?d   BROKEN utf8

`M-j` is `\xea`, the orphaned lead byte of `가`, and it is the `?` in the error line above.

The two faults need different answers. For `-a 3` the word and the mask are **one string**, so a
split inside a character still reassembles to the right bytes once the tail is literalized, and
the boundary needs no separate fix. `-a 1`, `-a 6`, `-a 7` and `-a 12` hand the two halves over
as **separate buffers**, and a UTF-16 mode converts each on its own, so half a character is two
invalid fragments and the candidate is dropped. Those need the split moved onto a boundary as
well, which is what `test.sh`'s `utf8_split_point()` does.

## What changed

`mask_dots()`, `mask_literalize()` and `utf8_split_point()` are copied from `tools/test.sh`
unchanged. `mask_for()` is new: it builds a mask covering exactly the bytes one side now spans,
and runs it through `mask_literalize()` when the token is `?d`.

Both the single-target and multi-target blocks are patched for `-a 3`, `-a 1`, `-a 6`, `-a 7`
and `-a 12`. The `pt_hex` and `pt_base58` paths keep their `?b` and `?a` shapes and are not
touched: those alphabets are ASCII and the fault does not reach them.

`-a 1` did not fail in any run here, having no mask at all, and is changed for the boundary
alone. It also loses two subshells: `word_1=$(echo $word | cut -c -${word_1_cnt})` left `$word`
unquoted, so a password carrying whitespace or a glob character was already at risk there.

## Testing

    ./tools/test_edge.sh -m <mode> -V 1 -D 1 -f --skip-clean-cache

    mode    master                                 this branch
    0       Errors detected: 10   (-a 3, 6, 7, 12)  Errors detected: 0
    100     Errors detected: 10   (-a 3, 6, 7, 12)  Errors detected: 0
    1400    Errors detected: 10   (-a 3, 6, 7, 12)  Errors detected: 0

34 attack, kernel-type and target-type combinations processed per mode, all green.

## Deviations

The three helpers are copied into `test_edge.sh` rather than shared with `test.sh`. The two
scripts are independent today and neither sources the other, so factoring them out is a change to
both and a wider one than this. Worth doing, and not here.

No GPU on this machine, so everything above is the CPU backend, `-D 1 -f`, and `-V 1` only.

Three modes swept, not all of them. They were picked as plain fast hashes with no salt, one salt
and a slow KDF; the code path is shared by every mode, so a fourth adds little, but a full
`./tools/test_edge.sh` on a GPU box is the run that would settle it.

The reproduction contains a euro sign and other multi-byte characters. They are inside pasted
output where the password is the entire point, so the ASCII check flags them and they stay.

### DIFF
--- a/tools/test_edge.sh
+++ b/tools/test_edge.sh
@@ -180,6 +180,84 @@ function clean_cache()
 export LC_CTYPE=C
 export LANG=C
 
+function mask_dots()
+{
+  # A mask of <count> '?d' groups, the shape the suite has always used for a run of digits.
+
+  local md_count="$1"
+  local md_out=""
+  local md_i
+
+  for ((md_i = 0; md_i < md_count; md_i++)); do
+    md_out="${md_out}?d"
+  done
+
+  printf '%s' "${md_out}"
+}
+
+function mask_literalize()
+{
+  # Rewrite a mask so that every position it covers spells the byte that belongs there. The
+  # generated passwords used to be digits from end to end, which is what makes a mask of '?d'
+  # groups work; tools/test.pl can now seed them with multi byte UTF-8, and no '?d' produces a
+  # byte above 0x7f. Those positions become literals, which costs the attack keyspace it was
+  # never searching anyway. Same function as the one in tools/test.sh.
+  #
+  # $1 = the mask, $2 = the exact bytes the mask has to spell. The mask is returned untouched
+  # unless it covers exactly that many bytes, so a caller that hands over the wrong slice, a
+  # mode with its own mask layout for instance, changes nothing.
+
+  local ml_mask="$1"
+  local ml_text="$2"
+
+  local ml_len=${#ml_mask}
+  local ml_pos=0
+  local ml_cnt=0
+  local ml_out=""
+  local ml_tok
+  local ml_byte
+
+  # count the positions first, a '?x' group covers one byte and anything else covers one byte
+
+  while [ ${ml_pos} -lt ${ml_len} ]; do
+    if [ "${ml_mask:${ml_pos}:1}" = "?" ]; then
+      ml_pos=$((ml_pos + 2))
+    else
+      ml_pos=$((ml_pos + 1))
+    fi
+
+    ml_cnt=$((ml_cnt + 1))
+  done
+
+  if [ ${ml_cnt} -ne ${#ml_text} ]; then
+    printf '%s' "${ml_mask}"
+    return
+  fi
+
+  ml_pos=0
+  ml_cnt=0
+
+  while [ ${ml_pos} -lt ${ml_len} ]; do
+    if [ "${ml_mask:${ml_pos}:1}" = "?" ]; then
+      ml_tok="${ml_mask:${ml_pos}:2}"
+      ml_pos=$((ml_pos + 2))
+    else
+      ml_tok="${ml_mask:${ml_pos}:1}"
+      ml_pos=$((ml_pos + 1))
+    fi
+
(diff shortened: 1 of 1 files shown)


jsteube:
Merged. CONTRIBUTING.md asks every contributor for test_edge.sh output on the modes they touch. Since #4813 that gate has been red on -a 3, -a 6, -a 7 and -a 12 for most modes, which taxed everyone sending a pull request and not only you. It had to go in whatever happens to the suite later.

Your deviations note offers to factor the 3 helpers out of test.sh and test_edge.sh so both share one copy. I would rather leave the duplication where it is. The two scripts are independent today, and how much of the suite the Python rework absorbs is not settled, so coupling them now commits to an answer I do not have yet.

test_edge.sh stays separate from that rework for the moment. The sharing question is worth reopening once the scope is clear, so keep the note rather than acting on it.

PR #4817, opened 2026-09-03
test_module_runner.py as alternative to test.pl

thatux:
# test.py: a python oracle engine beside test.pl, reached with test.sh -y

`tools/test.pl` loads `tools/test_modules/mNNNNN.pm` at run time and calls the hooks it defines.
This adds the same thing in python: `tools/test.py` loading `mNNNNN.py`, with two modules written
against it. `test.pl` and all 514 `.pm` files are untouched, both new modules sit beside their
`.pm`, and a run without `-y` is the run you get today.

The reason to want it: 33 of the `.pm` modules are python programs in a perl heredoc, shelled out
to with `qx{python3 -c ...}`. That arrangement is what produced #4815, where seven oracles built
their program around `b"$word"` and lost every candidate above 0x7f. A `.py` module computes in
process, so the class does not exist.

## Reproduction

Both engines on the same password, exact because m01000 is unsalted:

    $ echo 'hashcat' | perl    tools/test.pl passthrough 1000
    b4b9b02e6f09a9bd760f388b67351e2b

    $ echo 'hashcat' | python3 tools/test.py  passthrough 1000
    b4b9b02e6f09a9bd760f388b67351e2b

Through the suite, the two modes that have a python oracle:

    $ ./tools/test.sh -y -m all -a 0 -t single -D 1 -f -V 1 -P
    Init test for hash type 1000.   OK : 0/8 not found, 0/8 not matched
    Init test for hash type 5200.   OK : 0/8 not found, 0/8 not matched

and nothing else, which is the point of the flag.

## Architecture

These are the decisions the design rests on, in the order they matter. Each is a rule, what it
replaces in `test.pl`, and why the replacement is smaller.

**A1. The password is `bytes`. Everything else is `str`.**

    module_constraints()                                     -> list[list[int]]
    module_generate_hash(word: bytes, salt: str, iterations)  -> str | None
    module_verify_hash(line: bytes)                           -> tuple[str, bytes] | None

A password really is an arbitrary byte string: it carries multi-byte UTF-8, and after
`pack_if_HEX_notation` it can be bytes that are not text at all. A hash is text and a salt is
text. The one place the two meet is the crack line, `hash:password`, which is `bytes` and is split
on the separator before anything is decoded. *Replaces:* no rule at all, which is how `length`
came to silently mean bytes and 42 oracles needed a `decode()` added one at a time.

**A2. `IS_OPTIMIZED` is an input. Constraints are never substituted.** A mode with no kernel for
the family the run asked for is not applicable and says so. *Replaces:*
`get_module_constraints()`, which copies the constraints from whichever family exists into the one
that does not and rewrites `$IS_OPTIMIZED` to match, so a `-O` run reports on the pure kernel
under an optimized label. Applicability is decided by the word pairs, slots 0 and 2; slots 1 and 3
are the salt, where `[-1, -1]` keeps its own meaning of "no salt".

**A3. Failure is loud.** An exception exits non-zero; `None` from `module_generate_hash` means
only "this mode does not support that length". *Replaces:* a `.pm` shelling out and reading
stdout, where a dead dependency yields a wrong hash rather than an error. `install_modules.sh`
works around that today by importing every package itself.

**A4. Helpers by explicit import.** `from test_helpers import pack_if_HEX_notation`. *Replaces:*
Perl's `require` dropping `test.pl`'s subs into the module's namespace, which is why `m05200.pm`
can call `pack_if_HEX_notation` with nothing declaring it.

**A5. A module never spawns a process, and the password is a value, never a template.** Not
interpolated into python source, not into a shell command line, not into a format string. `"`,
`$`, `%` and a backtick are then ordinary bytes, exactly as a multi-byte character is. With no
process boundary there is nothing left to quote, so the class is removed rather than escaped.

Checked against an independently computed MD4 of the UTF-16LE encoding, every class at once:

    hashcat   1234€5678   a"b   a$HOME   a`id`b   100%s   a'b   a\b   a b

All nine agree. An engine that interpolates fails at least three of them. This is not theoretical:
fifteen `.pm` oracles still interpolate the password into a shell command line, and measured
against m31100, `a$HOME` and `` a`id`b `` come back as silently wrong digests.

**A6. The engine is a flag, not a per mode lookup.** Auto-detecting per mode is the right end
state and the wrong thing to build first: it makes every run a mixture, so a failure needs
bisecting before it can be read. A flag keeps a run entirely one engine or entirely the other.

**A7. The flag swaps the module list with the engine.** `HASH_TYPES` becomes `PM_MODES` alone, not
the union the perl side builds. The container families are tested by `test.sh` itself against a
real artifact and call no oracle, so running them under `-y` would spend hours on TrueCrypt and
VeraCrypt volumes to exercise an engine they never touch.

**A8. Python dependencies live in `tools/requirements.txt`.** `install_modules.sh` had five
`pip3 install` lines and a hand-written list of import names to verify; it now reads one file,
which is also where a contributor writing a module looks to see what is available. Nothing added,
nothing removed.

### A2 in practice

    ./tools/test.sh    -m 5200 -a 0 -t single -D 1 -f -V 1 -O
    OK   : 0/8 not found                        <- perl, having tested the pure kernel

    ./tools/test.sh -y -m 5200 -a 0 -t single -D 1 -f -V 1 -O
    Skip : no Optimized kernel for this mode

    ./tools/test.sh -y -m 1000 -a 0 -t single -D 1 -f -V 1 -O
    OK   : 0/8 not found                        <- m01000 has both families

`m05200` is the mode that shows it: `OpenCL/` carries only `m05200-pure.cl`. This is jsteube's
point on [#4813 ](https://github.com/hashcat/hashcat/pull/4813#issuecomment-5505612828) taken to the engine rather than to the oracles. `test.py` exits 2, `test.sh` turns
that into `record_skip` and leaves the mode out of the run rather than reporting `0/0`, which
would read as a failure rather than as a mode that was never applicable.

## The two modules

`m01000.py` sits beside the untouched `m01000.pm`, line for line, same `PW_CHARSET` branch, so the
two read side by side. It uses `Crypto.Hash.MD4`, because `hashlib` has no md4 wherever OpenSSL
ships without the legacy provider; `pycryptodome` was already a dependency.

`m05200.py` sits beside `m05200.pm`. Password Safe v3 earns the slot: a fixed 64 byte salt, the
optional `iterations` argument, a `module_verify_hash` that parses the artifact back, and no
optimized kernel. Its verify hook is better than the perl one in two ways. `m05200.pm` calls
`MIME::Base64::decode_base64` fully qualified with nothing loading the package, so it dies on any
call, and it regenerates the random tail, so the rebuilt hash could never equal the input. Here
the tail comes back off the artifact and the hash rebuilds byte for byte.

## Testing

    mode   engine  kernel  result
    1000   perl    -P      OK   : 0/8 not found
    1000   perl    -O      OK   : 0/8 not found
    1000   python  -P      OK   : 0/8 not found
    1000   python  -O      OK   : 0/8 not found
    5200   perl    -P      OK   : 0/8 not found
    5200   perl    -O      OK   : 0/8 not found
    5200   python  -P      OK   : 0/8 not found
    5200   python  -O      Skip : no Optimized kernel for this mode

    ./tools/test.sh [-y] -m <mode> -a 0 -t all -D 1 -f -V 1 <-P|-O>

Both engines were also run over the same mode and the same scope and their verdicts diffed:

    ./tools/test.sh    -m 1000 -a all -t all -D 1 -f -V 1 -P
    ./tools/test.sh -y -m 1000 -a all -t all -D 1 -f -V 1 -P

17 attack and target combinations each, identical verdicts, no differences. That covers -a 0, 1,
3, 4, 6, 7, 8, 9 and 12 in both single and multi.

`./tools/test.sh -m 0` still passes, `tools/test_coverage.sh` reports the same modes covered as
before, and `tools/test.py` handles its own entry points: `single`, `passthrough` and `potthrough`
work, `edge`, `password` and `verify` say they are not implemented and exit 1, a non-numeric mode
and a missing argument both exit 1.

## Deviations

`tools/test_edge.sh` is untouched and still calls `test.pl`. `-y` is a `test.sh` flag only.

`edge`, `password` and `verify` are not implemented in `test.py`. `edge` is what `test_edge.sh`
needs and `password` is what `-g` needs, so both follow when those callers do.

`test.py` generates numeric passwords. The non-ASCII generator in `test.pl` is about 150 lines
that scrape `src/modules/*.c` and `OpenCL/*.cl` to decide which modes may have one, and that is
policy rather than plumbing. It is the next step, and it is what makes `m01000.py`'s `PW_CHARSET`
branch mean something: until then a numeric password decodes identically as latin1 and as utf-8,
so the branch is carried for parity but dormant.

`test.py` is 261 lines rather than the 80 the plan estimated. The constraint handling and the
length pool are most of it.

The length pool reproduces `test.pl`'s `sort { length $a <=> length $b }`, which orders integers
by digit count. That looks like a bug and is not one: `single()` sorts the output again at print
time by the length of the actual password ([test.pl:541](tools/test.pl#L541)), and that sort is
numeric, so the pool order is overwritten before anything sees it. Measured over 12 runs across
both engines, the emitted lengths were non-decreasing every time. Reproducing it costs nothing and
keeps the diff free of a change with no observable effect.

No GPU here, so everything above is the CPU backend, `-D 1 -f`, `-V 1`.

_attached [plan_testpy.md](https://github.com/user-attachments/files/31769916/plan_testpy.md) which I prompted thi implementation with_


thatux:
Not sure if it makes sense to try and port all the older modules to Python (how to guarentee Python-test is the same as the Perl one?). Perl and Python can co-exist. But for show I've ported m1000 and refactored [m5200](https://github.com/hashcat/hashcat/blob/master/tools/test_modules/m05200.pm) as that is pretty much Python already (in a Perl wrapper).

I've tried to keep it MVP for now to allow for discussion / suggestions.

jsteube:
I see this is still in draft, but one note already. That stuff:

```
 word = pack_if_HEX_notation(word)
```

$HEX support was added shortly after the first test.pl was introduced, and these lines are not in all existing test modules to maintain this. But with this refresh, there's no more need to keep it int the module. If you see a design that would allow having this handled automatically, that might be worth a deviation from the existing layout.

jsteube:
The part I would preserve regardless of the remaining changes is the refusal to substitute constraints. If a mode has no kernel for the family requested by the run, we should report that directly instead of quietly assigning the wrong label.

I would move module_verify_hash higher on the priority list. The engine requires the hook, but nothing currently calls it, so every new module includes validation code that has never run. This is not test only plumbing. Contest scoring engines, and any other consumers that accept submitted hashes, rely on it to decide whether a hash is valid.

The Python files should also use 2 space indentation. CONTRIBUTING.md requires it, tools/cryptoloop2hashcat.py already follows it, and the first rule in that section is to match the surrounding code.

thatux:
Tnx, both comments should be addressed.

### $HEX in the engine 7a1de064c

Agreed, and it is out of the modules entirely. `$HEX[...]` is the password and the password is the
last field of a crack line, so `test.py` unwraps it before a module sees the line, without knowing
any module's hash format. Neither `.py` module mentions it, and the helper is not in
`test_helpers.py` either, so there is nothing to copy into the next one.

Your point about it never being maintained is measurable: 490 of the 511 `.pm` files carry the
call and 21 do not.

The outfile keeps the line hexified as hashcat wrote it, since that is what the caller compares
against. Only what the module is handed is unwrapped.

### module_verify_hash ba843dfce

You are right that it was dead. `verify` was already implemented in that same commit, but that
only matters if something calls it, and nothing in a suite run does.

So `single` now feeds every hash:password pair it is about to print back through
`module_verify_hash`, and refuses to print any of them if the module cannot reproduce its own
output. Sabotaging m01000.py's verify to return the wrong word:

    python3 tools/test.py single 1000
    module_verify_hash did not round trip module_generate_hash for mode 1000
    rc=1, nothing on stdout

Before, it printed eight vectors and said nothing.

Cost is one extra `module_generate_hash` per vector, since verify re-derives the hash to compare.
On a slow KDF that doubles generation time. If that matters it wants a flag rather than removal,
tell me which you prefer.


### 2 space indentation 7c4a51453

Done, all four files. I had used 4 out of python habit, no reason. Split into its own commit so it
stays reviewable: `git show -w` on it reports 0 non-whitespace changed lines, and both modules
produce identical output before and after.

### Constraints

Unchanged, and it stays. `-O` on a mode with no optimized kernel reports Skip rather than testing
the pure kernel under the wrong label.

thatux:
Some questions that'd guide the development direction of this:

1. Are we we going to let these two exist side-by-side or are we going to port everything to Python?
2. Are we choosing 100% Python? Quite a lot of logic is currently in test.sh; pytest is often used in Python projects. It can produce nicer output + we get all kinds of helpers that can assist us. For example, we can create a fixture for a test that does something like @isnvidiagpuavailable over your function.
3. If we chose 100% Python? How / who is going to migrate all the tests?
4. If we want to run them dual, how we want to do this to avoid having to maintain duplicate code bases?

I think I'd prefer to make one test.py instead of depending on test.sh, and leave test.sh and test.pl and the .pm be for the time being, and prefer new tests to use the .py version.

jsteube:
Points 1 and 4 have the same answer: move every oracle to Python and keep exactly one oracle per mode. When m01000.py is added, m01000.pm should be removed in the same commit, so nothing has to keep two implementations consistent and there is no duplicated code to maintain.

Point 3 does not require moving all 500 modules at once. A module can be converted whenever it is touched for another reason, and test.pl can check whether the Python version exists and say so when it does. The only sequencing requirement is that the code which selects the oracle engine for each mode must land before the first Perl oracle is removed, otherwise test.pl could silently stop covering that mode.

I would start with the 21 modules whose oracles invoke Python through a shell, 15 of which interpolate the password directly into the command line. Those modules are already incorrect rather than merely at risk.

Point 2 is also why I prefer moving the control logic to Python instead of keeping part of it in a shell script. @matrix @philsmd, the question is whether the shell script should continue to own any of this path.

The oracle side is wrong because a password crosses a shell boundary and can be expanded as shell text. The test.sh side is wrong for the same reason in the opposite direction, since it treats a password as shell text: the length it measures is a byte count and cut -c works on bytes, where the test means characters. That is what produced #4815 and #4816. Python can keep the password as bytes from end to end, which prevents both classes of failure. These are two forms of the same underlying defect, so I do not see an advantage in retaining the shell implementation for the control path.

I would still split the work into two changes, with the oracle engine first and the runner second, so each can be reviewed independently.

thatux:
To be sure we're on the same page what the 'oracle engine' and which is the 'oracle runner' is.

`test.py` will replace `test.sh` and be the oracle engine which will call
`generate_test_hash.py` (being the oracle runner replacing `test.pl`) which calls `m01000.py`


Replacing all this in one go, and removing the orginal files might end up being difficult to compare the two implementations, but I'll try.

I'll revert the changes in `test.sh` and make the Python tests completely seperate then.

jsteube:
Wait I think the plan was completely misunderstood.

- test.pl is the oracle engine: it loads mNNNNN.pm and calls module_constraints, module_generate_hash and module_verify_hash
- test.sh is the manager: it builds the dictionaries and the container families, picks devices, invokes hashcat and compares the result

Your test.py is already the test.pl replacement, not the test.sh replacement, and its interface says so, since "python3 tools/test.py passthrough 1000" mirrors "perl tools/test.pl passthrough 1000". So generate_test_hash.py would be a new file doing the job test.py already does. You do not need it.

Nothing is replaced in one go. The unit of migration is one hash mode: add mNNNNN.py, delete mNNNNN.pm in the same commit, and the engine for that mode follows from which file exists. Start with 2 or 3, and the .pm files that are still there stay untouched and keep working through test.pl.

That also removes the comparison problem you are worried about. You never have to show that a python oracle and a perl oracle agree, because after the commit only one of them exists for that mode.

Whether test.sh itself eventually becomes python is a separate question and it does not block any of this.

thatux:
Tnx, that's why I double checked.

IMHO test.pl should then be named test_module_runner.py; test.sh and test.pl always confuse me..

Moving test.sh to python has some large benefits as you suggest (no shell interpretation). But also quite some implications, bigger commit changing more stuff. Will do that later, first have this land.

jsteube:
I think you meant this:

test.pl should be named test_module_runner.pl
test.py should be named test_module_runner.py

If so: OK. Make sure to update all existing callers and documentations. Feel free to send in a separate PR for this if it blocks you.

test.sh: rework -> later

thatux:
exactly. I've renamed test_module_runner.py in this PR. I think this PR is OK to merge?

or do you want me to rename test.pl in this PR? otherwise I'd opt to merge this and I'll try to rename test_module_runner.pl in a separate PR.

jsteube:
Merged. The rename of test.pl to test_module_runner.pl as its own pull request is the right call, and you did not need to hold this one for it.

For the one after that, I would take the modules that shell out. Fourteen are left:

```
8800 9000 14500 17010 17020 17030 17040 17050 31100 32500 32700 33400 34300 34301
```

Every one of them runs python3 through a shell to produce the hash, with the password on the command line, so a password carrying a shell metacharacter is mangled before it is ever hashed and the oracle then disagrees with hashcat for a reason that has nothing to do with the mode. Porting those removes the shell from the path rather than only changing the language, which makes them worth more per module than the rest. You have already done one of them here, 5200, so the shape is set.

The two dependencies installed from git in tools/requirements.txt are worth keeping in view while you work through those. Once the modules that need pygost and pycryptoplus are on python we can point them at whatever the maintained package is now, instead of carrying a fork of a project that last released in 2015.

PR #4819, opened 2026-09-03
Add kwallet

jessevz:
Added kwallet as a module. Most code has been vibe coded but verified by me


jsteube:
The hash-mode 36400 and 36410 are available on master and across all open PRs. The kernels and modules look sound. The build is clean, both self tests pass, and the new BF_DECRYPT macro in inc_cipher_blowfish.cl does not modify any existing code, so bcrypt and the other Blowfish modes remain unaffected.

The blocking issue is m36400.pm, which currently fails on every invocation:

```
$ perl tools/test.pl single 36400
FATAL: iv must be string/buffer scalar at Crypt/Mode/CBC.pm line 14
```

Passing substr directly as the IV gives CryptX a magic lvalue scalar that its XS layer rejects. Store the IV in a regular scalar before passing it, both in the generation path:

```
my $iv = substr ($plain, 0, 8);

my $ct_sw = $iv . $cbc->encrypt (substr ($plain, 8, 56), $key, $iv);
```

Mode 36410 does not encounter this because ECB does not use an IV.

There are also two minor cleanup items. The branch changes an unrelated line in docs/changes.txt from "decoded from the start" to "decoded from the star". 

The oracle issue is the only substantive blocker.  Everything else is in good shape.

jessevz:
> The hash-mode 36400 and 36410 are available on master and across all open PRs. The kernels and modules look sound. The build is clean, both self tests pass, and the new BF_DECRYPT macro in inc_cipher_blowfish.cl does not modify any existing code, so bcrypt and the other Blowfish modes remain unaffected.
> 
> The blocking issue is m36400.pm, which currently fails on every invocation:
> 
> ```
> $ perl tools/test.pl single 36400
> FATAL: iv must be string/buffer scalar at Crypt/Mode/CBC.pm line 14
> ```
> 
> Passing substr directly as the IV gives CryptX a magic lvalue scalar that its XS layer rejects. Store the IV in a regular scalar before passing it, both in the generation path:
> 
> ```
> my $iv = substr ($plain, 0, 8);
> 
> my $ct_sw = $iv . $cbc->encrypt (substr ($plain, 8, 56), $key, $iv);
> ```
> 
> Mode 36410 does not encounter this because ECB does not use an IV.
> 
> There are also two minor cleanup items. The branch changes an unrelated line in docs/changes.txt from "decoded from the start" to "decoded from the star".
> 
> The oracle issue is the only substantive blocker. Everything else is in good shape.

All issues should be resolved :)

PR #4825, opened 2026-09-05
pcfg: large rulesets in far less memory and a much faster start

matrix:
The pcfg feed cannot load a large ruleset. It keeps every row of two per-structure tables for every
structure at once, and on a grammar of 27770870 structures the kernel kills the run at 59.8 GiB,
with the ruleset already read and the feed still starting. This keeps the row the rest of the run
reads, rebuilds the others into a per-thread scratch buffer, loads and builds in parallel, and
caches what the first run worked out.

Deviation from CONTRIBUTING.md, as AGENTS.md asks for: this is more than one problem per pull
request. Memory, loading, the cache, the walk, threading and messages are one change to one file,
because the memory work reshapes the tables and the rest load, store, walk and report that shape.
Splitting them would mean rewriting the same structures once per pull request.

### Memory

There are two tables per structure, the suffix counts and the device engine's unit tables, and only
their first row is read outside the unranking. Row zero is kept and the rest is rebuilt into a
per-thread scratch buffer for the one structure a thread is taking apart. Positions inside a
structure are contiguous, so the rebuild happens once per structure rather than once per candidate,
and the buffer is reused. A structure's slots were three arrays of PCFG_MAXSLOT entries each, where
the average structure uses far fewer, and now come out of an arena at the size it turned out to need.

    /usr/bin/time -v ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable --runtime 1 -d 1

Maximum resident set size, highest of three runs a side, on one Linux machine at twenty threads:

    shipped ruleset,     23154 structures    2.90 GiB  ->   1.38 GiB
    trained grammar,     93958 structures    7.29 GiB  ->   1.69 GiB
    trained grammar,  27770870 structures    killed at 59.8 GiB  ->  33.7 GiB

with the same 23154 structures and the same 12747516022634 base words counted on both sides of the
first. A peak measured this way carries the backend runtime with it, so the figure to read is the
pair and not either half of it. The run is bounded because the peak is reached while the feed
starts, before the first candidate leaves it.

### Loading

The terminal lists a grammar names are gathered in one cheap pass and read in parallel, rather than
one at a time from inside the parse. Resolving a token to a list was a search over every list loaded
so far, once per token of every structure, and is a lookup in a (type, length) map now. The two
level indexes are built a level at a time in parallel, which each level allows because it does not
look at any other. The probe that chooses maxword and kbits produces nothing, so it works on the
head of the grammar, which ranks the configurations the same way; the real build after it is forced
only when the probe did use a sample. Ordering a terminal list by length inside each run of equal
cost is a counting sort, and it was written as a comparison sort over the whole run.

The unit tables are the longest step the start has. The row the recurrence reads from is nearly all
zeros, and the search for the few costs that are reachable sat inside the bucket loops, so every
pair of buckets walked the whole span again to find the same handful; they are gathered once per
slot instead. Bucket costs come out in ascending order, so a pair that costs more than the span
means every pair after it does too, and the search stops there. And sat_add () tested for the carry
with a branch, inside a loop these tables run hundreds of billions of times. Together they take that
step from 59.6s to 42.4s on the 27770870-structure grammar, medians of two runs each, with the
candidates unchanged.

Every step prints what it took, so the figures below are read off an ordinary run:

    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable --runtime 1 -d 1

None of this shows on the shipped ruleset, which loads in about two seconds either way and is a
fifth of a second slower here: on a grammar that small the gathering and the maps cost more than
they save. It shows where the load is long. On the 27770870-structure grammar, PCFG_BUILD_THREADS=1
against the twenty threads the machine has, medians of two runs each:

    terminal lists      38.6s  ->  16.9s
    suffix tables       56.9s  ->  10.4s
    host level index     9.0s  ->   3.0s
    device level index   8.6s  ->   2.7s

PCFG_BUILD_THREADS is the number of workers to create and every stage also runs the work function on
the calling thread, so the left column is a two thread baseline rather than a single threaded one.

### The cache

Those tables are a function of the ruleset and of a few numbers, so the second run on a ruleset can
read what the first one worked out. They go under hashcat's cache directory, in a pcfgdbs folder
beside the seekdbs one the wordlist feed keeps.

The key is what the tables were built from, hashed as the ruleset was parsed: every terminal list's
bucket costs, starts, counts and lengths, its entry offsets and its order flag, and every
structure's cost, span, slot kinds, list handles and slot lengths. An edit that moves an entry to
another cost bucket moves the key, wherever in the ruleset it is. An edit that leaves the buckets
alone does not, because the tables it would rebuild are the ones already on disk; the terminal text
is deliberately not in the key, since the pool is read fresh on every run and the counts do not
depend on it. Nothing is read twice for this, since the ruleset is read in full anyway: on the
shipped ruleset the load is 0.67 to 0.69 seconds either way, and on a 16 GB grammar of 27770870
structures 109.86 seconds against 106.82, for the same 252137942098230 base words.

A cache that answers with the wrong tables does not fail, it enumerates something else and looks
like it worked, so the header carries every input and all of it is compared: the ruleset, scale,
costmax, maxword, kbits, PCFG_BUCKETCAP, PCFG_LENSPLIT, the structure count and whether the run is
varlen. It is written to a temporary named after the process and moved onto the real name, with the
old one unlinked first because rename refuses an existing target on Windows.

What is stored is the row the rest of the run reads, and on a large grammar 95 of every 100 of its
counts are zero, so a row goes down as how many counts it has and then each one as a position and a
value in the bytes it needs. On the 27770870-structure grammar the file is 130 MiB against a ruleset
of 16 GB, and the step it replaces goes from 42.4s to 2.1s. The second run says "unit tables read
from cache" where the first said "unit tables built", and cache=0 turns the whole thing off and does
not create the folder:

    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable --runtime 1 -d 1
    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable --runtime 1 -d 1 cache=0

make clean removes pcfgdbs and .gitignore covers it. There is no --pcfgdb-path and nothing prunes
the directory; both would touch files outside the feed, so they are left for a separate change.

### The walk

This is host side work, and it decides the rate on the device engine, where one call has to place a
base word for every cell the device is given. On the host engine every candidate is built here
anyway and the walk below is barely entered.

advance_unit () could only step the last token of a cell and gave up when that token was spent,
which on the device engine is nearly every candidate, because the last token belongs to the device
and its bucket seldom has a successor that fits the same budget. It carries now: the last token that
can move is stepped and everything behind it is laid out again from its first candidate, the way an
odometer turns the digit before the one that ran out. Where the token that opened devmode is the one
that is spent, the carry enters the host half of the same bucket at rank dblk with devmode closed,
before the bucket advances. The placement both it and unrank_unit () need is one function now, so
there is one walk rather than two that have to agree, and most base words are carried rather than
unranked from the start. bitlen () found the width of a bucket by shifting a one along until it
passed, a step per bit, in the innermost thing the walk does; the count of leading zeros answers it
in one instruction, and it also terminates, where the shift ran past 64 for anything above 2^63,
which sat_mul () can produce because it saturates.

--stdout runs on the host engine, so it cannot reach any of this. What shows it is a device run over
a window wide enough to carry, checked against digests the host engine produced:

    ./hashcat --stdout -a 4 pcfg/default-passwords.tar.xz --limit 3000000 | python3 -c 'import sys, hashlib
for l in sys.stdin.buffer:
    l = l.rstrip (b"\n")
    if l: print (hashlib.md5 (l).hexdigest ())' | sort -u > walk.hash

    ./hashcat -m 0 -a 4 walk.hash pcfg/default-passwords.tar.xz --potfile-disable -s 0 -l 50000000 -d 1 --status

    master     Exhausted, rc=1   138835703245 candidates   2520633/2790759 recovered   0 rejected
    this PR    Exhausted, rc=1   138835703245 candidates   2520633/2790759 recovered   0 rejected

The two agree on the candidate count, on the digests they reach and on the rejects, over a window
the walk carries through many times.

### Threading

The feed carried its own: eighty-two lines naming a mutex, a condition variable and a thread twice
over, once for Windows and once for pthread. types.h already carries those three types, one set per
platform, and thread.h carries the operations on them, so nothing is linked for them and Windows
gets its own API rather than a pthread it does not have; the processor count the feed worked out by
hand is hc_get_processor_count (). The wrappers are gone, and PCFG_BUILD_THREADS reaches every
parallel step, where the terminal list load answered only to the processor count.

### Messages

Every step of the load reports what it took, and the long steps report progress while they run: a
grammar that takes minutes to read said nothing at all before, which from outside is a hang. Every
one of those lines asks --quiet, which the feed is handed in its own context; the check was written
out at eleven of the forty-two sites and left out at the other thirty-one, so most of the load
talked through --quiet and through --stdout, which turns --quiet on.

### PCFG_MAX_STRUCTS

Stops the parse after that many structures. grammar.txt is written most probable first, so this
keeps the head of the distribution and drops its tail, which is the trade anyone makes by hand when
a ruleset outgrows the machine.

### Fixes

Two, both defects in master.

OMEN levels run from most probable to least, so level 0 holds the best guesses the escape has. Both
passes over them started at one, and the test for a model with no levels read the highest level
seen, which cannot tell a file that named none from one that named only level 0. The level was
dropped, and dropped in silence. No ruleset we ship names level 0, which is why the omission never
showed; a trainer that does name it was losing the guesses its model rates above all the others.

A grammar can put its whole mass on the escape, and then it holds no structures: one M line and
nothing else, which is what a ruleset trained without coverage looks like. The load refused it on
the count of structures alone, before the escape had been read. Structures and M lines are two ways
for a grammar to carry mass, so it is empty only when it has neither, and such a ruleset loads and
enumerates now. A grammar whose structures were all dropped, whether because a terminal list could
not be read or because every structure is priced above costmax, still fails and says which, and so
does one with no structures and no escape, which is what omen=0 or a missing Omen directory leaves:
neither engine has anything to enumerate there.

    mkdir -p /tmp/esc && tar xf pcfg/default-passwords.tar.xz -C /tmp/esc
    printf 'M\t1.0\n' > /tmp/esc/ruleset/Grammar/grammar.txt

    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 /tmp/esc/ruleset --potfile-disable --runtime 5 -d 1 --status

    master    rc=255  /tmp/esc/ruleset: no usable structures
    this PR   rc=0    OMEN escape carried, 15 levels over 1 model, 26386266585 guesses
                      no base words for the device engine, the host engine takes the run
                      Status Cracked, Progress 100290816/26386266585, 0 rejected

Two things had to move for the host engine to take it. The escape is dropped when the device engine
is on, because that engine cannot walk a trellis, and that is decided while the grammar is read; a
grammar with no structures gives the device engine no base word either, so the run will move to the
host engine and the escape is the only thing left to carry. And global_dev_init () refused a pool
with no terminal entries, where an empty index is reported as an empty inner loop instead and the
core moves the run, which is the answer #4827 gives.

### Checked

The candidates are unchanged. The first 20000 and the first 500000 candidates of a run hash to the
same digest as master's, for the shipped ruleset and for a trained one of 93958 structures, and the
shipped ruleset reports the same 23154 structures and the same 12747516022634 base words on both
sides.

    ./hashcat --stdout -a 4 pcfg/default-passwords.tar.xz --limit 20000  | md5sum
    ./hashcat --stdout -a 4 pcfg/default-passwords.tar.xz --limit 500000 | md5sum

    20000    43b848a00665a0e0220e64760a247760   on both
    500000   52c6693e92e780700c5766cc7662fc98   on both

The suite runs with -K 0 because -a 4 has no optimized kernel: generic.c refuses -O rather than
ignoring it, since the hashes were parsed under the flag on the way in, so the full suite reports
that refusal on master as well.

    ./tools/test_edge.sh -m 0 -K 0

    [ test_edge_1788707179 ] > All tests done in 0d:00h:02m:08s
    [ test_edge_1788707179 ] > Errors detected: 0

It builds on Linux and macOS in both the static and the shared arrangement, and cross compiles for
Windows, with the same six warnings master already has and none of them new.


jsteube:
The PR was reviewed in depth, by reading it and by running it against master, and every issue below is reproducible. The parts of the PR not mentioned here were tested as well and showed no problem, so this reply is limited to the defects and review issues that remain.

Finding 1/19, the project rules. This PR is much broader than the one problem rule in CONTRIBUTING.md allows, because it rewrites memory handling, loading, the walk, threading, progress reporting, and messages, adds a cache and an environment cap, and names six fixes. No issue was opened for any of those fixes. There is no test_edge.sh output even though the PR changes status.c and hashcat.c, which are not pcfg only files. AGENTS.md asks for one sentence naming any rule that could not be followed, and there is no such statement, so none of this is acknowledged. The description is nearly 3000 words, while the commit message body is empty, so after the merge commit the log carries only the subject and every reason for the change lives here.

Finding 2/19, the CI is red. All four Shared jobs of the PR's own CI run fail, while all four Static jobs pass, and the description states that it builds in both arrangements. The build itself compiles. What fails is the package self check, which holds every plugin in feeds/ to one of two exact export lists at tools/test_package.sh line 255, one for a host only feed and one for a feed that also runs on the device. The PR adds global_dev_span to feed_pcfg.so without adding it to either list, so the feed now matches neither:

```
      feeds/feed_pcfg.so exports: GENERIC_PLUGIN_OPTIONS GENERIC_PLUGIN_VERSION global_dev_init global_dev_span global_init global_keyspace global_term thread_init thread_next thread_next_dev thread_seek thread_term
FAIL  1 of 6 plugins in feeds/ do not export exactly: GENERIC_PLUGIN_OPTIONS GENERIC_PLUGIN_VERSION global_init global_keyspace global_term thread_init thread_next thread_seek thread_term
```

That check runs only where a core library exists, which is why the static arrangement is unaffected. It reproduces with bash tools/test_package.sh --no-device . in the build directory, and master passes it.

Finding 3/19, code style. The added code has several issues against CONTRIBUTING.md: two function signatures and one call are wrapped across lines, five locals made read only by the refactor did not gain const, three trailing comments were added to a file that had none, five returns contain a computation, *p++ was copied into a new parse loop, and two new comments join clauses with a semicolon.

Finding 4/19, the device engine walk. The device engine no longer produces the candidate range described by its own cost index. A foldable token's bucket is stored by place_from() as two adjacent blocks at line 6156, with the device partitions first and the host entries after them, but advance_unit() has no transition from the device half to the host half of the same bucket after carrying across tokens. A terminal is advanced only when th->tdev[ti] is false at line 6665, so the token that enabled devmode never takes that path and instead reaches the bucket advance at line 6713, skipping the host half of the bucket it is leaving.

Expanding each cell as the kernel does for 300 units of example.dict starting at unit 311700 gives:

```
build    candidates   duplicates
master   23,429,423   0
PR       33,586,001   11,907,096
```

More than a third of the device work in that region repeats candidates that were already hashed, while the candidates assigned to the skipped host partitions are never produced. Across 3000000 units, the walk and unrank paths disagree at 1088 positions for example.dict and 461 for rockyou-1M, while master disagrees at none. A single device run uses the walk, while --skip, restore, and multi device splitting use unrank, so the same command now describes a different attack depending on how the work is divided, which breaks the guarantee the cost index is meant to provide.

The two md5sum commands in the description cannot expose this defect because generic.c clears dev_enable for --stdout at line 399, sending both commands through the host engine while the affected walk exists only in the device engine. The missing transition belongs in advance_unit() before the bucket step: once the token that opened devmode is exhausted, the next unit must start at rank dblk in the same bucket with devmode disabled and the tail rebuilt from rank zero. With that transition, the walk agrees with unrank on both grammars and the expanded stream matches master byte for byte.

Finding 5/19, the cache key. The cache key can match a different ruleset and return its unit tables. Recursion into subdirectories happens in pcfg_ident_dir() only when scan_directory() returns a directory, but scan_directory() never does because folder.c opens every entry at line 297 and keeps only the entries that fail to open, leaving the recursion at line 1219 unreachable and reducing the identity to top level files. A pcfg_cracker ruleset keeps its grammar in the Grammar, Alpha, Digits, Capitalization, Context, Keyboard, Other, Years, and Omen subdirectories, so none of those files contributes to the key, which leaves only config.ini in the rulesets checked here.

Two copies of the same ruleset that differ only in Alpha/5.txt, while keeping the same config.ini and structure count, produce:

```
populated from the first     317,645,881,895 base words
second, cache on             317,645,881,895 base words
second, cache=0              316,128,734,899 base words
```

The second ruleset therefore reuses the first ruleset's unit tables and reports a keyspace that is 1,517,146,996 base words too large, without any warning. Editing a terminal list in place is the normal way to work on a ruleset, and doing so leaves config.ini unchanged.

Finding 6/19, the cache on an archive. The cache is disabled for archived rulesets because their identity is zero. The identity is computed by pcfg_ident_roots() at line 7628, before arc_load() runs at line 7661, so roots[i].arc is still NULL at that point and the archive path at line 1250 cannot run. Identification then falls back to pcfg_ident_dir() on the archive as a plain file, which fails and causes pcfg_cache_path() to return NULL. A directory with no loose files at its top level produces the same zero identity, which is the layout created when the shipped ruleset is unpacked. Every reproduction command in the description uses that ruleset, and none of them creates or reads a pcfgdbs directory, so the reported improvement from 42.4 seconds to 2.1 seconds cannot be reached by any command shown there.

Finding 7/19, a stack overflow at scale=2. Using scale=2 writes beyond a stack buffer. The reachable costs for one row are stored by build_unit_rows() in u32 nz[PCFG_COSTCAP + 1] at line 5278, which provides 65 entries, and one entry is then written for every nonzero cost across span = pg->costmax - s->cost + 1. Because pg->costmax is costmax * scale at line 7556 and both parameters accept values from 1 through 64, the span can reach 4096 without any bound on the write.

```
./hashcat -m 0 -a 4 pcfg/default-passwords.tar.xz scale=2 --keyspace

before: 3928969384847, rc=0
after:  *** stack smashing detected ***: terminated, rc=134
```

A crack run aborts in the same way, and the stack protector is what turns this into an abort instead of silent corruption in six worker threads at once.

Finding 8/19, the progress total. The span function is not monotonic, and the status calculation can underflow when it subtracts two values returned by global_dev_span(). On the shipped ruleset, span(0, 1000010000) is 2,480,559,523,464 while span(0, 1000020000) is 2,479,368,244,706, which is a drop of 1,191,278,758 as the base word increases. One of those values is used as the total by status_get_progress_end() and the other as the skipped amount by status_get_progress_skip(), with no guard around the subtraction.

```
./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable -s 1000010000 -l 10000 -d 1 --status

before: Progress 2739133834276/2739054790000 (100.00%)
after:  Progress 106444276/18446744072518272858 (0.00%)
```

The value 18446744072518272858 is 2^64 minus the same drop. The cause is that the span walk adds a device order block prefix to an offset that pcfg_rank() measures in host order, even though the two orders do not agree inside a block.

Finding 9/19, the quoted residual. The 0.003% residual in the description does not describe the error a user sees, because it refers to a cumulative count while the status line displays the difference between two cumulative counts, and that error has no bound. For example, span(0, 2000000) is only 0.0096% above the walk, but over windows of 20000 base words the ratio of span to emitted candidates is 6.63 at base word 10000000, 2.15 at 500000000, 0.51 at 700000000, and 0.00 at 1000000000, where the window still contains 230 million candidates. A run with --skip 500000000 --limit 20000 reaches Exhausted at 46.48% of its own total, while master reports 100.00%.

Finding 10/19, weights in the cache key. The weights= setting reaches neither the cache filename nor its header. Each root's weight is folded into its hash by pcfg_ident_roots() at line 1259, but the identity is computed at line 7628 before root_weights() parses weights= at line 7630, so every root still has the default weight of 1.0 and different weightings share one cache file.

Finding 11/19, the cache row format. The cache format repeats the same assumption that a row has at most 65 entries. A row is written by pcfg_cache_save() as a one byte count followed by one byte per position, while PCFG_CACHE_ROW reserves (PCFG_COSTCAP + 1) * 11 + 1 bytes, so a span above 256 wraps both the count and the positions while leaving a payload whose checksum still validates and is accepted by the loader.

Finding 12/19, an unstated ordering requirement. The bucket loop in build_unit_rows() now breaks where master continued, which is valid only when b_cost is ascending for every terminal list. Nothing creates, verifies, or documents that ordering because it comes directly from the order of lines in the terminal file, so a list that is not written from most probable to least probable silently loses device coverage.

Finding 13/19, a data race. Parallel terminal list loading races on a lazily initialized global. The chain from preload_worker() reaches tlist_build() and then pcfg_lensplit(), which initializes the file scope pcfg_lensplit_state through an unsynchronized read followed by a write that every worker and the calling thread can reach on the first list.

Finding 14/19, the fix list. Three of the six listed fixes do not repair defects in master. The pool size in master is already a multiple of 4, so rounding it has no effect. A terminal list that fails to load but still contributes to lists_cnt is a defect introduced by this PR, because the new parallel preload derives lists_cnt from the number of names rather than the number of successful loads. The unit scratch buffer with an untested second dimension is also new here, because master allocates and releases that storage per structure instead of reusing one buffer. This distinction matters because the fix list is how a reviewer determines the purpose and scope of the PR.

Finding 15/19, the emptiness test. The new emptiness test turns a load error into silence. A grammar is now accepted when it contains only an M line, which also accepts a grammar whose structures were all discarded because their terminal lists could not be read, while master rejected that input and named the file it could not open.

Finding 16/19, a clamp that never fires. The clamp added to status_get_progress_end() never fires. Its comment treats --limit as the point where a run ends, but user_options_preprocess() at user_options.c:2958 has already applied limit += skip before the feed sees it, so words_limit cannot be below words_skip there.

Finding 17/19, the thread count. PCFG_BUILD_THREADS=1 uses two threads rather than one. The count returned by pcfg_workers() is the number of worker threads to create, and every parallel stage also runs the work function on the calling thread, which is why the default is one less than the processor count. The loading figures described as single thread results are therefore two thread baselines.

Finding 18/19, the probe rounds. Probe rounds are cached despite the description saying they are not. The cacheable flag is pg->probe_n == 0 at line 5851, and probe_n is set only when the probe operates on a sample, so a grammar small enough to avoid sampling writes every configuration tried by the probe. One run with example.dict creates six cache files this way.

Finding 19/19, the cache directory. The pcfgdbs directory is placed like kernels but handled differently in four important ways. make clean removes kernels at src/Makefile:1586 but leaves pcfgdbs behind. The directory is created by pcfg_cache_path() during every path calculation, including a load that is about to miss, while kernels is created once in folder_config_init(). There is no --pcfgdb-path option even though the comment at line 5541 names seekdb.c as the model, seekdb provides such an option, and test_package.sh uses it to keep a seek database out of the package under test. Nothing prunes pcfgdbs, so unlike the kernel cache, which is bounded by modes and devices, this cache grows by one file per ruleset and configuration while each probe can add several files for one ruleset.

matrix:
All nineteen are correct. I reproduced the ones carrying numbers and checked the rest against the source. The changes are a delta on this PR's branch; after it the PR touches three files instead of seven, all under the feed.

1. Correct, and only partly answered. The progress work is gone entirely (8, 9, 16), so status.c, hashcat.c, generic.c, feed.h and types.h are untouched and there is nothing for test_edge.sh to cover. It is still more than one problem. The sentence AGENTS.md asks for is now in the description, and the commit carries a body.

2. Correct. global_dev_span went with the progress work, so the feed exports the device list again. `bash tools/test_package.sh --no-device .` now ends "every check passed".

3. Correct on every item, including the two comments, which were five.

4. Correct, and the one that mattered. advance_unit () now enters the host half of a folded bucket at rank dblk with devmode closed before the bucket advances. Over 50000000 base words on the shipped ruleset, against 2790759 digests taken from the first 3000000 host candidates, three runs each:

```
                candidates emitted     digests recovered
master             138835703245              2520633
this PR            141543759115              2520630
PR plus delta      138835703245              2520633
```

You are also right that the two md5sum commands cannot show this, since generic.c clears dev_enable for --stdout. The description now uses the run above.

5. Correct. The identity walks the ruleset's subdirectories itself and requires Grammar to be there.

6. Correct, and it took two attempts: moving the identity after arc_load () was not enough, because it opened the ruleset with hc_fopen (), which decodes a compressed file and then cannot seek to its end. It reads raw bytes now. First run writes one file, second run says "unit tables read from cache", cache=0 turns it off.

7. Correct. The buffer is sized on costmax times scale. `scale=2 --keyspace`: 3928969384847, rc=0.

8. Correct. The span function is removed and the calculation is master's again, for the reason you name: the walk adds a device order prefix to a host order offset, so it is not monotonic and no guard repairs it. `-s 1000010000 -l 10000` goes from 106444276/18446744072518272858 to 2739133834276/2739054790000.

9. Correct, and withdrawn with the feature. `-s 500000000 -l 20000` goes from 46.48% to master's 100.00%.

10. Correct. The identity is computed after root_weights ().

11. Correct. Counts and positions are varints sized on the same span cap, and the version is bumped.

12. Correct. Each list carries a flag saying whether its bucket costs never decrease, and the early stop happens only where it holds.

13. Correct. pcfg_lensplit () is settled on the calling thread before any worker exists.

14. Correct on all three, and the item I most needed. Master already rounds the pool; the failing terminal list and the unit scratch are defects this PR introduces. The list now names two fixes.

15. Correct. A grammar whose structures were all dropped now fails and says a terminal list could not be read; only one with neither structures nor M lines is called empty.

16. Correct, gone with the progress work.

17. Correct. The worker count is the request minus one, so PCFG_BUILD_THREADS=1 is one thread.

18. Correct. The flag is the probing state, not probe_n, so only the final build is written: the run under 6 does four builds and leaves one file.

19. Correct on all four, answered on two: make clean removes pcfgdbs and .gitignore covers it. There is still no --pcfgdb-path and nothing prunes the directory. Both would touch files outside the feed again, so I would rather do them separately. Say if you want them here.

jsteube:
Thanks for the update. Everything checks out except the following.

Finding 1/8, the description. It was not updated. Three separate fetches return the same body as round 1, byte for byte, so it still carries the whole Progress section for a feature this PR no longer contains, still says under Checked that "Bounded runs end on their own total" where the command it offers now prints master's 2832947/13700000 (20.68%), still lists six fixes including the three that repair nothing in master, and still offers the two --stdout md5sum commands as the check, which cannot reach the device engine. The reply says the sentence AGENTS.md asks for is now in the description and that the commit carries a body. The description contains no such sentence, and all three commit message bodies are empty.

Finding 2/8, the cache loader accepts a file it should reject. The count varint at the head of each row is read at feed_pcfg.c:5800 with a break that sits in the body of the if that tests row and at < len, which is not a loop, so the nearest enclosing loop is the structure loop. It exits that loop and skips the row == false guard eight lines below, landing on the checksum test and returning true. The three breaks inside the inner k loop are correct and reach the guard. Only this one escapes. Ordinary corruption does not get through, because a damaged file fails the payload checksum, but a file crafted so the checksum still matches does. Overwriting eleven bytes at payload offset 1500 with 0xff makes cache_varint fail on its shift bound with at < len, and recomputing the header sum leaves the file otherwise valid:

```
example.dict, intact cache      unit tables read from cache, 317,645,881,895 base words
example.dict, crafted file      unit tables read from cache,      10,028,928 base words
example.dict, cache=0                                        317,645,881,895 base words
```

The fix is one token:

```
-      if (cache_varint (buf, &at, len, &cnt) == false) { row = false; break; }
+      if (cache_varint (buf, &at, len, &cnt) == false) row = false;
```

cnt is still 0 at that point, so the k loop runs zero times and control reaches the guard that every other failure already uses.

Finding 3/8, PCFG_BUCKETCAP reaches the unit tables and not the cache key. The header carries the ruleset, scale, costmax, maxword, kbits, the structure count and varlen, and the filename carries five of those. The bucket cap re-cuts every terminal list, which is what build_unit_rows sums over. It normally moves the probe's choice of kbits or maxword, and those are in the filename, which is what hides it. Pin them and it does not:

```
BUCKETCAP=64, kbits=20 maxword=16, cache=0      289,432,191,820 base words
BUCKETCAP=8,  kbits=20 maxword=16, cache=0    1,961,498,688,085 base words
BUCKETCAP=8,  kbits=20 maxword=16, cache on   read from cache, 289,432,191,820
```

Finding 4/8, the identity reads only the two ends of a file. pcfg_ident_file at feed_pcfg.c:1210 folds in the size, the first PCFG_IDENT_ENDS bytes and the last PCFG_IDENT_ENDS bytes, and that constant is 4096. Terminal lists are large, so almost none of one is in the key: Alpha/6.txt of the shipped ruleset is 9122236 bytes, of which 8192 are read. Changing one character in the middle of it, e-07 to e-03 on a single line, keeps the length and both ends identical and moves that entry to another cost bucket:

```
unpacked shipped ruleset, cache=0               12,747,516,022,634 base words
the same with one middle character changed, cache=0   12,734,410,419,324
the same with one middle character changed, cache on  read from cache, 12,747,516,022,634
```

Round 1 fixed which files are hashed. How much of each file is hashed is the half that is left, and editing a terminal list in place is the ordinary way to work on a ruleset.

Finding 5/8, the identity ignores the order the rulesets were named in. pcfg_ident_roots folds each root in with an addition, so two orders give one identity and share one file, while the tables behind it are indexed by a structure order that does depend on the order given:

```
example.dict rockyou-1M costmax=30, cache=0   24,689 base words
rockyou-1M example.dict costmax=30, cache=0   24,689 base words
example.dict rockyou-1M, populates the cache  24,689
rockyou-1M example.dict, cache on             23,604
```

Finding 6/8, a refusal that master still has was removed. A structure dropped because its cost exceeds costmax takes a bare continue at feed_pcfg.c:2701 and is never counted in dropped_t, so a grammar whose structures are all above the cap reaches the emptiness test at 2824 with dropped_t at zero, passes the M line exemption and loads. Master guards on the structure count alone and refuses:

```
costmax=2   master rc=255 "no usable structures"      this PR rc=0, no output
costmax=3   master rc=255 "no usable structures"      this PR rc=0, no output
costmax=4   master rc=0,  no output                   this PR rc=0, no output
```

From costmax 4 up both produce nothing and exit 0, so that shape is master's and not this PR's. What is this PR's is the two values where master still refused. The escape reaches the same test from the other side: on a grammar whose only mass is an M line, omen=0 -S gives Status Exhausted and Progress 0 here where master refuses.

Finding 7/8, two arrays outlived their reader. uls_cpref and ulvl_cpref are the candidate-prefix half of the device level index and global_dev_span was their only consumer. The progress work took the consumer and left the producer, so both are still declared at 304 and 305, allocated at 6068, 6140 and 6142, written at 6089, 6183 and 6192, and freed at 7980, 7985 and 7989, and nothing reads either one. Neither field exists in master, so this is the PR's own weight, and uls_cpref is byte for byte as large as uls_pref, the array the unranking does read.

Finding 8/8, one extra whole-grammar table build. The delta made pg->built = false unconditional at feed_pcfg.c:8295, so the early return in unit_suffix_build can no longer fire and the final build repeats the last probe round. Measured with cache=0, three runs each, round 1 against this one:

```
round 1   TRYALL 1.95 1.97 1.97s, 3 builds     shipped 2.02 2.04 2.04s, 3 builds
round 2   TRYALL 2.04 2.02 2.02s, 4 builds     shipped 2.86 2.05 2.85s, 4 builds
```

It is 0.06s on TRYALL and does not matter on any grammar to hand. It matters where the build it repeats is long, and a warm cache serves the repeat, so the cost falls on a cold cache and on cache=0.

matrix:
All eight are correct. Seven are fixed here, and the eighth is the description, which my last reply described wrongly as well.

Finding 1. Correct, and my reply made it worse: I wrote about the delta as though it had reached the pull request text. The title is corrected now; the description and the commit messages are not, so here is what those statements should have said. The rule I could not follow is CONTRIBUTING's one problem per pull request: this is memory, loading, the cache, the walk, threading, messages and two fixes. The fix list still names six; two of them repair defects in master, the OMEN level 0 omission and the escape only grammar, while the pool is already a multiple of 4 there, the failing terminal list and the unit scratch are defects this PR introduces, and the level ladder message is not a fix. The Progress section describes work no longer here, along with the 0.003% residual it quotes. The evidence for the walk is the 50000000 base word run in my last reply, not the two md5sum commands, which run on the host engine. And the PCFG_BUILD_THREADS=1 loading figures are two thread baselines, as you said under finding 17 of round one.

Finding 2. Correct, and the break was the whole of it. It is gone, so cnt stays 0, the loop under it runs no times, and control reaches the row check every other failure already uses. A cache file with eleven bytes overwritten is no longer read.

Findings 3, 4 and 5 have one answer between them. The identity is no longer a function of the files on disk. It is a function of what the tables are built from, hashed as it was parsed: every terminal list's bucket costs, starts, counts and lengths, its entry offsets, its order flag, and every structure's cost, span, slot kinds, list handles and slot lengths. That answers all three at once and without reading anything twice, since the ruleset is read in full on every run anyway.

  - The bucket cap re-cuts the lists, so it moves the buckets and moves the key. Pinned at kbits=27 maxword=16 on the shipped ruleset, cap 64 gives 13576326691338 and cap 8 gives 60431938249774, each keeps its own number on a second run, and the two leave two files. It is in the header and the filename as well, with PCFG_LENSPLIT beside it.

  - An edit in the middle of a terminal list moves that entry's cost, so it moves a bucket and the key. On a small ruleset built for this, changing one probability on line 100 of Alpha/3.txt while keeping the byte length takes the keyspace from 2 to 6, the cached run returns 6 rather than the old 2, and the directory holds two files.

  - The order two rulesets are named in reaches the key through the structure sequence, which is where it reaches the tables: merge_cmp () orders by probability and breaks ties by the order the entries were read. Two rulesets whose probabilities tie now produce two identities, 08daa6c975e92a93 and 6c0cf80f3abb8f63, where before both orders produced 116731b7ca5643f2.

  The terminal text is deliberately not in it: the pool is read fresh every run and the counts do not depend on it, so an edit that leaves the buckets alone does not throw away a cache that is still right. The cost is the reason to mention the whole thing: on the shipped ruleset it is not measurable, 0.67 to 0.69 seconds either way over three runs a side, and on a 16 GB grammar of 27770870 structures it is 109.86 seconds against 106.82, so 3 seconds on a load of nearly two minutes, with the same 252137942098230 base words on both sides.

Finding 6. Correct. A structure priced above the cap is counted now, and a grammar whose structures are all above it is refused, naming costmax. costmax=2 and costmax=3 exit 255 on both engines.

Finding 7. Correct. Both arrays are gone, with the accumulators that fed them and the second half of the per level buffer they were summed into.

Finding 8. Correct, and worth more than the 0.06s. The loader frees the tables only once the header matches, so a miss leaves what the caller had and the final build keeps it rather than making it again. With cache=0 the shipped ruleset builds three times rather than four, and the second run still reads from the cache. It also stopped a leftover of round one's finding 18: two runs of the same attack now leave two cache files, one per configuration, where they left three.

Checked after each of these: the shipped ruleset still reports 12747516022634 base words, the first 20000 and 500000 candidates hash to master's digests, and tools/test_edge.sh -m 0 -K 0 detects 0 errors.

The empty device index has been fixed in the separate pull request, #4827

jsteube:
The generic.c split is the right move, so the omen=0 route is out of scope here and #4827 is where it belongs.

On seekdb, it is the same technique but not the same issue, and the difference is the reason the pcfg one was a defect. A seek database stores the byte offsets of line starts, and the comment above seekdb_path says why sampling is sound for that: a byte that changes without moving a line ending leaves every offset correct. For a compressed source, where that does not hold, seekdb reads the whole file instead. The pcfg cache stores unit counts, and a byte in the middle of a terminal list that moves an entry to another cost bucket changes those counts, so the same sampling was not sound there. Keying on the parsed tables answers it properly and is a better answer than reading the files whole would have been.

What is left is the description, and it is the one thing a merge makes permanent. It still carries the Progress section for work that is no longer here, along with the 0.003% residual it quotes. Its Checked section still says bounded runs end on their own total, where the command it offers now prints master's 2832947/13700000 (20.68%). The fix list still names six, where two of them repair defects in master, two are defects this PR introduces, one is already correct in master and one is a message rather than a fix. And it still offers the two --stdout md5sum commands as the evidence for the walk, which run on the host engine and cannot reach the code that changed.

Those corrections are all in your last two comments and they read well. They just need to be in the description rather than under it, because the body is what is read on GitHub afterwards and every commit message on the PR lands in master through the merge commit. The deviation sentence AGENTS.md asks for belongs there too, naming one problem per pull request as the rule this one does not follow. With that done there is nothing else outstanding.

matrix:
The description is rewritten, and it now carries the corrections rather than the comments under it.

The Progress section and the 0.003% residual are gone with the work they described. Checked no longer claims a bounded run ends on its own total, and the two --stdout md5sum commands are the digest check they actually are; the walk is shown by the 50000000 base word device run instead. The fix list names two, both defects in master. And the deviation sentence AGENTS.md asks for is at the top, naming one problem per pull request as the rule this does not follow.

#4827 landed first, so this branch carries the merge. There is one conflict, the timing line here against the comment block there, and both sides are kept.

The other half of your note on #4827 is in as well, because this pull request is what makes the case reachable. Widening global_dev_init () alone was not enough: omen_load () drops the escape whenever the device engine is on, and it decides that while the grammar is read, so the run reached the host engine with no escape tables and nothing to enumerate. A grammar with no structures gives the device engine no base word either, so the escape is kept for it, and then the empty pool is reported as an empty inner loop rather than refused:

    mkdir -p /tmp/esc && tar xf pcfg/default-passwords.tar.xz -C /tmp/esc
    printf 'M\t1.0\n' > /tmp/esc/ruleset/Grammar/grammar.txt

    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 /tmp/esc/ruleset --potfile-disable --runtime 5 -d 1 --status

    before: /tmp/esc/ruleset: no usable structures, rc=255
    after:  OMEN escape carried, 26386266585 guesses
            no base words for the device engine, the host engine takes the run
            Status Cracked, Progress 100290816/26386266585, rc=0

So the host engine is seen picking the run up rather than a keyspace of zero on both sides.

The shipped ruleset is untouched by it: 23154 structures, so the escape is dropped there as before, the same 12747516022634 base words, and the first 20000 and 500000 candidates still hash to 43b848a00665a0e0220e64760a247760 and 52c6693e92e780700c5766cc7662fc98. tools/test_edge.sh -m 0 -K 0
detects 0 errors; -K 0 because -a 4 has no optimized kernel and generic.c refuses -O rather than ignoring it, which the full suite reports on master as well.

jsteube:
Everything checks out except the following, and all but the last of it sits in the new commit.

Three answers are settled from dev_enable inside global_init, and the fallback that turns the device engine off runs after global_init has returned. A run that ends on the host engine therefore keeps three answers that were given for the device engine. None of the three can be seen on master, where units at zero always means keyspace zero and the fallback never carries a run. Carrying the escape is what puts a live run behind them.

Finding 1/6, the fallback run generates base words on one thread. The worker count is left at PCFG_PF_WORKERS_AUTO whenever dev_enable is true, at feed_pcfg.c:7894, and the device half settles it at 8404, which is below the units == 0 return at 8379. The prefetch pool is then never started, because thread_init returns at 8031 ahead of pf_start, so the host engine builds inline. Measured on the escape only grammar, one RX 9070 XT, --runtime 12, one hash that never cracks:

```
default    29578.4 kH/s
threads=8  44084.9 kH/s
threads=4  46010.5 kH/s
```

Finding 2/6, the status line names the engine the run is not on. The label is written at 7908 with "device" and nothing rewrites it, so Guess.Base reads "(scale 1, device)" while Candidate.Engine on the same screen reads "Host Generator + PCIe". The "host, OMEN" label describes this run exactly and is the one it cannot reach.

Finding 3/6, lookup= is answered by neither half. The host half at 7892 is gated on dev_enable == false and the device half at 8387 is below the units == 0 return, so a question asked with -d 1 gets an attack instead of an answer, and described is never set:

```
-d 1      no lookup output at all, the attack runs
-d 1 -S   lookup: no structure derives it, the OMEN escape does, at level 1 and cost 16
```

One fix covers the three. The run is known to be going to the host engine at the units == 0 branch, so the three answers can be given again there:

```
    global_ctx->dev_enable = false;

    pcfg_pick_workers (pg);

    snprintf (global_ctx->guess_base, sizeof (global_ctx->guess_base), "%s (scale %" PRIu64 ", %s)", pg->named, pg->scale, (pg->omen_lvl_cnt > 0) ? "host, OMEN" : "host");

    if (pg->lookup != NULL) lookup_report (global_ctx, pg);
```

That needs the worker choice at 7894 to 7903 lifted into pcfg_pick_workers(), and the joined ruleset name kept in pg, because roots is freed before this point.

Finding 4/6, a grammar with nothing left to enumerate is accepted. The M line exemption at 2777 is what lets an escape only ruleset load, and it runs before omen_load, so it cannot see whether the escape survives. Two things drop it, omen=0 and a Grammar that names an M line where the Omen directory is absent:

```
master  --keyspace                     rc=255  no usable structures
this PR --keyspace omen=0              rc=0    0
this PR --keyspace, no Omen directory  rc=0    0
this PR a run, either way              rc=1    no candidates, Progress 0
```

rc=1 is the ordinary "not found", so an attack that cannot produce a single candidate reports what an attack that ran and missed reports. A keyspace of 0 with rc=0 is worse, because that is the number a distribution script divides by. This is the other half of finding 6 of the last round, where the costmax half was fixed. The test belongs after omen_load, where roots is still alive:

```
  if ((pg->structs_cnt == 0) && (pg->omen_lvl_cnt == 0))
  {
    char named[256];

    roots_join (named, sizeof (named), roots, nroots);

    gerr (global_ctx, "%s: nothing to enumerate, no structures and the escape is not carried", named);

    roots_free (roots, nroots);

    return false;
  }
```

Finding 5/6, a comment whose reason the new condition removed. The block above pcfg_omen_lookup at 5105 says the escape is dropped when dev_enable is set, so omen_lvl_cnt is zero and the very first test refuses. The conclusion still holds, because a grammar with no structures has no units and the core turns the device engine off, but that is no longer the reason. The omen setting's help text has the same shape, where the escape is carried for "every hash the device engine is off for", and it is now also every grammar the device engine has no base word for.

Finding 6/6, a comment left behind by the round before. At 6093 in lvl_one, removing the candidate prefix took the accumulator with it and left the comment that explained it, so the text reads "the two prefixes advance together" over a doubled blank line where the statement was, and one prefix remains.

matrix:
All six are correct, and five of them are the new commit's own: carrying the escape put a live run
behind answers that were given for an engine it no longer uses.

Findings 1, 2 and 3 have one fix, at the units == 0 branch, where the run is known to be going back
to the host engine. The worker count moves into pcfg_pick_workers () and is asked for again there,
so the base words are no longer built on one thread. The label is rewritten, and reads "host, OMEN"
beside Candidate.Engine rather than "device". lookup_report () is called there as well, so a
question asked with -d 1 is answered instead of run. The joined ruleset name is kept in pg, since
roots is freed before global_dev_init ().

    ./hashcat -m 0 -a 4 00000000000000000000000000000000 /tmp/esc/ruleset --potfile-disable --runtime 12 -d 1 --status

    before   30739.9 kH/s   threads=8 35167.9   Guess.Base "(scale 1, device)"      no lookup output
    after    34885.0 kH/s   threads=8 34997.8   Guess.Base "(scale 1, host, OMEN)"  lookup answers

The gap here is 14 per cent rather than the third you measured, on an RTX 4080 against your 9070 XT,
and the default now sits with the explicit settings either way. lookup= with -d 1 answers "no
structure derives it, the OMEN escape does, at level 2 and cost 19 of costmax 64".

Finding 4 is right and the test is after omen_load () now, where roots is still alive. Both routes
are refused:

    ./hashcat -m 0 -a 4 /tmp/esc/ruleset --keyspace omen=0
    ./hashcat -m 0 -a 4 /tmp/esc_noomen --keyspace

    before   0, rc=0                                                              both
    after    nothing to enumerate, no structures and the escape is not carried, rc=255

A grammar that does carry the escape still answers 26386266585 and the shipped ruleset still answers
12747516022634, with the same 20000 and 500000 candidate digests and the same Guess.Base "device"
there. tools/test_edge.sh -m 0 -K 0 detects 0 errors.

Findings 5 and 6 are corrected. The comment above pcfg_omen_lookup () now says why the conclusion
still holds, that a grammar the device engine can amplify has structures and loses the escape, while
one without structures keeps it and reaches that code from the host engine. The omen help text says
the escape is carried for every hash the device engine is off for and every grammar it has no base
word for. And the comment in lvl_one () that outlived its accumulator is gone.

jsteube:
Merged. All six are fixed, and the shape they were fixed in is the right one: the three answers are given again at the point the run is known to be going back to the host engine, rather than guessed at earlier.

One thing is left, and it is on this page rather than in the code, which is why it is not another round. The walk section offers a command to build the digest set and the python in it does not run. The body of the for loop is flat, so it ends at an IndentationError before hashcat is reached, and the two lines under it need one level:

```
./hashcat --stdout -a 4 pcfg/default-passwords.tar.xz --limit 3000000 | python3 -c 'import sys, hashlib
for l in sys.stdin.buffer:
  l = l.rstrip (b"\n")
  if l: print (hashlib.md5 (l).hexdigest ())' | sort -u > walk.hash
```

The evidence behind it is sound. Rebuilt with the indentation put back, the set is 2790759 unique digests out of 3000000 candidates, which is the number the section quotes, and both sides reach 138835703245 candidates with 2520633 recovered and 0 rejected. It is worth correcting because that is the one command in the body a reader would paste to check the walk claim, and the description outlives the merge even though it does not enter git.

One more, and it belongs in a pull request of its own rather than here. A grammar that keeps its structures but has no reachable units is still accepted with nothing to do, and that one is master's now rather than yours, because 4827 is what opened it:

```
master with 4827 applied

./hashcat -a 4 pcfg/default-passwords.tar.xz --keyspace costmax=4        rc=0    0
./hashcat -m 0 -a 4 <hash> pcfg/default-passwords.tar.xz -d 1 costmax=4  rc=1    Exhausted, Progress 0

before 4827

the same run                                                            rc=255  device engine index is empty
```

That is the same shape as finding 4 of the last round, one step further along. The refusal added there covers the grammar that has no structures, and this one has them, so it reaches the emptiness test from the other side. A keyspace of 0 at rc=0 is the number a distribution script divides by, and a run that reports Exhausted at Progress 0 is indistinguishable from one that ran and missed.

PR #4828, opened 2026-09-06
Refuse a pcfg attack that has nothing to enumerate

matrix:
A pcfg attack whose candidates are all priced out of reach is accepted rather than refused, so it
reports a keyspace of 0 at rc 0 and a run of it reports Exhausted at Progress 0. This refuses it and
says which of the two ways it arrived with nothing.

Deviation from CONTRIBUTING.md, as AGENTS.md asks for: no issue was opened for the bug first. All of
it is the two commands below, so the issue would carry the same text this description does.

    ./hashcat -a 4 pcfg/default-passwords.tar.xz --keyspace costmax=4

    before: 0, rc=0
    after:  pcfg/default-passwords.tar.xz: nothing to enumerate, no level of the index lands at or below costmax 4, and the escape is not carried, rc=255

    ./hashcat -m 0 -a 4 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --potfile-disable -d 1 costmax=4 --status

    before: Status Exhausted, Progress 0, rc=1
    after:  the same refusal, rc=255

Both engines are affected and the second command behaves the same under -S, because neither engine
has anything to walk.

build_index () counts the structures and the escape into one ladder, and assigns pg->keyspace only
where that ladder holds a level, so a keyspace of zero there means both are empty. The refusal #4825
added for a grammar with no structures sat above build_index (), before the structures had been
priced, so it saw only one of the two ways to arrive with nothing to do. It moves below the index
build and is keyed on the keyspace, which covers both, and it names the one it found: no structures
and no escape, or no level at or below costmax.

Refusing is what master did here before #4827, which turned "device engine index is empty" into a
fallback to the host engine. That fallback is right for an engine with no base words, and this is
the case it left over, where neither engine has any.

A keyspace of zero is the number a distribution script divides by, and an Exhausted run at Progress 0
cannot be told apart from one that ran and missed, so a mistyped costmax reads as an attack that did
not find the password.

### Checked

Nothing that has a keyspace loses it. Against master at e2a7ab834, on the shipped ruleset:

    costmax   master                  this PR
    2         refused, rc=255         the same
    3         refused, rc=255         the same
    4         0, rc=0                 refused, rc=255
    5         0, rc=0                 refused, rc=255
    6         0, rc=0                 refused, rc=255
    7         0, rc=0                 refused, rc=255
    8         1, rc=0                 the same
    9         2, rc=0                 the same
    12        14, rc=0                the same
    unset     12747516022634, rc=0    the same

2 and 3 are refused on both sides, where every structure is priced above the cap and the load says
so. From 8 up nothing that has a keyspace is turned away.

A grammar whose whole mass sits on the escape still answers 26386266585, and one that has lost the
escape to omen=0 is refused as before, naming no structures. The first 20000 and 500000 candidates
still hash to 43b848a00665a0e0220e64760a247760 and 52c6693e92e780700c5766cc7662fc98.

    ./tools/test_edge.sh -m 0 -K 0

    [ test_edge_1788721754 ] > All tests done in 0d:00h:02m:18s
    [ test_edge_1788721754 ] > Errors detected: 0

-K 0 because -a 4 has no optimized kernel: generic.c refuses -O rather than ignoring it, since the
hashes were parsed under the flag on the way in, so the full suite reports that refusal on master as
well.

It builds on Linux with the same six warnings master already has and none of them new.

### DIFF
--- a/src/feeds/feed_pcfg.c
+++ b/src/feeds/feed_pcfg.c
@@ -7877,23 +7877,6 @@ bool global_init (generic_global_ctx_t *global_ctx, MAYBE_UNUSED generic_thread_
     return false;
   }
 
-  // The M line that let a grammar with no structures load was counted before the escape was read,
-  // so it could not tell whether the escape survives. omen=0 drops it, and so does a Grammar that
-  // names an M line beside no Omen directory, and either way nothing is left to enumerate.
-
-  if ((pg->structs_cnt == 0) && (pg->omen_lvl_cnt == 0))
-  {
-    char named[256];
-
-    roots_join (named, sizeof (named), roots, nroots);
-
-    gerr (global_ctx, "%s: nothing to enumerate, no structures and the escape is not carried", named);
-
-    roots_free (roots, nroots);
-
-    return false;
-  }
-
   char display[32];
 
   if (global_ctx->quiet == false) pmsg (pg, "pcfg: OMEN tables loaded in %s", pcfg_duration ((hc_timer_get (t_omen) / 1000.0), display, sizeof (display)));
@@ -7917,6 +7900,32 @@ bool global_init (generic_global_ctx_t *global_ctx, MAYBE_UNUSED generic_thread_
     {
       pmsg (pg, "pcfg: level index built in %s, %u levels", pcfg_duration ((hc_timer_get (t_ix) / 1000.0), display, sizeof (display)), pg->lvl_cnt);
     }
+  }
+
+  // build_index () counts the structures and the escape into the same ladder, so a keyspace of zero
+  // here is both of them empty: a grammar carrying no structures and no escape, or one whose
+  // structures are all priced out of reach. Either is a run with nothing to do, and reporting it as
+  // a success is worse than refusing, because a keyspace of zero is the number work is divided by
+  // and an Exhausted run at Progress 0 reads like one that ran and missed.
+
+  if (pg->keyspace == 0)
+  {
+    char named[256];
+
+    roots_join (named, sizeof (named), roots, nroots);
+
+    if (pg->structs_cnt == 0)
+    {
+      gerr (global_ctx, "%s: nothing to enumerate, no structures and the escape is not carried", named);
+    }
+    else
+    {
+      gerr (global_ctx, "%s: nothing to enumerate, no level of the index lands at or below costmax %" PRIu64 ", and the escape is not carried", named, pg->costmax);
+    }
+
+    roots_free (roots, nroots);
+
+    return false;
   }
 
   // lookup= is answered here and not earlier, because it reads the grammar grammar_load () parsed,
--- a/src/feeds/feed_pcfg.c
+++ b/src/feeds/feed_pcfg.c
@@ -2795,7 +2795,7 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
 
     if (dropped_c > 0)
     {
-      gerr (global_ctx, "%s: all %u structures cost more than costmax %" PRIu64 ", nothing is reachable", named, dropped_c, pg->costmax);
+      gerr (global_ctx, "%s: all %u structures cost more than costmax %" PRIu64 ", nothing is reachable", named, dropped_c, pg->costmax / pg->scale);
 
       return -1;
     }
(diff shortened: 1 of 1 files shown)


jsteube:
Looks good, but build_index () counts the structures and the escape into one ladder, so a keyspace of zero is the only test that covers both ways of arriving with nothing, and moving it below the index build is what lets it see the second one.

Finding 1/2, the costmax it names is the internal one. pg->costmax is the setting times scale and the message prints pg->costmax, which is the same number only while scale is 1, and scale is 1 in every example in the description:

```
costmax=5  scale=2   refused, "no level of the index lands at or below costmax 10"
costmax=10 scale=2   keyspace 4
```

So the run that is refused is told the threshold is 10, and 10 is the setting that makes the same attack run. Anyone acting on the message doubles their own value a second time, because costmax=10 at scale=2 is an internal 20. The message added in #4825 at feed_pcfg.c:2795 quotes the same scaled value, "all 23158 structures cost more than costmax 16" for costmax=4 scale=4, so both want the same division. pg->costmax / pg->scale is exact, since pg->costmax is the product and scale has a floor of 1.

Finding 2/2, the branch is chosen on the structure count, where costmax can be the cause on either side. A grammar with no structures whose escape is priced out by the cap falls into the first branch, which names neither the cap nor what the escape is actually doing:

```
/tmp/esc/ruleset, one M line and an Omen directory

costmax unset   26386266585
costmax=8       nothing to enumerate, no structures and the escape is not carried
```

The escape is carried in the sense the user means. The ruleset has the model, omen is 1, and omen_load () read it. What is empty is the index, because every level of the model is priced above the cap, which the load already says when it is not quiet, at 3721, "OMEN escape dropped, no level of it lands at or below costmax". Under --keyspace that line is suppressed and the refusal is the only thing the user sees, so it points at the ruleset where the answer is the setting.

pg->omen_cnt separates the two, because it counts models that were read where omen_lvl_cnt counts levels that survived the cap:

```
    if ((pg->omen_cnt == 0) && (pg->structs_cnt == 0))
    {
      gerr (global_ctx, "%s: nothing to enumerate, no structures and no escape to carry", named);
    }
    else
    {
      gerr (global_ctx, "%s: nothing to enumerate, no level of the grammar or of the escape lands at or below costmax %" PRIu64, named, pg->costmax / pg->scale);
    }
```

That gives each of the four cases the right half. Escape only with omen=0 and escape only with no Omen directory both leave omen_cnt at 0 and say there is no escape to carry. An escape priced out by the cap leaves it above 0 and names the cap. The shipped ruleset at costmax=4 on the device engine, where the escape is dropped for the engine rather than by the cap, has structures and names the cap as it does today.

matrix:
Both are correct, and the first one is worse than a wording problem: the message named the value that makes the same attack run.

Finding 1. pg->costmax is the setting times scale, so the message quoted the internal product. It divides by pg->scale now, which is exact because scale has a floor of 1, and the message #4825 added does the same:

    ./hashcat -a 4 pcfg/default-passwords.tar.xz --keyspace costmax=5 scale=2

    before: no level of the index lands at or below costmax 10, rc=255
    after:  no level of the grammar or of the escape lands at or below costmax 5, rc=255

    ./hashcat -a 4 pcfg/default-passwords.tar.xz --keyspace costmax=4 scale=4

    before: all 23158 structures cost more than costmax 16, rc=255
    after:  all 23158 structures cost more than costmax 4, rc=255

and costmax=10 scale=2 still answers 4, so the number the message names is no longer the number that would have worked.

The three lookup lines quote pg->costmax as well, at 7574, 7590 and 7597, and I left them: each of them prints hit.cost beside it and that is the internal cost too, so the pair is consistent with itself. Say if they should be divided as well, and hit.cost with them.

Finding 2. Correct, and omen_cnt is the right separator. The four cases now each get their own half:

    /tmp/esc/ruleset, one M line and an Omen directory

    costmax=8            no level of the grammar or of the escape lands at or below costmax 8
    omen=0               no structures and no escape to carry
    no Omen directory    no structures and no escape to carry
    shipped, costmax=4   no level of the grammar or of the escape lands at or below costmax 4

Nothing that has a keyspace lost it: the escape only ruleset still answers 26386266585, the shipped one 12747516022634, and costmax=8 still answers 1. tools/test_edge.sh -m 0 -K 0 detects 0 errors.

PR #4829, opened 2026-09-07
Lookup with rules

jessevz:
### --lookup with the rules in play for attack mode 0

--lookup reported where a run reaches a candidate, but in attack mode 0 with
-r it only ever searched the base words, so a candidate that some rule makes
was reported as absent. It now searches the rules as well, names the base word
and the rule that make the candidate, and a miss becomes a proof rather than an
answer about half the attack.

### Reproduction
The password merche03123 is one that this exact run tries. Both files ship
with hashcat, so there is nothing to create first.

```
$ ./hashcat --lookup=merche03123 -a 0 example.dict -r rules/best66.rule
```
before:
```
lookup: 'merche03123'
lookup: this wordlist does not hold it as a word, in any of its 128416 lines
lookup: whether one of the 66 rules makes it out of some other word was NOT checked, and is not something this can answer
lookup: so this is not a proof that the run misses it, only that the word itself is not there
rc=0
```
after:
```
lookup: 'merche03123'
lookup: trying all 66 rules on each of the 128416 words, in the order the run does
lookup: word 89999 of 128416
lookup: that word is 'merche03', and rule 27 of 66 is what makes the candidate out of it: $1 $2 $3
lookup: 70.0839% into the run
lookup: this run reaches it at -s 89999, because -a 0 counts -s in words
lookup: -s 89999 -l 1 runs that word with all 66 rules applied to it, and the candidate is one of them
lookup: no earlier word and no earlier rule on this word makes it, so this is the first time the run tries it
rc=0
```
The offset it reports is one a real attack cracks the hash at, and the word
before it is not:

```
$ ./hashcat -m 0 -a 0 -s 89999 -l 1 --quiet --potfile-disable 1dcae95666d13ba5836bc14314a4aa86 example.dict -r rules/best66.rule
1dcae95666d13ba5836bc14314a4aa86:merche03123
rc=0
$ ./hashcat -m 0 -a 0 -s 89998 -l 1 --quiet --potfile-disable 1dcae95666d13ba5836bc14314a4aa86 example.dict -r rules/best66.rule
rc=1
```

A candidate the run really does not reach is now said to be out of reach,
instead of being left open:

```
$ ./hashcat --lookup=ZZnotpresentZZ -a 0 example.dict -r rules/best66.rule
```
before:
```
lookup: this wordlist does not hold it as a word, in any of its 128416 lines
lookup: whether one of the 66 rules makes it out of some other word was NOT checked, and is not something this can answer
lookup: so this is not a proof that the run misses it, only that the word itself is not there
rc=0
```
after:
```
lookup: trying all 66 rules on each of the 128416 words, in the order the run does
lookup: nothing in this run produces it. every one of the 128416 words was tried with all 66 rules and none of them makes it
rc=0
```

### What it does
A rule is a program and is not invertible, so the answer is found forwards, by
building the candidates the run builds and comparing them. generic.c gets one
new reader, generic_ctx_word_index_rules (), which walks the feed once and
applies every kernel rule to each word it transforms. The two loops are nested
the way the run nests them, base words outside and the rule set inside, which is
what makes the first pair it finds the earliest the run reaches. It stops there.
Which rule engine it uses follows hashconfig->opti_type, so -O is answered by
the same implementation that would run it.

The reported offset stays in --skip units, because attack mode 0 counts those
in base words whether or not a rule multiplies each of them. Nothing else about
the option changes, and no other attack mode is touched.

A word longer than the rule engine holds is passed over rather than cut to fit.
A run rejects such a base word before any rule touches it, so cutting it here
would answer about a candidate the run never builds.

The cost is one pass over the wordlist. Proving the miss above is the worst case
for that command line, 128416 words against 66 rules, which is 8.5 million rule
applications on one thread:

```
$ time ./hashcat --lookup=ZZnotpresentZZ -a 0 example.dict -r rules/best66.rule >/dev/null
real    0m0.114s
```

No hash mode is touched by this change, which is host side only, so what it adds
instead is tools/test_lookup.sh:

```
$ ./tools/test_lookup.sh --crack
## the --stdout oracle, 20000 words of example.dict with best66.rule
## the same, with -O, which runs the other rule engine
## what the report says
## a miss with rules is a proof
## without rules, unchanged
## candidates and words no shell can pass
## several dictionaries
## the window, and where the answer sits in the run
## generated rules
## a word longer than the rule engine can hold
## cracking at the offset the lookup reports

passed: 63
failed: 0
```

Its oracle is --stdout, which shares no code with the lookup. For attack mode
0 with R rules the candidate on stream line L comes from base word (L - 1) / R
under rule (L - 1) % R, so the script builds the stream once, probes seven points
in it, and asserts that --lookup names that word, that rule and that offset for
the first line equal to the candidate. It repeats four probes under -O. Without
--crack it opens no device and reads no hashes, and takes 1.9 seconds.

The code has mostly been generated by AI, but verified by me


jsteube:
I tested this with example.dict and best66.rule, and it correctly identifies the word and rule while reporting the same offset at which a normal run cracks the hash.

The problem is how quickly that result becomes impractical, because proving that no word and rule combination produces the candidate requires walking the complete product of both inputs. The example in the description takes 1.9 seconds, and scaling that measurement gives 3 minutes for rockyou with best64.rule, 88 hours for rockyou with dive.rule, and longer still for a wordlist several gigabytes in size.

A size limit does not make this useful, because any sensible limit accepts only the small dictionary and rule combinations people use for testing while refusing the workloads they would actually use. That behavior would make the feature appear unreliable, since it works during evaluation and stops working as soon as the user supplies real data.

Every other --lookup path costs at most a single pass over the input, which is what the run itself has to read anyway, and a mask attack does not even need that because it ranks the candidate directly. Adding rules turns one pass into one pass per rule. If the first common use someone tries is -a 0 with -r and the command appears never to return, they will conclude that --lookup itself is slow and may never try the configurations that answer in a single pass.

I would therefore reject -a 0 with -r before starting the search and explain that this is the only combination that cannot be answered directly, while every other supported configuration answers within a single pass. Hashcat already reports that rules were not checked, so the remaining work is mainly to make the message clear enough that users understand the limitation without losing confidence in the option.

Working backwards from the candidate could make this case fast, but we should not add that implementation because hashcat already has three versions of the rule semantics in rp.c and the two device engines, and an inverse would add two more implementations that must remain consistent forever. Every future rule engine change would then require matching inverse logic, and an omitted or incorrect inverse would not merely make --lookup slower but could falsely report that a reachable candidate cannot be produced.

The --stdout oracle in your test file is the part I would like to keep. Checking a --lookup answer by replaying the same window with --stdout, where the two sides share no code, works for every attack mode that already answers, and none of -a 1, -a 3, -a 6 or -a 7 has a --lookup test today. That would be a smaller change built on the same idea, and I would take it.

jessevz:
> I tested this with example.dict and best66.rule, and it correctly identifies the word and rule while reporting the same offset at which a normal run cracks the hash.
> 
> The problem is how quickly that result becomes impractical, because proving that no word and rule combination produces the candidate requires walking the complete product of both inputs. The example in the description takes 1.9 seconds, and scaling that measurement gives 3 minutes for rockyou with best64.rule, 88 hours for rockyou with dive.rule, and longer still for a wordlist several gigabytes in size.
> 

Sorry I think my message was not very clearly worded, the test script takes around 1.9 seconds, not the single lookup, when i test it for example with rockyou and best 66 i get 12 seconds instead of 3 minutes:

```
time ./hashcat --lookup=merche03123 -a 0 rockyou.txt -r rules/best66.rule
lookup: 'merche03123'
lookup: trying all 66 rules on each of the 14344391 words, in the order the run does
lookup: nothing in this run produces it. every one of the 14344391 words was tried with all 66 rules and none of them makes it

real	0m12.489s
user	0m12.448s
sys	0m0.040s
```

Where as doing it with just rockyou I get 0.3 seconds:

```
time ./hashcat --lookup=merche03123 -a 0 ~/wordlists/rockyou.txt
lookup: 'merche03123'
lookup: nothing in this run produces it. the wordlist does not hold it, and without rules the wordlist is the whole attack

real	0m0.294s
user	0m0.276s
sys	0m0.019s
```

So I think it scales relatively well. But yes big attacks will take long, but the same goes for very big wordlists. Maybe it could make sense to add a progress bar to show the progress to the user with perhaps a warning for big keyspaces? If not I will remove the code and only improve the test script with the modes  -a 1, -a 3, -a 6 and -a 7

jsteube:
You are right about the 1.9 second figure, because it measures the complete test script while one lookup of the same case takes 0.143 seconds here, which means my earlier extrapolation was off by about 17 times.

I rebuilt your PR and measured worst case misses, which cannot exit early, against a real wordlist and subsets of it: 31.7 GB and 2,482,103,743 words.

    10,000,000 words x best66.rule         7s
    2,482,103,743 words, no rules      1m 10s    
    200,000 words x dive.rule          9m 18s

The first result assumes a warm seek database, while the initial run took 2m 25 s because it also had to build one.

The important comparison is between the two rule cases, because the largest wordlist I have answers in 70 seconds without rules while only 200,000 words take more than 9 minutes with dive.rule. A wordlist costs one pass and rules cost one pass per rule, so the rule count is the multiplier here, which is why this is not equivalent to accepting a long scan over a very large file.

At the measured rates, the complete 31.7 GB wordlist would take about 31 minutes with best66.rule and about 80 days with dive.rule. The rate does not scale linearly either, because 66 rules reach about 89 million applications per second while the 98,670 multi command rules in dive.rule reach about 35 million.

A progress bar would make the wait visible, but it would not make the lookup usable, and someone who waits hours for an answer will still remember --lookup as the option that takes too long. A warning for a large keyspace is closer to the right behavior, but once we know enough to warn that the configuration is impractical, we also know enough to reject it immediately and explain why.

A progress indicator for --lookup under -a 0 is still a useful improvement, because even the plain wordlist case gives no sign of life while reading and 70 seconds of silence on a large file is worth avoiding.

I would also take the proposed test harness for -a 1, -a 3, -a 6, and -a 7, since none of those attack modes currently has --lookup coverage and the --stdout oracle you built is the right way to test them.

jessevz:
Makes sense. For the early rejection, do you have a rough runtime/keyspace threshold in mind for what should be considered impractical? I don't want to arbitrarily reject lookups that are merely slow, especially since the plain wordlist case can take ~70s and you suggested keeping that with a progress indicator.

I will add a progress indicator for -a 0 lookup and add the -a 1/3/6/7 test coverage using the oracle.


Edit: perhaps it could also make sense to make the treshold configurable in the makefile and keep the default value conservative. This way for most people the --lookup will be fast. Only if you really configure it by hand can you make the --lookup work for very big keyspaces.

jsteube:
Wait there is no size threshold, and that is intentional. Plain wordlists remain supported at any size. The example shows a 31.7 GB file taking 70 seconds calls for a progress indicator, not a refusal. The progress indicator for -a 0 is worth adding.

Just for background again: Rules turn that single pass into one pass per rule. We therefore reject --lookup with -a 0 and -r at startup, regardless of the wordlist or rule file. A size limit would let people test --lookup on small inputs, then discover that it refuses their real workloads. I would rather give them a clear restriction upfront. Your 12 seconds for rockyou with best66 is not in question, but setting a threshold to accommodate it only moves the argument to dive.rule. Restricting the configuration avoids an arbitrary number that we then have to defend. I would avoid a Makefile setting too.

jessevz:
Reworked this along the lines you suggested. Six commits on top of the branch:

The rules search is reverted, so src/generic.c, src/straight.c and include/feed_ctx.h are back to where they were. -a 0 with -r or -g is now refused in the --lookup validation block, before any file is opened, rather than answering about the base word alone. I did not word it as the only combination that cannot be answered directly, because it is not: -a 4 with -r has the same gap and still answers. Rules and --lookup can only meet in modes 0 and 4, so -a 4 is the whole remainder, and I left it alone since its answer comes from the pcfg feed rather than a wordlist. Happy to look at it separately if you want it closed too.

The --stdout oracle is kept and pointed at the modes that already answer. It no longer builds the whole candidate stream; instead it replays the window each report names. The report says which cell holds the answer and where in that cell it is, so the check runs --stdout -s N -l 1 and asserts the cell is that size with the candidate at that position, and that the window before it does not hold the candidate at all. One cell rather than a walk of the keyspace, and it reads the same for every mode. -a 1, -a 3, -a 6 and -a 7 are covered now, along with the round number in an --increment queue, the three ways a mask can miss, the hybrid split reports and the folder segment naming. --stdout forces -m 2000, and the hash mode decides how a -a 3 base word is split and so how large a cell is, so every replayed check passes -m 2000 to --lookup as well. --stdout refuses -S, so that one branch has its wording checked instead of its offset. 77 checks, and ten deliberate one line breaks of the reports were each caught by between 1 and 11 of them.

Writing those turned up a small bug: lookup->has_q was only assigned where a hit is recorded, and a miss reaches none of those assignments, so -a 1 and -a 12 with a ?q always described a miss as a wordlist and a mask instead of two wordlists. Fixed by recording it where it is read. That one is unrelated to the rest and I am happy to split it out.

Last two commits are the progress feedback. A wordlist lookup is one pass over the whole file and cannot stop at the first hit, so on a large wordlist it was minutes of silence. It now stays quiet for two seconds and then reports, using event_log_info_nn so the line erases itself and the answer lands clean, and only when stdout is a terminal. The round is named for the hybrids, since they read the wordlist once per mask. Underneath that, seekdb_build only reported progress for compressed sources: a plain wordlist arrives as one window, so the report at the bottom of the loop was reached once, after the counting had finished. It reports from inside the scan now. That half benefits every attack mode, not just --lookup, so it is a separate commit if you would rather take only the first.

jsteube:
Could you please refuse -a 4 the same way, in this PR? Its answer comes from the pcfg feed, but the rule set is applied to what that feed produces, so the offset still names a word the run then changes, which is what -a 0 is refused for. With both refused the attack mode test can go: -r and -g need 0, 4, 8 or 9 and --lookup takes 0, 1, 3, 4, 6, 7 and 12, so the rule condition on its own is the whole check, and the comment above it loses its third paragraph.

I would keep both of the parts you offered to split out. The has_q assignment is three lines in the function the reports are built in, and the seekdb one is a report that a plain wordlist reached once, after the counting had finished, which is a defect with or without --lookup. Either as its own PR costs more to read than it saves.

jessevz:
Changes have been added

PR #4833, opened 2026-09-09
Hand the pcfg pool over in up to four buffers, and without a copy

matrix:
Two things go wrong on the same path, the one the pcfg pool takes from the feed to the device, and they are in one change because the second settles the arithmetic of the first.

## The pool is refused whole

A device does not always allocate the pool in one block. The largest single allocation it offers is often a fraction of the memory it has, so a pool that fits the card is refused whole and the run falls back to the host engine on a card with memory to spare.

Two real cases, both OpenCL, both on cards with far more free memory than the pool needs:

```
NVIDIA GeForce RTX 4080, 15680 MiB free, 3993 MiB in one allocation
Apple M4 Pro,            36800 MiB free, 3456 MiB in one allocation
```

A grammar trained with hpcfg on hashmob's combined founds packs more terminals than either will take in one piece: 4325 MiB of them on the Linux box and 4269 MiB on the Mac, trained from the same corpus. Neither device would take it.

That one is also past what an offset into the pool reaches, which is a separate matter and a separate pull request, #4835. The runs further down use the same grammar at a `costmax` that keeps it inside.

## And where the device's memory is the host's, it is held twice

The pool is the largest buffer a run has, it is read only, and it lives for the whole run. `hc_clCreateBuffer_ext ()` and `hc_mtlCreateBuffer ()` already turn a host pointer into `CL_MEM_USE_HOST_PTR` and `newBufferWithBytesNoCopy`, and nothing in `backend.c` ever passed one, so the pool was copied even where the copy lands in the same memory it came from.

## Why the two are one change

They meet in the same handful of lines, and more than that, the second settles what the first is allowed to decide.

- **The budget.** `pcfg_pool_budget ()` decides how many parts a device needs, and on a unified memory device a copied pool costs twice, once in the feed and once on the device. So the budget has to know which of the two it is looking at. On the Apple M4 Pro under OpenCL it is 13824 MiB if the pool is handed over and 6912 MiB if it is copied, a factor of two, and between those two figures the question is not how the pool is cut but whether the device runs at all.
- **The alignment.** Both wrappers want a page aligned pointer. The feed aligns the pool for that, and the parts are cut on the same 64 KiB grain so that the pointer handed over for a part is aligned too. One grain answers both: splitting alone would not need it, and handing over alone would not survive a pool that had to be split.
- **The same devices.** Of the two cards that cannot take the pool whole, one is unified memory. It is the same device that would otherwise hold the pool twice in the memory it shares with the host.

Done as two changes, the second would have to reopen the same four lines of `backend.c` and rewrite the budget the first had just put there.

## What changes

The pool now travels to the kernel as up to `PCFG_POOL_PARTS` buffers instead of one, and a read finds the part its index falls in.

- `pcfg_pool_split()` decides how many parts a device needs and cuts them equal, as few as the ceiling on one allocation allows.
- A part is a whole number of pages rather than a power of two. Rounding a part down to a power of two threw away up to half of what the device offered.
- `pcfg_pool_budget()` answers two different bounds: what the parts come to together, and what one part may be. Reading the second as if it were also the first splits a pool the device would have taken whole.
- A device that cannot hold the pool drops out of the run with a warning, the way a device that cannot hold its launch buffers does, rather than taking the session down with it.

Where the pool went over in one piece the search is compiled out, because the number of parts is settled before the kernel is built.

## How the pool is handed over

- The feed allocates the pool page aligned and a whole number of pages long, which is what both wrappers ask of a host pointer. Parts are cut on the same grain, so the pointer handed over for a part is aligned too.
- A device that cannot be handed the bytes still takes a copy, and `pcfg_pool_budget ()` halves its budget, because on unified memory that copy comes out of the same pool of memory.
- An AMD device under Metal is one of those. `hc_mtlCreateBuffer ()` gives it a Managed buffer where Shared was asked for, and then refuses a host pointer on anything but Shared, and that refusal ends the session rather than the device.
- The pool moved from `hccalloc ()` to `hc_alloc_aligned ()`. `hccalloc ()` asks the kernel for huge pages on a large block and `hc_alloc_aligned ()` did not, so the hint moved with it rather than being lost on the largest allocation hashcat makes.

What that is worth, on an Apple M4 Pro running the same attack twice, the second time with `pcfg_pool_shared ()` forced to answer no so the pool is copied the way it was before:

```
              peak memory footprint     speed
handed over          3.49 GiB        3470.2 MH/s
copied               3.91 GiB        3472.2 MH/s
```

433 MiB apart on a pool of 462 MiB, and the same speed either way. Both runs covered the same `Progress 120757569592`.

## What it is groundwork for

The change that follows this one gives the device engine the OMEN escape, which today it cannot walk: on a grammar trained at the default coverage the escape holds about 40% of the model's mass, and a run either drops it or goes to the host engine to keep it.

The escape's tables go into this same pool. On the grammar used above that takes the pool from 3870 MiB of terminals to 6484 MiB with the escape beside them, against a largest single allocation of 3993 MiB on one of these cards and 3456 MiB on another. So the escape on the device cannot exist at all until the pool can travel in more than one buffer, which is what this change is.

The 4 GiB ceiling on a pool offset, #4835 against master, is the other half of the same groundwork. That one hands such a grammar to the host engine; the follow-up turns the bound into a cut instead, shortening the least probable terminal lists until they fit, which is what leaves the escape its tables in

```
pcfg: terminals cut at cost 28 instead of 64, 7 lists shorter, 3870 MiB instead of 4325 MiB, which leaves the escape its tables
```

## The parts travel as arguments, not behind a struct

This is not a style choice. A local struct whose address is taken goes to private memory, and every read of the descriptor then becomes a scratch access.

It was measured on an RX 6900 XT under HIP, with the ceiling on one buffer forced down so that the shipped ruleset's 17 MiB pool is cut into four and the search is compiled in. Same binary otherwise, same pinned launch, and both runs exhausted the same 229014641206 candidates:

| Arrangement | Private memory | `buffer_load` | Speed |
| --- | ---: | ---: | ---: |
| Four buffers behind a struct | 144 B | 86 | 8368.7 MH/s |
| Four buffers as arguments | 96 B | 38 | 9801.0 MH/s |

The 48 bytes between them are the struct itself, four pointers and three offsets.

The comparisons that find the part nest rather than chain, which is the other shape a reader might ask about. That one measures the same either way here, because a part is found once per slot and not once per byte:

| | nested | chained |
| --- | ---: | ---: |
| ISA lines | 13296 | 13112 |
| `s_and_saveexec` | 34 | 44 |
| `s_cbranch_execz` | 364 | 338 |
| Speed | 9801.0 MH/s | 9906.5 MH/s |

## What it costs when it does not split

This is the case that matters for everyone who is not hitting the limit, so it was measured rather than argued. Same ruleset, same parameters, kernels disassembled for gfx1030 and compared against the same commit without this change:

| Mode | ISA lines | vgpr | scratch | `buffer_load` |
| --- | --- | ---: | ---: | --- |
| 0 MD5 | 11853 -> 11849 | 53 | 96 | 38 |
| 23003 SecureZIP AES-256 | 12056 -> 12068 | 248 | 464 | 20 |
| 22951 PEM AES | 29736 -> 29748 | 256 | 272 | 88 |
| 17225 PKZIP | 23025 -> 22985 | 256 | 77456 | 627 -> 631 |
| 31400 SCRT v2 | 12840 -> 12836 | 120 | 352 | 20 |
| 26000 Mozilla 3DES | 36666 -> 36742 | 248 | 176 | 20 |
| 13100 Kerberos TGS | 13634 -> 13710 | 148 | 528 | 120 |

Registers and scratch are identical on every mode, and so are the memory instruction counts except on 17225, which is already at the register limit on its own (it spills 68 without this change and 69 with it) and takes four more `buffer_load`. The instruction count moves by 0.1 to 0.2 per cent, in both directions.

## Seeing it

The change is inert on the shipped ruleset, whose pool is 17 MiB. It shows up on a grammar trained on a large corpus, on a device whose single allocation ceiling is below the pool size. The grammar below is the one from the top of this page at `costmax=28`, which keeps its terminals inside what a `u32` reaches:

```
* Device #02: Apple M4 Pro, GPU, 36800/36864 MB (3456 MB allocatable), 20MCU
pcfg: device pool packed in 0s 218ms, 3839 MiB
* Device #2: pcfg pool of 3839 MiB in 2 buffers of 1919 MiB, 1919 MiB
Guess.Base.......: Feed (... (scale 1, device))
Speed.#02........: 813.8 MH/s @ Accel:711 Loops:1024 Thr:64 Vec:1
```

Without this change that device does not start the device engine at all.

## Reaching the split without that grammar

The change is inert on the shipped ruleset, whose 17 MiB pool every device here takes in one piece, and the grammar above is not something a reviewer has. A ceiling forced on one buffer makes the split reproducible on the shipped ruleset in a few seconds. In `pcfg_pool_budget ()`, immediately after

```c
u64 part = device_param->device_maxmem_alloc;

if (total < part) part = total;
```

add

```c
const char *env = getenv ("PCFG_POOL_ONE");

if (env != NULL) part = strtoull (env, NULL, 10);
```

The budget is clamped to `part_max` times `PCFG_POOL_PARTS` a few lines below, so bounding one part bounds the pool budget with it, and the 17 MiB pool comes out in two, three or four buffers.

Same binary, the same 9998 digests taken one in forty out of the first 400000 candidates, the same `--limit`, and the outfile sorted and hashed. On an RTX 4080 under CUDA:

```
                        recovered            progress              outfile
unchanged tree          9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
this change, 1 buffer   9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 2 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 3 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 4 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab

* Device #1: pcfg pool of 17 MiB in 4 buffers of 4 MiB, 4 MiB, 4 MiB, 4 MiB
```

The same recipe reaches the split on the pocl CPU device. That one reports unified memory, so the pool goes through `CL_MEM_USE_HOST_PTR` rather than a copy, and the handover is exercised at the same time:

```
unchanged tree          9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
this change, 1 buffer   9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 4 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
```

## The candidates do not change

On the shipped ruleset, whose 17 MiB pool no device has to split, the change should be invisible. Against the same commit without it, at the same `--limit`:

```
                     unchanged tree                        with this change
-d 1 (CUDA)  1774580599978/2285160000000  Rejected 0   1774580599978/2285160000000  Rejected 0
-d 2 (HIP)   1774580599978/2285160000000  Rejected 0   1774580599978/2285160000000  Rejected 0
```

All four runs exhausted, and identical to the digit.

On the device where the split fires the unchanged tree has nothing to compare against, because it produces no run at all. What can be compared is the split against no split, on one machine that does both: an Apple M4 Pro holds that 3839 MiB pool whole under Metal and takes it in two buffers of 1919 MiB under OpenCL. Same binary, same grammar, and both runs left to exhaust the whole of it rather than bounded:

```
-d 1 Metal    one buffer    Exhausted   Progress 2657077/2657077   Rejected 0
-d 2 OpenCL   two buffers   Exhausted   Progress 2657077/2657077   Rejected 0
```

Every candidate the grammar reaches, built once out of one buffer and once out of two, and the same count both ways.

## Checks

```
$ ./tools/test_edge.sh -m 0 -K 0                                        # Linux
[ test_edge_1788979196 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1788979196 ] > All tests done in 0d:00h:02m:27s
[ test_edge_1788979196 ] > Errors detected: 0

$ ./tools/test_edge.sh -m 0 -K 0                                        # macOS
[ test_edge_1788978259 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1788978259 ] > All tests done in 0d:00h:02m:48s
[ test_edge_1788978259 ] > Errors detected: 0

$ ./tools/test_package.sh --no-device                                   # both
```

The native, static and Windows builds compile without a warning.

`-K 0` is deliberate. Without it the suite also runs attack type 4 with `-O`, and the device engine has no optimized kernel, so hashcat refuses the combination and exits 255. That is unchanged from before this patch:

```
$ ./hashcat -m 0 -a 4 -O 8743b52063cd84097a65d1633f5c74f5 pcfg/default-passwords.tar.xz --runtime 1
The device engine has no optimized kernel. Run this without -O.
rc=255
```

The shipped ruleset still answers `12747516022634` and its first 20000 candidates still hash to `43b848a00665a0e0220e64760a247760`.

A CPU device is worth naming separately, because it has unified memory and so takes the pool through `CL_MEM_USE_HOST_PTR` rather than a copy. On pocl it covers the same candidates as the unchanged tree and rejects the same ones:

```
                 unchanged tree                  with this change
-d 3 (pocl CPU)  5621650743/5480000000  Rej 0    5621650743/5480000000  Rej 0
```

### DIFF
--- a/OpenCL/inc_common.h
+++ b/OpenCL/inc_common.h
@@ -99,10 +99,16 @@
 #define _KERN_ATTR_PCFG()                  KERN_ATTR (GLOBAL_AS,   GLOBAL_AS   const bf_t      *g_bfs_buf,     void, void, void),  \
   MAYBE_UNUSED GLOBAL_AS const pcfg_cell_t *pcfg_cells,                                                                            \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool,                                                                            \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool1,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool2,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool3,                                                                           \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_wmap
 #define _KERN_ATTR_PCFG_ESALT(e)           KERN_ATTR (GLOBAL_AS,   GLOBAL_AS   const bf_t      *g_bfs_buf,     void, void, e),     \
   MAYBE_UNUSED GLOBAL_AS const pcfg_cell_t *pcfg_cells,                                                                            \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool,                                                                            \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool1,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool2,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool3,                                                                           \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_wmap
 #define _KERN_ATTR_TMPS(t)                 KERN_ATTR (GLOBAL_AS,   GLOBAL_AS   const bf_t      *g_bfs_buf,     t,    void, void)
 #define _KERN_ATTR_TMPS_ESALT(t,e)         KERN_ATTR (GLOBAL_AS,   GLOBAL_AS   const bf_t      *g_bfs_buf,     t,    void, e)
@@ -119,10 +125,16 @@
 #define _KERN_ATTR_PCFG()                  KERN_ATTR (GLOBAL_AS,   CONSTANT_AS const bf_t      *bfs_buf,       void, void, void),  \
   MAYBE_UNUSED GLOBAL_AS const pcfg_cell_t *pcfg_cells,                                                                            \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool,                                                                            \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool1,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool2,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool3,                                                                           \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_wmap
 #define _KERN_ATTR_PCFG_ESALT(e)           KERN_ATTR (GLOBAL_AS,   CONSTANT_AS const bf_t      *bfs_buf,       void, void, e),     \
   MAYBE_UNUSED GLOBAL_AS const pcfg_cell_t *pcfg_cells,                                                                            \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool,                                                                            \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool1,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool2,                                                                           \
+  MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_pool3,                                                                           \
   MAYBE_UNUSED GLOBAL_AS const u32          *pcfg_wmap
 #define _KERN_ATTR_TMPS(t)                 KERN_ATTR (GLOBAL_AS,   CONSTANT_AS const bf_t      *bfs_buf,       t,    void, void)
 #define _KERN_ATTR_TMPS_ESALT(t,e)         KERN_ATTR (GLOBAL_AS,   CONSTANT_AS const bf_t      *bfs_buf,       t,    void, e)
--- a/OpenCL/inc_pcfg.cl
+++ b/OpenCL/inc_pcfg.cl
@@ -9,13 +9,34 @@
 #include "inc_common.h"
 #include "inc_pcfg.h"
 
-DECLSPEC u32 pcfg_pool_byte (GLOBAL_AS const u32 *pool, const u32 off)
+DECLSPEC u32 pcfg_pool_byte (PCFG_POOL_ARGS, const u32 off)
 {
-  GLOBAL_AS const u8 *pb = (GLOBAL_AS const u8 *) pool;
+  // A byte is loaded as a byte. Reading the word around it and shifting the byte out, together with a
+  // copy that accumulates words, measured 54.3 ms of kernel time against 48.5 on an RX 6900 XT under
+  // HIP at identical work, and the same either way on CUDA.
+
+  #if PCFG_POOL_SPLIT == 0
+
+  GLOBAL_AS const u8 *pb = (GLOBAL_AS const u8 *) pool0;
 
   const u32 b = pb[off];
 
   return b;
+
+  #else
+
+  u32 base = 0;
+  u32 end  = 0;
+
+  GLOBAL_AS const u32 *p = pcfg_pool_span (PCFG_POOL_PASS, off / 4, &base, &end);
+
+  GLOBAL_AS const u8 *pb = (GLOBAL_AS const u8 *) p;
+
+  const u32 b = pb[off - (base * 4)];
+
+  return b;
+
+  #endif
 }
 
 DECLSPEC void pcfg_put_byte (PRIVATE_AS u32 *w, const u32 off, const u32 b)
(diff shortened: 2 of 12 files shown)


jsteube:
I rebuilt this and checked the pool split that the shipped ruleset cannot reach. It is correct. Below is what I ran, since that recipe belongs in the pull request, and then the one thing I would move out of it.

To reach the split without using a large grammar, I capped the allocation size from an environment variable in pcfg_pool_budget (), immediately after:

    u64 part = device_param->device_maxmem_alloc;

    if (total < part) part = total;

Because the budget is clamped to part_max times PCFG_POOL_PARTS a few lines later, limiting part also limits the pool budget. This turns the 17 MiB pool from the shipped ruleset into 2, 3, or 4 buffers.

Using the same binary, the same hash file with 4998 digests, one in 40 of the first 400000 candidates, the same limit, and an RTX 4090 under CUDA:

    unchanged tree        4974/4998   Progress 12492692017   outfile 5dd5cd9f97578be2
    this change, 1 buffer 4974/4998   Progress 12492692017   outfile 5dd5cd9f97578be2
    forced to 2 buffers   4974/4998   Progress 12492692017   outfile 5dd5cd9f97578be2
    forced to 3 buffers   4974/4998   Progress 12492692017   outfile 5dd5cd9f97578be2
    forced to 4 buffers   4974/4998   Progress 12492692017   outfile 5dd5cd9f97578be2

    * Device #1: pcfg pool of 17 MiB in 4 buffers of 4 MiB, 4 MiB, 4 MiB, 4 MiB

The same setup also reaches the split on an OpenCL CPU device. Since that device reports unified memory, the pool goes through CL_MEM_USE_HOST_PTR instead of a copy, which exercises the handover path at the same time:

    unchanged tree        4974/4998   Progress 55133336   outfile 5dd5cd9f97578be2
    this change, 1 buffer 4974/4998   Progress 55133336   outfile 5dd5cd9f97578be2
    forced to 4 buffers   4974/4998   Progress 55133336   outfile 5dd5cd9f97578be2

tools/test_edge.sh -m 0 -K 0 detects 0 errors here too, and the native build is clean with gcc.

This recipe should be part of the pull request. With the shipped ruleset the change is otherwise inert, and the grammar from "Seeing it" is not available to reviewers. A forced ceiling makes the split reproducible on the normal 17 MiB pool in a few seconds. You already used the same kind of override for the gfx1030 disassembly, so this only needs to be documented.

## The 4 GiB ceiling should be a separate pull request

The need > 0xffffffff guard fixes a bug that already exists in master. Once the terminal data exceeds 4 GiB, the running offset wraps while the pool is packed and candidates are built from the wrong bytes without an error. Your own readback shows the same problem on the stock tree, with 2 of 921 lists wrong and none of the pool splitting changes applied, and you say yourself that one of the four cards here will allocate such a pool in one piece.

That fix does not depend on this pull request. I moved those 34 lines onto the merge base by themselves and they build cleanly. Splitting the pool makes grammars of this size more practical, which is a good reason to address the limit now, but it does not make the two changes dependent on each other.

CONTRIBUTING.md asks for one problem per pull request and for bugs to have an issue behind them, and this also matches 4827 and 4828, where the fallback was separated from 4825 and merged independently. I would open the 4 GiB fix against master on its own. It can go in immediately, where this pull request still has the device name question below open, and the refusal path then gets discussed on its own terms: whether handing the run to the host engine is the right answer, or whether the least probable terminal lists should be trimmed until the pool fits.

## pcfg_pool_shared () should not decide this from the device name

    if (strstr (device_param->device_name, "AMD") != NULL) return false;

The device name is being used here as a proxy for a capability, and that can be wrong in both directions. A unified memory Metal device with a different name could miss the handover, while any device containing "AMD" in its name would lose it regardless of whether it is affected.

The behavior you describe originates in hc_mtlCreateBuffer (), which returns a Managed buffer when Shared was requested and then refuses a host pointer on anything but Shared. I would solve it at the wrapper level: let the wrapper take the pointer or fall back to a copy and report which path it used, so pcfg_pool_shared () decides from the buffer it actually got rather than the device name.

I am also not sure how the AMD branch is reached. hc_mtlDeviceGetAttribute () derives MTL_DEVICE_ATTRIBUTE_UNIFIED_MEMORY from hasUnifiedMemory, which is NO for a discrete AMD GPU, so the first check in pcfg_pool_shared () should already return false before the name is read. If you saw this on a device where hasUnifiedMemory is YES and the name contains AMD, that is worth documenting. If the branch was inferred from surrounding behavior instead, the comment above it should say so rather than read like an observed failure.

## A smaller one

The checks table currently summarizes the test suite output. CONTRIBUTING.md asks for the tail of the actual run through the line that reports the number of detected errors. That is the part a reviewer needs to see, so I would paste it directly rather than describe it.

The rest looks good to me. The arithmetic remains valid for every part count: base times four stays within a u32 because the final part begins three quarters into a pool capped at 4 GiB, and the 0xffffffff sentinel for unused parts falls out correctly for two and three part pools. kernel_params reaches index 30 with PARAMCNT at 64. inc_types.h is part of the checksummed set, so growing the structure invalidates the kernel cache, and bit 19 keeps split and unsplit kernels separate. The buffers are released in backend_session_destroy () before the feed releases the pool in global_term (), which is the required order when the device holds a handed over host pointer.

matrix:
Thank you for rebuilding it and for the detail. All four points are done. Below is what changed and what I found while checking the two technical ones, including a place where you were right about something I had got wrong.

## The 4 GiB ceiling is now its own pull request

You are right, and the reasoning I had for keeping it here was the wrong one. I had it in this change because splitting the pool takes the bug from needing a card with an unusually large single allocation to fitting almost any card, and I treated that as a dependency. It is not one: it is a reason to fix it now, which is a different thing, and your own test settles it, since the guard applies to the merge base by itself.

It is out of this pull request and open separately against master as #4835, with the readback that shows the corruption on the stock tree, and with the question you raise answered there rather than here: the separate change hands the run to the host engine, and says plainly that trimming the least probable terminal lists until they fit is the better cure and belongs beside the change that already has to trim them to fit a device's single buffer.

Removing it puts the large grammar back on the device engine in this pull request, reading the pool at wrapped offsets, exactly as master does today. That is the state the separate fix addresses.

## pcfg_pool_shared () no longer reads the device name

You were right twice here, and the second one matters more than the first.

The name is gone. The wrapper now decides, as you suggested. `hc_mtlCreateBuffer ()` falls back to a buffer of its own instead of failing when it cannot take the pointer, and says which of the two it did:

```c
mem->buf_host = 0;

if ((ptr != NULL) && (bufferOptions == MTLResourceStorageModeShared))
{
  mem->buf_ptr  = [metal_device newBufferWithBytesNoCopy: ptr length: size options: bufferOptions deallocator: nil];
  mem->buf_host = 1;
}
else
{
  mem->buf_ptr = [metal_device newBufferWithLength: size options: bufferOptions];
}
```

and the caller reads the answer instead of predicting it:

```c
if (device_param->metal_d_pcfg_pool[i].buf_host == 1) continue;
```

`pcfg_pool_shared ()` is now only about the memory:

```c
static bool pcfg_pool_shared (const hc_device_param_t *device_param)
{
  if (device_param->device_host_unified_memory == 0) return false;

  if (device_param->is_metal  == true) return true;
  if (device_param->is_opencl == true) return true;

  return false;
}
```

On your second question: the branch was inferred, not observed. I read it out of `hc_mtlCreateBuffer ()` and never saw it happen, because there is no Mac with an AMD GPU here. The comment stated it as a fact anyway, which is the part you should not have had to catch. And you are right that it was unreachable: `MTL_DEVICE_ATTRIBUTE_UNIFIED_MEMORY` comes straight from `hasUnifiedMemory`, so the first test in that function already returns false for a discrete AMD GPU and the name was never read.

The comment on `pcfg_pool_budget ()` now also says what it can and cannot know, since it runs before any buffer exists: it assumes the offer is taken, and a device that is offered the pointer and declines it holds a copy that the budget did not count.

## The recipe is in the pull request

It is, with your override, and I ran it here. Same binary, the same 9998 digests taken one in forty out of the first 400000 candidates, the same limit, and the outfile sorted and hashed. On an RTX 4080 under CUDA:

```
                        recovered            progress              outfile
unchanged tree          9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
this change, 1 buffer   9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 2 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 3 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 4 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab

* Device #1: pcfg pool of 17 MiB in 4 buffers of 4 MiB, 4 MiB, 4 MiB, 4 MiB
```

And on the pocl CPU device, which reports unified memory and so exercises the handover at the same time:

```
unchanged tree          9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
this change, 1 buffer   9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
forced to 4 buffers     9974/9998 (99.76%)   601105114/548000000   34ad50e9efb387ab
```

## The checks section is the run, not a description of it

Replaced with the tail of each, through the line that reports the count.

PR #4834, opened 2026-09-09
Filter pcfg candidates the hash mode cannot accept

matrix:
The pcfg feed enumerates the same keyspace whatever the hash mode takes, so a run against a kernel that stops at 8 bytes still builds every candidate the grammar spells and then throws most of them away. This teaches the feed to read the hash mode's length bounds and not produce that work at all.

## Reproduction

```
$ ./hashcat -m 1500 -a 4 --keyspace pcfg/default-passwords.tar.xz

before: 12747516022634, rc=0
after:      14930777890, rc=0

$ ./hashcat -m 0 -a 4 --keyspace pcfg/default-passwords.tar.xz

before: 12747516022634, rc=0
after:  12747516022634, rc=0

```

`-m 1500` is descrypt, whose kernel takes 8 bytes. `-m 0` takes 256 and is the control: nothing is bounded there, and the keyspace is unchanged to the digit.

What the difference is made of, on one RTX 4080. The target is descrypt of `q7Xz!9Kw`, a password the shipped ruleset does not contain, so the run does not stop before the limit:

```
$ ./hashcat -m 1500 -a 4 --potfile-disable --runtime 60 --status --status-timer 58 -d 1 zzSpQ80B2atvk pcfg/default-passwords.tar.xz

before: Rejected 93816284787/137096760921 (68.43%), Speed 699.0 MH/s, Status Aborted (Runtime), rc=4
after:  Rejected           238/52915120166  (0.00%), Speed 904.6 MH/s, Status Aborted (Runtime), rc=4

```

Two thirds of everything the run produced was built and refused. The candidates that survive are the same ones, so the same password still cracks, and at the same place:

```
$ ./hashcat -m 1500 -a 4 --potfile-disable -d 1 aaAN1ZUwjW7to pcfg/default-passwords.tar.xz

before: aaAN1ZUwjW7to:123456, Status Cracked, rc=0
after:  aaAN1ZUwjW7to:123456, Status Cracked, rc=0

```

## What was wrong

`fill_generic ()` measures every candidate against `hashconfig->pw_min` and `pw_max` and rejects it there, after the feed has built it. The feed never asks for those bounds, so it cannot avoid the work.

`global_init ()` now reads them. `hashconfig_init ()` settles them in `outer_loop ()` before `generic_ctx_init ()` loads the feed, and for this attack mode nothing writes them afterwards.

Three places use them:

* A structure is admitted only where the bytes it can produce can land inside the bounds. The test is on the range, not on one length, so a structure holding a flat `X` or `Y` token is dropped only when even its shortest entry is too long, or its longest still too short.
* The OMEN escape does not go through the grammar, so it gets the same test where it enumerates its own lengths. `k` there is the length less the context, and the model's own `ip_len` and `tr[].clen` give the bytes, so the test is in the same unit as the bounds.
* Each slot carries the byte bounds its buckets have to sit in: `pw_max` less what the other slots spend at their shortest, `pw_min` less what they can reach at their longest. A bucket outside them cannot appear in an acceptable candidate whatever the other slots choose. One predicate is read by everything that counts and everything that unranks, so the two cannot disagree.

Lengths come from the terminal lists, not from the token length the grammar spells: that number counts characters, `pw_min` and `pw_max` count bytes, and `Alpha/8.txt` can hold an entry of 8 characters and 16 bytes.

## What it does not change

Where the hash mode bounds nothing, no bounds array is allocated, `pcfg_ident_tables ()` hashes what it hashed before, and cached unit tables stay valid. That is the `-m 0` row above.

Where it does bound something, the keyspace changes, so `--skip`, `--restore` and a distributed split are not interchangeable between two hash modes over one ruleset. They already were not between two values of `scale` or `costmax`.

## Effect elsewhere

Two rulesets trained with hpcfg on hashmob dumps, `-m 1500`, one RTX 4080:

| ruleset | structures | rejected before | rejected after |
|---|---|---|---|
| 21.8M structures, 13 GB | 21798964 to 15439 | 42.58% | 0.01% |
| 368573 structures, 1.6 GB | 368573 to 11285 | 100.00%, 0 H/s | 0.00%, 147.3 MH/s |

The second one is the case that motivated the last part of the patch. Its `Alpha` lists hold ASCII and Cyrillic entries of the same character length, so no per-structure bound separates them, and on the device engine the head of the run produced nothing the hash mode would take.

`-m 14000`, whose kernel takes exactly 8 bytes, exercises the lower bound: 77.50% rejected before, 0.00% after, 4614.8 MH/s to 5760.9 MH/s.

## tools/test_edge.sh

```
$ ./tools/test_edge.sh -m 1500 -d 1

before: All tests done in 0d:00h:03m:12s, Errors detected: 85
after:  All tests done in 0d:00h:03m:12s, Errors detected: 85

```

The 85 are the same on both sides and none of them is new. Every one of them, compared line by line out of `test_edge.details.log`, differs from what the harness expected only in the top bit of one or more bytes, and in nothing else. That bit is the one descrypt does not use: it takes seven bits of each key byte, so hashcat reports a password that hashes to the target without being the byte string the harness generated, and the harness compares bytes. Both of these are the same hash:

```
$ perl -e 'use bytes; print crypt(pack("H*","39303131eab08039"), "52"), "\n"'
52BcfoVxnAFtc

$ perl -e 'use bytes; print crypt(pack("H*","393031316a308039"), "52"), "\n"'
52BcfoVxnAFtc

$ echo '$HEX[39303131eab08039]' | ./hashcat --quiet --potfile-disable --self-test-disable -m 1500 52BcfoVxnAFtc -a 0
52BcfoVxnAFtc:$HEX[393031316a308039]

```

The two byte strings differ only in the top bit of two bytes. The five failures under `-a 4` are the same thing: the attack finds the word, and the plain it reports back is the equivalent one. This is not part of this pull request and is left alone.

### DIFF
--- a/src/feeds/feed_pcfg.c
+++ b/src/feeds/feed_pcfg.c
@@ -283,6 +283,14 @@ typedef struct
 
   u32 pwmin;
   u32 pwmax;
+
+  // Structures those bounds took out of the grammar, and whether any bucket is bounded at all. The
+  // first names the reason where a run ends up with nothing to enumerate. The second keeps a run the
+  // hash mode does not bound on exactly the path it was on before.
+
+  u32  out_of_range;
+  bool bounded;
+
   u32 kbits;
   u32 threads;
   bool walk;
@@ -2927,6 +2935,8 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
         memcpy (sc, cbuf, (size_t) s.nslot * sizeof (u16));
 
         s.cap = sc;
+
+        pg->bounded = true;
       }
 
       if (fbind == true)
@@ -2938,6 +2948,8 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
         memcpy (sf, fbuf, (size_t) s.nslot * sizeof (u16));
 
         s.flr = sf;
+
+        pg->bounded = true;
       }
     }
 
@@ -2967,19 +2979,24 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
 
     roots_join (named, sizeof (named), roots, nroots);
 
-    if (dropped_t > 0)
-    {
-      gerr (global_ctx, "%s: all %u structures were dropped, most likely a terminal list that could not be read", named, dropped_t);
-
-      return -1;
-    }
-
-    // A hash mode that excludes the whole grammar is not a broken ruleset, and saying so as one
-    // would send the user looking for a problem in the wrong place.
+    // The hash mode's own bounds took every structure. That is not a ruleset that cannot be read and
+    // not a grammar priced past costmax: the escape enumerates its own lengths, it may still hold
+    // some this mode accepts, and omen_load () has not run yet. So the run goes on. Where the escape
+    // has nothing in range either, the keyspace comes out zero and the attack is refused there, with
+    // the bounds named as the reason. It comes first because a ruleset carrying a few unusable
+    // structures is normal, and the shipped one does, so testing that first would answer every
+    // bounded run with a terminal list that could not be read.
 
     if (dropped_p > 0)
     {
-      gerr (global_ctx, "%s: all %u structures lie outside the %u to %u bytes this hash mode accepts", named, dropped_p, pg->pwmin, pg->pwmax);
+      if (global_ctx->quiet == false)
+      {
+        pmsg (pg, "pcfg: every structure lies outside the %u to %u bytes this hash mode accepts, so the escape is all that is left", pg->pwmin, pg->pwmax);
+      }
+    }
+    else if (dropped_t > 0)
+    {
+      gerr (global_ctx, "%s: all %u structures were dropped, most likely a terminal list that could not be read", named, dropped_t);
 
       return -1;
     }
@@ -2988,14 +3005,14 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
     // holds, and master refuses it. The M line exemption below is for a grammar that never had
     // structures, not for one whose structures were all put out of reach.
 
-    if (dropped_c > 0)
+    else if (dropped_c > 0)
     {
       gerr (global_ctx, "%s: all %u structures cost more than costmax %" PRIu64 ", nothing is reachable", named, dropped_c, pg->costmax / pg->scale);
 
       return -1;
     }
 
-    if (dropped_m == 0)
+    else if (dropped_m == 0)
     {
       gerr (global_ctx, "%s: nothing to enumerate, no structures and no M lines", named);
 
@@ -3020,7 +3037,8 @@ static int grammar_load (generic_global_ctx_t *global_ctx, pcfg_global_t *pg, co
 
   if (global_ctx->quiet == false) pmsg (pg, "pcfg: suffix tables built in %s", pcfg_duration ((hc_timer_get (t_sweep) / 1000.0), display, sizeof (display)));
 
-  pg->m_lines = dropped_m;
+  pg->m_lines      = dropped_m;
+  pg->out_of_range = dropped_p;
 
   if (global_ctx->quiet == false)
   {
@@ -8227,7 +8245,11 @@ bool global_init (generic_global_ctx_t *global_ctx, MAYBE_UNUSED generic_thread_
 
     roots_join (named, sizeof (named), roots, nroots);
 
-    if ((pg->omen_cnt == 0) && (pg->structs_cnt == 0))
+    if ((pg->structs_cnt == 0) && (pg->out_of_range > 0))
+    {
+      gerr (global_ctx, "%s: nothing to enumerate, the %u to %u bytes this hash mode accepts leave the grammar empty and the escape holds no length in range", named, pg->pwmin, pg->pwmax);
+    }
+    else if ((pg->omen_cnt == 0) && (pg->structs_cnt == 0))
     {
       gerr (global_ctx, "%s: nothing to enumerate, no structures and no escape to carry", named);
     }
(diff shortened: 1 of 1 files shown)


jsteube:
Filtering the grammar against the hash mode's bounds is the right idea, and the parts I could check hold up. One thing has to be fixed before it can go in.

A hash mode that ran before now fails:

```
$ ./hashcat -m 22001 -a 4 --keyspace pcfg/default-passwords.tar.xz

before: 34935457766817424, rc=0
after:  pcfg/default-passwords.tar.xz: all 4 structures were dropped, most likely a terminal list that could not be read
        rc=255
```

-m 22001 has a pw_min of 64, so every structure in the shipped ruleset falls under the floor and structs_cnt comes out zero. Two things then go wrong.

The message is wrong, and it is the one everyone will see. The dropped_t branch is tested before the new dropped_p branch, and the shipped ruleset carries exactly 4 unusable structures, so dropped_t is never zero there and wins every time. Your own new message is unreachable on the ruleset hashcat ships with, and the user is sent looking for a terminal list that could not be read when nothing was unreadable. The dropped_p test wants to come first, or the two want to be one message that names both counts.

The refusal itself is the larger question. It fires in grammar_load (), and omen_load () does not run until afterwards, so the escape is never asked. That matters here because this same patch teaches the escape to bound itself in bytes, and because the structs_cnt == 0 block it sits in already treats an escape only grammar as a real run: its only fatal case is dropped_m == 0, no structures and no M lines. The shipped ruleset has M lines. So a grammar whose structures the hash mode excludes but whose escape still holds acceptable lengths is refused, where master ran it. That is the same shape as 4827 and 4828.

I do not think -m 22001 against this ruleset is a run anybody wants. But it is a run that worked, and the reason it stops working should be the escape having nothing in range rather than the structure loop returning early.

Everything else I checked came back clean. The keyspace figures reproduce exactly here, and -m 14100, whose pw_min is 24, is another good row:

```
                       before               after
-m 1500        12747516022634          14930777890
-m 14000       12747516022634          14481523132
-m 14100       12747516022634           1608688849
-m 0           12747516022634       12747516022634
```

Where nothing binds the change is provably inert. Exhausting the grammar at costmax=24 and comparing the candidate streams, both trees produce the same 1958217 candidates, byte identical as a set.

Where it does bind, the surviving candidates are the same ones. I took that same stream, split it at descrypt's 8 bytes, and asked the feed's own lookup= for a sample of each side under -m 1500. All 50 of the sample over 8 bytes are gone. 99 of the 100 under it are still reached, and the hundredth is identically unreachable on master for an unrelated reason, cost 27 against costmax 24. aaAN1ZUwjW7to:123456 cracks on both.

The cache key is right, which is the thing I looked at first. pcfg_ident_tables () hashes cap and flr only where they exist, so an unbounded run keys as it did before. That also covers the suffix count cache that just landed, since build_suffix () is one of the walks that now reads bucket_out ().

Two smaller things. --stdout refuses -m, so a -a 4 preview runs with no hash mode and therefore no bounds. Before this patch that did not matter, because the stream did not depend on the mode. Now --stdout shows a stream the real run will not produce for any bounded mode, and that is worth a line in the description, since --stdout is how people check what an attack is going to do.

And the probe cell built at the end of global_dev_init () picks the widest bucket in the whole grammar without asking bucket_out (), so on a bounded run it can be sized on a bucket the run never uses. It is autotune sizing rather than enumeration, so nothing is wrong with the candidates, but it is the only bucket walk outside the diagnostics that did not get the predicate.

The rest reads well. The byte versus character distinction is the part that would have been easy to get wrong and it is handled in all three places, including the k to length conversion in the escape. Reading the per slot minimum and maximum straight after list_get () rather than keeping a pointer is the right call given that list can move. The bounds arrays are only allocated where a bound actually excludes something, which is what keeps the unbounded run identical rather than merely equivalent.

matrix:
Both fixed, and you were right about which one was the larger question.

The refusal is gone from `grammar_load ()`. The count of structures the bounds excluded is tested first now, and it does not return `-1`: it says what happened and the load carries on, so `omen_load ()` runs and the escape gets asked. Where the escape holds nothing in range either, the attack is refused where the keyspace is counted, and the reason named is the one you wanted:

```
$ ./hashcat -m 22001 -a 4 -S --potfile-disable 'WPA*01*5ce7ebe97a1bbfeb2822ae627b726d5b*27462da350ac*accd10fb464e*686173686361742d6573736964***' pcfg/default-passwords.tar.xz

pcfg: grammar read in 0s 001ms, 0 structures
pcfg: every structure lies outside the 64 to 64 bytes this hash mode accepts, so the escape is all that is left
pcfg: 0 structures, 146 terminal lists, 4 unusable dropped, 23154 outside 64-64 bytes
pcfg: OMEN escape held to 64-64 bytes, 18 lengths left out
pcfg: OMEN escape dropped, no length it holds is one this hash mode accepts
pcfg/default-passwords.tar.xz: nothing to enumerate, the 64 to 64 bytes this hash mode accepts leave the grammar empty and the escape holds no length in range
rc=255
```

I took the first of the two options you offered rather than one message naming both counts, because once `dropped_p` stops being fatal the two are no longer the same kind of statement: one is a note about what the load dropped, the other is a refusal. Merging them would have meant either refusing again or turning `dropped_t` into a note, and a terminal list that could not be read should still stop the run.

A grammar the bounds empty out but whose escape still holds acceptable lengths now runs, which is the case master ran and this branch had been refusing.

The probe cell reads `bucket_out ()` now. It walks the head of the grammar the way `probe_n` already bounds itself, and only where something is actually bounded; where nothing is, the list walk is untouched and the probe comes out the same bucket it did before.

`--stdout` is in the description now, under its own heading.

One more thing, since it landed while this was open: #4832 is in master, so this branch is behind it. The two touch adjacent lines in `grammar_load ()` and plain `git apply` fails in both orders, but a three way merge is clean and I have carried the branch onto current master locally and re-run everything on it. The suffix count cache and this change agree: a bounded run and an unbounded one over the same ruleset key to different `.suf` files, verified byte for byte on both an RTX 4080 and an M4 Pro.

```
162811c28a80fce9-1-64-0-1.suf        -m 0
192289205869b32e-1-64-0-1.suf        -m 1500
```

PR #4839, opened 2026-09-11
Add the sm3crypt pure kernel so mode 35100 takes a password over 15 bytes

matrix:
`-m 35100` cannot crack a password longer than 15 bytes. The module says it can.

## Reproduction

The hash is what `tools/test.pl passthrough 35100` produces for the password used as the mask, which is 20 characters.

```
$ ./hashcat -m 35100 -a 3 -d 1 --potfile-disable '$sm3$995050$B8DvnZ8bKWnPIRcIDIoNfUSAybQNmD.uwbPGxlU9rT2' AbcdefghijKlmnopqrst

before: Skipping mask 'AbcdefghijKlmnopqrst' because it is larger than the maximum password length.
        rc=255

after:  Kernel.Feature...: Pure Kernel (password length 0-256 bytes)
        Status...........: Cracked
        Guess.Mask.......: AbcdefghijKlmnopqrst [20]
        $sm3$995050$B8DvnZ8bKWnPIRcIDIoNfUSAybQNmD.uwbPGxlU9rT2:AbcdefghijKlmnopqrst
        rc=0
```

No `-O` is passed. That is the path where the module promises the full length, and the mask is 20 characters against the 15 the optimized kernel takes.

## What was wrong

`module_35100.c` chooses the limit the way the other crypt modes do:

```c
u32 module_pw_max (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra)
{
  u32 pw_max = PW_MAX;

  const bool optimized_kernel = (hashconfig->opti_type & OPTI_TYPE_OPTIMIZED_KERNEL);

  if (optimized_kernel == true)
  {
    pw_max = 15;
  }

  return pw_max;
}
```

md5crypt and sha256crypt carry the same `OPTS_TYPE` and the same choice, sha512crypt the same choice with one flag fewer, and each of the three ships two kernels:

```
m00500-optimized.cl  m00500-pure.cl
m01800-optimized.cl  m01800-pure.cl
m07400-optimized.cl  m07400-pure.cl
m35100-optimized.cl
```

sm3crypt shipped one. With no `m35100-pure.cl` to load, every run takes the optimized kernel whatever the user asked for, so the `PW_MAX` arm is unreachable and a password over 15 bytes is rejected instead of tried.

## What this adds

`OpenCL/m35100-pure.cl`, and nothing else. No module changes.

The `sm3crypt_tmp_t` it uses is the struct `m35100-optimized.cl` already declares, under a comment that reads `// pure version`:

```c
typedef struct sm3crypt_tmp
{
  // pure version

  u32 alt_result[8];
  u32 p_bytes[64];
  u32 s_bytes[64];

} sm3crypt_tmp_t;
```

So `module_tmp_size` already returns the right size for it, and the two kernels agree on the layout the way `m00500`'s two do.

Benchmarks do not move, because a benchmark always takes the optimized kernel. Three runs each on an Apple M4 Pro, `-m 35100 -b -d 1`:

```
before: 173.2  172.7  173.7 kH/s      median 173.2, peak 173.7
after:  173.7  173.6  173.7 kH/s      median 173.7, peak 173.7
```

The command above cracks on both backends this machine has, each reporting `Cracked`, `Pure Kernel (password length 0-256 bytes)` and `Guess.Mask AbcdefghijKlmnopqrst [20]`: Metal as `-d 1` and OpenCL as `-d 2` on an Apple M4 Pro.

## tools/test_edge.sh

```
$ ./tools/test_edge.sh -m 35100 -d 1      # Metal, Apple M4 Pro

before: rc=0, 10m59s, Errors detected: 0
after:  rc=0, 20m19s, Errors detected: 0
```

The run takes about twice as long because the suite now has two kernel types to cover for this mode instead of one. Every `Kernel-Type pure` line in the log is a combination that did not exist before.

### DIFF
--- a/OpenCL/m35100-pure.cl
+++ b/OpenCL/m35100-pure.cl
@@ -1,0 +1,370 @@
+/**
+ * Author......: See docs/credits.txt
+ * License.....: MIT
+ */
+
+#ifdef KERNEL_STATIC
+#include M2S(INCLUDE_PATH/inc_vendor.h)
+#include M2S(INCLUDE_PATH/inc_types.h)
+#include M2S(INCLUDE_PATH/inc_platform.cl)
+#include M2S(INCLUDE_PATH/inc_common.cl)
+#include M2S(INCLUDE_PATH/inc_hash_sm3.cl)
+#endif
+
+#define COMPARE_S M2S(INCLUDE_PATH/inc_comp_single.cl)
+#define COMPARE_M M2S(INCLUDE_PATH/inc_comp_multi.cl)
+
+typedef struct sm3crypt_tmp
+{
+  // pure version
+
+  u32 alt_result[8];
+  u32 p_bytes[64];
+  u32 s_bytes[64];
+
+} sm3crypt_tmp_t;
+
+KERNEL_FQ KERNEL_FA void m35100_init (KERN_ATTR_TMPS (sm3crypt_tmp_t))
+{
+  /**
+   * base
+   */
+
+  const u64 gid = get_global_id (0);
+
+  if (gid >= GID_CNT) return;
+
+  /**
+   * init
+   */
+
+  const u32 pw_len = pws[gid].pw_len;
+
+  u32 w[64] = { 0 };
+
+  for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+  {
+    w[idx] = pws[gid].i[idx];
+  }
+
+  for (u32 i = 0, idx = 0; i < pw_len; i += 4, idx += 1)
+  {
+    w[idx] = hc_swap32_S (w[idx]);
+  }
+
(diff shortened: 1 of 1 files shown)


jsteube:
From the description:

```
Benchmarks do not move, because a benchmark always takes the optimized kernel.
```

True of "-b", but the default path did move. Before, every run took the optimized kernel because there was no other one to take. Now a run without "-O" takes the new pure kernel, on an RTX 4090 with a seven character mask:

```
pure        1042329 H/s
optimized   2953360 H/s
```

Worth a line in the description, because somebody who upgrades and does not pass "-O" loses 2.8x and nothing warns them.

Those two numbers come from a mask run rather than from "-b", because the benchmark sets optimized_kernel unconditionally and cannot reach the new kernel at all.

PR #4845, opened 2026-09-11
Size the autotune launch by the bridge's own power, not by threads

matrix:
`-m 73000` does not run on CUDA or HIP. It reports an illegal memory access and 0 H/s, because the autotune sizes its measuring launches by threads on a mode whose work is not counted in threads.

## Reproduction

`-R` picks the physical device for a bridged mode, which `docs/hashcat-assimilation-bridge.md` explains: *"`-d` selects which units run. ... `-R` selects the physical device that generates the candidates, which is a separate question."*

```
$ ./hashcat -b -m 73000 -R 1                     # CUDA, RTX 4080

before: cuStreamSynchronize(): an illegal memory access was encountered
        Speed.#*.........:        0 H/s (0.00ms) @ Accel:141 Loops:1
        rc=255
after:  Speed.#*.........:     2743 H/s (0.00ms) @ Accel:141 Loops:1
        rc=0

$ ./hashcat -b -m 73000 -R 2                     # HIP, RX 6900 XT

before: hipEventSynchronize(): an illegal memory access was encountered
        hipLaunchKernel(): an illegal memory access was encountered
        * Device #1: skipped, due to kernel autotune failure (-5).
after:  Speed.#*.........:     3046 H/s (0.02ms) @ Accel:141 Loops:1
        rc=0
```

Three runs each after the change: CUDA 2739, 3048, 2743 H/s, median 2743. HIP 3050, 2746, 3046 H/s, median 3046.

```
$ compute-sanitizer --tool memcheck ./hashcat -b -m 73000 -R 1

before: ========= Invalid __global__ read of size 4 bytes
        =========     at m73000_init+0xf0
        =========     by thread (0,0,0) in block (5,0,0)
        ========= ERROR SUMMARY: 9 errors
after:  ========= ERROR SUMMARY: 0 errors
```

## What was wrong

A bridge counts its work in units rather than in waves of threads: `kernel_accel` there is a candidate count and `hardware_power` is `bridge_workitem_multiple ()`, so the thread count does not enter. `size_pws` is sized by that rule at `backend.c:19534`, and so is every other launch size, including `autotune.c:90` and `autotune.c:931`. Three places in `autotune.c` did not branch:

```c
const u32 kernel_power_full = ((hashconfig->opts_type & OPTS_TYPE_MP_MULTI_DISABLE) ? 1 : device_param->device_processors)
                            * ((hashconfig->opts_type & OPTS_TYPE_THREAD_MULTI_DISABLE) ? 1 : threads) * accel;
```

On this bridge that is `1 * 32 * 160`, which is 5120 against a buffer of 160. They feed the measuring launches of the init and comp kernels, of init2 and loop2p, and of loop2. All three now ask one function for that number, so one place is left that can be wrong about it instead of three.

The comment at `backend.c:19490` describes the same shape from the buffer side, where it was found and fixed once: *"On a 496 core unit that is 66 candidates' worth of buffer for a launch of 32,736, which the GPU reports as an illegal memory access from inside the init kernel, with nothing to say the size was the problem."* This is the launch side of that rule.

## Why only two backends report it

The buffer the oversized launch runs off the end of is device memory:

```
$ grep -n 'd_pws_buf' src/backend.c | grep MemAlloc
19973:      if (hc_cuMemAlloc (hashcat_ctx, &device_param->cuda_d_pws_buf,      size_pws)      == -1) return -1;
20017:      if (hc_hipMemAlloc (hashcat_ctx, &device_param->hip_d_pws_buf,      size_pws)      == -1) return -1;
```

CUDA and HIP check the bounds of their own allocations and refuse the read. The CPU runtimes on the same machine do not, so with the bug still in place the mode reports a speed and finishes:

```
$ ./hashcat -b -m 73000 -R 3 -D 1     # PoCL 6.0, unpatched:  rc=0, 3009 H/s
$ ./hashcat -b -m 73000 -R 6 -D 1     # Intel OpenCL, same:   rc=0, 2795 H/s
```

Apple is the same: on an M4 Pro the mode gives 382 H/s before the change and 380 after, on Metal. The launch is wrong there too, it is simply not reported.

This also rules out `-fsanitize=address`. A `make DEBUG=2` build of both trees reports no overflow on either side, because AddressSanitizer instruments the host heap and `cuMemAlloc` and its siblings do not allocate from it. `compute-sanitizer` instruments device memory, which is why it is the instrument above.

## What it does not change

Where no bridge is active the expression is the one that was there. Three runs each, same machine:

```
$ ./hashcat -b -m 0 -d 1
before: 94092.9  94552.0  94624.7 MH/s     median 94552.0
after:  94041.4  94390.8  94555.5 MH/s     median 94390.8

$ ./hashcat -b -m 72000 -R 1        # pyenv 3.14.7t
before: 553  582  559 H/s                  median 559
after:  582  564  573 H/s                  median 573
```

On an Apple M4 Pro `-m 0` is 11093.0 MH/s before and 11093.4 after, as medians of three.

## Four copies of the same rule, folded in

Review asked for the four places that still spelled the expression out to call the function instead. The thread count was the only difference between them, and it is the function's parameter, so all four substitute verbatim:

```
try_run ()    hardware_power                  kernel_threads
autotune ()   device_param->hardware_power    kernel_threads_min
autotune ()   hardware_power_max              kernel_threads_max
autotune ()   hardware_power                  device_param->kernel_threads
```

`hardware_power_max` is the one that matters on its own: it sizes the `pws_buf[]` the three launches were overrunning, so the rule and the buffer that depends on it now come from one place rather than two.

The function moves above `try_run ()` to be in scope there, next to `autotune_kernel_power ()`. That also leaves the comment describing `autotune2_fixed_msec ()` attached to it again, which it was not while the function sat between them.

`bridge_workitem_multiple ()` is now named once in the file instead of five times, and the diff is eight lines shorter than before.

Substituting verbatim is checkable, so it was checked. The else branch of the open coded form is `(A) * (B)`, because `*` binds tighter than `? :`, which is what the function computes after its early return; enumerating both branches over the four flags and a range of processor, thread and workitem counts gives 216 cases and no disagreement. On the device, the autotune picks the same parameters before and after, on both branches and both machines, three runs each:

```
                                   before                              after
macOS   -m 0       Accel:474 Loops:1024 Thr:256 Vec:1   Accel:474 Loops:1024 Thr:256 Vec:1
macOS   -m 73000   Accel:94 Loops:1                     Accel:94 Loops:1
Linux   -m 0       Accel:141 Thr:1024 Vec:8             Accel:141 Thr:1024 Vec:8
Linux   -m 73000   Accel:141 Loops:1                    Accel:141 Loops:1
```

`Loops` on `-m 0` under Linux lands anywhere between 772 and 841 from run to run, on both sides alike, which is the loop search and not this change.

## The fault this used to report is gone

An earlier version of this description said that with `-m 73000` running, a second fault became visible on Linux: on Python 3.14 the process faulted during `_Py_Finalize` after the run, in a few runs out of twenty, from a reference counting error in `bridge_python_generic_hash_mp.c`.

#4840 and #4846 fixed that. Twenty runs on master with this branch merged into it, Python 3.14.7, the same machine and the same command as before:

```
$ for i in $(seq 1 20); do ./hashcat -b -m 73000 -R 1 >/dev/null 2>&1; echo -n "$? "; done
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0

Speed.#*.........:  3273  3259  3217  3179  3230 H/s
```

Nothing is left over for a separate report.

## tools/test_edge.sh

`-m 73000` needs the bridge and a Python, and the suite has no case for it. The mode is exercised by the runs above instead, on two vendors and two CPU runtimes.


jsteube:
In "autotune.c" this statement is still open coded four times:

```
  const u32 hardware_power = bridge_active (hashcat_ctx, device_param->bridge_link_device) ? bridge_workitem_multiple (hashcat_ctx, device_param->bridge_link_device)
                           : ((hashconfig->opts_type & OPTS_TYPE_MP_MULTI_DISABLE)     ? 1 : device_param->device_processors)
                           * ((hashconfig->opts_type & OPTS_TYPE_THREAD_MULTI_DISABLE) ? 1 : kernel_threads);
```

Before, the three launch sites computed the same thing without the bridge branch, which is the bug. Now they call autotune_hardware_power(), but these four still spell it out: once each for kernel_threads, kernel_threads_min, kernel_threads_max and device_param->kernel_threads. That thread count is the only difference between them and it is the function's parameter, so all four substitute verbatim.

The hardware_power_max one is worth folding in even on its own. It sizes pws_buf[] that the three launches were overrunning, so the rule and the buffer that depends on it end up in one place instead of two.

From the description:

```
It is a reference counting error in bridge_python_generic_hash_mp.c
and it has nothing to do with this file. It is reported separately.
```

That may already be gone. PRs #4840 and #4846 landed today, both on reference counts in that file. Worth re-running your twenty on current master before opening anything for it.

matrix:
Both done, and the branch is merged up to master, which it was 18 commits behind.

**The four copies.** Folded, and `autotune_hardware_power ()` moves above `try_run ()` to be in scope there, next to `autotune_kernel_power ()`. `bridge_workitem_multiple ()` is now named once in the file instead of five times, and the diff is eight lines shorter than it was. `hardware_power_max` is in, so the rule and the `pws_buf[]` it sizes come from one place.

You are right that all four substitute verbatim, and since that is checkable I checked it rather than reading it. The else branch of the open coded form is `(A) * (B)`, because `*` binds tighter than `? :`, which is what the function computes after its early return; enumerating both forms over the four flags and a range of processor, thread and workitem counts gives 216 cases and no disagreement. On the device the autotune picks the same parameters before and after, on both branches of the function and on two machines, three runs each:

```
                                   before                              after
macOS   -m 0       Accel:474 Loops:1024 Thr:256 Vec:1   Accel:474 Loops:1024 Thr:256 Vec:1
macOS   -m 73000   Accel:94 Loops:1                     Accel:94 Loops:1
Linux   -m 0       Accel:141 Thr:1024 Vec:8             Accel:141 Thr:1024 Vec:8
Linux   -m 73000   Accel:141 Loops:1                    Accel:141 Loops:1
```

`Loops` on `-m 0` under Linux lands anywhere between 772 and 841 from run to run, on both sides alike, which is the loop search rather than this change.

**The fault in the description.** Gone, as you thought. Twenty runs on master with this branch merged into it, Python 3.14.7, same machine and same command as the original twenty:

```
$ for i in $(seq 1 20); do ./hashcat -b -m 73000 -R 1 >/dev/null 2>&1; echo -n "$? "; done
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0
```

That section is out of the description, and there is nothing left to open separately for it.

jsteube:
In "autotune.c" the four are still spelled out, and the last commit on the branch is the master merge:

```
  const u32 hardware_power = bridge_active (hashcat_ctx, device_param->bridge_link_device) ? bridge_workitem_multiple (hashcat_ctx, device_param->bridge_link_device)
                           : ((hashconfig->opts_type & OPTS_TYPE_MP_MULTI_DISABLE)     ? 1 : device_param->device_processors)
                           * ((hashconfig->opts_type & OPTS_TYPE_THREAD_MULTI_DISABLE) ? 1 : kernel_threads);
```

In the same file, autotune_hardware_power() still sits below try_run(), so the fold looks to be still local. Could you push it?

PR #4852, opened 2026-09-12
Add -M to test_edge.sh, and fix three faults in the same file

matrix:
Review on #4651 asked for the mode loop here to stop globbing `.pm` only, since 1000 and 5200 got a `.py` oracle in #4817 and have been outside this suite from that commit on. That is what this branch now carries beyond `-M` itself, and it turned out to need more than the glob.

## `-M`

Unchanged from the first commit here, and review settled that it stays that way: the benchmark array
in `src/benchmark.c` answers which modes a user wants measured, while `-M` answers the other end, one
mode per distinct code path. `tools/test.sh` now reads six of its 24 out of the family lists defined
in that same file, which cannot be done here because this script has no such lists.

## The glob

```
$ ls tools/test_modules/*.pm | wc -l
511
$ ls tools/test_modules/m[0-9][0-9][0-9][0-9][0-9].pm tools/test_modules/m[0-9][0-9][0-9][0-9][0-9].py | wc -l
513
```

#4817, merged as `731f2ed8c`, gave 1000 and 5200 a `.py` oracle. The loop globbed `*.pm`, so from that commit both modes have been outside this suite, and nothing said so:

```
$ ./tools/test_edge.sh -m 1000 -d 1 -D 2      # before
! cleaning cache ...
[ test_edge_1789300896 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789300896 ] > Errors detected: 0

$ ./tools/test_edge.sh -m 1000 -d 1 -D 2      # after
! cleaning cache ...
[ test_edge_1789300896 ] > Skip processing Hash-Type 1000 (edge is implemented in tools/test.pl only, and this mode's oracle is a .py)
[ test_edge_1789300896 ] > All tests done in 0d:00h:00m:01s
[ test_edge_1789300896 ] > Errors detected: 0

$ ./tools/test_edge.sh -m 5200 -d 1 -D 2      # before
! cleaning cache ...
[ test_edge_1789300897 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789300897 ] > Errors detected: 0

$ ./tools/test_edge.sh -m 5200 -d 1 -D 2      # after
! cleaning cache ...
[ test_edge_1789300897 ] > Skip processing Hash-Type 5200 (edge is implemented in tools/test.pl only, and this mode's oracle is a .py)
[ test_edge_1789300897 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789300897 ] > Errors detected: 0
```

`Errors detected: 0` on all four, and that is the point: before, the zero is over nothing at all, and
nothing in the run says the mode was asked for and not reached.

Anchoring the glob to `mNNNNN` keeps `test_helpers.py` out of it. The two modes cannot be run yet, because an edge case run needs the oracle's `edge` entry point and the Python runner does not implement it:

```
$ python3 tools/test_module_runner.py edge 1000
edge is not implemented in test_module_runner.py yet, use tools/test.pl
$ python3 tools/test_module_runner.py edge 5200
edge is not implemented in test_module_runner.py yet, use tools/test.pl
```

So they are named and skipped with the reason, the way the other skips in that loop read, rather than entering the loop and failing. `edge ()` and `edge_format ()` in `tools/test.pl` are 309 lines of per attack type length constraints, so porting them is a change of its own, and the skip goes away on its own once it exists.

## The skip for 73000

The `continue` sat outside the test for Darwin while the message sat inside it, so on every other platform the mode was skipped in silence. Measured on a Raspberry Pi 4, Linux aarch64, with `pyenv local` set to a free-threaded build so the branch is taken:

```
$ ./tools/test_edge.sh -m 73000 -f
```

before:

```
! cleaning cache ...
[ test_edge_1789260441 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789260441 ] > Errors detected: 0
```

after:

```
! cleaning cache ...
[ test_edge_1789260442 ] > Skip processing Hash-Type 73000 (needs a Python without the 'free-threaded' library support)
[ test_edge_1789260442 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789260442 ] > Errors detected: 0
```

The bridge refuses a free-threaded library wherever it runs, not only on Apple, which is what it says itself:

```
src/bridges/bridge_python_generic_hash_mp.c:541
  event_log_info (hashcat_ctx, "This mode wants an ordinary Python built as a shared library, not the free-threaded one.");
```

so the message names the Python rather than the platform.

## No other mode changes

The new skip fires for exactly the two modes whose oracle is a `.py`, and every other one of the 513 has a `.pm` beside it:

```
modes the glob lists                     513
of those, with no .pm                      2    1000 5200
```

A sample of five modes across the range, run in full on both sides, to show that nothing else moves:

```
mode    before                 after
0       Errors detected: 0     Errors detected: 0     160 combinations each side
100     Errors detected: 0     Errors detected: 0     160
400     Errors detected: 0     Errors detected: 0     180
1400    Errors detected: 0     Errors detected: 0     160
3000    Errors detected: 0     Errors detected: 0      85
```

Same count of attack, kernel, target and vector width combinations on both sides for every one of the
five, so the glob change adds the two modes and touches nothing else.

### DIFF
--- a/tools/test_edge.sh
+++ b/tools/test_edge.sh
@@ -50,6 +50,8 @@ function usage()
   echo ""
   echo "     --skip-clean-cache             : Skip cleaning the kernel caches before starting the tests"
   echo ""
+  echo "-M / --minimal                      : test only 24 hash types covering all distinct code paths, vector-width 1"
+  echo ""
   echo "-f / --force                        : run hashcat using --force"
   echo ""
   echo "-v / --verbose                      : show debug messages (supported: -v or -vv)"
@@ -374,6 +376,7 @@ BACKEND_DEVICES_KEEPFREE=0
 ALL_ATTACKS=0
 SELF_TEST_DISABLE=1
 CLEAN_CACHE_DISABLE=0
+MINIMAL=0
 
 OPTS="--quiet --potfile-disable --machine-readable --logfile-disable"
 
@@ -425,6 +428,12 @@ while [[ $# -gt 0 ]]; do
         usage
       fi
       shift 2
+      ;;
+    --minimal)
+      MINIMAL=1
+      HASH_TYPE="all"
+      VECTOR_WIDTHS="1"
+      shift
       ;;
     --allow-all-attacks)
       ALL_ATTACKS=1
@@ -795,6 +804,11 @@ while [[ $# -gt 0 ]]; do
 
             break
             ;;
+          M)
+            MINIMAL=1
+            HASH_TYPE="all"
+            VECTOR_WIDTHS="1"
+            ;;
           *)
             echo "Unknown option: -$opt"
             usage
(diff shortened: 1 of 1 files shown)


matrix:
Both done. `-M` reads the array here too, the hardcoded list is gone, and the glob no longer takes `.pm` only.

It parses `DEFAULT_BENCHMARK_ALGORITHMS_BUF` the same way `tools/benchmark_deep.pl` does, through `TDIR` the way the rest of this file reaches `src/`, and gives the same answer: 107 modes, the same 107. The help text no longer names a count. The `tools/test.sh` side of the same request is in #4651.

What that widened run turned up is in the description, and one of the three is worth flagging here because it changes what the suite reports rather than only what it covers.

On the glob: #4817, merged as `731f2ed8c`, gave 1000 and 5200 a `.py` oracle instead of a `.pm`, so from that commit both modes have been outside this suite, with nothing said. The glob now takes either extension and is anchored to `mNNNNN`, so `test_helpers.py` does not come along with it, and 511 modes become 513.

The two cannot be run yet. An edge case run needs the oracle's `edge` entry point and `tools/test_module_runner.py` does not implement it:

```
edge and password are not implemented in test_module_runner.py yet, use tools/test.pl.
```

So rather than have them enter the loop and fail, they are named and skipped with the reason, which is how the other skips in that loop read. The skip goes away on its own the day that entry point exists, and porting `edge` looks like a change of its own: `edge ()` and `edge_format ()` in `tools/test.pl` are 309 lines of per attack type length constraints.

matrix:
Reverted here too. The list is the one the first commit had, mode for mode, and the help string goes back with it.

`tools/test.sh` reads six of its 24 out of the family lists defined in that file, which is the part of your first comment that survives. It cannot be done here, because this script has no such lists: it globs the oracles in `tools/test_modules/` and never names a container family.

```
$ grep -cE '^(TC_MODES|VC_MODES|CL_MODES|LUKS1_LEGACY_MODE|LUKS1_MODES|LUKS2_MODES)=' tools/test_edge.sh
0
```

What is left on this branch beyond `-M` is the glob, which stands on its own: 1000 and 5200 got a `.py` oracle in #4817 and the loop globbed `.pm`, so from that commit both were outside the suite with nothing said. They are in it now, and named and skipped rather than run, because an edge case run needs the oracle's `edge` entry point and the Python runner does not have one. The 73000 message that only printed on Darwin is in there too. All three are in the description.

PR #4857, opened 2026-09-13
Fix device temperature readings for Intel GPUs with xe driver

RealEnder:
Older i915 driver exposes temp1_input for device temperature readings.
The new xe driver splits the temperature readings like this:
  temp2_input : Package temperature (the main GPU die/package reading)
  temp3_input : VRAM temperature
  temp4_input : Memory controller temperature
  temp5_input : GPU PCIe temperature
For our case, first try to open the xe package temperature (temp2_input); if it fails, fallback to i915 (temp1_input)


jsteube:
In "ext_sysfs_intelgpu.c" the fallback reassigns path over an allocation that is still live:

```
  hc_asprintf (&path, "%s/temp2_input", syspath);

  if (hc_fopen_raw (&fp, path, "r") == false)
  {

    hc_asprintf (&path, "%s/temp1_input", syspath);
```

Every byte of the temp2_input string is lost there, since hc_asprintf() reaches vasprintf() and allocates. That branch is the common one rather than the rare one: on an A770 under i915 the hwmon directory carries temp1_input and no other temperature input at all, so a machine on the older driver takes the fallback on every reading.

The monitor thread reads one per second per GPU whenever --hwmon-temp-abort is not 0, which is the default, and the status display adds more on top. On that box the path is 52 characters, so a single Intel GPU on i915 loses around 4.5 MB a day, and a run left going for a week reaches 30 MB.

Would you free it before the second call, the way "ext_sysfs_cpu.c" does in the same situation?

```
    if (hc_fopen_raw (&fp, path, "rb") == false)
    {
      hcfree (path);

      continue;
    }
```

One line covers it, and it takes out the blank line that currently sits under the opening brace.

PR #4859, opened 2026-09-14
Add module 00009 for Salted MD5

M0ht3p:
This PR adds mode 00009 (Salted MD5).

- Created src/modules/module_00009.c
- Added corresponding OpenCL kernels (m00009_a0-pure.cl and m00009_a0-optimized.cl)
- Verified functionality using example hashes and dictionary attacks on Intel Compute Runtime / OpenCL.


M0ht3p:
This Pull Request adds custom module 00009 for Salted MD5.

Summary of changes:
- Created module logic in `src/modules/module_00009.c`
- Added corresponding OpenCL kernel files (`OpenCL/m00009_a0-pure.cl` and `OpenCL/m00009_a0-optimized.cl`)
- Successfully tested locally against example hashes and dictionary attacks using OpenCL

jsteube:
This one is already in hashcat as -m 10. The example hash from your own module cracks with it unchanged:

```
$ ./hashcat -m 10 -a 0 --quiet 01dfae6e5d4d90d9892622325959afbe:7050461 wordlist
01dfae6e5d4d90d9892622325959afbe:7050461:hashcat
```

The kernel here computes the same digest:

```
    md5_update (&ctx, tmp.i, tmp.pw_len);
    md5_update (&ctx, s, salt_len);
```

Password then salt is md5($pass.$salt), which is what -m 10 is. 

One difference worth knowing about for a future module. This branch carries a0 kernels only, so -a 1, -a 3 and the rest would have no kernel to run on, where a shipped raw hash mode carries a0, a1 and a3 in both optimized and pure form.

The quickest way to see whether a scheme is already covered is the example hash list, which gives every mode with its format:

```
$ ./hashcat --example-hashes | grep -i -B2 -A2 'md5($pass'
```

Closing this one for that reason. Thanks for writing it up, and the module and kernel layout you used is the right shape if you come across a scheme that is genuinely missing.

PR #4862, opened 2026-09-14
Case a pcfg terminal by character, so a single byte list cracks

matrix:
A capitalization mask is applied one byte at a time, so a terminal that is not UTF-8 gets the wrong characters cased and the word the ruleset holds is never guessed. The shipped `pcfg/hints.tar.xz` shows it with no ruleset to build.

```
$ rm -rf cache/kernels/
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    f56deeb4cde69477f5b46607cf0104fc pcfg/hints.tar.xz hintwords=$'NA\x8aTAK' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

The kernel cache has to go between the two, because the change touches `OpenCL/inc_pcfg.cl`. The limit is there because the keyspace is 2733385897658 and the word's case forms all fall inside the first 200000 candidates. On a machine with no GPU, add `-D 1 --force`.

The hint word is "nastak" with a caron on the S, in windows-1252, given in capitals. The hash is md5 of its lower case form.

`pcfg_lower_ascii ()` lowers A-Z and nothing else, so every lower case form of a word that is not ASCII keeps a capital. `pcfg_upper_image ()` uses the full Unicode table, so the two are used as a pair while only one of them is complete. There is no `pcfg_upper_ascii ()` and there never was.

Nothing has to be declared to tell the two encodings apart. The mask has one position per character and the terminal is that word, so the two lengths already in hand say what a character is: equal means one byte per character. A hint word has no list behind it to give a character count, so its own bytes answer instead, and a word that does not decode as UTF-8 is a single byte encoding.

Two more faults sit in the same code and need the same answer:

* `assemble ()` advances the mask index once per character and decides what a character is by testing `(b & 0xc0) == 0x80`. In a latin-1 or windows-1252 list those byte values are letters, so one mask position swallows two characters and the rest of the mask slides. The same walk is in `pcfg_case_slot ()` in `OpenCL/inc_pcfg.cl`, in `pcfg_expand ()` in `src/shared.c`, and in `mask_match ()`, and the comment on `pcfg_expand ()` says the three have to stay the same walk.
* `pcfg_cp_upper ()` has no uppercase for U+009A, because Unicode has a control there.

`tmp.ln = len` moves above `tlist_load ()` because the uppercase image is built inside it and needs the list's character count.

```
$ ./tools/test_edge.sh -m 0 -K 0

[ test_edge_1789410527 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 8, Target-Type single
[ test_edge_1789410527 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 8, Target-Type multi
[ test_edge_1789410527 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type single
[ test_edge_1789410527 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1789410527 ] > All tests done in 0d:00h:02m:50s
[ test_edge_1789410527 ] > Errors detected: 0
```

That run is the pure kernels. On this machine's Metal device, `./tools/test_edge.sh -m 0 -d 1` covers the optimized ones as well and also ends with 0 errors.

```
$ bash tools/test_package.sh --no-device .

ok    a .xz wordlist reads back the same 128416 words

```

The native, static (`SHARED=0`) and Windows (`make win`) builds are warning free.


jsteube:
In "feed_pcfg.c" the lowering and the uppercase image that follows it test different bytes:

```
        pcfg_lower_word (lo, pg->hw[i].buf, len);

        pcfg_upper_image (up, lo, len, pcfg_is_utf8 (lo, len));
```

The lowering reads the word as it arrived, the image reads the word the lowering just wrote, and pcfg_byte_lower() moves 0xC0-0xDE to 0xE0-0xFE, so a two byte lead becomes a three byte one and a word can change side in between. The same pair is in assemble() and in hint_mask_apply(), both testing the lowered word.

```
src      c0 9a 9e   is_utf8 false   byte path, as intended
lowered  e0 9a 9e   is_utf8 true    codepoint path

all capitals should be   c0 8a 8e
all capitals come out    e0 9a 9e
```

Three cp1252 letters, and every capitalised form of that hint word is gone, which is the loss your reproduction shows one layer further up. Of the three byte words, 63504 sit on the wrong side of it. Could you compute pcfg_is_utf8() once on the word as it arrived and pass that down, instead of testing it again after the lowering?

The second one is in the test itself:

```
    if (c < 0x80)            { at++; continue; }
    else if ((c & 0xe0) == 0xc0) need = 1;
    else if ((c & 0xf0) == 0xe0) need = 2;
    else if ((c & 0xf8) == 0xf0) need = 3;
    else return false;
```

An overlong sequence passes that, and the two bytes which can never open a UTF-8 sequence, 0xC0 and 0xC1, are the capitals A-grave and A-acute in the encoding this test has to separate. A cp1252 word with either of them in front of a byte in 0x80-0xBF takes the codepoint path, decodes to a control character and keeps its capital. That is 128 of the two byte words, and the surrogate range and the 0xF5-0xF7 leads do the same. A minimum codepoint check for each of the three widths, plus the surrogate and U+10FFFF bounds, closes it. I would take both of these before it goes in, since they are the same fault the branch is fixing.

matrix:
Both are real, and both are in. Thank you for the two reproductions, they were enough to build a case for each on its own.

**The test now runs once, on the word as it arrived.** `hint_mask_apply ()` takes the answer as a parameter rather than asking it of `lo`, the hint loop computes it from `pg->hw[i].buf`, and `assemble ()` keeps it in a `hint_wide` set beside `hint_up_ready` when the word is lowered. No call to `pcfg_is_utf8 ()` is left on a lowered word: the three that remain read `src`, `pg->hw[i].buf` and `th->hint[h].buf`.

**`pcfg_is_utf8 ()` now decodes the codepoint and checks it** rather than trusting the lead byte: the lowest value each of the three widths is allowed to encode, the surrogate range, and U+10FFFF. That is what takes 0xC0 and 0xC1 out as leads.

Both are needed. With only the second one in, 59535 of the 63504 three byte words still change side when they are lowered, because a valid three byte lead is still reachable from a two byte one.

Nine hint words follow. Each hash is the md5 of the all capitals form of its word, so the run is asking for the one candidate the fault loses. The words are written as `$'\xNN'` because they are single byte encodings and pasting them as characters would hand hashcat their UTF-8 forms instead. The first six are the fault. The last three are ordinary UTF-8 and are there to show the stricter test still accepts a word it should.

Your own example, `c0 9a 9e`, all capitals `c0 8a 8e`:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    4c21e7938066f24f4ba0d15e46feb01b pcfg/hints.tar.xz hintwords=$'\xc0\x9a\x9e' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

A two byte word behind 0xC0, all capitals `c0 8a`, one of the 128:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    9ce37f7943f07b4da9e9138d14a41a68 pcfg/hints.tar.xz hintwords=$'\xc0\x9a' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

The same behind 0xC1, all capitals `c1 8c`:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    37d3b3690c94a5528db4a8ef9269b0f2 pcfg/hints.tar.xz hintwords=$'\xc1\x9c' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

`c1 9a 9e`, all capitals `c1 8a 8e`. This one isolates the first fault: it reads as a single byte encoding under both the old test and the new one, so only the lowering can move it:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    2345aeae9cad547c00a92e9961cd45cf pcfg/hints.tar.xz hintwords=$'\xc1\x9a\x9e' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

A surrogate, `ed a0 80`, all capitals `cd a0 80`:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    47f86ae641bcb4bb7ecc6538932e8ed1 pcfg/hints.tar.xz hintwords=$'\xed\xa0\x80' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

An 0xF5 lead, `f5 9a 9e 9c`, all capitals `d5 8a 8e 8c`. It needs four bytes, because at three the old test rejects it on length before it ever reaches the lead:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    12a23de45a6b2a875d710df54c7e3d3a pcfg/hints.tar.xz hintwords=$'\xf5\x9a\x9e\x9c' omen=0

before: Recovered 0/1, Status Exhausted, rc=1
after:  Recovered 1/1, Status Cracked,   rc=0
```

And three words that are UTF-8, cracked on both sides:

```
$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    6b6c9c8443255e5f7a8828482ccdd648 pcfg/hints.tar.xz hintwords=$'caf\xc3\xa9' omen=0

before: Recovered 1/1, Status Cracked, rc=0
after:  Recovered 1/1, Status Cracked, rc=0

$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    80c7577814b0b935a9f43a9a48fce139 pcfg/hints.tar.xz hintwords=$'na\xc3\xafve' omen=0

before: Recovered 1/1, Status Cracked, rc=0
after:  Recovered 1/1, Status Cracked, rc=0

$ ./hashcat -m 0 -a 4 -d 1 --potfile-disable --limit 1000000 \
    eec0a85be4008b485df445f68dda7bbb pcfg/hints.tar.xz hintwords=$'\xc3\xbcber' omen=0

before: Recovered 1/1, Status Cracked, rc=0
after:  Recovered 1/1, Status Cracked, rc=0
```

`rm -rf cache/kernels/` is not needed between the two here, because neither change touches a kernel.

```
$ ./tools/test_edge.sh -m 0

[ test_edge_1789416428 ] > All tests done in 0d:00h:07m:26s
[ test_edge_1789416428 ] > Errors detected: 0
```

160 blocks, every attack type, kernel type and vector width. The reproduction in the description still cracks, and the trainer's own suite is unchanged at 82 passed, 0 failed either side of the change.

PR #4864, opened 2026-09-15
Give -a 4 a device engine on the modes that ship only an optimized kernel

matrix:
43 hash modes have no device engine for `-a 4`: the refusal that stops them reads a flag `src/interface.c` sets for them whether or not `-O` was passed, and the kernel file the engine then looks for does not exist for any of them. This adds the 43 kernels and moves the refusal to the point where the missing file is what it tests.

## Reproduction

`rm -rf cache/kernels/` between the two runs, because hashcat reuses a compiled kernel.

```
$ ./hashcat -a 4 -m 9700 '$oldoffice$0*55045061647456688860411218030058*e7e24d163fbd743992d4b8892bf3f2f7*493410dbc832557d3fe1870ace8397e2' pcfg/default-passwords.tar.xz --potfile-disable --runtime 10

before: The device engine has no optimized kernel. Run this without -O.   rc=255
after:  pcfg: device engine, one base word becomes many candidates inside the hash kernel
        Status...........: Aborted (Runtime)                              rc=4
```

There is no `-O` on that command line. A run that does pass it, on a mode with no optimized kernel to load, is still refused and now names the mode; and `-a 0` still picks its kernel by password length, which is what the six sites this touches decide:

```
$ ./hashcat -a 4 -m 100 b89eaac7e61417341b710b727768294d0e6a277b pcfg/default-passwords.tar.xz --potfile-disable --runtime 10 -O
The device engine has no optimized kernel for this hash mode. Run this without -O.   rc=255

$ echo hashcat | ./hashcat -a 0 -m 0 8743b52063cd84097a65d1633f5c74f5 --potfile-disable -O
Kernel.Feature...: Optimized Kernel (password length 0-31 bytes)                     rc=0
```

## What was wrong

`src/interface.c:485` sets `OPTI_TYPE_OPTIMIZED_KERNEL` for a mode whose `a0` pair has no pure kernel, whatever the user asked for, so `hashconfig->opti_type` does not say whether `-O` was passed and the refusal read exactly that. `user_options->optimized_kernel` does say, so the test moved into the branch where `hc_path_read ()` finds no file. The flag cannot simply be dropped there instead, because the hashes were parsed under it: 2 of the 43 subtract the initial state out of the stored digest in `module_hash_decode ()` when it is set.

`generate_source_kernel_filename ()` reads that same flag, so for those 43 modes the engine asks for `mNNNNN_a4-optimized.cl`, and not one was in the tree. Counting every mode whose kernel runs inside, 46 of the 299 had no `-a 4` device kernel of either kind before this and 3 have none after: `-m 2000` is `STDOUT` and every entry point it has is empty, `-m 5100` compares three times per candidate where the engine hands back one set of four words, and `-m 20510` has `NOT AVAILABLE` where its multi hash entry point would be.

The name is the digest convention rather than the candidate layout. The engine hands the candidate over as a word array plus a byte length either way, so each file is the mode's own optimized kernel with the base word paste taken out and the four hooks of `OpenCL/inc_pcfg_kernel.cl` around what is left. What the flag changes is what the kernel owes, because `src/interface.c:495` clears a set of `OPTS_TYPE` flags when it is absent and 19 of the 43 modules declare one of them, so the padding and widening the host does for a wordlist attack has to happen inside the kernel. The entry points stay `_sxx`, so the 6 sites that chose between the two kernel shapes ask a new `is_opti_kernel_no_pcfg ()`, which reads the attack kern as well as the flag: `src/generic.c` moves `attack_kern` to `ATTACK_KERN_PCFG` once the device engine is the one that will run, and a run that fell back keeps `ATTACK_KERN_STRAIGHT` and the optimized shape.

Both test suites refused the `-a 4` optimized round for every mode whose kernel runs inside, which takes all 43 of these. Each guard now asks whether the mode's `kern_type` has such a file, read out of the module the way both scripts already read `attack_exec` so that neither spends a run of hashcat on it. That `sed` is the shape the two already use on `src/modules`, and the shape had no end anchor, so `module_00100.cpp` came out as `100pp`: the ten patterns of it now anchor the end, and the five that also left the dot unescaped escape it. Every list they build is unchanged, on GNU `sed` 4.9 and on BSD.

What the kernels buy is why the fallback is worth avoiding. On `-m 0 -a 4` on an RTX 4080, over the one second status samples of three `--runtime 20` runs after a discarded first pass, the device engine's median is 29525.5 MH/s and the host engine's 67.1 MH/s, with peaks of 33365.6 and 71.3.

## tools/test_edge.sh

`./tools/test_edge.sh -m 9700 -a 4 -K 1 -d 1 -D 2` on an RTX 4080, and the same for each of the other 42. This is the round these kernels are for, and the one master refuses:

```
[ test_edge_1789419211 ] # Processing Hash-Type 9700, Attack-Type 4, Kernel-Type optimized, Vector-Width 16, Target-Type multi
[ test_edge_1789419211 ] > All tests done in 0d:00h:00m:12s
[ test_edge_1789419211 ] > Errors detected: 0
```

430 rounds over the 43, 490 hashcat runs, 0 errors, and `tools/test.sh -a 4` over the same 43 gives 86 OK with no Skip. One tail rather than 43 is the one place this deviates from CONTRIBUTING.md, because no other round of either suite reaches these files.

Each mode also cracks its own `ST_HASH` out of a grammar holding its `ST_PASS`, 43 of 43 on CUDA and on HIP, and falls back to the host engine with the kernel file moved away, 1/1 each. Every kernel builds and passes its self test on all six back end and device slots here, 43 of 43 in each, cache cleared first: Metal and OpenCL on an M4 Pro, CUDA and OpenCL on an RTX 4080, HIP and OpenCL on an RX 6900 XT. Apple is where a kernel of this shape has failed before, which is what #4823 was about, so both of its back ends are in that list.

`docs/hashcat-pcfg.md` said the device engine needs an `_a4-pure.cl` and that `-O` is refused on a fast hash, and `docs/hashcat-plugin-development-guide.md` said attack mode 4 has no optimized form. Both now say what the layout is.


dalmatele:
@matrix I don't know where to report bug but we have a case that we trained data with coverage = 0 (only use OMEN).
When we injected this rule to hashcat with pcfg option, we got error: `no usable structures`.
This error happened because in grammar file we only have one line with content:
`M	1`

jsteube:
Could you pull the latest master and try that again or use a beta binary from hashcat.net/beta/

If it still happens on a current tree, please place an issue for it, with the grammar.txt attached. This thread is a merged pull request about something else, so a report here is easy to miss.

PR #4867, opened 2026-09-15
Give a bridge the salt attack mode 9 pairs with the candidate

matrix:
Attack mode 9 pairs one candidate with one hash, and on a mode that uses a bridge it recovers only the hash in the first position. `run_bridge_loop ()` hands the bridge the salt index the host loop carries, and attack mode 9 collapses that loop to a single iteration because the kernel is supposed to read the salt from the global id instead. A bridge never reaches that macro, so every candidate is computed against the first hash's salt.

### Reproduction

Tested on v7.1.2-676-ge6a7f642c, Ubuntu 24.04.5, RTX 4080 and RX 6900 XT. master has moved to 210523483 since, and `git diff --stat e6a7f642c..master -- src/backend.c src/bridges/ src/bridges.c OpenCL/inc_types.h` is empty, so none of the three commits in between touches any of this. `-m 72000` needs a free threaded interpreter and `-m 73000` refuses one, so `pyenv local` selects 3.14.7t for every mode below except 73000, which takes 3.14.7.

The suite reaches it with no files to prepare:

```
$ ./tools/test_edge.sh -m 72000 -a 9
```

before:

```
[ test_edge_1789490039 ] > All tests done in 0d:00h:00m:47s
[ test_edge_1789490039 ] > Errors detected: 5
[ test_edge_1789490039 ] !> Details on test_edge_1789490039/test_edge.details.log
```

The five are one per vector width. Each is `error (1)`, `STATUS 5`, `RECHASH 1 2`, on the multi hash case.

after:

```
[ test_edge_1789490087 ] # Processing Hash-Type 72000, Attack-Type 9, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1789490087 ] > All tests done in 0d:00h:00m:47s
[ test_edge_1789490087 ] > Errors detected: 0
```

The same command on every mode that uses a bridge:

| mode | bridge | master | with this |
| --- | --- | --- | --- |
| 70000 | argon2id | 5 | 0 |
| 70100 | scrypt-jane | 0 | 0 |
| 70200 | scrypt-yescrypt | 0 | 0 |
| 72000 | python, no GIL | 5 | 0 |
| 73000 | python, GIL | 5 | 0 |
| 74000 | rust | 5 | 0 |

70100 and 70200 pass on both sides because the only thing they take from that index is the cost of the hash, and the suite gives every hash in a run the same cost. Two yescrypt hashes with different costs show it there as well:

```
$ python3 -c 'import hashlib, base64
def mk(pw, salt, N, r, p):
    dk = hashlib.scrypt(pw.encode(), salt=salt.encode(), n=N, r=r, p=p, dklen=32, maxmem=1<<30)
    return "SCRYPT:%d:%d:%d:%s:%s" % (N, r, p, base64.b64encode(salt.encode()).decode(), base64.b64encode(dk).decode())
print(mk("passaaaa", "saltaaaa", 16384, 8, 1))
print(mk("passbbbb", "saltbbbb", 32768, 8, 1))' > y.hashes
$ printf 'passaaaa\npassbbbb\n' > y.words
$ ./hashcat -m 70200 -a 9 --potfile-disable --self-test-disable --outfile y.out y.hashes y.words
```

before:

```
Status...........: Exhausted
Recovered........: 1/2 (50.00%) Digests (total), 1/2 (50.00%) Digests (new), 1/2 (50.00%) Salts
rc=1

$ cat y.out
SCRYPT:16384:8:1:c2FsdGFhYWE=:LBtIzKJCQelJkwD2QoXVo6GVqrtL
```

after:

```
Status...........: Cracked
Recovered........: 2/2 (100.00%) Digests (total), 2/2 (100.00%) Digests (new), 2/2 (100.00%) Salts
rc=0

$ cat y.out
SCRYPT:16384:8:1:c2FsdGFhYWE=:LBtIzKJCQelJkwD2QoXVo6GVqrtL
SCRYPT:32768:8:1:c2FsdGJiYmI=:QBhZ71gof6on7hFHfjlRwqRJxrXj
```

### Cause

`backend.c` collapses the salt loop for this attack mode:

```c
u32 salts_cnt = hashes->salts_cnt;

if (user_options->attack_mode == ATTACK_MODE_ASSOCIATION)
{
  // We will replace in-kernel salt_pos with GID via macro
  salts_cnt = 1;
}
```

The replacement is in `OpenCL/inc_types.h`, and it covers the digest as well as the salt:

```c
#if ATTACK_MODE == 9
#define SALT_POS_HOST            (kernel_param->pws_pos + gid)
#define DIGESTS_OFFSET_HOST      (kernel_param->pws_pos + gid)
#define DIGESTS_CNT              1
#endif
```

A mode that uses a bridge still runs that kernel, and its `_comp` compares candidate `gid` against digest `gid` correctly. What it does not do is compute the hash: `_loop` is empty and the bridge does the work, so the bridge is the one that reads the salt, and it reads it at the index `run_bridge_loop ()` passes. That index is the collapsed `salt_pos`, which is 0 for the whole run. Candidate 0 is computed with salt 0 and compared against digest 0, so it matches. Candidate 1 is computed with salt 0 and compared against digest 1, so it cannot.

The host already has this problem solved for hooks, which run on the host for the same reason. `hook_thread_param` carries `salt_per_pw` and `pws_pos`, and `backend.c` builds the index per candidate:

```c
const u32 salt_pos = (hook_thread_param->salt_per_pw == true) ? (u32) (hook_thread_param->pws_pos + pw_pos) : hook_thread_param->salt_pos;
```

The bridge path is written the same way. It needs no `pw_pos` term because a bridge launch carries one candidate: adding one `fprintf` to `launch_loop ()` and running 64 hashes gives 64 calls, every one with `pws_cnt` of 1 and a distinct `pws_pos`, with and without a rule file.

That is measured rather than enforced. Nothing in the code says a bridge launch cannot carry more than one candidate, and if one ever did, this would give every candidate in it the salt of the first, which is the fault being fixed. Covering that needs what the hook has, a base index the plugin adds the candidate position to, and that is a change to the bridge interface rather than to this call.

`kernel_param.pws_pos` is the value the kernel would use. `backend.c` already reduces it modulo the salt count for this attack mode, in the same function that sets it.

Which bridges show it depends on what they take from the index. The two Python bridges and the Rust one hand the whole `salts_buf` and the index to the plugin, which indexes it. `bridge_argon2id_reference.c` reads `esalts_buf[salt_pos]`, which holds that hash's own salt. `bridge_scrypt_jane.c` and `bridge_scrypt_yescrypt.c` read only `scrypt_N`, `scrypt_r` and `scrypt_p` from it and take the salt itself out of the `tmps` buffer the kernel filled, so they answer correctly as long as every hash in the run carries the same cost.

### Checks

Nothing else moves. The same commands on master and on this branch:

| case | master | with this |
| --- | --- | --- |
| 70200, two hashes, same cost | 2/2 recovered, rc=0 | 2/2 recovered, rc=0 |
| `-m 10 -a 9`, no bridge | 2/2 recovered, rc=0 | 2/2 recovered, rc=0 |
| `-m 10 -a 0`, no bridge | 2/2 recovered, rc=0 | 2/2 recovered, rc=0 |

`-m 10` is there because it is the same attack on a salted mode that runs no bridge, which is the path this must not touch.

### tools/test_edge.sh

The modes this touches are the six that use a bridge, and the table above is `./tools/test_edge.sh -m <mode> -a 9` on each of them, on master and on this branch.

The full run of one of them, every attack type and vector width, on this branch:

```
[ test_edge_1789490272 ] # Processing Hash-Type 72000, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1789490272 ] > All tests done in 0d:00h:07m:35s
[ test_edge_1789490272 ] > Errors detected: 0
```


jsteube:
> A bridge launch carries one candidate, so the position of that candidate inside the chunk is zero and the base is the whole index.

That is the one part I cannot follow from the code. The same pws_cnt goes to run_kernel() beside this call, autotune.c reaches run_bridge_loop() with kernel_power_try, and the bridge on the other side builds a list of that size:

```
  PyObject *pws = python->PyList_New (pws_cnt);
```

The hook path carries the position rather than the base for that reason:

```
  const u32 salt_pos = (hook_thread_param->salt_per_pw == true) ? (u32) (hook_thread_param->pws_pos + pw_pos) : hook_thread_param->salt_pos;
```

Could you put whatever pins pws_cnt to 1 into that comment? Your runs show it is 1 in practice, over four modes and two cards, so I would take the fix on that. What I would rather not leave unwritten is the reason, because above 1 every candidate after the first in a chunk takes the first one's salt again, and a later change to bridge accel would put that back, with only that comment standing against it. The same salt_pos reaches launch_loop2 a few lines down, which no bridge implements today, so that one is only the same shape left inconsistent.

matrix:
# Attack mode 9 on the bridge path: replacing the design in this pull request

You stopped at the right sentence. Nothing pins `pws_cnt` to 1, so the comment you asked for cannot be written and the design that rests on it has to go. What follows is what I would put in its place.

Every run below is on `150a0e61a`, the upstream head at the time of writing, with both the unpatched tree and the patched one built from it.

## The problem, on two hashes

Two argon2id hashes, two passwords, one paired with each:

```
printf '%s\n' \
  '$argon2id$v=19$m=1024,t=1,p=1$AAAAAAAAAAAAAAAAAAAAAA$xOHpt3uXlaY0aYbYda2yD/Ee6E3M9kIO+KfFCqdDOr0' \
  '$argon2id$v=19$m=1024,t=1,p=1$AAAAAAAAAAAAAAAAAAAAAQ$0loQU+/2Jd+zOQhElb1vf7nuNeZC24KsH1sZLzvoes8' > a9.hash
printf '%s\n' pw0 pw1 > a9.words

./hashcat -m 70000 -a 9 --potfile-disable a9.hash a9.words
```

```
before: Recovered 1/2, Status Exhausted, rc=1
after:  Recovered 2/2, Status Cracked,   rc=0
```

Attack mode 9 reads its candidates from a file, so the two files have to exist. Everything in them is above.

## Why the first design does not fix it

The first design is right only for the candidate at position 0 of a launch, and a launch carries more than one. Forcing the launch size shows it. The lists are `n` passwords paired one to one with `n` distinct salts, which is what the mode wants: hand it a wordlist of another length and hashcat refuses the run with `Number of words in wordlist is not in sync with number of unique salts`, so the work is set by the hash count and never by the word count.

Modes 72000, 73000 and 74000 hash the salt followed by the password with sha256 and then run 10000 more rounds of it, so one generator serves the three, and the scrypt modes take `SCRYPT:N:r:p:base64(salt):base64(digest)`:

```python
import hashlib, base64

n = 64                  # 400000 for the speed table
N, r, p = 16384, 8, 1   # 1024, 8, 1 for the speed table

pws   = ["pw%05d" % i for i in range (n)]
salts = ["salt%05d" % i for i in range (n)]

def gh (pw, salt):
    h = hashlib.sha256 (salt.encode () + pw.encode ()).digest ()
    for _ in range (10000): h = hashlib.sha256 (h).digest ()
    return h.hex ()

def scr (pw, salt):
    d = hashlib.scrypt (pw.encode (), salt = salt.encode (), n = N, r = r, p = p, dklen = 32, maxmem = 1 << 30)
    return "SCRYPT:%d:%d:%d:%s:%s" % (N, r, p,
        base64.b64encode (salt.encode ()).decode (), base64.b64encode (d).decode ())

open ("h_sha.txt",    "w").write ("".join ("%s*%s\n" % (gh (a, b), b) for a, b in zip (pws, salts)))
open ("h_scrypt.txt", "w").write ("".join ("%s\n" % scr (a, b)       for a, b in zip (pws, salts)))
open ("pw.txt",       "w").write ("".join (p + "\n" for p in pws))
open ("nomatch.txt",  "w").write ("".join ("zz%05d\n" % i for i in range (n)))
```

Argon2id comes from the test module already in the tree, which needs `Crypt::Argon2`, the same one `tools/test.sh` needs. The digest has to land in a variable before it is printed. `require` defines the sub at run time, so a bareword straight after `print` is parsed as a filehandle and the run writes nothing while still exiting 0.

```
perl -e 'require "./tools/test_modules/m70000.pm";
  for my $i (0 .. 399999) {
    my $h = module_generate_hash (sprintf ("pw%05d", $i), sprintf ("%032x", $i), 1024, 1, 1, 32);
    print "$h\n";
  }' > h_argon.txt
```

70000 takes `h_argon.txt`, 70100 and 70200 take `h_scrypt.txt`, 72000, 73000 and 74000 take `h_sha.txt`. The speed table below is the only one that takes 70000 from a generated list, so the perl above is needed for that table and for nothing else here. The two hash reproduction at the top carries its hashes inline, and `test_edge.sh` generates its own.

With `n = 64`, on master, on the branch as it stands, and on the one below. `wc -l` on the outfile is the recovered count. The first line is the auto row, the second is run once per pinned accel value, `2`, `8`, `32` and `256`:

```
./hashcat -m 74000 -a 9 --potfile-disable --quiet                -o out.txt h_sha.txt pw.txt
./hashcat -m 74000 -a 9 --potfile-disable --quiet --force -n <N> -o out.txt h_sha.txt pw.txt
```

```
                  master      first design      this design
  auto            1/64        64/64             64/64
  --force -n 2    1/64        32/64             64/64
  --force -n 8    1/64        16/64             64/64
  --force -n 32   1/64        16/64             64/64
  --force -n 256  1/64        16/64             64/64
```

Master recovers one hash out of the 64 and nothing else, whatever the launch size, which is the bug. The one it recovers is the hash whose salt sorts first, `salt00000` here, which with this list is also the first line of the file. The first design recovers one per batch: one in two at a batch of two, one in four above it, which is your reading exactly, since every candidate after the first in a batch takes the first one's salt again. Autotune settles on a batch of one for this bridge on the hardware I have, which is the only reason the four modes came back clean, and `-n` is all it takes to move it.

The bridge reports the same number from the other side. One line at the top of `process_batch` in `Rust/bridges/dynamic_hash/src/interop.rs`, then a rebuild, which carries the crates with it:

```rust
eprintln!("DIAG pws_cnt={} salt_id={} stride={}", io.len(), salt_id, stride);
```

That build is the patched one, since `stride` is what the patch introduces, and the algorithm reads the salt whole so the run reaches the end rather than stopping on a field a salt does not have:

```
./hashcat -m 74000 -a 9 \
  --bridge-parameter1 ./bridges/subs/dynamic_hash.so \
  --bridge-parameter2 'md5($p.$s)' \
  --self-test-disable --potfile-disable --quiet \
  --force -n 256 -o out.txt h_sha.txt pw.txt
```

```
DIAG pws_cnt=4 salt_id=0 stride=1
DIAG pws_cnt=4 salt_id=12 stride=1
DIAG pws_cnt=4 salt_id=28 stride=1
DIAG pws_cnt=4 salt_id=24 stride=1
DIAG pws_cnt=4 salt_id=48 stride=1
```

Sixteen of those for 64 hashes, every one of them `pws_cnt=4`, and the sixteen `salt_id` values are `0, 4, 8 ... 60`, the batch starts, arriving in whatever order the launches are dispatched rather than in order.

## What the core already does

Under attack mode 9 the kernel does not read a salt index that was handed to it. It builds one, in `OpenCL/inc_types.h`:

```c
#define SALT_POS_HOST       (kernel_param->pws_pos + gid)
```

The core gives the position the batch starts at in the salt array, and the work item adds its own place in that batch. The hook path is the same rule written for the host, which is the line you quoted:

```c
const u32 salt_pos = (hook_thread_param->salt_per_pw == true) ? (u32) (hook_thread_param->pws_pos + pw_pos) : hook_thread_param->salt_pos;
```

A bridge is a third consumer of that rule. It runs on the host, so it never reaches the macro, and it is handed the collapsed `salt_pos`, which is 0. The first design answered that by substituting the base for the index in `run_bridge_loop ()`:

```c
const u32 bridge_salt_pos = (salt_per_pw == true) ? (u32) device_param->kernel_param.pws_pos : salt_pos;
```

One index for the whole batch, which is the table above.

## The design

The core keeps handing over the position the batch starts at. The consumer adds its own place, in the loop over candidates it already has:

```c
static inline u32 bridge_salt_pos (const hashcat_ctx_t *hashcat_ctx, const hc_device_param_t *device_param, const hashes_t *hashes, const u32 salt_pos, const u64 pw_pos)
{
  // The self test hands a hashes_t holding the one self test salt and names it by index, so there is
  // no batch to place a candidate in.

  if (hashes->salts_buf == hashes->st_salts_buf) return salt_pos;

  const user_options_t *user_options = hashcat_ctx->user_options;

  if (user_options->attack_mode == ATTACK_MODE_ASSOCIATION)
  {
    const u32 pos = (u32) (device_param->kernel_param.pws_pos + pw_pos);

    return pos;
  }

  return salt_pos;
}
```

It is `static inline` in `include/bridges.h` because a bridge is a separate shared object and the core exports none of its own symbols to it. The tree builds with 0 errors and 0 warnings, and `rustfmt --check` over `git ls-files '*.rs'` exits 0.

The three bridges that compute in C call it from inside the candidate loop they already had, so the salt is read per candidate rather than once above the loop. The three that hand the batch to a plugin cannot add the position on the C side, because the plugin is the one that walks the candidates: they pass the base and the plugin adds its own index. Whether to add is decided once when the context is built and travels under the name `salt_per_pw`, the same name the core already uses for a hook. Both shapes are in the diff.

The Python plugin guide documents `salt_id` for the author who walks the batch instead of delegating, so it gets the entry that goes with this: `salt_per_pw`, what `salt_id` becomes under it, and that the helpers do the addition for you.

## What to expect from it

At any batch size the candidate at position `i` of a launch is computed against the salt at `pws_pos + i`: the index the kernel builds for its work item, and the one the core already hands a hook. The table above becomes 64 of 64 at every accel value rather than at the one autotune happened to pick. Outside attack mode 9 the index is the `salt_pos` the core passed, unchanged, and the added work is one branch and one add per candidate. The self test is unaffected: it hands a bridge a `hashes_t` holding the single self test salt and names it by index, so there is no batch to place a candidate in, which is the first thing the helper checks.

Nothing splits a batch. The bridge callback interface does not change, so `BRIDGE_INTERFACE_VERSION` stays at 720 and a bridge written against it still compiles. The Rust `new_context` and the Python `_worker_batch` do take one argument more, but both are the interop layer this tree ships and `make` rebuilds, not what a plugin author writes: `generic_hash.rs` and a Python `calc_hash` are untouched.

One thing reaches an author, and it is narrower than it sounds. The Python guide documents `salt_id` as the salt a calculation is about, and invites a plugin to walk the batch itself rather than delegate to `handle_queue`. Under this attack mode such a plugin is already wrong on master, where `salt_id` is 0 on every launch and every candidate is computed against one salt. It is now the salt the batch starts at, which is the base it needs: adding the position of the candidate it is on makes it right, the same one line the C side does in `bridge_salt_pos ()`. Outside this attack mode nothing it sees changes.

## launch_loop2

It needs no conversion and it is not left inconsistent. `run_kernel (hashcat_ctx, device_param, KERN_RUN_LOOP2, pws_pos, pws_cnt, ...)` sets `kernel_param.pws_pos` a few lines above the `launch_loop2` call in the same function, and `launch_loop2` takes the same `hashcat_ctx`, `device_param`, `hashes` and `salt_pos`. A bridge that implements it calls `bridge_salt_pos ()` and gets the right answer for the same reason `launch_loop` does. All six in-tree bridges set it to `BRIDGE_DEFAULT`, so there is nothing to change today.

## Range

`pws_pos + pw_pos` stays inside the salt array because `get_work ()` cuts an association batch at the round boundary:

```c
const u64 round_left = salts_cnt - (words_off % salts_cnt);

work = MIN (work, round_left);
```

and `run_cracker ()` reduces the offset it was given:

```c
if (salts_cnt > 0) pws_pos = pws_pos_arg % salts_cnt;
```

so `pws_pos` is a position in the salt array, not in the keyspace, and `pws_pos + pws_cnt <= salts_cnt` holds for the whole batch.

`run_bridge_loop ()` has a second caller, `autotune.c`, which reaches it for a `BRIDGE_TYPE_REPLACE_LOOP` mode with no `run_kernel ()` in front of it, so nothing has written `kernel_param.pws_pos` for that call. It holds there for two other reasons: the field is zeroed in `backend_session_begin ()` and autotune runs before the first launch, and `autotune_kernel_power ()` caps the candidate count at `salts_cnt` under this attack mode. On the autosplit rounds after the first, `backend_ctx_devices_tuning_restore ()` reuses the tuning and autotune does not run again. If that ever stops being true the helper would be reading a `pws_pos` from an earlier launch, so it is worth saying out loud rather than leaving to be discovered.

## What this asks of a consumer

One thing, and it is worth stating because it caught out one of the six: with the salt now moving from one candidate to the next, any state a bridge derives from the salt has to move with it.

The Rust `dynamic_hash` plugin binds `$s1 .. $sn` from the `*` separated fields of the salt into an evaluation context. Reusing one context across a batch was correct when the batch had one salt. It is not now: a salt with fewer fields than the one before it leaves the extra names holding the previous candidate's data.

Hash 42 below gets a salt of two fields where every other salt has three, and its digest is built from the third field of the salt in front of it. 42 rather than 40 because the batch is four candidates wide, so a multiple of four opens a batch and starts clean:

```python
import hashlib
md5 = lambda s: hashlib.md5 (s.encode ()).hexdigest ()
rows = []
for i in range (64):
    pw = "pw%02d" % i
    if i == 42:
        rows.append ((md5 (pw + "S%02d" % (i - 1)), "x%02da*x%02db"       % (i, i)))
    else:
        rows.append ((md5 (pw + "S%02d" % i),       "x%02da*x%02db*S%02d" % (i, i, i)))
open ("h_bug.txt",  "w").write ("".join ("%s*%s\n" % r for r in rows))
open ("pw_bug.txt", "w").write ("".join ("pw%02d\n" % i for i in range (64)))
```

```
./hashcat -m 74000 -a 9 \
  --bridge-parameter1 ./bridges/subs/dynamic_hash.so \
  --bridge-parameter2 'md5($p.$s3)' \
  --self-test-disable --potfile-disable --quiet \
  --force -n 256 -o out.txt h_bug.txt pw_bug.txt
```

```
before:  rc=0    64/64 recovered, hash 42 among them, no error
after:   rc=255  hash 42 not recovered, "Undefined variable 's3'"

1d9b6ef2e8b7a2f8d5f53ff9ddccc034*x42a*x42b:pw42     <- produced by the first
```

The second run stops at the error, so how many of the other 63 reach the outfile first varies between runs. Hash 42 is the row that matters and it is stable: recovered on the first, never on the second, across the runs I made.

That salt has no third field. The first build cracks it anyway because `$s3` still holds `S41`, a fragment of a different hash's salt. The context is now rebound whenever the salt moves.

## test_edge.sh

On each build, once per mode, ten combinations executed on each side of every line:

```
./tools/test_edge.sh -m 70000 -a 9
```

```
  mode     master               with the fix
  70000    Errors detected: 5   Errors detected: 0
  70100    Errors detected: 0   Errors detected: 0
  70200    Errors detected: 0   Errors detected: 0
  72000    Errors detected: 5   Errors detected: 0
  73000    Errors detected: 5   Errors detected: 0
  74000    Errors detected: 5   Errors detected: 0
```

Yescrypt with two hashes that carry different costs, the case where the cost taken from the first salt is what goes wrong rather than the salt itself. The generator above holds `N` fixed for the whole list, so this pair takes one of its own:

```python
import hashlib, base64
def scr (pw, salt, N, r = 8, p = 1):
    d = hashlib.scrypt (pw.encode (), salt = salt.encode (), n = N, r = r, p = p, dklen = 32, maxmem = 1 << 30)
    return "SCRYPT:%d:%d:%d:%s:%s" % (N, r, p,
        base64.b64encode (salt.encode ()).decode (), base64.b64encode (d).decode ())

open ("mixed.hash",  "w").write (scr ("pw0", "salt0", 16384) + "\n" + scr ("pw1", "salt1", 32768) + "\n")
open ("mixed.words", "w").write ("pw0\npw1\n")
```

```
./hashcat -m 70200 -a 9 --potfile-disable --quiet -o out.txt mixed.hash mixed.words
```

```
  master   rc=1  recovered 1/2
  fix      rc=0  recovered 2/2
```

It is reachable there and not on scrypt-jane because `module_70200.c` sets `salt_iter` to 1, while `module_70100.c` sets it to `scrypt_N * 2 * scrypt_p` and `hashes.c` refuses attack mode 9 outright when two salts disagree on it.

The self test left enabled, which `test_edge.sh` disables. The same commands with no `--self-test-disable`, which is the point of the row, `n = 64`. The last row is the auto, `-n 8` and `-n 256` runs of the accel table above, which carry no `--self-test-disable` either:

```
./hashcat -m 70100 -a 9 --potfile-disable --quiet -o out.txt h_scrypt.txt pw.txt
./hashcat -m 70200 -a 9 --potfile-disable --quiet -o out.txt h_scrypt.txt pw.txt
```

```
  70100  rc=0  64/64
  70200  rc=0  64/64
  74000  rc=0  64/64 at auto, at -n 8 and at -n 256
```

## Speed

Not a performance patch, so this is here only to show the change does not cost throughput.

Every mode gets a list of the same size, 400000 hashes. Not the same list: the six modes take three, one per format, argon2id for 70000, scrypt for 70100 and 70200, and the generic one for 72000, 73000 and 74000. The size is what has to match, because the launch size autotune settles on depends on how many hashes it is handed, and the launch size is how many candidates a bridge is given at once, which is exactly what the change loops over. Two modes measured with different list sizes are not exercising it equally. Costs are lowered, argon2id at `m = 1024, t = 1` and scrypt at `N = 1024`, so that one list size can serve all six and because a cheaper primitive is where a branch per candidate would show first. The wordlist cracks nothing, so both arms compute the same digests.

One bridge unit, `-d 1`, which is what selects a bridge unit. The geometry is not left to chance: each mode is probed on both arms with `--runtime 1` and the pair autotune settles on is read back, and it is pinned for the measured passes only when it came out the same on every probe and on both arms. It did, for all six.

    ./hashcat -m 70000 -a 9 h_argon.txt nomatch.txt \
      --keep-guessing --self-test-disable --markov-disable --restore-disable \
      --outfile-autohex-disable --wordlist-autohex-disable \
      --potfile-disable --logfile-disable \
      --status --status-timer 1 --runtime 60 \
      --force -n 28 -u 1 -d 1 > m-70000-p1.log 2>&1

Five passes of 60 s per arm, the arms alternating inside each pass so that drift lands on both, the first pass discarded because it pays the kernel build and the bridge start-up. The speed of a run is hashcat's own figure, the median of its samples after the first five. The speed of a block is the median of the four kept runs, with their min and max beside it.

The hash count in the table is the one hashcat reports having loaded, not the line count of the file:

    Hashes: 400000 digests; 400000 unique digests, 400000 unique salts

```
  mode     hashes   master                     this design                pin
  70000    400000   6367 H/s  (6266 - 6406)    6391 H/s  (6375 - 6502)    Accel:28 Loops:1
  70100    400000   1022 H/s  (1020 - 1029)    1020 H/s  (1020 - 1029)    Accel:6 Loops:2048
  70200    400000    979 H/s  ( 978 -  984)     978 H/s  ( 977 -  978)    Accel:6 Loops:1
  72000    400000    354 H/s  ( 354 -  355)     355 H/s  ( 354 -  356)    Accel:6 Loops:1
  73000    400000     15 H/s  (  15 -   15)      15 H/s  (  15 -   15)    Accel:141 Loops:1
  74000    400000   2655 H/s  (2573 - 2656)    2627 H/s  (2626 - 2632)    Accel:6 Loops:1
```

Only one unit runs, which the log shows rather than the command implying it: a single `Speed.#*` and a single `Hardware.Mon.#01`, and none of the other units the bridge exposes.

Three modes come out nominally slower with the change, two faster, one level. The widest gap is 74000 at 1.1 percent, and master's own four passes there span 3.1 percent, so the gap is inside the spread of the arm it is measured against. That is what this measurement can say: no gross regression, and nothing finer. The change adds one comparison and one addition per candidate, orders of magnitude below what a run of this kind resolves.

One row is worth reading with care rather than at face value. 73000 runs at 15 H/s on a list of 400000 where it ran at 190 H/s on a list of 50000, so something in that bridge scales with the number of salts rather than with the work. Both arms carry it identically, so the row still compares them, but it is not measuring the hashing.

PR #4869, opened 2026-09-16
Added seperate rule lookup tool

jessevz:
In this pull request I added a separate tool to check for a plaintext password, if it is inside the keyspace of a wordlist + rule file. We did not want to include this inside the --lookup of hashcat because traversing the dictionaries plus rules can take a lot longer than the other lookups.

I am able to go through rockyou + dive.rule in around 1h and 30 minutes, which is faster than my previous implementation in --lookup, by using multiple threads:

`
└─>>> tools/rule_lookup 'doesnotexist~~~~abcd' rockyou.txt -r dive.rule
rule_lookup: 14344391 words in 1 file, 98670 rules from 1 file, pure engine
rule_lookup: 1415361059970 candidates to build, 1 candidate to look for, 16 threads
doesnotexist~~~~abcd                                                          
  not produced: no rule makes it out of any of the 14344391 words
rule_lookup: 0 of 1 found, 1415361059970 rule applications in 5126.14s, 276.1M/s
`

You can compile the tool as follows:

`make -f tools/rule_lookup.mk`

And this is how it looks when it is able to find a password:

`
└─>>> ./rule_lookup merche03123 ../example.dict -r ../rules/best66.rule 
rule_lookup: 128416 words in 1 file, 66 rules from 1 file, pure engine
rule_lookup: 8475456 candidates to build, 1 candidate to look for, 16 threads
merche03123                                                                   
  word 89999      merche03  (../example.dict:90000)
  rule 27         $1 $2 $3  (../rules/best66.rule:33)
  reached at -s 89999 of 128416, 70.0839% into the run
rule_lookup: 1 of 1 found, 8475456 rule applications in 0.05s, 170.0M/s
`

I am not sure if the tools directory is for tools like this, but for me this is a helpful tool which i wanted to share.

Most code is generated by AI but verified by me.


PenguinKeeper7:
You can use `-m 99999`, which is the plaintext hash mode. As an example:
```
hashcat -m 99999 -a 0 'doesnotexist~~~~abcd' rockyou.txt -r dive.rule
```
or
```
hashcat -m 99999 -a 0 merche03123 rockyou.txt -r dive.rule
```
This will also get far, far higher speeds than your 170MH/s

jessevz:
> You can use `-m 99999`, which is the plaintext hash mode. As an example:
> 
> ```
> hashcat -m 99999 -a 0 'doesnotexist~~~~abcd' rockyou.txt -r dive.rule
> ```
> 
> or
> 
> ```
> hashcat -m 99999 -a 0 merche03123 rockyou.txt -r dive.rule
> ```
> 
> This will also get far, far higher speeds than your 170MH/s

Yes, I am aware of that, but this way you do not know which rule was responsible for your password, which was the main thing i wanted to solve with this. And it is not always trivial to find out which rule in combination with which candidate of your dict was able to crack the password

jsteube:
> Yes, I am aware of that, but this way you do not know which rule was responsible for your password, which was the main thing i wanted to solve with this.

It does, through --debug-mode. Mode 5 records Original-Word:Finding-Rule:Processed-Word:Wordlist, so your own example comes back whole:

```
$ hashcat -m 99999 -a 0 merche03123 example.dict -r rules/best66.rule --debug-mode=5 --debug-file=dbg.log
merche03123:merche03123

$ cat dbg.log
merche03:$1 $2 $3:merche03123:example.dict
```

It is also the faster of the two by an order of magnitude. Both tools built from the same tree, the same 12670806720 candidates, a candidate neither one finds, your tool on 24 threads and hashcat on one RX 9070 XT:

```
  rule_lookup   29.78 s    425.5 M/s
  hashcat        3 s      4845.1 MH/s
```

Scaled to the run in your description, the 5126 seconds for rockyou and dive.rule comes out near 5 minutes.

What is left after that is the position, "reached at -s 89999 of 128416", and the position is what #4829 settled. Rules turn one pass over the base words into one pass per rule, so --lookup refuses -r outright rather than answering for a small input and refusing the real one. A separate binary answering the same question is that decision reached by another route.

So I am closing this one. Thanks for the work in it, and #4829 is in master now, which is the half of this that did land.

jessevz:
Ah that is perfect! I was not aware of solving this through the debug mode

PR #4880, opened 2026-09-17
Remove the duplicated IOKit init block in hwmon_ctx_init ()

matrix:
On macOS, `hwmon_ctx_init ()` runs the same IOKit initialisation twice, and the second run makes the first unreachable. It is 98% of what a short run leaks there, and it costs the same whatever the run does afterwards.

### Reproduction

Tested on v7.1.2-697-g34f12d4c1: macOS 15.7.9 on an Apple M4 Pro, Apple clang 17.0.0.

AddressSanitizer is not the tool here. The Makefile builds with `cc`, which is Apple clang, and Apple clang ships no LeakSanitizer: `detect_leaks=1` answers `detect_leaks is not supported on this platform` and aborts. `leaks` is the platform's own tool, it needs no special build, and what it measures is therefore the binary everyone else runs.

```
$ git clone https://github.com/hashcat/hashcat && cd hashcat
$ git checkout 34f12d4c1
$ make clean -s && make -s -j
$ MallocStackLogging=1 leaks --atExit -- ./hashcat -m 0 -a 3 --potfile-disable \
    --self-test-disable --logfile-disable 5f4dcc3b5aa765d61d8327deb882cf99 '?d?d?d'
```

Four runs of each:

```
                    leaks              bytes
master    83562 (82466-84069)   4133400 (4091456-4168976)
branch       28 (25-29)           73944 (69808-78064)
```

Medians, with the spread of the four in brackets. What the branch is left with belongs elsewhere: the status strings, `backend_ctx_devices_init ()`, an `hcstrdup ()`, and blocks with no frame of hashcat's in them.

The run is three digits of a mask against one MD5 and finds nothing. The size of the leak has nothing to do with the attack: `hwmon_ctx_init ()` runs once at startup, and what it loses is the same whatever comes after.

Almost all of the difference is CoreFoundation, not hashcat's own: `NSNumber`, `NSMutableDictionary`, `CFString` and `NSMutableArray` by the ten thousand. That is what an IOReport subscription is made of. `IOReportCopyAllChannels ()` and `IOReportCreateSubscription ()` are called from one place in the tree, `iokit_init ()` at `src/ext_iokit.m:520` and `533`, so a second subscription is the only thing that can account for them.

### Cause

The `#if defined (__APPLE__)` block that allocates `hwmon_ctx->hm_iokit`, opens the SMC connection and starts the IOReport subscription appears twice in `hwmon_ctx_init ()`, byte for byte, in straight-line code with no branch between them. Both run whenever `need_iokit` is true. The second `hcmalloc ()` overwrites the pointer the first stored, so the first struct, its connection and its subscription become unreachable, and `iokit_close ()` reaches only the second.

The two copies have different origins. The lower one is the original, from bc4ce4cbe in 2021. The upper one was added by 1978231c8, the change that split `hwmon_ctx_init ()` into per-library functions and introduced `hwmon_ctx_init_sysfs_amdgpu_iokit ()`: that helper reads `hm_iokit` and skips the IOKit capabilities when it is NULL, so the init had to move above it. The copy was added and the original was left where it was.

That is also what decides which one goes. `hwmon_ctx_init_sysfs_amdgpu_iokit ()` sits between the two and needs the first to have run, so the second is the one to remove.

Nothing between them writes to `hm_iokit`. That helper is the only code there that names it, in three reads and no assignment, and the only `hcfree (hwmon_ctx->hm_iokit)` in the file are the two inside the blocks themselves, on the path where `iokit_init ()` failed. `iokit_close ()` is reached once, from `hwmon_ctx_destroy ()`.

One thing does change. `iokit_init ()` ends by taking the baseline the power reading is a delta from:

```
  iokit->pwr_t1 = mach_absolute_time ();
  iokit->pwr_e1 = hm_IOKIT_IOReport_get_gpu_energy (iokit->sub, iokit->subscribed);
```

That baseline used to be the second block's and is now the first's, taken 41 lines earlier, across the NVML, ADL and sysfs initialisation. The first power sample of a run therefore covers a slightly longer window. Both points are before any kernel is built, let alone run.

`hwmon_ctx_destroy ()` calls `iokit_close ()`, which releases the SMC connection and the two CoreFoundation objects and returns. It does not free the `IOKIT_PTR` around them, and nothing else does. That is the second change.

### Regressions

What the removed block was there to do is read the hardware, so the question is whether the copy that stays still does it. A short run on each tree, `-m 0 -a 3 --runtime 12` over `?d?d?d?d?d?d?d?d`, which ends on its own:

```
                 SMC fans            IOReport power
master      Fan0: 0%, Fan1: 0%          12527 mW
branch      Fan0: 0%, Fan1: 0%          12787 mW
```

Both subsystems answer on both sides. The power differs because it is a live reading of what the machine is drawing, not a property of the build.

The same command under `--machine-readable`, where the status line carries the fields as numbers:

```
master   TEMP -1   UTIL 0   POWER 8361
branch   TEMP -1   UTIL 0   POWER 8202
```

`TEMP` and `UTIL` are `-1` and `0` on both. This Mac reports neither, before the change and after it, so they are not evidence either way; `POWER` is, and it is there on both.

### Test suites

```
tools/test_edge.sh -m 0     master   160 combinations   Errors detected: 0
                            branch   160 combinations   Errors detected: 0
```

10 of the 160 are skipped on each side, the `-a 9` with a single salt.

`bash tools/test_package.sh --no-device .` ends the same way on both:

```
```


jsteube:
In "hwmon.c", in the branch of hwmon_ctx_destroy() that shuts the IOKit subscription down, this is the new part:

```
  if (hwmon_ctx->hm_iokit)
  {
    iokit_close (hashcat_ctx);

    // iokit_close () gives back what the subscription holds. The IOKIT_PTR around it is ours.

    hcfree (hwmon_ctx->hm_iokit);

    hwmon_ctx->hm_iokit = NULL;
  }
```

The branch just above it, the one for hm_sysfs_cpu, calls its close function and stops there, because that function gives the struct back itself. In "ext_sysfs_cpu.c":

```
void sysfs_cpu_close (void *hashcat_ctx)
{
  hwmon_ctx_t *hwmon_ctx = ((hashcat_ctx_t *) hashcat_ctx)->hwmon_ctx;

  SYSFS_CPU_PTR *sysfs_cpu = (SYSFS_CPU_PTR *) hwmon_ctx->hm_sysfs_cpu;

  if (sysfs_cpu)
  {
    hcfree (sysfs_cpu);
  }
}
```

Could you put the hcfree() inside iokit_close() instead? In "ext_iokit.m" the four failure paths of iokit_init() already free that pointer, so the success path's free lands in the same file as the rest of the ownership, and the branch in hwmon_ctx_destroy() then reads like the five above it. The assignment to NULL can go with it, since hwmon_ctx_destroy() ends on a memset over the whole struct. That is the only thing I would ask for before this goes in.

PR #4884, opened 2026-09-18
Accept the salt field in -m 19800 and -m 19900

### LINKED ISSUE #2783: Kerberos 5, etype 18, Pre-Auth (19900) with salt
Hello,

Is it possible to add support for salt when using Kerberos 5, etype 18, Pre-Auth?

$krb5pa$18$user$domain$salt$token

I can crack such a format using John, but I'd rather use Hashcat.

Thank you.

bishwabikash:
Fixes #2783.

### Problem

`-m 19800` and `-m 19900` derive the PBKDF2 salt as `uppercase(realm) + user` and accept only the four field line:

```
$krb5pa$18$user$realm$enc_timestamp+checksum
```

RFC 4120 section 4 makes the default salt the realm followed by every component of the principal name, with no separators, and the real salt reaches the client in PA-ETYPE-INFO2 (RFC 4120 section 5.2.7.5) rather than being derived - so it is observed, and it can be anything. `uppercase(realm) + user` is right for an Active Directory user account and wrong for:

- a multi-component principal: `host/foo.example.com@EXAMPLE.COM` salts as `EXAMPLE.COMhostfoo.example.com`, with the `/` absent, since it is a printing convention rather than part of the name
- an AD computer or trust account ([MS-KILE] 3.1.1.2)
- an MIT or Heimdal principal given a non-default keysalt (`norealm`, `onlyrealm`, `special`, or a literal salt in Heimdal's `default_keys`)
- an account renamed since its password was last set

In those cases hashcat derives a key from the wrong salt and the hash never cracks, with nothing said - the attack runs to exhaustion as though the password were not in the wordlist.

krb5pa-sha1 always writes a salt field, empty when the salt is the default and filled when it is not (`src/krb5pa-sha1_fmt_plug.c`, the `if (p[-1] == '$')` branch in `get_salt`), so its output is one of:

```
$krb5pa$18$user$realm$$enc_timestamp+checksum
$krb5pa$18$user$realm$salt$enc_timestamp+checksum
```

Neither loaded. Every one of the format's own test vectors ended in a token length exception, not only the ones carrying a salt.

This is also a correctness fix beyond John compatibility. Responder writes `$krb5pa$18$` with no salt field while its own code derives `salt = realm + cname.lower()`, so a capture whose cname has an uppercase letter is silently uncrackable today.

### Approach

The decoder counts the fields before it builds the token spec and keeps which of the three layouts it read in the esalt, numbered the way the other krb5 modules number theirs (1 for hashcat's own layout, as in module_32100.c). An empty salt field means the default salt, so it parses as the five field layout and derives the salt the way it always has.

jsteube suggested module_09800.c in the issue, which tokenizes once and retries with a longer spec. That does not transfer here: 09800's two shapes differ by a whole fixed-length token, so the first attempt fails with exactly `PARSER_TOKEN_LENGTH` and the retry is diagnostic. Both krb5pa shapes end in a variable 104 to 112 character token, so a failed first attempt says nothing about which layout was meant, and retrying blind replaces the error a malformed line should report:

```
etype 17 line fed to -m 19900   Signature unmatched
truncated line                  Token length exception
```

Those are what you get now. With a retry they both became `Separator unmatched`, and the token length count that drives hashcat's "wrong hash type?" advice stopped firing.

The encoder reproduces the layout that was read, so `--show`, `--left` and the potfile give back the line that was supplied. The field is added to the kernel copies of the struct as well, unused there, so the two definitions keep the same length.

The salt is taken as raw ASCII and used verbatim, which is what krb5pa-sha1 does with it:

```c
pbkdf2_sha1((const unsigned char*)saved_key[index], len[0],
       cur_salt->salt, strlen((char*)cur_salt->salt),
       4096, tkey[0], key_size, 0);
```

### Verification

All six test vectors from `krb5pa-sha1_fmt_plug.c` recover, including the two AD ones:

```
etype 17 user=user1   -> openwall
etype 18 user=user1   -> openwall
etype 18 user=user1   -> openwall
etype 18 user=l33t    -> openwall
etype 18 user=aduser  -> password@123
etype 18 user=aduser  -> password@12345678
```

A key with a machine account salt cracks only when the salt field carries it, and runs to exhaustion when it is stripped:

```
19800  with salt -> hashcat      stripped -> not cracked
19900  with salt -> hashcat      stripped -> not cracked
```

The existing example hashes rewritten into five field form, with `HASHCATDOMAIN.COMhashcat` spelled out, recover unchanged, so the explicit salt reaches the KDF and agrees with the computed one. All three layouts round-trip byte identical through `--show`.

The test modules generate all three layouts, alternating on the salt:

```
$ ./tools/test_edge.sh -m 19800
> Errors detected: 0
$ ./tools/test_edge.sh -m 19900
> Errors detected: 0
```

Both modules compile with no warnings under `-W -Wall`.

### Deviations and limits

CONTRIBUTING.md asks for one problem per pull request and this changes two hash modes. 19800 and 19900 carry the same defect in the same code, and fixing one leaves the other silently wrong, so I treated it as one problem. Happy to split it if you would rather.

A principal name containing `$` still cannot be expressed in this format, which matters because an AD machine account's sAMAccountName ends in one. The producers write it verbatim and the field count then shifts, so the line is rejected - John the Ripper rejects it too, for the same reason. Dropping the trailing `$` from the account field and passing the real salt cracks it, and before this change there was no way to crack a machine account at all.

hashcat uppercases the realm when it derives the default salt where krb5pa-sha1 does not. RFC 4120 puts the realm in verbatim; uppercase is convention, not rule. Changing it would alter what existing hashes derive, so I left it.

`krb2john.py` writes a `user:` login prefix, so its output needs `--username`.

Not part of this change: `-m 19600`, `-m 19700`, `-m 28800` and `-m 28900` build the same `uppercase(realm) + user` salt with no way to override it, and John's matching formats do accept an explicit salt - `krb5-17`/`krb5-18` accept *only* one. I will raise those separately rather than widen this.


bishwabikash:
Force-pushed. The first version only handled the salt field when it was filled, which missed the more common case: krb5pa-sha1 always writes the field and leaves it empty for the default salt, so all six of its own test vectors still failed to load. The parser now takes an empty field to mean the default salt, and all six recover. Test modules and description updated to match.

bishwabikash:
Force-pushed again: format values renumbered to 1/2/3 so 1 means hashcat own layout, as in module_32100.c and module_13100.c, and a stale layout comment copied in from krb5tgs removed. No behaviour change. Description now carries the sourced motivation and says why module_09800.c does not transfer here.

PR #4886, opened 2026-09-19
Fix five leaks in src/hashcat.c, and make the teardown safe

matrix:
`src/hashcat.c` loses buffers in five places, and the one an ordinary run reaches is `autodetect_hashmode_test ()`, which every `--identify` walks once per hash mode: it gives back nothing on three of its returns, and on no path at all does it give back the four blocks `hash_info` owns. That is 2144 allocations on any `--identify` that finds a mode, 256 MB when the input hash is empty, and up to 373 MB on the four that need a refusal or a failure to reach.

### Reproduction

Tested on v7.1.2-725-ge4df80905, on Ubuntu 24.04.5, kernel 6.8.0-139, with an RTX 4080 on driver 615.71.09 and an RX 6900 XT on ROCm 7.2.4, and on macOS 15.7.9 on an Apple M4 Pro.

```
$ git clone https://github.com/hashcat/hashcat && cd hashcat
$ git checkout e4df80905
$ make clean -s && make DEBUG=2 -j -s
$ export ASAN_OPTIONS=protect_shadow_gap=0:use_sigaltstack=0:detect_odr_violation=0
```

The figures below are the SUMMARY line LeakSanitizer prints at exit. A run that leaks nothing prints no such line, which is what the first `after` is.

```
$ ./hashcat --identify ""
before: 255888825 byte in 4516 allocations
after:  no SUMMARY line, nothing reported

$ ./hashcat --identify 5f4dcc3b5aa765d61d8327deb882cf99
before:    177752 byte in 2144 allocations
after:        132 byte in    4 allocations

$ ./hashcat --stdout -m 0 -a 3 '?d?d?d?d?d?d' --limit 1
before: 373216724 byte in   48 allocations
after:     148794 byte in   19 allocations
```

What the branch is left with are four blocks the dynamic linker and the backend runtime own, with no frame of hashcat's in them.

Exit codes cannot be read off that build, because ASAN replaces the status of any run it reports leaks for. Taken from an ordinary `make clean -s && make -s -j` instead, the same three commands give 255, 0 and 255 on both sides, the first printing `No hash-mode matches the structure of the input hash.` and the third `Use of -m/--hash-type is not supported with --stdout.`. Behaviour does not change.

Grouped by the first frame above `src/memory.c`, the empty-hash run names two of the five:

```
autodetect_hashmode_test    253488825 byte   4515 obj
hashcat_session_execute       2400000 byte      1 obj
```

and the `--stdout` refusal names what `outer_loop ()` was holding:

```
                             master                  this branch
mask_ctx_init          337407076 byte   8 obj                  -
hashes_init_stage4      33554432 byte   2 obj                  -
cpt_ctx_init             2097152 byte   1 obj                  -
mask_append_final           8013 byte   2 obj                  -
hashes_init_stage2           596 byte   4 obj                  -
hashes_init_selftest         588 byte   2 obj                  -
bitmap_ctx_init               64 byte   8 obj                  -
hashconfig_init                9 byte   2 obj                  -
call_init                 145408 byte   2 obj     145408 byte   2 obj
posix_memalign              1536 byte   1 obj       1536 byte   1 obj
operator new                 216 byte   3 obj        216 byte   3 obj
no hashcat frame            1634 byte  13 obj       1634 byte  13 obj
```

Both columns add up to their SUMMARY line. Master reported 373216724 on eight of ten runs in one sweep and nine of ten in another, and the rest came out lower by exactly one allocation, because a block inside `mask_ctx_init ()` is sometimes not made. The branch reported 148794 bytes in 19 allocations ten times out of ten, in both sweeps. The two places left leak only when a thread or an autotune fails, so no ordinary run reaches them.

### The five places

`outer_loop ()` destroys eleven contexts before its normal return, twelve in a `WITH_BRAIN` build. It has 47 returns, and on 42 of them it destroys none of the eleven. The `--stdout` refusal above is one of those, taken where eight of the eleven exist. The sequence moves into `outer_loop_destroy ()`, and all 47 returns call it. Each destroy is safe on a context that was never built: six return on a flag such a context still has false, `generic_ctx_destroy ()` does the same one level down, `hashconfig_destroy ()` has a branch for a module that never loaded, and `hashes_destroy ()` and `status_progress_destroy ()` free a zeroed one. Being reached twice would matter, because `hashconfig_destroy ()` ends on `module_unload ()`, which calls `hc_dlclose ()` without clearing the handle. It does not arise: each of the 47 calls is immediately followed by the return it belongs to. Upstream had written the sequence out four times inside this one function, of 5, 11, 12 and 12 calls, and all four become the call.

`autodetect_hashmode_test ()` allocates `digest`, `salt`, `esalt`, `hook_salt`, a `hash_info` with four blocks of its own, and a `hashes_buf`. Three returns above the end give back none of them, and the end gives back the `hash_info` struct without the four blocks it owns, which is what `hash_info_destroy ()` does for the loader. `autodetect_hashmode_test_destroy ()` now does both at all four exits, and clears the five pointers `hashes` keeps into those buffers.

`inner2_loop ()` loses `threads_param` and `c_threads` when autotune fails, `hashcat_session_execute ()` loses `usage_sort_buf` when `--identify` matches nothing, and `outer_loop ()` loses `inner_threads` on three error returns.

### Two things the shared teardown needed first

`bridges_salt_destroy ()` does not zero anything of its own. `bridges_init_late ()` sets `bridge_ctx->enabled` before it loads the library, so a bridge whose `platform_init ()` gave up leaves the flag true, every hook resolved, and `platform_context` NULL, and the three C bridges dereference that context in `salt_destroy` without checking. Master never reaches it there, because two of the four copies leave `bridges_salt_destroy ()` out and that failure path is one of them. With `platform_init ()` made to return NULL, on an ordinary build:

```
master                          exit 255, "Platform initialization failed" on stderr, no benchmark line
this branch without the guard   exit 139, segmentation fault, core dumped
this branch                     exit 255, same answer as master
```

Built with ASAN instead, the middle row names the read. The line numbers below are that tree's, with the guard taken out and `platform_init ()` forced, so they do not line up with the patch:

```
==1665203==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000008
==1665203==The signal is caused by a READ memory access.
    #0 salt_destroy         src/bridges/bridge_argon2id_reference.c:251
    #1 bridges_salt_destroy src/bridges.c:432
    #2 outer_loop_destroy   src/hashcat.c:944
    #3 outer_loop           src/hashcat.c:1558
```

Frame 0 is `for (int unit_idx = 0; unit_idx < bridge_argon2id->units_cnt; unit_idx++)`, and address 8 is where `units_cnt` sits in a `bridge_argon2id_t` whose first member is a pointer. The guard puts frame 1 back to a return.

Three of the 47 returns are in the block that starts the monitor and outfile-check threads. The first two are reached with nothing running, because the block above wants `--stdout` and the block below wants its absence. The third is the outfile-check thread failing to start, and by then the monitor thread of the same block is up, reading the contexts a teardown gives back. With only that one `hc_thread_create_ok ()` made to fail:

```
without the wait, 5 runs of 10:
==1666628==ERROR: AddressSanitizer: heap-use-after-free on address 0x514000000055
READ of size 1 at 0x514000000055 thread T34

this branch, 0 runs of 10:
no AddressSanitizer error of any kind
```

Half the runs rather than all of them, because the read has to land after the free and the thread does not always get there first. Two sweeps of ten gave four and five on the unfixed side, and none on this one either time. The address and the thread number are the same in every run that reports, so it is one object and one reader. The report stops at those two lines because the symbolizer deadlocks against the threads still holding locks, which leaves a header and no frames, so what names the reader is the code rather than the trace: at that return the monitor thread started by the block above is running, and `outer_loop_destroy ()` gives back what it reads.

`inner_threads_destroy ()` carries the wait that stood only at the end of the run to all four exits. `outcheck_ctx` is on by default, so that third return is the ordinary configuration rather than a corner of it.

### No regression

Ten of the eleven contexts have `outer_loop_destroy ()` as their only caller anywhere in `src/`. The eleventh is `hashconfig_destroy ()`, and each of its six other callers pairs a `hashconfig_init ()` of its own with the destroy that follows it, so none of them shares a live `hashconfig` with `outer_loop ()`.

`./hashcat -b --benchmark-min 0 --benchmark-max 1500` calls `outer_loop ()` 59 times in a row, which is the shape that would show a context given back too early or twice. On an ordinary build both sides benchmark the same 59 modes, print the same 177 speed lines, write nothing to stderr and exit 0, and the list of mode headers is identical line for line.

`./hashcat -m 0 -a 3 --potfile-disable --self-test-disable 4a7d1ed414474e4033ac29ccb8653d9b '?d?d?d?d'` cracks on both and reports `Recovered........: 1/1 (100.00%)` on both, with no `ERROR: AddressSanitizer` on either. Under ASAN the leaks it does report are the same on both sides down to the caller, 151007 bytes in 60 objects, and none of the eleven contexts is among them.

`--keyspace`, `--show` and `--left` all leave `outer_loop ()` early, and all three write byte for byte the same output as master on stdout and on stderr, with an empty potfile and with one that holds the hash.

### Test suites

```
tools/test_edge.sh -m 0        160 processed   Errors detected: 0
tools/test_edge.sh -m 70000     90 processed   Errors detected: 0
tools/test_edge.sh -m 70100     90 processed   Errors detected: 0
tools/test_edge.sh -m 70200     90 processed   Errors detected: 0
tools/test_edge.sh -m 72000     90 processed   Errors detected: 0
tools/test_edge.sh -m 73000     90 processed   Errors detected: 0
tools/test_edge.sh -m 74000     90 processed   Errors detected: 0
```

This changes no mode, it changes code every mode runs through, so `-m 0` stands for all of them. `Errors detected: 0` is what the suite prints. The middle column is not, it is the count of `# Processing Hash-Type` lines in the run, and it is there because a verdict of 0 on its own cannot tell a clean run from a skipped one. The six bridge modes are here because this changes `src/bridges.c` as well, and two of them want opposite interpreters: `-m 72000` runs only against a free-threaded Python and `-m 73000` only against one without it. One build covers both, because the makefile for the 72000 plugin compares the version with the free-threaded suffix stripped, so an ordinary 3.14.7 builds it. Only the run needs the free-threaded interpreter, and it goes around that one command.

`bash tools/test_package.sh --no-device .` ends with `## ./hashcat: every check passed` on both sides, on macOS.

Where this bends the one problem per pull request rule: the first of the five forces a shared teardown, and that teardown cannot be sent without the two guards above, so those three have to travel together. The other two, in `inner2_loop ()` and `hashcat_session_execute ()`, are the same defect in the same file found by the same read, and each is a two line fix. Splitting them would be three pull requests against one function's neighbours. They can go out separately if you would rather have them that way.


jsteube:
In "hashcat.c" the new autodetect_hashmode_test_destroy() gives back the 4 blocks a hashinfo_t owns:

```
  if (hash_info != NULL)
  {
    hcfree (hash_info->dynamicx);
    hcfree (hash_info->user);
    hcfree (hash_info->orighash);
    hcfree (hash_info->split);
  }
```

Master grew a function for that in #4878, which is in your base. In "hashes.c":

```
static void hash_info_destroy (hashinfo_t *hash_info)
{
  if (hash_info == NULL) return;

  user_t *user = hash_info->user;

  if (user != NULL)
  {
    // user_name, user_gecos and user_home all point into one allocation and user_name is its front.

    hcfree (user->user_name);

    hcfree (user);
  }

  dynamicx_t *dynamicx = hash_info->dynamicx;

  if (dynamicx != NULL)
  {
    hcfree (dynamicx->dynamicx_buf);

    hcfree (dynamicx);
  }
```

That difference is harmless today: the probe allocates both structs itself through hcmalloc(), which zeroes, and no module_hash_decode() in the tree writes either inner pointer, so both of those frees are no-ops. What it leaves is 2 functions that have to stay in step, already differing on arrival.

Would you drop the static from hash_info_destroy() and call it here instead? It does not block anything, the copy is correct as it stands.

matrix:
Done in the commit above: `hash_info_destroy ()` loses the `static` and the probe calls it, so there is one copy.

You are right that the two extra frees are no-ops here. The probe allocates both inner structs with `hcmalloc ()`, which zeroes them, and the only writers of `user_name` and `dynamicx_buf` in the tree are `hashes_dynamicx_line ()` and `hashes_init_stage1 ()`, both on the loader's side.

`--identify` over a file holding one MD5 hash gives the same mode list and the same leak report on both sides, three runs each, and no double free. On the branch with this commit `./tools/test_edge.sh -m 0` ends `Errors detected: 0` and `bash tools/test_package.sh --no-device .` ends `## ./hashcat: every check passed`.

PR #4891, opened 2026-09-19
More Python tests

thatux:
Follow up on https://github.com/hashcat/hashcat/pull/4817#issuecomment-5615950245

9 commits. Each stands on its own and they are in the order they have to land, so they can
also go as separate pull requests.

    cb76447d0  Implement the edge entry point in tools/test_module_runner.py
    36a537fa0  Port the test oracles that shell out to python3
    72bc5b4bd  Replace the pygost and pycryptoplus checkouts with PyPI packages
    71df31c74  Rename tools/test.pl to tools/test_module_runner.pl
    cca0fa921  Generate the same passwords in the python engine as in the perl one
    c98f4cb66  Move the shared oracle code into tools/test_modules/lib
    c56ba5910  Point the plugin development guide at the python test engine
    df8b756f2  Say in CONTRIBUTING.md that a line ends at its last visible character
    894afd856  Report a kernel type for a mode whose kern_type is dynamic

## 1. The edge entry point

`tools/test_edge.sh` generates its vectors through the oracle's `edge` subcommand and only the
perl engine had one, so a mode whose oracle is a `.py` was skipped and nothing was tested:

```
git checkout master
python3 tools/test_module_runner.py edge 1000 0 1
edge is not implemented in test_module_runner.py yet, use tools/test.pl
rc=1
```

After:

```
python3 tools/test_module_runner.py edge 1000 0 1
1000,0,1,0,0,,,3331643663666530643136616539333162373363353964376530633038396330
1000,0,1,27,0,f09f988036e4b8ad3931e382ab35343135e3818b37eab080eab080,,3833643538643165396137323338353438396632373664633436623233396439
rc=0
```

The length matrix, the corners it covers and the hex encoded output are the perl engine's, and so
is the non ASCII password generator behind them. The two agree:

| check | result |
| --- | --- |
| `non_ascii_supported ()` | the same answer for all 598 modes |
| `sprinkle_non_ascii ()` | byte for byte identical for every length 0 to 80 |
| `edge` word and salt lengths | 306 of 306, over 17 modes, nine attack types, both kernel families |

The second one is not cosmetic. The layout is seeded from the password length alone so the eight
passwords of a length share one mask, and `test_edge.sh` rewrites the `?d` at a position whose
byte is not a digit into that byte. A generator that placed the characters differently would
break every mask the suite builds.

`edge` keeps the perl engine's constraint substitution, where a mode with a kernel for only one
family has that family's word and salt pairs copied over the other. `single ()` refuses instead,
and deliberately, but `test_edge.sh` picks the family from hashcat's own `Kernel.Type(s)` and
counts an empty vector list as an error, so refusing there would report a mode as broken for
having nothing to say about a kernel it does not have.

`test_edge.sh` routes to the engine the mode has a file for, with the same `run_oracle` dispatch
`tools/test.sh` uses. Both directions run:

```
./tools/test_edge.sh -m 31100 -D 1 -f      # a .py mode:  Errors detected: 0, 160 processed, 0 skips
./tools/test_edge.sh -m 0 -D 1 -f          # a .pm mode:  Errors detected: 0, 160 processed, 0 skips
```

## 2. The fourteen oracles that shell out

8800, 9000, 14500, 17010, 17020, 17030, 17040, 17050, 31100, 32500, 32700, 33400, 34300 and
34301 built their hash by running `python3` through a shell with the password on the command
line, so a password carrying a shell metacharacter never reached the hash intact:

```
git checkout master

perl -Itools/test_modules -e 'require "m31100.pm"; my $h = module_generate_hash ("a\$(id -u)b", ""); print "$h\n"'
2a660fe372f1a092da0d17f97029d6e53fb5729ac2751bd93f9bdda39efe3ef9

perl -Itools/test_modules -e 'require "m17010.pm"; my $h = module_generate_hash ("a\"b", ""); printf "returned %d chars\n", length ($h)'
sh: 48: Syntax error: "(" unexpected
returned 0 chars
```

The first is SM3 of `a1000b`, the password after the shell ran `id`. A password with a double
quote produces no hash at all. After:

```
python3 -c 'import sys; sys.path.insert (0, "tools/test_modules"); import m31100; print (m31100.module_generate_hash (b"a$(id -u)b", None))'
0f7ef5d5d62705625551257a4b9a32fd2f7328955105dc82e6ccb37062a5e858

python3 -c 'import sys; sys.path.insert (0, "tools/test_modules"); import m17010; print (len (m17010.module_generate_hash (b"a\"b", None)))'
726
```

`module_verify_hash` could not round trip in 12 of the 14. Generating a hash and handing it
straight back, on master:

| behaviour | modes |
| --- | --- |
| returned undef | 14500 17010 17020 17030 17040 17050 34300 34301 |
| returned a different hash | 8800 9000 32500 33400 |
| reproduced the hash | 31100 32700 |

The eight split the crack line into fields the format does not have, and the four threw the hash
away and built a new one out of fresh randomness. Each module rebuilds the hash out of the hash
now, which also checks the integrity field the kernel decides on, and all 14 reproduce it.

Every module was checked against the one it replaces by feeding it hashes the `.pm` generated,
four per mode, all reproduced byte for byte.

## 3. The two git dependencies

`tools/requirements.txt` was the only file in the tree that installed from a git URL. pygost is
not on PyPI at all, not as an old release and not under another name:

```
curl -s -o /dev/null -w '%{http_code}\n' https://pypi.org/pypi/pygost/json
404

curl -s https://pypi.org/simple/ | grep -oiP '(?<=>)[^<]+(?=</a>)' | grep -ix pygost
(no output, exit 1)
```

`gostcrypto` takes its place in the seven Streebog modes. MIT, pure python, nothing outside the
standard library, and 5.5 times faster: 0.68 s against 3.76 s for 1000 Streebog-256 digests of a
32 byte message, which the suite feels because these oracles run once per candidate. Feeding each
mode's eight vectors, generated under pygost, back through the patched module: 8/8 on all six of
11700, 11750, 11760, 11800, 11850 and 11860.

m35600 needed a change of its own, and it is why the dependency carries a comment in
`requirements.txt`. gostcrypto's `update ()` only keeps the state right while what it has been fed
is a whole number of 64 byte blocks:

```
python3 -c '
import gostcrypto
m = bytes (range (256)) * 2
h = gostcrypto.gosthash.new ("streebog512", data = bytearray (m[:100]))
h.update (bytearray (m[100:]))
one = gostcrypto.gosthash.new ("streebog512", data = bytearray (m))
print (bytes (h.digest ()) == bytes (one.digest ()))'
False
```

m35600 was the one module that streamed a message in pieces of other sizes, so its hash moved for
every password of 40 bytes or more, and `tools/test.sh -m 35600` reported 4/8 not found. It
assembles each message and hashes it once now. Against the pygost module over 9 passwords by 4
salts by 8 round counts: 288 of 288 identical, and the mode is back to 0/8. The standard library's
`hmac` was never affected, because the first thing it feeds the hash is the 64 byte pad: 200
random message lengths agree with pygost.

pycryptoplus is the other one, and there is no package to swap it for. Nothing else on PyPI offers
XTS over Twofish or Serpent:

* `cryptography` has AES, Camellia, SM4, SEED, IDEA, CAST5, Blowfish, ARC4, TripleDES and
  ChaCha20, and native AES-XTS. No Twofish, no Serpent.
* `pycryptodome` has the same gap.
* `twofish` 0.3.0 is the only Twofish on PyPI and it is dead on Python 3.12 and later, at
  `import imp`.
* Botan has all three plus XTS, but it is not on PyPI under any name.

So the three DiskCryptor modules build the mode themselves, twenty lines over a 16 byte block
cipher, with AES from pycryptodome and the other two from vendored files. Against CryptoPlus
`MODE_XTS`, 20 random 256 bit key pairs and 512 byte buffers per cipher:

```
aes      encrypt 20/20  decrypt 20/20
twofish  encrypt 20/20  decrypt 20/20
serpent  encrypt 20/20  decrypt 20/20
```

`pytwofish.py` is the file CryptoPlus itself ships, so that cipher does not change. `pyserpent.py`
is a different implementation of the same cipher: CryptoPlus's own Serpent is Frank Stajano's 1998
reference code, 1354 lines of bitstring arithmetic written for clarity rather than speed, at 9 ms
a block. This one comes from the pyserpent package on PyPI, Konstantin Vitkovskii's modernisation
of the Bjorn Edstrom code the Twofish file also descends from, both tracing back to Brian Gladman's
independent implementations. It matches CryptoPlus block for block over 20 random keys and is 190
times faster, 0.003 s against 0.577 s for 64 block encryptions. Two characters in its header were
normalised to ASCII; the notices are otherwise untouched.

m20011, m20012 and m20013 became `.py` in the same commit, because they were the last three users
of CryptoPlus and they reached it through a shell. Both directions were checked against the `.pm`
they replace, 12 hashes per mode each way, all reproduced byte for byte.

### The option not taken: pycryptoplus from PyPI with the one import filled in

pycryptoplus is on PyPI, and it ships the two cipher files this commit vendors. The reason the
tree carries a git URL for it instead is one line: its `__init__` reads `parse_version` out of
`pkg_resources`, and setuptools dropped `pkg_resources` in 82.0.0, released 2026-02-08. The newest
release does not carry it:

```
python3 -c "
import zipfile
z = zipfile.ZipFile ('setuptools-84.0.0-py3-none-any.whl')
print (any (n.startswith ('pkg_resources/') for n in z.namelist ()))"
False
```

and it is not published on its own either, so nothing can pull it back in:

```
curl -s -o /dev/null -w '%{http_code}\n' https://pypi.org/pypi/pkg_resources/json
404
```

`parse_version` is the whole of what CryptoPlus wants from it, in four files, so the name can be
supplied from `packaging` in about ten lines before the import. That does work. With
`pkg_resources` made to raise outright, the PyPI release produces what `matrix/pycryptoplus`
produces:

```
aes      f4405455438eab0715cd7d34d678955b  matches the fork
twofish  c601eb812fce0f4091ce95063f12178b  matches the fork
serpent  3d735a1ef711c9441c2e7589e31ef322  matches the fork
```

with 20 of 20 encrypt and decrypt round trips on each of the three ciphers, on 256 bit key pairs
and 512 byte buffers, and AES accepted at 128, 192 and 256 bits.

It is a real option and it is the smallest diff of the three: ten lines against this commit's
1139. It was not taken because those ten lines are the fork's patch written a second time, in this
tree rather than in the package, and because it keeps a package with no release since 2018 on the
dependency list of the test suite, still needing to be patched at import time to load at all.
Vendoring the two files it would have supplied ends the question instead. The trade is 1139 lines
of third party cipher, of which 1071 are the two ciphers and 68 are the chaining mode, against one
dependency that has to be repaired before it can be imported.

## 4. Rename test.pl to test_module_runner.pl

The suite has two oracle engines behind one interface and only one of them was named after it.
Every live reference moves with the file: the `run_oracle` dispatch in `tools/test.sh` and
`tools/test_edge.sh`, `tools/test_coverage.sh`, `tools/install_modules.sh`, both READMEs, the
plugin development guide, the 21 test modules that name it in a comment, and the usage text it
prints itself. The `docs/changes.txt` and v7.0.0 release note mentions are left alone: they say
what shipped under the old name.

## 5. The passwords the python engine generates

`single` and `multi` were the last entry points where the two engines disagreed. The perl one
sprinkles multi byte UTF-8 into the passwords it generates, per mode, and the python one produced
digits:

```
git checkout master
python3 tools/test_module_runner.py single 1000
0 of 8 words carry a multibyte character

python3 tools/test_module_runner.py single 1000
5 of 8 words carry a multibyte character
```

`tools/test.sh` builds its multi hash files out of what `single` prints, so the one change covers
both target types. `password` is implemented with it, which was the last subcommand missing:

```
python3 tools/test_module_runner.py password 1000 20 | xxd | head -1
00000000: e0a4 b930 e282 ac34 e9be 8de0 a4b9 33e6  ...0...4......3.

perl tools/test_module_runner.pl password 0 20 | xxd | head -1
00000000: e0a4 b931 e282 ac30 e9be 8de0 a4b9 31e6  ...1...0......1.
```

Same characters in the same places, which is what the masks need; the digits between them are per
password. `NO_NON_ASCII` still asks for ASCII, as `test.sh -g` does under `-O`.

## 6. The shared code moves into a package

`tools/test_module_runner.py` puts `tools/test_modules` first on `sys.path`, so every helper beside
the modules shadowed any installed distribution of the same name for the whole run. `gpg` is a real
one, the GPGME bindings, and the modules importing our `gpg.py` are the GPG modules:

```
import gpg  ->  tools/test_modules/gpg.py
```

As a package, neither hides the other:

```
import gpg            ->  site-packages/gpg.py           the GPGME bindings
from lib import gpg   ->  tools/test_modules/lib/gpg.py
```

`test_helpers.py`, `gpg.py`, `diskcryptor.py`, `xts.py`, `pyserpent.py` and `pytwofish.py` move
into `tools/test_modules/lib`, and the 23 import lines that reach them are rewritten. The suite
scripts glob `m[0-9][0-9][0-9][0-9][0-9]` with a `.pm` or `.py` suffix, so a subdirectory is
invisible to them.

## What this does not fix

Four things turned up on the way. None is caused by these commits and none is fixed by them.

**m20011, m20012 and m20013 disagree with the kernel on a password that is not ASCII.** They hand
it over as UTF-16, and hashcat cracks neither the widened form the oracle produces nor the decoded
one. I generated both and tried each: ASCII cracks, both non ASCII variants come back rc=1. It is
red on master today, for the same reason:

```
git checkout master
./tools/test.sh -m 20011 -D 1 -f -P -t all
[ Type 20011, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Pure, Vector-Width 1 ] > Error : 6/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
```

The `.py` port had been hiding it by not generating those passwords, which commit 5 puts back. The
conversion is left as the `.pm` had it so both engines say the same thing about it.

**`-m 34300` segfaults on a low Argon2 memory cost.** Every value I tried from 64 to 12288 KiB
gives rc=139 in the bridge; 32768 and 262144 KiB are fine. The oracle draws `1 << randint (12, 18)`
KiB, so 4096 KiB is in range and this is reachable on master today.

**Four modes fail in multi mode only**, and none of them because of anything here. 9000, 14500 and
32500 fail the same way with master's `.pm`. 5200 has no `.pm` to compare against, so it was run
both ways instead, and the password content makes no difference to it:

```
NO_NON_ASCII=1 ./tools/test.sh -m 5200 -D 1 -f -P -t multi
[ Type 5200, Attack 0, Mode multi, Device-Type Cpu, Kernel-Type Pure, Vector-Width 1 ] > Error : 0/1 not found, 1/1 not matched, 0/1 timeout, 0/1 skipped

./tools/test.sh -m 5200 -D 1 -f -P -t multi
[ Type 5200, Attack 0, Mode multi, Device-Type Cpu, Kernel-Type Pure, Vector-Width 1 ] > Error : 0/1 not found, 1/1 not matched, 0/1 timeout, 0/1 skipped
```

9000 behaves the same under that test. Both are OPTS_TYPE_BINARY_HASHFILE modes, where multi mode
concatenates the artifacts into one file.

## Test runs

`./tools/test.sh -m <mode> -D 1 -f -P -t all` on every mode with a Python oracle, and on the seven
Streebog modes, after all five commits. Each cell is single and multi at vector width 1 then 4, `o`
for OK and `E` for Error:

```
m1000  oooo   m17010 oooo   m20011 EEEE   m32500 oEoE   m11750 oooo   m11860 oooo
m5200  oEoE   m17020 oooo   m20012 EEEE   m32700 oooo   m11760 oooo   m35600 oooo
m8800  oooo   m17030 oooo   m20013 EEEE   m33400 oooo   m11800 oooo
m9000  oEoE   m17040 oooo   m31100 oooo   m34301 oooo   m11850 oooo
m14500 oEoE   m17050 oooo                                m11700 oooo
```

Every `E` is one of the four listed above. `./tools/test.sh -m 0 -D 1 -f -t all` passes on all four
combinations, which is the renamed perl engine going through the same dispatch.

`./tools/test_edge.sh -m <mode> -D 1 -f`, every vector width:

```
m31100   Errors detected: 0    160 processed, 0 skips     python engine
m0       Errors detected: 0    160 processed, 0 skips     perl engine
m17010   Errors detected: 0     90 processed, 0 skips
m20011   Errors detected: 90    90 processed, 0 skips     master's .pm gives 90 of 90 as well
m20012   Errors detected: 90    90 processed, 0 skips     master's .pm gives 90 of 90 as well
```

`./tools/test_edge.sh -m <mode> -V 1 -D 1 -f`, vector width 1 only, which is where the rest of the
modes this branch changes were run. The processed count is quoted beside the error count on
purpose, for the reason in the next section:

```
m8800    processed 18   errors 0
m9000    processed  9   errors 0
m11700   processed 32   errors 0
m11750   processed 18   errors 0
m11760   processed 18   errors 0
m11800   processed 32   errors 0
m31100   processed 32   errors 0
m32700   processed 18   errors 0
m14500   processed  0   errors 0   <- see below, this one tested nothing
```

Twelve of the modes this branch changes have no edge run of their own: 11850, 11860, 17020, 17030,
17040, 17050, 20013, 32500, 33400, 34300, 34301 and 35600. They are covered by the tools/test.sh
table above, and by the .pm cross checks, but not by tools/test_edge.sh. The family members that
were run are the ones that share their code: 17010 stands for the m170x0 group over lib/gpg.py,
20011 and 20012 for the m2001x group over lib/diskcryptor.py, and 11700 to 11800 for the Streebog
group. 17050 and 34300 are the two that do not finish in a sensible budget on this hardware.

**`tools/test_edge.sh -m 14500` tested nothing and reported success.** Fixed in 894afd856, kept
here because it is what the sweep above ran into. The mode loop asks hashcat which kernel families
a mode has and iterates over the answer:

```
kernel_types=$(./hashcat -m ${hash_type} -HH | grep 'Kernel.Type(s' | cut -d: -f2 | xargs | sed -e 's/,//g')
```

m14500 does not print that line:

```
./hashcat -m 14500 -HH | grep -i kernel
  Name................: Linux Kernel Crypto API (2.4)
  Kernel.Type.Filter..: pure

./hashcat -m 31100 -HH | grep -i kernel
  Kernel.Type(s)......: pure, optimized
  Kernel.Type.Filter..: pure
```

so kernel_types is empty, the loop body never runs, and the suite ends green with nothing done:

```
! cleaning cache ...
[ test_edge_1789855380 ] > All tests done in 0d:00h:00m:00s
[ test_edge_1789855380 ] > Errors detected: 0
```

`src/terminal.c:1403` prints that line only where `has_pure_kernel` or `has_optimized_kernel` is
set, and `src/interface.c` left both false for a mode that names its kernel from inside the parser.
Five modes do that: 14500, 14600, 16500, 16501 and 24600, and they were the five with no line.

They still declare a kern_type of their own and it points at a kernel that exists, so 894afd856
answers the flags from it, and `tools/test_edge.sh` gets a guard so the next mode to reach that
state is not silent either. After it, no mode of the 598 is without the line, and the five report:

```
m14500  18 processed   8 errors
m14600   0 processed   9 errors
m16500  18 processed   0 errors
m16501  18 processed   0 errors
m24600  18 processed   0 errors
```

Neither failure is new. 14500 gives the same 8 of 18 with master's own `m14500.pm` in place, so
they are the mode's, uncovered rather than caused. 14600 is LUKS and its oracle wants sudo to build
a container, which this machine does not grant without a terminal, so it emits records with an
empty hash field and the existing "empty test vectors" check rejects them.

## What I could not run

`tools/test.sh -m 34300` at vector width 4. Argon2 at up to 256 MiB with `--runtime 400` per
vector does not fit a sensible budget on the CPU backend; width 1 took over two hours on its own.
What it did report there is the segfault above, in 3 of the 8 vectors a run draws:

```
[ Type 34300, Attack 0, Mode single, Device-Type Cpu, Kernel-Type Pure, Vector-Width 1 ] > Error : 3/8 not found, 0/8 not matched, 0/8 timeout, 0/8 skipped
[ Type 34300, Attack 0, Mode multi,  Device-Type Cpu, Kernel-Type Pure, Vector-Width 1 ] > OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped
```

## TODO: the oracles that still shell out

Fifteen `.pm` modules still reach python3 through a subprocess. None of them puts the password
through the shell, so none carries the bug commit 2 removed, and there is no correctness reason to
pull any of them into this pull request. Twelve pass their arguments as a list:

```
cd tools/test_modules
grep -l '"-|", *"python3"' *.pm

m11700.pm  m11750.pm  m11760.pm  m11800.pm  m11850.pm
m11860.pm  m17200.pm  m17210.pm  m17220.pm  m17225.pm
m17230.pm  m35600.pm
```

and three use backticks, but interpolate only base64 and hex into the command, never the
candidate:

```
grep -l '`python3' *.pm

m27500.pm  m27600.pm  m35100.pm

grep -h 'python_code =~ s/' m27500.pm m27600.pm m35100.pm | sort -u
$python_code =~ s/c_data/"$c_base64"/;
$python_code =~ s/data/"$data"/;
$python_code =~ s/key_main/"$key_main"/;
$python_code =~ s/key_tweak/"$key_tweak"/;
$python_code =~ s/loops_data/$loops/;
$python_code =~ s/p_data/"$p_base64"/;
$python_code =~ s/s_data/"$s_base64"/;
$python_code =~ s/sm3_data/"$data_base64"/;
$python_code =~ s/tmp_data/"$tmp_base64"/;
```

which is the position m20011, m20012 and m20013 were in, and why jsteube's list was fourteen
rather than twenty. What is left in all fifteen is a subprocess per candidate and an oracle
written in two languages at once.

The seven Streebog modules in the first list are the ones this branch already touches for
gostcrypto, so they are the cheapest next port: the python is written and measured, it only has to
move out of the heredoc and into a `.py`.


matrix:
@thatux please do not remove .pm now. Thanks!

thatux:
I prefer not to keep them, that'd end up requiring to support two codebases (both the Python and Perl tests).

Discussed before:
"add mNNNNN.py, delete mNNNNN.pm in the same commit" https://github.com/hashcat/hashcat/pull/4817#issuecomment-5586453424

We can stall merging this commit, but could you please add some arguments why you prefer that?

matrix:
> I prefer not to keep them, that'd end up requiring to support two codebases (both the Python and Perl tests).
> 
> Discussed before: "add mNNNNN.py, delete mNNNNN.pm in the same commit" [#4817 (comment)](https://github.com/hashcat/hashcat/pull/4817#issuecomment-5586453424)
> 
> We can stall merging this commit, but could you please add some arguments why you prefer that?

Because the current testing pipeline relies on the PMs, and if you remove them, I can no longer test them. For example, in one of my local branches, I had to restore the first two that you removed

thatux:
With this PR all testing pipeline that used the .pm now chooses the .py when it's available, for example https://github.com/hashcat/hashcat/pull/4891/changes#diff-065379b962a3bcd1614b4fe4816bf31dd02ba2e18816a0700fee6506a7462d6cR368

Are there any places left that now only use the pm?

jsteube:
The plan is still to move every mode to Python and remove its ".pm" file in the same commit that adds the corresponding ".py" file. That keeps one implementation of each mode. As for what still requires a ".pm" file, the last check was in "test_edge.sh", which skipped edge cases without one. This PR adds the missing edge case entry point to the Python runner and removes that check. Everything else already looks for both file types.

The concern matrix raised is separate from that, and we need to address it before merging these conversions. The suite passes for the 17 modes converted here, but that only shows they still crack. It does not establish that the Python oracles behave the same as the Perl versions they replace.

To make that comparison part of every conversion, could you add a reusable tool under "tools/" that runs both engines for a selected mode? We need to compare hashes generated from identical inputs, check that both engines reject the same invalid inputs, and cover the options supported by "test.pl" rather than just one configuration. We should then require agreement between the engines before merging each conversion, rather than relying on the existing suite alone.

Two things stand in the way of that comparison, and both are worth doing regardless. "test.pl" gives an oracle 10 random helpers and the Python engine has 5, so "random_count()", "random_string()", "random_lowercase_string()", "random_uppercase_string()" and "random_mixedcase_string()" still have to come across. And neither engine seeds its random generation, so two runs of one mode produce different salts even within the same engine. Either a fixed seed on both sides, or a way to hand both the same salt and password, is what makes their output comparable at all.

thatux:
Sure, makes sense, like this?

## The py-pm-test-compare tool

    tools/test_engine_compare.py --ref master --all
    tools/test_engine_compare.py --ref master 17010

A converted mode has no `.pm` left in the tree, so the perl side comes out of git and `--ref` says
where to look. Both engines then run over the entry points `tools/test.sh` and `tools/test_edge.sh`
drive, for the optimized and the unoptimized kernel, and a list of eleven malformed lines has to be
refused by both.

Two comparisons, because one of them is not always possible.

**Cross verification, always.** Each engine verifies what the other generated, through its own
`module_verify_hash ()`. A hash one engine produced and the other re-derives out of its own fields
is the same algorithm in both, whatever salt was drawn. This is the check that means anything for
the fourteen modes whose `.pm` shells out to `python3`: that embedded script calls
`get_random_bytes ()` and `randint ()`, and no seed reaches inside it.

**Byte comparison, where both engines are seedable.** `HCTEST_SEED` puts both on one generator, so
the same salts and the same words come out and the output can be compared line for line. The tool
works out which case it is in by running each engine twice under one seed, and says so rather than
failing where a seed cannot reach.

Three outcomes rather than two. A mode where the engines disagree fails. A mode whose `.pm` cannot
verify its own output is reported separately, because that is a defect the conversion inherited
rather than caused. A mode with no kernel for the family under test is not applicable: the python
engine exits 2 there, which `test.sh` turns into a Skip, while the perl engine substitutes the
other family's constraints and reports as if the kernel were there, which is the deliberate
difference from #4817.

## What it says about the conversions in #4891

    tools/test_engine_compare.py --ref master --all

    engines agree on all 19 modes
    the perl oracle cannot verify its own output for 12 of 19 modes, which is a defect of the .pm
    rather than of the conversion: m08800 m09000 m14500 m17010 m17020 m17030 m17040 m17050 m32500
    m33400 m34300 m34301

| what | modes |
| --- | --- |
| identical byte for byte on every entry point | m20011 m20012 m20013 m31100 m32700 |
| cross verified, byte comparison not possible | m08800 m09000 m14500 m17010 m17020 m17030 m17040 m17050 m32500 m33400 m34300 m34301 |
| nothing to compare, converted before #4891 | m01000 m05200 |

Not one cross verification failed. Every python oracle re-derives every hash its `.pm` produced.

## The five helpers

`tools/test_module_runner.pl` hands a `.pm` ten random helpers and
`tools/test_modules/lib/test_helpers.py` had five, so a mode whose `.pm` draws a count, a mixed case
string or an alphanumeric one had nowhere to draw it from on the python side. `random_count ()`,
`random_string ()`, `random_lowercase_string ()`, `random_uppercase_string ()` and
`random_mixedcase_string ()` are the five that were missing.

`random_bytes ()` changes with them. The perl engine builds those out of a hex string of twice the
length rather than drawing bytes, and the two have to draw the same way or they are not on the same
stream.

## The seed

Neither engine seeded anything, so two runs of one mode gave different salts. That is what the
suites want, because a mode that only works for one salt is a mode that is broken, and it is also
why the two engines could not be compared at all.

`HCTEST_SEED` sets a seed for both. Under it every helper draws from the generator
`sprinkle_non_ascii ()` already uses, implemented the same way on both sides:

    HCTEST_SEED=7 perl    tools/test_module_runner.pl password 31100 16
    HCTEST_SEED=7 python3 tools/test_module_runner.py  password 31100 16

    both: 6435793334329671

`List::Util`'s `shuffle` draws from perl's own generator, which the seed does not reach, so the
length pools go through a Fisher Yates over the shared stream instead, matching what the python
engine does. Unseeded, both engines behave exactly as they did:

    python3 tools/test_module_runner.py password 31100 12   twice, two different passwords

jsteube (changes requested):
In "tools/test_engine_compare.py" the byte comparison is exact:

```
  if left == right:
```

Running it at "--seed 7" rather than the default 1 puts m32700 on the wrong side of that, in the optimized family only. Both engines emit the same 8 vectors, with one of them at position 7 on the perl side and position 8 on the python side, so sorted they are equal and compared as bytes they are not:

```
  m32700 IS_OPTIMIZED=0 seed=7   IDENTICAL
  m32700 IS_OPTIMIZED=0 seed=42  IDENTICAL
  m32700 IS_OPTIMIZED=1 seed=7   DIFFER
  m32700 IS_OPTIMIZED=1 seed=42  IDENTICAL
```

One seed therefore does not establish that the engines agree, and m32700 sits in the table as byte identical because 1 is a seed where it happens to. Could you run a handful of seeds instead of one, and carry per-mode output into the PR rather than the closing summary? The order needs a decision with it. If the order of the vectors carries no meaning, sorting both sides before the comparison is the fix and the mode is clean. If it does carry meaning, one of the engines is emitting them in the wrong order and that is the thing to find.

In "docs/hashcat-plugin-development-guide.md" the suite is now described as 3 files, and there are 4. The paragraph above the list already covers the case: a mode written before the Python engine still carries a ".pm", and the suite still runs it. That leaves "tools/test_module_runner.pl" belonging in the list beside the other 3. Could you put it back?

In "docs/changes.txt", the entries describing what "tools/test_module_runner.py" gained are changes to something no release has shipped, since v7.1.2 carries "tools/test.pl" alone and the Python engine arrived after it in 731f2ed8c. The entry "The 14 test oracles that ran python3 through a shell" has a second problem, which is the count itself. The TODO at the end of your description puts 15 more still reaching pythhon3 that way, so the number moves at the next conversion and will not be 14 by release. A changelog line is read long after the work, so the shape to use is one that needs no recounting. Could you drop the unreleased ones and take the count out of whatever stays? The rename and the move from the pygost and pycryptoplus checkouts to PyPI both touch files that shipped, so both of those belong. 

The change to "src/interface.c" reaches past the test framework. It sets "has_pure_kernel" and "has_optimized_kernel" for teh 5 modes with a dynamic kern_type, so "-HH" prints a Kernel.Type(s) line for them where it printed none, and the 6 lines that do it are the same 6 the else branch below already has. What reads the flags afterwards did not move with them: "OPTI_TYPE_OPTIMIZED_KERNEL" is set in 2 places and both sit in that else branch, so "-O" on m14500 still takes no effect and still prints no warning, while the new line advertises an optimized kernel. Could you take this out into a PR of its own? It carries a decision about whether "-O" should work for a dynamic kern_type, and that is separate from moving oracles to Python.

thatux:
All four taken. The branch was force pushed, because one of the four was to take a commit out of
it, and the last of the four landed on top of that.

    d34d32af0  Compare the engines over several seeds, and on content not order
    cff0728a5  Keep the unreleased engine out of the changelog
    87c4f4288  Put the perl engine back in the test suite file list
    8fd8c6131  Settle seedability per mode, and check a difference before reporting it

and `894afd856`, the `src/interface.c` one, is gone from here. It is #4900 now, on its own, with the
`-O` decision you asked for settled in it.

## The order carries no meaning

You found m32700 differing at `--seed 7` in the optimized family, with one vector at position 7 on
one side and 8 on the other, and asked which of the two it was. It is neither: the perl engine has
no order of its own to be wrong about.

`single ()` collects its vectors in a hash and emits them at tools/test_module_runner.pl:621 with

```perl
  for my $word (sort { length $a <=> length $b } keys %{$db_prev})
```

which orders on the byte length of the word and ties for every pair of equal length. Perl
randomises hash key order per process, so the tie breaks differently in every run:

    perl -e '%h=(aaa=>1,bbb=>1,ccc=>1,ddd=>1,eee=>1); print join(",", sort { length $a <=> length $b } keys %h), "\n"'

    bbb,ccc,aaa,ddd,eee
    ddd,eee,ccc,bbb,aaa
    bbb,eee,aaa,ccc,ddd

The two vectors you saw swap are `ह5中4Ａ4か` and `ह1中9Ａ2か`, both exactly 15 bytes, so they tie.
It shows without the comparison tool at all:

    HCTEST_SEED=7 IS_OPTIMIZED=1 perl    tools/test_module_runner.pl single 32700
    HCTEST_SEED=7 IS_OPTIMIZED=1 python3 tools/test_module_runner.py  single 32700

    perl,   six runs of one seed: 2 distinct outputs, identical once sorted
    python, six runs of one seed: 1 distinct output

So the perl side disagrees with itself between runs, at one seed, and the difference you saw
between the engines is that same instability seen across the two rather than across two runs of
one. `compare ()` sorts both sides now, which hides nothing, and the diff it prints for a mode that
really differs is over the sorted lines too.

I have left the perl comparator alone. Giving it a tie break, `length $a <=> length $b or $a cmp
$b`, would make the perl engine reproducible at a seed and is one line, but it changes what
`test.sh` and `test_edge.sh` emit for every mode that still has a `.pm`, which is not this pull
request. Say if you want it and it can go in its own.

## Several seeds, and what the extra ones found

`--seed` takes a comma separated handful now, `1,2,3,4,5` by default, and every one is run. A check
is reported once per mode, with the line from the seed that had something to say against it, so the
output stays the size it was:

    !> single                             DIFFERS  (seed 4)

You were right that one seed establishes very little. Five of them found two modes, neither of
which seed 1 has anything to say about, and getting from the first to a tool I would trust took two
goes. The whole of it, because the second attempt is the interesting one:

**The first sweep over five seeds** reported m20011 differing at seed 4. Chasing it: the perl side
is the unsteady one, and it disagrees with itself at a fixed seed.

    HCTEST_SEED=4 IS_OPTIMIZED=0 perl tools/test_module_runner.pl single 20011

    fourteen runs of one seed: two distinct outputs, seven each
    the same eight words both times, 16 bytes differing on two of them

`m20011.pm` shells out to `python3 -c` six times, which is the family the tool already describes as
one no seed reaches. It got past the seedability probe because the probe ran `potthrough`, which is
stable for that mode, while the byte comparison went on to cover `single`, which is not. So the
probe covered both entry points, three runs each rather than twice, and m20011 started reporting
what it should.

**The second sweep, with that probe,** reported m20012. Same family, same shape, and the fix above
does not catch it, because its coin is weighted rather than fair:

    HCTEST_SEED=4 IS_OPTIMIZED=0 perl tools/test_module_runner.pl single 20012

    eight runs of one seed: 6 one answer, 2 the other

A probe can only sample, and sampling cannot prove the absence of a coin. The m20012 block in that
sweep carries no "byte comparison skipped" line anywhere, so three runs came up the same way at all
five seeds and the comparison then differed at one of them. Eight samples do not pin the weighting
down closely enough to argue from a number, and the number is not the point: however many runs the
probe makes, a weighted coin can get past all of them.

`8fd8c6131` is the answer to that, in two parts. The verdict is settled for the mode before any
comparison runs, so a mode can no longer be seedable at one seed and not at the next; the probe
still runs per seed, since a mode can be steady under one and not another, but the answers are
combined. And a difference is checked before it is reported: when a byte comparison comes back
different, both engines are probed again and the line says skipped rather than DIFFERS if either has
moved. That costs nothing on a run where the engines agree, which is every run on a mode that really
is seedable, and it is what catches a coin this weighted.

What both DiskCryptor modes should report is

    byte comparison    skipped, the perl oracle draws from a generator HCTEST_SEED does not reach

rather than a difference that is neither engine's fault, and the table below is the run that says
whether they do.

## Per mode output

PENDING

## What the sweep says about the .pm files

One number in the closing summary is worth reading, and it is not about this branch:

    the perl oracle cannot verify its own output for 12 of 19 modes, which is a defect of the .pm
    rather than of the conversion: m08800 m09000 m14500 m17010 m17020 m17030 m17040 m17050 m32500
    m33400 m34300 m34301

That is the first check the tool runs, before it compares anything: hand an engine its own generated
pairs and ask it to re-derive them. Twelve of the nineteen `.pm` files cannot, which means their
`module_verify_hash ()` draws something fresh instead of reading it back out of the hash. It is why
the cross verification exists at all, and why a mode failing "perl verifies python" is only counted
against the conversion where the perl side passed "perl verifies perl" first. The python oracles
replacing them all verify their own output.

## The suite is four files

Fixed, and you are right that the prose below the list already carries the case:

    * tools/test_module_runner.py: ...
    * tools/test_module_runner.pl: The same program for a mode whose unit-test stub is still a
      m[hash_mode].pm. Both scripts take the same arguments and the two shell scripts below pick
      whichever one the mode has a stub for.
    * tools/test.sh: ...
    * tools/test_edge.sh: ...

## The changelog

The section is `changes v7.1.2 -> v7.1.x`, v7.1.2 shipped `tools/test.pl` alone, and
`tools/test_module_runner.py` arrived after it in `731f2ed8c`, which carries no tag. Three entries
were about what that unreleased thing gained, and they are out:

    - "tools/test_module_runner.py" implements the edge entry point ...
    - The 14 test oracles that ran python3 through a shell are Python modules now ...
    - "tools/test_module_runner.py" seeds its passwords with the same multibyte ...

The two you named as belonging are the two that stay, the requirements moving to PyPI and the
rename, and neither carries a count.

I read "take the count out of whatever stays" as applying to a line that survived, and none did, so
nothing needed rewording. If you would rather keep the oracle conversion as a changelog line in
some count-free shape, say so and it goes back in that form.

thatux:
The `## Per mode output` section of my comment above says PENDING. The sweep has finished, and it is
clean.

    tools/test_engine_compare.py --all --seed 1,2,3,4,5

    engines agree on all 19 modes, over 5 seeds each

| mode | byte comparison | perl verifies perl | python verifies python | python verifies perl | perl verifies python |
| --- | --- | --- | --- | --- | --- |
| m01000 | no .pm here or at origin/master, nothing to compare | | | | |
| m05200 | no .pm here or at origin/master, nothing to compare | | | | |
| m08800 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m09000 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m14500 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m17010 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m17020 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m17030 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m17040 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m17050 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m20011 | skipped | 7 of 7 | 7 of 7 | 7 of 7 | 7 of 7 |
| m20012 | skipped | 7 of 7 | 7 of 7 | 7 of 7 | 7 of 7 |
| m20013 | skipped | 7 of 7 | 7 of 7 | 7 of 7 | 7 of 7 |
| m31100 | same | 7 of 7 | 7 of 7 | 7 of 7 | 7 of 7 |
| m32500 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m32700 | same | 7 of 7 | 7 of 7 | 7 of 7 | 7 of 7 |
| m33400 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m34300 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |
| m34301 | skipped | 0 of 7 | 7 of 7 | 7 of 7 | 0 of 7 |

"byte comparison" is `potthrough`, `single` and `edge` for attack modes 0, 1 and 3, in both kernel
families, and "same" means every one of them matched under every seed. Where it says skipped, an
engine draws from a generator `HCTEST_SEED` does not reach and the comparison would mean nothing;
the cross verification in the four columns beside it is what covers those modes, and it needs no
seed. m01000 and m05200 were converted before this branch, so there is no `.pm` at the reference to
compare against.

The three DiskCryptor modes are the ones that needed `8fd8c6131`, and they now say

    m20011
         byte comparison    skipped, the perl oracle draws from a generator HCTEST_SEED does not reach

rather than reporting a difference that is neither engine's fault.

Twelve modes carry `0 of 7` in the "perl verifies perl" column, which is the section on the .pm
files in the comment above.

PR #4895, opened 2026-09-20
Parser fixes the fuzz targets found

thatux:
Seventeen module parsers and three shared helpers, one class of bug in three of the four commits: a
function is given a length and then reads past it, because what it actually stops at is a NUL. The
fourth is a leak on every error path of the four pkzip parsers.

Found with the fuzz targets in https://github.com/hashcat/hashcat/pull/4774.

    7ffb1f5e6  Bound the operand read in rule_utf8_len ()
    2ecbc9396  Measure the rounds field against the length the caller gave
    ad0770813  Stop the hash line parsers reading past the length they were given
    e9a17cc9e  Free the line copy on every pkzip parser error path

Rebased onto 836f11de1, which fixed one of the five findings while this was open: m22000 and
m22001 built the current line format from the legacy one with `snprintf ("WPA*01*%s***", line_buf)`,
and `%s` ran to the first NUL rather than to `line_len`. That commit rebuilds the line from its
parsed fields, which bounds the read and also fixes the colon separated legacy form, which did not
load before it and does now. The commit that bounded the `%s` here is dropped.

## How these reproduce

Two of the five show through `./hashcat` on its own, and both of those are below with nothing but
hashcat commands. The other three need the targets from #4774, and that is worth being precise
about rather than glossing:

* hashcat reads a hash line into a buffer of `HCBUFSIZ_LARGE` and NUL terminates it, so the byte a
  parser reads past its length is the terminator, or the next field of the same line. It is inside
  the allocation either way, so an AddressSanitizer build of hashcat says nothing.
* a hash given on the command line is not copied either: `hashes.c:3678` hands the parser
  `user_options_extra->hc_hash`, which points into `argv`, and ASan does not track that region. A
  DEBUG=2 build handed a hash shorter than its own signature prints the ordinary parse error and no
  report:

      make clean && make DEBUG=2 -j$(nproc)
      ./hashcat -m 13100 -a 3 -D 1 --force --potfile-disable '$krb5tgs$' '?d'

      Hash parsing error: '$krb5tgs$': Separator unmatched (expected ':')
      no AddressSanitizer report, before or after

  The fuzz target is what makes it visible, because it hands the parser an allocation of exactly
  the length it passes with it.

That is not an argument for leaving them: the byte read is not part of the hash, the function was
told how long its input is, and `hashes.c` already hands parsers slices of a longer line, which
section 2 shows reaching the wrong bytes in a way `./hashcat` prints.

Every reproduction that needs the targets starts the same way, from the branch of #4774:

    tools/fuzz/build.sh

## 1. rule_utf8_len () reads the operand of a rule that has none

Needs the target.

    printf '$' > /tmp/rule_dollar
    ./fuzz_out/fuzz_rule /tmp/rule_dollar

    before: AddressSanitizer: heap-buffer-overflow, READ of size 1,
            rule_utf8_len src/rp.c:292, cpu_rule_to_kernel_rule src/rp.c:384, rc=1
    after:  rc=0

```c
u32 rule_utf8_len (const char *rule_buf, const u32 rule_len, const u32 rule_pos)
{
  const u8 c = (u8) rule_buf[rule_pos];
  ...
  if ((rule_pos + need) >= rule_len) return 1;   // the continuation bytes are checked
```

Every caller passes `rule_pos + 1`, the position of the operand, so a rule whose last command is
`$`, `^`, `i` or `o` is read one byte past its length. The byte decides nothing: a lead byte there
fails the continuation check on the next line, so the function returns 1 whatever is in it. Same
comparison, one line earlier, and the compiler emits the same rules:

    printf 'hashcat\n' > /tmp/w1.txt
    ./hashcat --stdout -a 0 /tmp/w1.txt -r rules/best66.rule | md5sum

    before: 916edf248cbfd931608ea4257979a0be
    after:  916edf248cbfd931608ea4257979a0be

Introduced by 38dc95f4b, "Use UTF-8 for Windows text handling and literal rule characters".

## 2. rounds_count_length () looks for the end of the field past the end of the input

Reproduces in hashcat, no build flags and no harness:

    printf 'user:$5$rounds=5000:19000:0:99999:7:::$x\n' > /tmp/shadow1.txt
    printf 'hashcat\n' > /tmp/w1.txt
    ./hashcat -m 7400 -a 0 -D 1 --force --username --potfile-disable /tmp/shadow1.txt /tmp/w1.txt

    before: Hash parsing error ... Separator unmatched (expected '$')
    after:  Hash parsing error ... Separator unmatched near 'rounds=5000' (expected '$')

The line is a passwd or shadow one, where the hash is the field between the first two colons and
the rest of the line follows it in the same buffer. `hlfmt_hash_passwd ()` hands the parser a
pointer into the middle of the line with a length, and then:

```c
static int rounds_count_length (const char *input_buf, const int input_len)
{
  if (input_len >= 9)
  {
    if (memcmp (input_buf, rounds, 7) == 0)
    {
      const char *next_pos = strchr (input_buf + 8, '$');   // stops at a NUL, not at input_len
```

`strchr ()` runs to the first NUL, which is the end of the whole line rather than the end of the
hash, so it finds the `$` in the last field. The rounds field is then measured as longer than the
field it lives in, the token advances past the end of the hash, and `len_left` goes negative. The
error message is where that surfaces: `tokenizer_error_capture ()` prints the snippet it choked on
only `if (len > 0)`, so before the fix there is no snippet, and after it there is, because the
parser is still inside the hash.

`hc_strchr_next ()` is the bounded search the rest of the tokenizer already uses. Where the buffer
is NUL terminated at `input_len` the two agree, which is why the ordinary hashes are unchanged:

    ./hashcat -m 7400 -a 0 -D 1 --force --quiet --potfile-disable '$5$rounds=5000$saltsalt$WzTQ6v8Nn6pZzmdm56RDydzHM/eup8eaDAZMSm.V6y7' /tmp/w1.txt
    ./hashcat -m 1800 -a 0 -D 1 --force --quiet --potfile-disable '$6$rounds=5000$saltsalt$I/OCh7dg1sUTsQhKVLNqX0F4YWILXsVvLHfzsq5YhZAHi2nltij.ZSP9zEoS3v8Thx3mgcDsFzl57RKNh4ZwK0' /tmp/w1.txt
    ./hashcat -m  500 -a 0 -D 1 --force --quiet --potfile-disable '$1$38652870$DUjsu4TTlTsOe/xxZ05uf/' /tmp/w1.txt

    before: all three cracked, rc=0
    after:  all three cracked, rc=0

## 3. m22000 and m22001 printed the line with %s, and master has fixed it

Left here because the finding is part of the story and the reproduction still runs, against a tree
from before 836f11de1:

    printf '4d4fe7aac3a2cecab195321ceb99a7d0*fc690c158264*f4747f87f9f4*686173686361742d6573736964' > /tmp/wpa_legacy
    ./fuzz_out/fuzz_parse_22000 /tmp/wpa_legacy

    before 836f11de1: AddressSanitizer: heap-buffer-overflow, READ of size 86,
                      module_hash_decode src/modules/module_22000.c:848, rc=1
    after:            rc=0

Both legacy forms load on master now, which the star separated one did and the colon separated one
did not:

    printf 'hashcat!\n' > /tmp/w2.txt
    ./hashcat -m 22000 -a 0 -D 1 --force --quiet --potfile-disable '4d4fe7aac3a2cecab195321ceb99a7d0*fc690c158264*f4747f87f9f4*686173686361742d6573736964' /tmp/w2.txt
    ./hashcat -m 22000 -a 0 -D 1 --force --quiet --potfile-disable '4d4fe7aac3a2cecab195321ceb99a7d0:fc690c158264:f4747f87f9f4:686173686361742d6573736964' /tmp/w2.txt

    both cracked as hashcat!, rc=0

## 4. Two shapes across eighteen parsers

Needs the target. One reproduction for each shape:

    printf '$krb5tgs$17$a' > /tmp/krb
    ./fuzz_out/fuzz_parse_19600 /tmp/krb

    before: AddressSanitizer: heap-buffer-overflow, READ of size 1,
            module_hash_decode src/modules/module_19600.c:113, rc=1
    after:  rc=0

    printf '$krb5asrep$23$user@domain.com:3e156ada591263b8aab0965f5aebd837$007497cb51b6c8116d6407a782ea0e1c5402b17db7afa6b05a6d30ed164a9933c754d720e279c6c573679bd27128fe77e5fea1f72334c1193c8ff0b370fadc6368bf2d49bbfdba4c5dccab95e8c8ebfdc75f438a0797dbfb2f8a1a5f4c423f9bfc1fea483342a11bd56a216f4d5158ccc4b224b52894fadfba3957dfe4b6b8f5f9f9fe422811a314768673e0c924340b8ccb84775ce9defaa3baa0910b676ad0036d13032b0dd94e3b13903cc738a7b6d00b0b3c210d1f972a6c7cae9bd3c959acf7565be528fc179118f28c679f6deeee1456f0781eb8154e18e49cb27b64bf74cd7112a0ebae2102a' > /tmp/k18odd
    ./fuzz_out/fuzz_parse_18200 /tmp/k18odd

    before: AddressSanitizer: heap-buffer-overflow, READ of size 1,
            module_hash_decode src/modules/module_18200.c:252, rc=1
    after:  rc=0

The second input is m18200's own example hash with one character taken off the end of its last
field.

**A search, or a comparison, that runs past the end of the line.** m13100, m18200, m19600, m19700,
m28800, m28900, m35300 and m35400 all do

```c
const char *spn_info_start = strchr (line_buf + 12 + 1, '*');
```

most of them twice, because bounding the first search is what lets the next one be reached. m19600
and m19700 already carry a comment about it: the length check above the call is there so the search
*starts* inside the line, which says nothing about where it *ends*. `hc_strchr_next ()` bounds both.
m13100, m18200, m35300 and m35400 also read the bytes that tell their two formats apart before
knowing those bytes are there, so they get the length check that comparison needs.

**A hex decode that steps two characters at a time and stops at the length.**

```c
for (int i = 0, j = 0; i < in_len; i += 2, j += 1)
{
  out_buf[j] = hex_to_u8 (&in_buf[i]);   // reads in_buf[i] and in_buf[i + 1]
}
```

An odd length reads one past the token, which is the next field of the line. That byte was never
part of the answer: the count is `in_len / 2`, which already ignores the trailing character.
`hex_decode ()` in src/convert.c is where m24420 reaches it, and m05500, m05600, m06600, m08200,
m13100, m13600, m18200, m19600, m19700, m19800, m19900, m27000, m27100, m32100, m32200, m35300 and
m35400 spell the same loop out by hand. The targets proved it in m05600, m13100, m18200, m19600,
m19700, m27100, m32100 and m32200; the rest are the identical loop in a sibling module.

## 5. The pkzip parsers leak the line on every error path

Reproduces in hashcat, with nothing but a sanitizer build:

    make clean && make DEBUG=2 -j$(nproc)
    python3 -c "open('/tmp/pk100.txt','w').write(('\$pkzip2\$1*1*2*0*' + 'A'*8000 + '*\n')*100)"
    printf 'hashcat\n' > /tmp/w1.txt
    ./hashcat -m 17225 -a 0 -D 1 --force --quiet --potfile-disable /tmp/pk100.txt /tmp/w1.txt

    before: SUMMARY: AddressSanitizer: 814165 byte(s) leaked in 104 allocation(s)
              Direct leak of 801800 byte(s) in 100 object(s)
                hcmalloc src/memory.c:50
                hashes_init_stage1 src/hashes.c:4385
    after:  SUMMARY: AddressSanitizer:  12365 byte(s) leaked in   4 allocation(s)

801,800 bytes is exactly the hundred rejected lines, 8,018 each. The four that remain are
`hwmon_ctx_init ()` and `backend_ctx_init ()` at session level and have nothing to do with parsing.

`module_hash_decode ()` in m17210, m17220, m17225 and m17230 copies the line so it can walk it with
`strtok_r ()` and frees the copy once, on the success path at the bottom. The 23 error returns in
between do not, 24 in m17230, so every line hashcat rejects leaks its own length, and hashcat
parses every line of the hash file. A macro rather than a brace around each return, because the
point is that they all release the same thing.

One byte of input is enough to see it through the target instead:

    printf 'x' > /tmp/pk_bad
    ./fuzz_out/fuzz_parse_17225 /tmp/pk_bad

    before: LeakSanitizer: detected memory leaks, direct leak of 2 bytes in 1 object,
            hcmalloc src/memory.c:50, module_hash_decode src/modules/module_17225.c:222, rc=1
    after:  rc=0

and the example hash still cracks:

    ./hashcat -m 17225 -a 0 -D 1 --force --quiet --potfile-disable '$pkzip2$3*1*1*0*0*24*3e2c*3ef8*0619e9d17ff3f994065b99b1fa8aef41c056edf9fa4540919c109742dcb32f797fc90ce0*1*0*8*24*431a*3f26*18e2461c0dbad89bd9cc763067a020c89b5e16195b1ac5fa7fb13bd246d000b6833a2988*2*0*23*17*1e3c1a16*2e4*2f*0*23*1e3c*3f2d*54ea4dbc711026561485bbd191bf300ae24fa0997f3779b688cdad323985f8d3bb8b0c*$/pkzip2$' /tmp/w1.txt

    before: cracked as hashcat, rc=0
    after:  cracked as hashcat, rc=0

## What I could not show

A hash file that turns one of the over reads in sections 1 and 4 into a crash, a wrong crack or a
changed message. The reasons are the two at the top, and I tried both: an ASan build
reading a hash file sees the reads land inside the line buffer, and an ASan build given the hash on
the command line sees nothing at all, because that buffer is `argv`. Section 2 is as close as it
comes, and what it changes is an error message.

So sections 1 and 4 are latent: correct today only because of what the callers happen to do, and one
caller away from mattering. Section 5 needs no such argument.

## Test runs

    make -j12                                             rc=0 before and after
    ./hashcat --stdout -a 0 w1.txt -r rules/best66.rule   identical md5 before and after
    29 modes, each cracked with the password its own module declares, before and after

All of it against master at 836f11de1, where all four of these are still present. The fifth, the
WPA `%s`, is the one that commit fixed.

The findings came out of runs of the targets in #4774: ten minutes each on the rule compiler and
the tokenizer, and forty seconds each on the parsers of the 74 modes that either walk the line with
`strtok`, `strchr`, `sscanf` or a `%s`, allocate with more returns than frees, or carry a hex loop
that steps two characters at a time. Fifteen of those reported something, every one is fixed here,
and fixing the first read in a parser is what let the fuzzer reach the second in four of them.

The rerun, against this branch, is clean:

    28 parser targets at 60 seconds each, plus 3 minutes on the rule compiler and the tokenizer
    30 targets, 330,819,873 executions, no crashes, no leaks, no artifacts written


thatux:
Two more commits. The sweep behind this branch was widened from the four modes the targets build by
default to all 599, which found five more bugs in seven modules.

    38cad1cd0  Bound four more parsers a wider sweep reached
    1cc1e36e0  Convert the last numeric token of m36800 from a copy that ends where it does

| mode | what |
| --- | --- |
| m33400 | reads `line_buf[1]` before the line is known to hold two bytes. One byte of input is enough |
| m07300, m13500 | the hex pair loop in a third spelling, `hex_to_u8 (salt_pos + i)`, where the two reads happen inside the helper rather than being named in the loop, so the pass that bounded the other spellings walked past these |
| m10600, m10700 | validate and then read a fixed 80 characters of a field the tokenizer allows to be shorter, taking the length the line declares for itself as the bound |
| m36800 | `if ((ct_offset + 16) > (u32) data_len)`, where `ct_offset` comes off the line: a value near `UINT32_MAX` wraps, passes a check it cannot pass, and the read starts that far into the blob |
| m36800 | and then `hc_strtoul ()` on the last token, which runs to the end of the line, so nothing stops it reading past the hash |

m10600 is in hashcat's own benchmark list. m36800 is two days old, from e352c2422, and is the one
that reported a second bug once the first was fixed.

The `hc_strtoul ()` shape is worth a word, because it is not only here: 14 modules call it straight
on a token buffer, and the other 13 are fine for the reason this one was not, that their numeric
token has a separator behind it and `hc_strtoul ()` stops at the first character that is not a
digit. A bounded conversion in convert.c would settle the shape for all of them, and that belongs
in a pull request of its own rather than in this one.

Each finding carries its own before and after in the commit that fixes it, with the target from
#4774 and the input that reproduces it.

Verification of these two commits:

    build                                        rc=0
    m07300 m10600 m10700 m13500 m33400 m36800    each cracks its own example hash, before and after
    the same six as fuzz targets, 3 minutes each, 4 workers apiece, all clean

`tools/test_edge.sh` over every mode this branch touches runs next, on an otherwise idle machine,
and I will post what it says.

thatux:
One more commit, and a different class from the six above it: not a read past a length, but
undefined behaviour in a shift, and found with UndefinedBehaviorSanitizer rather than
AddressSanitizer.

    b0fd60373  Pack a MAC address into a u32 without shifting into the sign bit

The four WPA parsers build their PMKID input out of the two MAC addresses a byte at a time:

```c
wpa->pmkid_data[4] = (mac_sta[2] <<  0)
                   | (mac_sta[3] <<  8)
                   | (mac_sta[4] << 16)
                   | (mac_sta[5] << 24);
```

Each byte promotes to `int`, so a MAC byte of 0x80 or more shifts into the sign bit of a signed 32
bit type. The compilers hashcat is built with all produce the value the code wants; the cast says
so rather than relying on it.

It reproduces on the example hash, before any fuzzing starts, so the seed corpus of the target is
enough:

    CFLAGS="-g -O1 -fsanitize=fuzzer-no-link,address,undefined -fno-omit-frame-pointer" \
      tools/fuzz/build.sh
    UBSAN_OPTIONS=halt_on_error=1 ./fuzz_out/fuzz_parse_22000 fuzz_out/work/seeds/parse_22000

    before: src/modules/module_22000.c:1007:38: runtime error: left shift of 244 by 24 places
            cannot be represented in type 'int'
    after:  no report

UBSan reported m22000 and m22001. m16800 and m16801 carry the identical three lines each, so all
four modes are cast, twelve sites in total.

Verification:

    build                                          rc=0
    m16800 m16801 m22000 m22001                    each cracks its own example hash, before and after
                                                   (the two deprecated ones with --deprecated-check-disable)
    tools/test_edge.sh -m 22000 / -m 22001 -V 1 -f  0 errors, 18 vectors each
                                                   16800 and 16801 are skipped by the suite as deprecated
    32 targets under ASan and UBSan, one minute each, nothing else reported

A sweep of all 599 modes under both sanitizers is running now.

jsteube (changes requested):
The odd length looks worth refusing rather than tolerating. Bounding a loop with "(i + 1) < len" stops the read leaving the token, but the line is still accepted, with the field one byte shorter than it declared and the parse reporting success. A length check ahead of the loop refuses it instead, and it is one line per token:

```
  if (token.len[1] & 1) return (PARSER_TOKEN_LENGTH);
```

I would keep the bounds you already added and put the check in front of them, rather than swap one for the other. The check could also sit in "input_tokenizer()" instead of in each module, since a VERIFY_HEX field that decodes to bytes has an even length by definition. Three tokens would not fit that rule as it stands: m20011, m20012 and m20013 each mark a 1 character version field VERIFY_HEX and read it as a character rather than decoding it, so carrying the rule centrally means a separate attribute for hex that decodes to bytes.

Either way two modules are still open. "src/modules/module_05300.c" and m05400 carry eight loops each in the third spelling, with both reads inside the helper:

```
  for (int i = 0; i < token.len[0]; i += 2) *ptr1++ = hex_to_u8 (token.buf[0] + i);
```

Tokens 0 to 7 in both take "len_min" 0 against a "len_max" of 1024 or 128, so an odd length reaches those loops today.

The cast lands on the last line of each group of four, in twelve groups across m16800, m16801, m22000 and m22001. Taking one from "src/modules/module_16800.c":

```
  wpa_pmkid->pmkid_data[2] = (wpa_pmkid->orig_mac_ap[0]  <<  0)
                           | (wpa_pmkid->orig_mac_ap[1]  <<  8)
                           | (wpa_pmkid->orig_mac_ap[2]  << 16)
                           | ((u32) wpa_pmkid->orig_mac_ap[3] << 24);
```

Each group is one expression lined up on the shift column, and the two spaces after "orig_mac_ap[0]" are there to hold that column against the longer "orig_mac_sta" in the group below it. The last line drops the padding and gains a parenthesis, so the column breaks in all twelve. Could you cast all four operands in each group instead?

The same class of shift sits outside the WPA modules. In "src/modules/module_28503.c" the reassembly works on "u8 t[64]" holding 5 bit values, and shifts it by 27, 29 and 31:

```
  digest[2] = (t[13] << 31) | (t[14] << 26) | (t[15] << 21) | (t[16] << 16)
```

A shift of 31 is undefined for any non-zero "t[13]", so this one does not need an unusual value the way a MAC byte of 0x80 does. m28504, m30903 and m30904 carry the same three lines. Would you take those here as well?

thatux:
All four taken, and the `input_tokenizer ()` suggestion with them. Five commits land with this
push; the first two were written before the review arrived.

    6731e6cc7  Shift unsigned where the value reaches the sign bit
    cb5d76210  Bound a shift exponent and a salt terminator taken from the line
    fb06254f4  Refuse a hex token whose length is odd
    37ba3fedd  Cast every operand in the WPA MAC packs, not only the one that overflows
    78457af13  Refuse an odd hex token in the tokenizer, not in each module

## The odd length, in input_tokenizer ()

You were right that the bound alone leaves the line accepted with a field one character shorter than
it declared, and right that the tokenizer is where the rule belongs. `78457af13` puts it there, so
it covers every mode rather than the ones a fuzz target walked into.

    ./hashcat -m 16100 -a 3 -D 1 --force --potfile-disable '$tacacs-plus$0$5fde8e68$4e13e8fb33df0$c006' '?d'

    before: Status Exhausted, Recovered 0/1, rc=1      six bytes decoded, the thirteenth character dropped
    after:  Token length exception: 1/1 hashes, rc=255 the line is refused

m16100 is not one of the modules this branch had touched. Its token 2 takes a range of 12 to 256, so
thirteen hex characters sat inside it.

`TOKEN_ATTR_VERIFY_HEX` keeps its name and its bit. It is part of what a plugin outside this tree
compiles against, and renaming it would break those for a rule that is the meaning it already
carries at 1000 of the 1005 tokens that set it. It gains the length half beside the character check:

```c
    if (token->attr[token_idx] & TOKEN_ATTR_VERIFY_HEX)
    {
      if (token->len[token_idx] & 1) return (PARSER_TOKEN_LENGTH);

      if (is_valid_hex_string (token->buf[token_idx], token->len[token_idx]) == false) return (PARSER_TOKEN_ENCODING);
    }
```

The tokens that are hex characters read as characters take a new bit, `TOKEN_ATTR_VERIFY_BASE16`,
which checks the characters and says nothing about the length. It sits with `VERIFY_BASE58`,
`VERIFY_BASE64A` and `VERIFY_BECH32` because that is what those mean too, and adding a bit is
additive, so nothing outside the tree stops building.

Five tokens in four files rather than the three in the review:

| file | token | the field |
| --- | --- | --- |
| module_20011.c, module_20012.c, module_20013.c | `token[1]` | one character version, compared against `'0' + DISKCRYPTOR_VERSION` and written back with `%i` |
| module_13600.c | `token[5]` | `verify_bytes`, read with `sscanf ("%4x")`, `len_min` 1 |
| module_13600.c | `token[6]` | `compress_length`, read with `hc_strtoul (..., 16)`, `len_min` 1 |

m13600 is not in the review and matters more than the other three, because its own example hash
carries an odd value in that field:

    $zip2$*0*1*0*0675369741458183*5dc5*0**36b85538918416712640*$/zip2$

`compress_length` is that bare `0`, so the mode stops loading its own example hash the moment the
rule lands, which is how it turned up. `TOKEN_ATTR_VERIFY_DIGIT` would have done for the three
DiskCryptor fields, whose only legal value is `'0'`, but not for this one, where `5dc5` has letters
in it.

Those five are the whole set. A sweep of all 1005 `TOKEN_ATTR_VERIFY_HEX` sites for a token that can
take an odd length, then a run of every one of the 599 modes against its own example hash, finds
nothing else. The rest divide into 99 sites reaching `generic_salt_decode ()`, which already refuses
an odd length at src/parser.c:567, and 169 that decode with `hex_decode ()`, `hex_to_u32 ()` or
`hex_to_u8 ()`, where an even length is what they want.

## The per module checks, and why they are still there

`fb06254f4` is the form you asked to see first, 27 checks in front of the bounds they already had,
and the bounds stay as you said. m18200's own example hash with the last character removed:

    echo hashcat > /tmp/one.txt
    ./hashcat -m 18200 -a 0 -D 1 --force --potfile-disable '$krb5asrep$23$user@domain.com:3e156ada591263b8aab0965f5aebd837$007497cb51b6c8116d6407a782ea0e1c5402b17db7afa6b05a6d30ed164a9933c754d720e279c6c573679bd27128fe77e5fea1f72334c1193c8ff0b370fadc6368bf2d49bbfdba4c5dccab95e8c8ebfdc75f438a0797dbfb2f8a1a5f4c423f9bfc1fea483342a11bd56a216f4d5158ccc4b224b52894fadfba3957dfe4b6b8f5f9f9fe422811a314768673e0c924340b8ccb84775ce9defaa3baa0910b676ad0036d13032b0dd94e3b13903cc738a7b6d00b0b3c210d1f972a6c7cae9bd3c959acf7565be528fc179118f28c679f6deeee1456f0781eb8154e18e49cb27b64bf74cd7112a0ebae2102a' /tmp/one.txt

    before: Status Exhausted, Recovered 0/1, rc=1      the line is loaded and cracked against
    after:  Token length exception: 1/1 hashes, rc=255 the line is refused

Most of those 27 cannot fire any more, because the tokenizer refuses the line before the module sees
it. Not all of them: a module can take a length off a token the tokenizer did not mark `VERIFY_HEX`.
m32100 and m32200 are the clear case. Their JtR branch reads the blob from token 2, which that branch
declares as

```c
    token.len_min[2] = 1;
    token.len_max[2] = 512;
    token.attr[2]    = TOKEN_ATTR_VERIFY_LENGTH;
```

with no `VERIFY_HEX` on it, and then hex decodes it. The check in the module is the only thing
standing in front of that one.

Where the token does carry `VERIFY_HEX`, the check is dead weight, and I am happy to take those back
out if you would rather have the rule in one place only. Say which way you want it.

There is no hashcat only reproduction of the read itself, for any of these. hashes.c hands the parser
a NUL terminated line buffer, so the byte one past an odd token is the next field of the same
allocation and AddressSanitizer has nothing to report. The exact sized buffer in the fuzz target of
#4774 is what turns it into a finding, and that reproduction is in ad0770813. The commands above show
the other half of your point, that the line was taken at all.

## m05300 and m05400

Both, eight loops each, and the diagnosis in the review is exact: tokens 0 to 7 took a `len_min` of 0
against a `len_max` of 1024 or 128, so an odd length reached them, and the two reads are inside
`hex_to_u8 ()` where the pass that bounded the other spellings did not look. They take the bound as
well as the check, since they never had the bound:

```c
  if (token.len[0] & 1) return (PARSER_TOKEN_LENGTH);
  ...
  for (int i = 0; (i + 1) < token.len[0]; i += 2) *ptr1++ = hex_to_u8 (token.buf[0] + i);
```

m05300's own example hash with the last character removed from token 0 goes from `Status Exhausted,
rc=1` to `Token length exception: 1/1 hashes, rc=255`; the whole line is in the commit message.

One more of the same class that the review does not name: m24420 reaches the shape through
`hex_decode ()` in convert.c, which returns the whole pairs it could read and says nothing about the
character it left behind, so its two variable length tokens take the check at the call site.

## The cast column

All twelve groups, all four operands:

```c
  wpa_pmkid->pmkid_data[2] = ((u32) wpa_pmkid->orig_mac_ap[0]  <<  0)
                           | ((u32) wpa_pmkid->orig_mac_ap[1]  <<  8)
                           | ((u32) wpa_pmkid->orig_mac_ap[2]  << 16)
                           | ((u32) wpa_pmkid->orig_mac_ap[3]  << 24);
```

The padding after `orig_mac_ap[0]` is back, so the column holds against the longer `orig_mac_sta` in
the group below. The shift the sanitizer reported is the one at 24 and it keeps the cast it had, so
what UndefinedBehaviorSanitizer sees is unchanged from b0fd60373; the other three shift a byte by 0,
8 and 16, which is defined for any value a MAC byte can hold, so this is the shape leaving the file
rather than a second finding.

## m28503, m28504, m30903 and m30904

Already done, in `6731e6cc7`, which was written before the review and is in this push. It casts every
term of the digest reassembly in all four rather than the three lines the sanitizer reached, so
`<< 31`, `<< 30`, `<< 29`, `<< 28` and `<< 27` are all covered:

```c
  digest[2] = ((u32) t[13] << 31) | ((u32) t[14] << 26) | ((u32) t[15] << 21) | ((u32) t[16] << 16)
            | ((u32) t[17] << 11) | ((u32) t[18] << 6) | ((u32) t[19] << 1) | (t[20] >>  4);
```

The `checksum` block above it is left alone on purpose: `t[]` holds base32 values, so 31 at most, and
`31 << 25` is 0x3e000000, which is inside a signed 32 bit type. Say so if you would rather have it
uniform anyway.

That commit came out of a sweep of all 599 modes under UndefinedBehaviorSanitizer, 51 reports from 11
distinct places, 38 of them one line in `OpenCL/inc_common.cl`, the helper every `append_0x80_*`
reaches:

```c
  const u32 r = 0xff << ((offset & 3) * 8);
```

`cb5d76210` is the other half of that sweep, the two findings that are bugs rather than promotions: a
shift exponent taken from the line in m14200, and a write past the end of a `u32[64]` in m29100.

## Test runs

PENDING

thatux:
The `## Test runs` section of my comment above says PENDING. The suite has run since.

    build                                             rc=0
    599 modes, each parses its own example hash       0 failures
    30 modes, each cracks its own example hash        rc=0

`tools/test_edge.sh` over every mode this branch touches, at vector width 1:

    tools/test_edge.sh -m <mode> -V 1 -D 1 -f

    05300 0   05400 0   05500 0   05600 0   06600 0   07300 0   08200 0   13100 0
    13500 0   13600 0   16100 0   18200 0   19600 0   19700 0   19800 0   19900 0
    22000 0   22001 0   24420 0   27000 0   27100 0   32100 0   32200 0

    16800 skipped, deprecated          16801 skipped, deprecated

    20011 18   20012 18   20013 18     the same 18 on master
    35300 16   35400 18                master 18 and 13, same suite, same flags

Three things in that worth saying out loud.

Vector width 1 rather than the whole matrix is a deviation from what CONTRIBUTING.md asks for, and
the reason is that everything on this branch is in the hash line parser. A line is read once, before
any kernel runs, and reading it does not vary with the vector width, so the other four widths re-run
the same parse behind a different kernel. The tokenizer change in `78457af13` is the one piece with
wider reach, and the 599 mode parse above is what covers that rather than the edge suite.

m16800 and m16801 report zero because the suite skips a deprecated mode, not because they were
tested. They do crack their own example hash with `--deprecated-check-disable`, which is what the
30 mode line above is.

m20011, m20012 and m20013 fail 18 of 18, and so does master, to the vector. I chased that one
because the coincidence looked bad: those are exactly the three modules whose `token[1]` this branch
retypes to `TOKEN_ATTR_VERIFY_BASE16`. They are not it, and the baseline is the evidence rather than
the argument. It is a kernel bug, and it now has its own branch and its own reproduction.

`m20011-pure.cl` calls `sha512_hmac_init_global_utf16le_swap ()`, which decodes the UTF-8 password
into a private `enc_buf` and builds an HMAC key out of it. That goes wrong twice:

* `u32 enc_buf[256];` is never zeroed, while `hc_enc_next_global ()` writes only as far as the
  decoded text reaches. The branch for a key below the 128 byte block reads a whole block out of it
  regardless, so everything past the password is stack contents.
* the branch for a key above the block hashes it down with `sha512_update_utf16le_swap ()`, which
  runs the UTF-8 decoder over its input, while `enc_buf` already holds the UTF-16LE the helper
  decoded fifteen lines above. The key is encoded a second time.

Between them those two cover the whole length range, so no non-ASCII password cracks in these three
modes today. The identical pair of lines is in the SHA-1 and SHA-256 helpers, which m31200 and
m33700 use.

An ASCII password of the same 256 bytes cracks, because `hc_enc_scan_global ()` sends it to a
separate branch of the same function that widens the bytes and never touches a decode buffer. The
runs that show both halves, and the ones for the SHA-1 and SHA-256 helpers, are on the other branch.

A correction to something I wrote earlier in this thread. I put the failure down to the maximum
length, and that was reading a pattern into two data points: `tools/test.pl edge 20011 0 0` emits
exactly two vectors, word length 0 and word length 256. The longest is the only one that carries a
password at all, so there is nothing shorter that passes.

Second thing worth recording: the suite cannot see the fix either. All 60 test modules that build a
UTF-16LE password call `encode ("UTF-16LE", $word)` on a byte string, which widens the bytes instead
of decoding the UTF-8. A decoding kernel and that oracle disagree on any non-ASCII password whether
the kernel is right or wrong, so the 18 stay 18 on both sides of the kernel fix, and the reproduction
on the other branch is built against Python's `hashlib` instead.

None of this is a parser question, so none of it is on this branch.

m35300 and m35400 are the Kerberos kernel problem #4899 fixes, not the parser. This branch does
touch those two modules, but only their hex loops, and the suite says much the same about them on
master.

jsteube:
In "src/modules/module_18100.c" a token takes VERIFY_HEX under OPTS_TYPE_ST_HEX and is then read as a number rather than decoded:

```
      const u8 *salt_pos = token.buf[2 * i + 1];

      // convert ascii timestamp to ulong timestamp
      u64 timestamp = hc_strtoull ((const char *) salt_pos, NULL, 10);
```

It is the only module that marks a token VERIFY_HEX and calls no hex decoder anywhere, so by the split this adds it is the BASE16 case rather than the HEX one. Its own example hash is the decimal variant, so a pass over the 599 does not enter that branch. Could you take it with the other 4?

thatux:
Ok, taken, as `db8b430a9`.

    token.attr[i + 1]     = TOKEN_ATTR_VERIFY_LENGTH
                          | TOKEN_ATTR_VERIFY_BASE16;

You are right that it is the BASE16 case and not the HEX one, and right that a pass over the 599
does not enter it. It is narrower than that: the branch cannot be entered at all today.

    ./hashcat -m 18100 --hex-salt --potfile-disable '597056:3600:613004:1234567890:322664:9876543210'
    Parameter --hex-salt not valid for hash-type 18100

`--hex-salt` sets `OPTS_TYPE_ST_HEX` only for `SALT_TYPE_GENERIC` (src/interface.c:405), m18100
declares `SALT_TYPE_EMBEDDED`, and the module does not carry the flag itself. So there is no run
that shows the difference and the commit message says so rather than inventing one. It is still the
wrong attribute for what the module does with the token.

The mode's own example hash, the decimal variant, is unchanged:

    echo hashcat | ./hashcat -m 18100 -D 1 --force --potfile-disable --quiet '597056:3600:613004:1234567890:322664:9876543210'
    597056:3600:613004:1234567890:322664:9876543210:NBQXG2DDMF2A====
    rc=0

jsteube (changes requested):
The casts in "src/modules/module_16800.c" and its 3 siblings went to every operand of each group. In 3 other places this PR takes one operand out of a run and leaves the rest:

```
  oldoffice01->rc4key[0] |= (u32) hex_convert (rc4key_pos[0]) << 28;
  oldoffice01->rc4key[0] |= hex_convert (rc4key_pos[1]) << 24;
  oldoffice01->rc4key[0] |= hex_convert (rc4key_pos[2]) << 20;
```

That is m09720 and m09820 at one line out of 8, and m10420 at one out of 16. Could they take the same treatment the WPA ones did?

In "src/modules/module_28503.c" it goes the other way: the cast operands lost the padding that lined the run up, and the operand left alone kept it.

```
  digest[0] = ((u32) t[1] << 27) | ((u32) t[2] << 22) | ((u32) t[3] << 17) | ((u32) t[4] << 12)
            | ((u32) t[5] << 7) | ((u32) t[6] << 2) | (t[ 7] >>  3);
```

On master that reads "t[ 1]" and "<<  7", so a single digit index and shift line up with a double. m28504, m30903 and m30904 carry the same 3 lines. Could the padding come back with them?

thatux:
Both done in 498d4c3ef.

m09720, m09820 and m10420 cast all ten lines of the rc4key run now. The run is ten lines in each of
the three, rc4key[0] and the first two of rc4key[1], and there were two `<< 28` lines in each, so
the one I had left alone was as much a signed overflow as the one I cast.

m28503, m28504, m30903 and m30904 have the padding back, `((u32) t[ 1] << 27)` and `<<  7`, so a one
digit index or shift lines up with a two digit one again.

The self-test vector of each of the seven cracks:

    for m in 9720 9820 10420 28503 28504 30903 30904; do tools/test.sh -S -m $m -D 1 -f; done

    7 x OK : 0/1 not found, 0/1 not matched, 0/1 timeout, 0/1 skipped

PR #4896, opened 2026-09-21
Let the vector width lookup see a module's own tuning row

matrix:
A module can add rows to the tuning database through `module_extra_tuningdb_block`. 30ac079ca gave those rows a name of their own, `MODULE_<device id>_<device name>`, and taught the lookup that feeds `kernel_accel` and `kernel_loops` to search under that name before the device's own. The lookup that feeds `vector_width` is in the same function and was not changed, so the `Vector_Width` column of a module row is never read.

The column itself is in active use. Of the 322 tuning rows `tunings/` ships, 287 ask for a width above 1, 13 ask for the native width and 22 ask for 1. A row a module generates is the only kind that cannot reach it.

### Reproduction

Deviating from the reproduction format in CONTRIBUTING.md, which asks for a command the reader can run without preparing anything first. Nothing in the tree as it stands shows the difference, and the reason is particular to the fourteen modes that register a block. All three functions behind them ask for a `Vector_Width` of `1`, no file in `tunings/` carries a row for any of those fourteen hash modes, and on a GPU the default path also gives 1. On a GPU what they ask for and what they get coincide, so the dropped column costs them nothing. The diff below makes MD5 ask for `4` instead, which is a value the device would not otherwise be given, and is the whole of what the reproduction needs.

```diff
--- a/src/modules/module_00000.c
+++ b/src/modules/module_00000.c
@@ -10,6 +10,7 @@
 #include "convert.h"
 #include "shared.h"
 #include "parser.h"
+#include "memory.h"

 static const u32   ATTACK_EXEC    = ATTACK_EXEC_INSIDE_KERNEL;
 static const u32   DGST_POS0      = 0;
@@ -120,6 +121,28 @@
   return out_len;
 }

+const char *module_extra_tuningdb_block (MAYBE_UNUSED const hashconfig_t *hashconfig, MAYBE_UNUSED const user_options_t *user_options, MAYBE_UNUSED const user_options_extra_t *user_options_extra, const backend_ctx_t *backend_ctx, MAYBE_UNUSED const hashes_t *hashes, const u32 device_id, MAYBE_UNUSED const u32 kernel_accel)
+{
+  hc_device_param_t *device_param = &backend_ctx->devices_param[device_id];
+
+  int   lines_sz  = 4096;
+  char *lines_buf = hcmalloc (lines_sz);
+  int   lines_pos = 0;
+
+  char *new_device_name = hcstrdup (device_param->device_name);
+
+  for (size_t i = 0; i < strlen (new_device_name); i++)
+  {
+    if (new_device_name[i] == ' ') new_device_name[i] = '_';
+  }
+
+  lines_pos += snprintf (lines_buf + lines_pos, lines_sz - lines_pos, "%s * %u 4 7 A\n", new_device_name, user_options->hash_mode);
+
+  hcfree (new_device_name);
+
+  return lines_buf;
+}
+
 void module_init (module_ctx_t *module_ctx)
 {
   module_ctx->module_context_size             = MODULE_CONTEXT_SIZE_CURRENT;
@@ -145,7 +168,7 @@
   module_ctx->module_esalt_size               = MODULE_DEFAULT;
   module_ctx->module_extra_buffer_size        = MODULE_DEFAULT;
   module_ctx->module_extra_tmp_size           = MODULE_DEFAULT;
-  module_ctx->module_extra_tuningdb_block     = MODULE_DEFAULT;
+  module_ctx->module_extra_tuningdb_block     = module_extra_tuningdb_block;
   module_ctx->module_forced_outfile_format    = MODULE_DEFAULT;
   module_ctx->module_hash_binary_count        = MODULE_DEFAULT;
   module_ctx->module_hash_binary_parse        = MODULE_DEFAULT;
```

The row it emits is the device name, any attack mode, hash mode 0, `Vector_Width` 4, `Kernel_Accel` 7, `Kernel_Loops` autotune. On an Apple M4 Pro through Apple's OpenCL:

```
$ make clean && make
$ rm -rf cache/kernels/
$ ./hashcat -m 0 -a 3 -O --force --potfile-disable --self-test-disable 8743b52063cd84097a65d1633f5c74f5 '?l?l?l?l?l?l?l'

before:
Status...........: Cracked
Speed.#02........:  9869.6 MH/s (0.07ms) @ Accel:7 Loops:1024 Thr:256 Vec:2
rc=0

after:
Status...........: Cracked
Speed.#02........:  9604.8 MH/s (0.08ms) @ Accel:7 Loops:1024 Thr:256 Vec:4
rc=0
```

`Accel:7` on both sides is the `Kernel_Accel` column of that same row, which the existing two pass lookup does read. It is what shows the row is registered, that it matches this device and this hash mode, and that `Vector_Width` is the one column being dropped.

`-O` is needed because `vector_width` is forced back to 1 further down for a kernel that is not optimized on a GPU, whatever the tuning database said:

```c
    if ((hashconfig->opti_type & OPTI_TYPE_OPTIMIZED_KERNEL) == 0)
    {
      if (device_param->opencl_device_type & CL_DEVICE_TYPE_GPU)
      {
        vector_width = 1;
      }
    }
```

### What answers instead

`Vec:2` before the change is not the device default. It is this row, which a search by device name does reach:

```
tunings/Modules_default.hctune:191:ALIAS_Apple_M                                   3       0       2       A       A
```

The module row is not outranked by it. It is never searched for. `tuning_db_search_real` looks up the exact name, then any alias, then `DEVICE_TYPE_CPU` or `DEVICE_TYPE_GPU`, then the `*` wildcard, and `MODULE_00_Apple_M4_Pro` is none of those under a search for `Apple M4 Pro`.

### The change

The block added is the `kernel_accel` one verbatim, one indentation level deeper. Stripped of comments and leading whitespace the two are 24 identical lines. It is duplicated rather than extracted into a shared helper so that this patch does not also rewrite the lookup that already works.

The first pass carries no vendor id. Without that, `tuning_db_vendor_alias` would answer it from the vendor id alone on an NVIDIA, AMD or Intel device, and an `ALIAS_NV` row would end the search before the module's own row had been tried.

Where a module registers a row, that row now outranks a hand written one for the same device and hash mode, as it already does for `kernel_accel` and `kernel_loops`. With `Apple_M4_Pro 3 0 8 A A` put in a file under `tunings/` and the module diff above applied:

```
$ ./hashcat -m 0 -a 3 -O --force --potfile-disable --self-test-disable 8743b52063cd84097a65d1633f5c74f5 '?l?l?l?l?l?l?l'

master:       Speed.#02........:  8532.3 MH/s (0.09ms) @ Accel:7 Loops:1024 Thr:256 Vec:8
this branch:  Speed.#02........:  9552.2 MH/s (0.08ms) @ Accel:7 Loops:1024 Thr:256 Vec:4
```

Revert the module diff and that same row wins on both sides again, so only a mode whose module registers a row of its own is affected. `--backend-vector-width` is read before the tuning database is consulted at all, so it overrides either row. What is left is that for those fourteen a `Vector_Width` can now be set from the command line but no longer from a file. `kernel_accel` keeps a way out that the vector width does not, since `backend.c` looks the row up by device name before calling the module and hands that one value over as `kernel_accel_user`. Putting the module pass second would invert the order, at the cost of no longer being the same block as the `kernel_accel` one, and of never reaching a module row wherever a shipped row answers first, which is the case the reproduction turns on.

### What it changes for the fourteen modes that ship a block

They register one through three shared implementations, in `scrypt_common.c`, `argon2_common.c` and `yescrypt_common.c`. A mode from the argon2 one and a mode from the scrypt one, same command against master and against this branch:

```
$ ./hashcat -m 34000 -a 3 --force --potfile-disable --self-test-disable --runtime 15 '$argon2id$v=19$m=65536,t=3,p=1$FBMjI4RJBhIykCgol1KEJA$2ky5GAdhT1kH4kIgPN/oERE3Taiy43vNN70a3HpiKQU' '?l?l?l?l?l?l?l'

master:       Speed.#02........:      204 H/s (12.17ms) @ Accel:104 Loops:12 Thr:32 Vec:1
this branch:  Speed.#02........:      204 H/s (12.17ms) @ Accel:104 Loops:12 Thr:32 Vec:1

$ ./hashcat -m 8900 -a 3 --force --potfile-disable --self-test-disable --runtime 15 --scrypt-tmto 3 'SCRYPT:16384:8:1:OTEyNzU0ODg=:Cc8SPjRH1hFQhuIPCdF51uNGtJ2aOY/isuoMlMUsJ8c=' '?l?l?l?l?l?l?l'

master:       Speed.#02........:      252 H/s (6.64ms) @ Accel:20 Loops:2048 Thr:32 Vec:1
this branch:  Speed.#02........:      251 H/s (6.65ms) @ Accel:20 Loops:2048 Thr:32 Vec:1
```

`--scrypt-tmto 3` because this device reports 3456 MB allocatable and the default TMTO does not fit. `36100` does not fit on this device at all, so `yescrypt_common.c` is the one of the three not covered by a run here.

On an OpenCL CPU device the default is whatever the runtime reports for `CL_DEVICE_NATIVE_VECTOR_WIDTH_INT`, and the `OPTI_TYPE_OPTIMIZED_KERNEL` clamp quoted under the reproduction is GPU only, so those fourteen modes are given that width instead of the 1 they ask for. None of their kernels reads `VECT_SIZE` and none of their modules sets an `OPTI_TYPE_SLOW_HASH_SIMD_*` flag, so no work item covers more candidates for it. It lands on the count the buffers are sized from instead. `device_param->vector_width` is written in one place, `backend.c:13686`, and it is both the `Vec` the status line prints and what the sizing reads. `kernel_power_raw` is `device_processors * kernel_threads * kernel_accel_max`, and:

```c
      const u64 vector_width = MAX (device_param->vector_width, 1);

      const u64 kernel_power_max = CEILDIV (kernel_power_raw, vector_width) * vector_width;
```

`size_pws`, `size_tmps`, `size_hooks` and every other buffer sized from `kernel_power_max` are rounded with it. That rounding is deliberate. ad0deb39d (#4889) put it there because a kernel built for a width above one reaches past a count that is not a whole number of vectors, and it costs at most `vector_width - 1` entries per buffer. At the width the module asks for it costs nothing, because the rounding is then by one.

What these fourteen pay on a CPU device is that cost at a width they did not ask for, and they have no slack for it. All three implementations set `kernel_accel` to `available_mem / size_per_accel`, which fills the device, `backend.c` pins `kernel_accel_min` to the same value, and the loop that walks the accel down until the buffers fit then has nowhere left to go.

The module cannot see the width when it computes that figure. It is called at `backend.c:13497`, `device_param->vector_width` is not assigned until 13686, and none of the three reads it. The width it puts in the row is the assumption the `Kernel_Accel` beside it was computed under, stated in the same row. Reading one of those two columns and not the other keeps the number and drops the assumption it was worked out from. I have no OpenCL CPU device and have not measured it.

### A mode with no module row

`module_00000.c` untouched, same command as the reproduction, master against this branch:

```
$ ./hashcat -m 0 -a 3 -O --force --potfile-disable --self-test-disable 8743b52063cd84097a65d1633f5c74f5 '?l?l?l?l?l?l?l'

master:       Speed.#02........: 11789.3 MH/s (0.93ms) @ Accel:474 Loops:1024 Thr:256 Vec:2
this branch:  Speed.#02........: 11803.0 MH/s (0.93ms) @ Accel:474 Loops:1024 Thr:256 Vec:2
```

### tools/test_edge.sh -m 0 and tools/test_package.sh

The change is in `src/backend.c` and no mode is touched, so the mode run here is the one the reproduction uses.

```
$ ./tools/test_edge.sh -m 0

[ test_edge_1789946284 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type optimized, Vector-Width 8, Target-Type single
[ test_edge_1789946284 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type optimized, Vector-Width 8, Target-Type multi
[ test_edge_1789946284 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type optimized, Vector-Width 16, Target-Type single
[ test_edge_1789946284 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type optimized, Vector-Width 16, Target-Type multi
[ test_edge_1789946284 ] > All tests done in 0d:00h:05m:39s
[ test_edge_1789946284 ] > Errors detected: 0
```

```
$ ./tools/test_package.sh --no-device

ok    every shipped folder is in the package
ok    example.dict is in the package, 128416 words
ok    --keyspace reads example.dict without a device, 128416 words
ok    a .gz wordlist reads back the same 128416 words
ok    a .xz wordlist reads back the same 128416 words

```


jsteube:
In "src/tuningdb.c", "tuning_db_search()" already owns the sequence of names a lookup walks:

```
  const char *NV_prefix = (const char *) "NVIDIA ";

  if (strncmp (device_name, NV_prefix, strlen (NV_prefix)) == 0)
  {
    entry = tuning_db_search_real (hashcat_ctx, device_name + strlen (NV_prefix), device_type, vendor_alias, attack_kern, hash_mode);

    if (entry) return entry;
  }
```

A module row is one more name in that sequence, tried first and with no vendor id, so it belongs beside these rather than in the caller. Putting it there takes the loop out of "backend.c" at both sites and closes the gap by construction, because a lookup cannot skip a pass that is not written in the caller. It needs the device id as an argument, which every call site already holds. Could you move it down there instead?

matrix:
done

PR #4898, opened 2026-09-21
Account for the launch time --pipeline-stats was dropping

matrix:
`--pipeline-stats` reports where a launch's time goes, and its stages do not add up to the wall clock. Three things are missing: the wait for a filled batch, which is time the launch thread spends not launching; whatever falls between two bracketed stages, which is dropped rather than booked; and the copy, on the paths where nothing brackets it. On a slow hash the second gap is the loop, and the report says 21 ms for a run of 45 seconds.

## Reproduction

The report prints once every 50 launches, so the run has to be long enough to reach one. `HASHCAT_PIPE` sets a different interval where a launch is slow enough to need it.

```
./hashcat -m 0 -a 4 00000000000000000000000000000000 pcfg/default-passwords.tar.xz --potfile-disable --logfile-disable --restore-disable --self-test-disable --markov-disable --force -d 1 --pipeline-stats --runtime 30
```

before, rc=4:

```
[host] device #1, 100 launches, 15483 ms total, feed 8031 (ahead, 80.31 ms), copy 2907 (18.8%, 29.07 ms), init 0 (0.0%, 0.00 ms), xfer 0 (0.0%, 0.00 ms), launch 12576 (81.2%, 125.76 ms), comp 0 (0.0%, 0.00 ms), effective 10972257 H/s, peak 1758 MB
```

after, rc=4:

```
[host] device #1, 100 launches, 17419 ms total, feed 8103 (ahead, 81.03 ms), copy 2917 (16.7%, 29.17 ms), init 0 (0.0%, 0.00 ms), xfer 0 (0.0%, 0.00 ms), launch 12524 (71.9%, 125.24 ms), comp 0 (0.0%, 0.00 ms), wait 1462 (8.4%, 14.62 ms), other 516 (3.0%, 5.16 ms), sort 0 (0.0% of copy, 0.00 ms), cells 2300 (78.9% of copy, 23.00 ms), pwsio 303 (10.4% of copy, 3.03 ms), decomp 3 (0.1% of copy, 0.03 ms), sync 310 (10.6% of copy, 3.10 ms), effective 9752400 H/s, peak 1809 MB
```

`rc=4` on both sides is the runtime limit, which is how the run ends. `-d 1` keeps it on one device, an RTX 5080 on CUDA 13.4.

Two figures the first report does not have. The 1462 ms of `wait` is time spent in `pw_pipe_take ()` with nothing to launch, and the 516 ms of `other` is the part of each launch that no bracket covers, 5.16 ms of every one. Neither was in any stage, so every percentage the first report prints is taken against a total smaller than the launches took.

The same option against a slow hash, where far more than 11 per cent is missing. `-m 23800` is one of the nineteen modes that ask for the length sort, and it runs no hooks and no bridge. The rule file multiplies the shipped dictionary so the run reaches a report, and `HASHCAT_PIPE=2` asks for one every other launch, because a launch there takes about a second:

```
HASHCAT_PIPE=2 ./hashcat -m 23800 -a 0 '$RAR3$*1*ad56eb40219c9da2*834064ce*32*13*1*eb47b1abe17a1a75bce6c92ab1cef3f4126035ea95deaf08b3f32a0c7b8078e1*33' example.dict -r rules/best66.rule --potfile-disable --logfile-disable --restore-disable --self-test-disable --markov-disable --force -d 1 --pipeline-stats --runtime 45
```

before, rc=4:

```
[host] device #1, 44 launches, 21 ms total, feed 2 (ahead, 0.05 ms), copy 3 (14.5%, 0.07 ms), init 14 (66.3%, 0.32 ms), xfer 0 (0.0%, 0.00 ms), launch 0 (0.0%, 0.00 ms), comp 4 (19.2%, 0.09 ms), effective 89708385 H/s, peak 2862 MB
```

after, rc=4:

```
[host] device #1, 42 launches, 42776 ms total, feed 2 (ahead, 0.05 ms), copy 3 (0.0%, 0.07 ms), init 14 (0.0%, 0.33 ms), xfer 0 (0.0%, 0.00 ms), launch 0 (0.0%, 0.00 ms), comp 4 (0.0%, 0.10 ms), wait 1 (0.0%, 0.03 ms), other 42754 (99.9%, 1017.95 ms), sort 1 (23.4% of copy, 0.02 ms), cells 0 (0.0% of copy, 0.00 ms), pwsio 2 (71.2% of copy, 0.05 ms), decomp 0 (0.5% of copy, 0.00 ms), sync 0 (4.9% of copy, 0.00 ms), effective 42228 H/s, peak 2916 MB
```

Both ran for 45 seconds, 44 launches on one side and 42 on the other. The first report claims 21 ms of that and says nothing about the rest. The second claims 42776 ms and puts 99.9% of it in `other`, which is the loop: `PIPE_LAUNCH` is filled in `choose_kernel ()` and in `run_bridge_loop ()`, and a slow hash whose loop is a plain kernel passes through neither, so everything between the init stage and the comp stage is discarded by the mark that starts the comp stage.

## The report changes, the run does not

`effective` is the candidates divided by the total, and the total now carries the wait and the residue, so the same work is divided by a larger denominator. Four runs a side of the first command, each read from its report at 100 launches so that the four are the same quantity, sorted:

```
before  10972257  10976022  10981900  11011369 H/s
after    9732247   9742201   9744820   9752400 H/s
```

The medians are 10978961 and 9743511 H/s, 11.3 per cent apart, which is what `wait` and `other` come to together in those four runs: 7.9 to 8.4 per cent and 3.0 per cent. The figure moved because the denominator did. What did not move is the speed the status screen reports for the device, from those same runs and sorted the same way:

```
before  22922.6  22925.0  22939.4  23032.6 MH/s
after   22907.9  22959.6  23028.9  23029.8 MH/s
```

Medians of 22932.2 and 22994.3 MH/s, three tenths of a per cent apart with the patched side the higher of the two, so the extra `pipe_mark ()` and `pipe_acc ()` calls per launch do not show. With the option off both return on their first line.

## What the stages are

`wait` is in the total, because a launch thread with no batch to launch is not launching. `feed` stays out of it, because it runs on the producer thread and costs the launch nothing. `other` is in it and is what makes the total trustworthy: the stages are bracketed one at a time, so anything between two of them would otherwise be dropped in silence.

`--task-time-breakdown` already reports this way, and `docs/hashcat-task-time-breakdown.md` says why: every section ends with an `Other` line, printed even when it is zero, "because unclassified time is the thing worth seeing".

The clock `other` is measured against starts when the first launch ends, together with what the stages had claimed by then. The two have to cover the same launches, or the subtraction comes out short by everything before the clock started, and with the default interval of 50 that is enough to report a residue of zero on a device whose stages are busy.

Five of the stages are quoted as a share of `copy` rather than of the total, because they are not beside the copy, they are what it is made of: the length sort, the cell upload, the `pws_idx` and `pws_comp` upload, the decompress enqueue, and the wait at the end of `run_copy ()` for what that enqueue left running. Counting them in the total would book the same milliseconds twice. In the first run above they come to 2916 ms against the copy's 2917, the millisecond between them being what rounding each figure for printing costs.

That document also sets the invariant the five have to hold to: an indented line can never be more than its parent. For that to be true the copy has to be timed wherever `run_copy ()` is called, and it was not. The pipelined loop timed it only when the host is not building the candidates itself, so `-S` and a brain client lost it, and the mask producer, which has a loop of its own, timed it nowhere. The second command with `-a 3` and a literal mask in place of the dictionary and the rule file shows both sides of that: `copy 0 (0.0%, 0.00 ms)` before the change, and `copy 4 (0.0%, 0.10 ms)` with `sync 4 (98.3% of copy, 0.10 ms)` after it.

The length sort is one of the five for the same reason the wait is a stage: it runs inside `run_copy ()` and before the first mark, so it was inside the copy and inside no share. It is zero on the first run, where `-m 0` does not ask for it, and 23.4% of a 3 ms copy on the second.

One line changes meaning rather than being added. `feed` is labelled `ahead` because it runs on the producer thread, which is true only where the producer has a thread of its own. A brain client runs it on the launch thread, and there the feed happens inside `pw_pipe_take ()`, so its time is already inside the wait and is on the critical path. `pw_pipe_start ()` tells the device which of the two it is, and the report says `inside wait` rather than `ahead` for that case, in the text form and as `in_critical_path` in the JSON form.

The new stage is called `PIPE_TAKE` rather than `PIPE_WAIT` because `PIPE_WAIT` is a named pipe constant in `winbase.h`, and an enumerator of that name compiles everywhere except Windows.

## What still lands in other

`other` names the residue rather than removing it. On a slow hash it holds the loop and its prepare, the extended loop, `init2` and the second loop, and the hook round trip, which runs host threads. On the fast path it is whatever the launch loop does between two brackets. Giving each of those a stage of its own is a separate change. This one makes the time visible instead of lost.

## tools/test_edge.sh

`./tools/test_edge.sh -m 0 -K 0`, on the tree the runs above came from. The change is in the dispatcher and in the report rather than in a mode, so one mode across every attack type it supports is what there is to cover:

```
[ test_edge_1790001909 ] # Processing Hash-Type 0, Attack-Type 12, Kernel-Type pure, Vector-Width 16, Target-Type multi
[ test_edge_1790001909 ] > All tests done in 0d:00h:03m:06s
[ test_edge_1790001909 ] > Errors detected: 0
```


jsteube:
In "src/backend.c", the m23800 report after the change reads:

```
[host] device #1, 42 launches, 42776 ms total, feed 2 (ahead, 0.05 ms), copy 3 (0.0%, 0.07 ms), init 14 (0.0%, 0.33 ms), xfer 0 (0.0%, 0.00 ms), launch 0 (0.0%, 0.00 ms), comp 4 (0.0%, 0.10 ms), wait 1 (0.0%, 0.03 ms), other 42754 (99.9%, 1017.95 ms), sort 1 (23.4% of copy, 0.02 ms), cells 0 (0.0% of copy, 0.00 ms), pwsio 2 (71.2% of copy, 0.05 ms), decomp 0 (0.5% of copy, 0.00 ms), sync 0 (4.9% of copy, 0.00 ms), effective 42228 H/s, peak 2916 MB
```

The 42776 ms is right, and "launch" reads 0 on a mode whose loop is all of it, with "other" holding the 99.9%. Could you fill "PIPE_LAUNCH" on the plain kernel slow path in this PR, so that "other" keeps the meaning it carries in "pipe_slot_t"?

The text report is one line, and the 5 shares of copy take it from 231 characters to 451. At that width it wraps across several rows of a terminal with no structure, and the shares sit inline among the stages although they are a different quantity, which "pipe_slot_t" and "PIPE_TOTAL_END" already separate. Could you break after "other" and put the shares on a second line? The accounting stays as it is, and the JSON already covers anything reading this by machine.

matrix:
Both done.

`PIPE_LAUNCH` was not missing a call, it was being thrown away. The `pipe_mark ()` that follows the salt-repeat loop reset the timer without booking it, so everything between `PIPE_INIT` and the comp kernel went unclaimed and landed in the residue. `pipe_acc (device_param, PIPE_LAUNCH, &timer_stage)` now runs after the loop kernels. The bridge keeps a `pipe_mark ()` of its own after `run_bridge_loop ()`, because that call already books its span against `PIPE_XFER` and `PIPE_LAUNCH` and would otherwise be counted in both.

The report breaks after `other`. The five shares of copy get the second line, with their text unchanged.

`-m 23800`, the same hash, on an Apple M4 Pro through Apple's OpenCL:

```
[host] device #2, 8 launches, 585603 ms total, feed 0 (ahead, 0.00 ms), copy 12 (0.0%, 1.44 ms), init 164 (0.0%, 20.47 ms), xfer 0 (0.0%, 0.00 ms), launch 584983 (99.9%, 73122.82 ms), comp 9 (0.0%, 1.07 ms), wait 0 (0.0%, 0.00 ms), other 436 (0.1%, 54.54 ms), effective 19724 H/s, peak 807 MB
[host] device #2, sort 0 (0.0% of copy, 0.00 ms), cells 0 (0.0% of copy, 0.00 ms), pwsio 0 (0.0% of copy, 0.00 ms), decomp 0 (0.0% of copy, 0.00 ms), sync 12 (99.9% of copy, 1.44 ms)
```

`launch` holds the share your run had under `other`, and `other` is back to the residue `pipe_slot_t` describes. The first line is 292 characters where yours was 451.

One thing you did not specify: `effective` and `peak` stayed at the end of the first line rather than moving with the shares, because they come from the total and not from copy. Moving them is one line if you would rather have the break fall immediately after `other`.

PR #4899, opened 2026-09-21
Kerberos bug: Give the CPU RC4 S box to the work item that uses it

thatux:
m35300 and m35400 miss a password they hold the candidate for, on a CPU device, as soon as the run
carries more than one candidate. This fixes the S box layout that causes it, and raises the ceiling
from one candidate to a work group; the section at the end says what is still wrong past that.

    ./hashcat -m 35300 -a 3 -D 1 --force --potfile-disable '$krb5tgs$23$*user$realm$test/spn*$b548e10f5694ae018d7ad63c257af7dc$35e8e45658860bc31a859b41a08989265f4ef8afd75652ab4d7a30ef151bf6350d879ae189a8cb769e01fa573c6315232b37e4bcad9105520640a781e5fd85c09615e78267e494f433f067cc6958200a82f70627ce0eebc2ac445729c2a8a0255dc3ede2c4973d2d93ac8c1a56b26444df300cb93045d05ff2326affaa3ae97f5cd866c14b78a459f0933a550e0b6507bf8af27c2391ef69fbdd649dd059a4b9ae2440edd96c82479645ccdb06bae0eead3b7f639178a90cf24d9a' 'b4b9b02e6f09a9bd760f388b67351e2?h'

    before: Recovered 0/1, Status Exhausted, rc=1, five runs out of five
    after:  Recovered 1/1, Status Cracked,   rc=0, five runs out of five

    ./hashcat -m 35400 -a 3 -D 1 --force --potfile-disable '$krb5asrep$23$user@domain.com:3e156ada591263b8aab0965f5aebd837$007497cb51b6c8116d6407a782ea0e1c5402b17db7afa6b05a6d30ed164a9933c754d720e279c6c573679bd27128fe77e5fea1f72334c1193c8ff0b370fadc6368bf2d49bbfdba4c5dccab95e8c8ebfdc75f438a0797dbfb2f8a1a5f4c423f9bfc1fea483342a11bd56a216f4d5158ccc4b224b52894fadfba3957dfe4b6b8f5f9f9fe422811a314768673e0c924340b8ccb84775ce9defaa3baa0910b676ad0036d13032b0dd94e3b13903cc738a7b6d00b0b3c210d1f972a6c7cae9bd3c959acf7565be528fc179118f28c679f6deeee1456f0781eb8154e18e49cb27b64bf74cd7112a0ebae2102ac' 'b4b9b02e6f09a9bd760f388b67351e2?h'

    before: Recovered 0/1, Status Exhausted, rc=1, five runs out of five
    after:  Recovered 1/1, Status Cracked,   rc=0, five runs out of five

`?h` is 16 candidates, one of them the example password. `rm -rf cache/kernels/` between the two
builds.

## The cause

Both kernels declare their RC4 state as

```c
LOCAL_VK u32 S[64 * FIXED_LOCAL_SIZE];
```

one 256 byte S box for every work item in the group. The GPU half of inc_cipher_rc4.cl addresses it
that way, interleaved so each work item keeps to one memory bank. The `IS_CPU` half takes `lid` and
throws it away:

```c
DECLSPEC u8 GET_KEY8 (LOCAL_AS u32 *S, const u8 k, MAYBE_UNUSED const RC4_LID_TYPE lid)
{
  LOCAL_AS u8 *S8 = (LOCAL_AS u8 *) S;

  return S8[k];
}
```

so every work item in the group shares the first 256 bytes. That holds only for a group of one, and
nothing in the header says so. Each work item gets its own 256 bytes now, which is the layout the
declaration already promises; there are no banks to dodge on a CPU, so the linear layout stays
linear.

Every other caller compiles with `FIXED_LOCAL_SIZE` 1 on a CPU, where `lid` is 0 and
`(lid * 256) + k` is `k`, so no address moves for them. m35300 and m35400 no longer need the CPU
exception either and ask for the 32 every other RC4 mode asks for.

## What is still wrong past 32 candidates

Both modes start failing again once the run holds more than a work group, and that is not the S box
layout: it survives this change, it survives a local size of 1, and it goes away when POCL is held
to one worker thread.

    awk 'BEGIN{for(i=0;i<99;i++)printf "b4b9b02e6f09a9bd760f388b67351%03d\n", i}' > /tmp/k100.txt
    echo 'b4b9b02e6f09a9bd760f388b67351e2b' >> /tmp/k100.txt

    ./hashcat -m 35300 -a 0 -D 1 --force --potfile-disable '$krb5tgs$23$*user$realm$test/spn*$b548e10f5694ae018d7ad63c257af7dc$35e8e45658860bc31a859b41a08989265f4ef8afd75652ab4d7a30ef151bf6350d879ae189a8cb769e01fa573c6315232b37e4bcad9105520640a781e5fd85c09615e78267e494f433f067cc6958200a82f70627ce0eebc2ac445729c2a8a0255dc3ede2c4973d2d93ac8c1a56b26444df300cb93045d05ff2326affaa3ae97f5cd866c14b78a459f0933a550e0b6507bf8af27c2391ef69fbdd649dd059a4b9ae2440edd96c82479645ccdb06bae0eead3b7f639178a90cf24d9a' /tmp/k100.txt

      12 POCL worker threads   : Status Exhausted, rc=1, four runs out of five
      POCL_MAX_PTHREAD_COUNT=1 : Status Cracked,   rc=0, five runs out of five

Building the module at 4, 8, 16 and 32 and walking the candidate count puts the ceiling at the group
size on both sides of this change. Correct word last, three runs each, 1 is cracked:

| FIXED_LOCAL_SIZE | last count that works | first count that fails |
| --- | --- | --- |
| 4  | N=4  `111` | N=6   `000` |
| 8  | N=8  `111` | N=12  `000` |
| 16 | N=16 `111` | N=24  `000` |
| 32 | N=32 `111` | N=64  `000` |

On master, where a CPU gets 1, that ceiling is one candidate, so the modes are wrong for anything
past a self test. This raises it to 32; a real run of millions of candidates is still wrong.

Two things say the header is not it on its own. m10500 is the other slow hash on it, compiles with
the same `FIXED_LOCAL_SIZE` 1 on a CPU, and cracks the same 100 candidate list five times out of
five either way. m13100 and m18200 run the same RC4 against the same etype 23 and differ in being
fast hashes rather than slow ones, and `test_edge.sh` reports 0 errors for both.

All of it was measured on POCL on a 12 core CPU. There is no GPU here, so whether the remainder
reproduces on one is untested, and if it does not, the ceiling this raises is the whole of what
hashcat can do about it. The one thing m35300 and m35400 do that m10500 does not is run RC4 in the
comp kernel rather than the loop kernel, which is where I would look next.

## test_edge.sh

PENDING


thatux:
The `## test_edge.sh` section of the description above still says PENDING. This is it.

Every mode that reaches `inc_cipher_rc4.cl`, the full matrix, against the same suite run on
836f11de1, which is what this branch sits on:

    tools/test_edge.sh -m <mode> -D 1 -f

| mode | master | this branch |
| --- | --- | --- |
| m07500 | | 0 |
| m09700 | | 0 |
| m09710 | 15 | 15 |
| m09800 | | 0 |
| m09810 | | 0 |
| m09820 | | 0 |
| m10400 | | 0 |
| m10410 | | 0 |
| m10500 | | 0 |
| m10510 | | 0 |
| m13100 | | 0 |
| m18200 | | 0 |
| m25400 | 100 | 100 |
| m33500 | | 0 |
| m33501 | | 0 |
| m33502 | | 0 |
| **m35300** | **79** | **6** |
| **m35400** | **82** | **3** |

All eighteen were run on this branch. The master column is filled in only for the four that are not
zero, which are the four that needed a baseline; the other fourteen are zero here and were not
re-run on master.

m35300 goes from 79 errors to 6 and m35400 from 82 to 3. That is the same fix the mask in the
description shows, seen across every attack type, kernel type, target type and vector width the
suite has, rather than on one hash.

m09710 and m25400 come back with the same number on both sides, to the error, so neither is this
change. They are what master does today, and both have a shape worth recording even though neither
belongs in this pull request:

* **m09710**, 15 errors, every one of them attack mode 3, at word length 5, across all five vector
  widths, 10 single target and 5 multi. The words are plain ASCII digits so this is not the non
  ASCII mask problem the suite has elsewhere. It is not really 15 failures either: hashcat refuses
  to start at all, and only the suite's `--self-test-disable` lets it get as far as failing to
  crack.

      ./hashcat -m 9710 -a 3 -D 1 --force -O --potfile-disable '$oldoffice$0*73755538499428852818012347339288*cb55c15298f070630c8ebfe0d5ce4aec*fc43b2f9adb65653a9134d1ba660ff2f' '54?d?d?d'
      * Device #1: ATTENTION! OpenCL kernel self-test failed.
      Aborting session due to kernel self-test failure.

      ./hashcat -m 9710 -a 0 -D 1 --force -O --potfile-disable '<the same hash>' <(echo 54005)
      Status...........: Cracked

  With `--self-test-disable` it runs and finds nothing, whichever position the mask varies: `?d4005`,
  `54?d05`, `540?d5` and `5400?d` all come back not cracked against a hash whose password is 54005.
  m09810, the same collider a rung along, passes its `-a 3` self-test. I could not localise it, and
  what I ruled out is worth recording so the next person does not repeat it:

  * not the declared length. Both modules pin `pw_min` and `pw_max` to 5.
  * not `OPTS_TYPE_PT_ADD80`, which is the single flag m09710 carries and m09810 does not. Taking
    it out and rebuilding leaves the self-test failing exactly as before.
  * not the kernel. m09710's `-a 3` and its own working `-a 0` are the same kernel written twice:
    the same md5 input, the same `w0[1] & 0xff`, the same `9 * 8` length, and a verification half
    that is identical line for line down to `COMPARE_M_SIMD`. The `-a 3` entry point fills `w0`
    from `pws[gid].i` exactly as m09810's working `-a 3` does.

  * not the missing vectorisation. Both kernels use a plain `u32 w0lr` where 104 of the other
    `_a3-optimized` kernels use `u32x`, but that is the convention across the 9 RC4 kernels that
    cannot vectorise a shared S box, and m09810 works under it.

  So the arithmetic is right and what reaches it is not. `selftest_init ()` puts the first byte of
  the password in `bfs_buf` and the rest in `pws` at offset 1, which is what the `w0l | w0r` in the
  kernel expects, so that much reads correctly too. Going further means instrumenting the kernel to
  dump what it actually receives, which is a session of its own rather than a note on a pull request
  about an S box layout.
* **m25400**, 100 errors, spread over attack modes 3, 4, 7, 8, 9 and 12 at word lengths 0, 1, 2 and
  15, five of each because that is the five vector widths. Attack mode 0 is clean. 100 is 20
  combinations times 5 widths rather than 100 separate faults.

Both are a mode's own business rather than the RC4 header's, and fixing either here would put an
unrelated change in a pull request about an S box layout, so they are noted rather than touched.

What is left on m35300 and m35400 is the part the description says this does not fix, and which
vectors fail says what shape it has. All nine are attack mode 3, the mask attack, and none is any
other attack mode:

    m35300, six:   vector width 2 (x2), 4, 8, 16 (x2)      none at width 1
    m35400, three: vector width 1, 4, 16

A mask attack hands the kernel many candidates at once and the vector width multiplies them again,
so this is the ceiling the description describes rather than anything the mask itself does.

jsteube:
Building each change on its own separates what they do. On pocl, m35300, 5 runs of the 16 candidate mask and 10 of the 100 candidate list:

```
                              16 cand    100 cand
  1 thread  + old S box        00000     00000
  1 thread  + new S box        00000     -
  32 threads + old S box       11111     0000000000
  32 threads + new S box       11111     0110010000
```

Raising "native_threads" to 32 is what lifts the 16 candidate miss. The S box change is what keeps 32 sound once the run outgrows one group, which is the second column. This is named for the second one.

The miss is pocl's rather than a CPU's. On master both modes crack at 16 candidates and at 100 under Intel OpenCL on 3 different CPUs, and on 5 GPUs across 3 vendors. pocl is the only device that misses, on the same i7-14700K that Intel OpenCL cracks on.

So what fails at a local size of 1 is still unaccounted for, and raising the size steps over it. Could you set out in the description what each of the 2 changes does? And is the local size 1 failure worth finding before this goes in, given that a later change to "native_threads" would bring it back with the S box fix still in place?

thatux:
Thanks, I've separated the two problems: I've changed the PR message above and moved the PoCL
problem to: https://github.com/hashcat/hashcat/pull/4905

Then I went after what was left here and could not make it fail on anything. Closing.

## Which of the two was which

Building each change on its own, as you asked, split them.

**The one that was making modes miss, now #4905.** PoCL loses the S box between two statements:
`rc4_init_128 ()` fills it and the very next call reads it back as zeros. A work group of one is
enough. It is one compiler mishandling a `__local` array across a call it does not inline, and it
is what stopped m35300, m35400, m25400 and m09710 from cracking. Folding the RC4 functions into the
kernel fixes all four at a work group of one, so `native_threads` never needs raising and this pull
request was not what fixed them.

**The one left here.** The buffer reserves one S box per work item,

    LOCAL_VK u32 S[64 * FIXED_LOCAL_SIZE];

and the `IS_CPU` code indexes the first one whatever `lid` is. Correct only for a work group of
one, which is all anything asks for.

## Nothing reaches it, and forcing it does not break either runtime

`native_threads` is set by `module_jit_build_options ()`, which overrides the tuningdb entry and
`-T` on purpose, so it is not a runtime input. All 18 RC4 modes pin a CPU to 1, and on a GPU this
branch of the header is not compiled. No hash, flag, device or wordlist reaches the wrong index.

So I forced it. m35300 with `native_threads` at 32, nothing else changed, 100 candidates, 10 runs:

    PoCL            1111111111
    Intel OpenCL    1111111111

Both give a real group of 32, and the group is full: gids are contiguous, so 100 candidates fill
groups of 32 with live work items rather than padding. A kernel printf on the Intel device
confirms `lsz=32`. All 32 index the same 256 bytes and it cracks every time on both.

A probe backs that up on PoCL, where each work item writes its own id into the shared byte, waits
and reads it back: zero clashes over a whole run. I would not lean on that one on the Intel device,
because the wait is a `volatile` loop and that is exactly what stops its vectoriser, so the ten runs
above are with no probe in the kernel at all.

## So there is no bug to fix

Two CPU OpenCL runtimes, a work group of 32, one S box shared between all of them, and neither
misses a password. I cannot produce a reproduction because there does not appear to be one. #4905
is the one with the four modes in it.

## What is left is tidiness, and I think it still counts for something

The declaration and the indexing disagree about what the buffer is:

    LOCAL_VK u32 S[64 * FIXED_LOCAL_SIZE];      // room for one per work item
    return S8[k];                                // one, shared

Only one of those can be the intent, and a reader of the header cannot tell which. At
`FIXED_LOCAL_SIZE` 1 the two readings coincide, which is why nothing is broken and why it has sat
there. Raise the number and they part: on the declaration's reading the group needs 8 KB, on the
code's it needs 256 bytes, and the code would quietly take the first 256 of the 8 KB it asked for.

That is worth three lines to me, because it is the sort of disagreement that is free to settle now
and awkward to work out later from a wrong result. It is also entirely a taste argument and I have
no measurement behind it, so it is yours to weigh.

Either form covers it. The change as it stands, or a comment in the header saying the `IS_CPU`
indexing is only correct at `FIXED_LOCAL_SIZE` 1 and I will send that instead. If neither appeals,
close it and nothing is lost.

jsteube:
In "OpenCL/inc_cipher_rc4.cl" the branch below this one keeps the addressing in a macro and leaves each function a single line:

```
  #define KEY32(t,k) (((k) * 32) + ((t) & 31) + (((t) / 32) * 2048))

  DECLSPEC void SET_KEY32 (LOCAL_AS u32 *S, const u8 k, const u32 v, const RC4_LID_TYPE lid)
  {
    S[KEY32 (lid, k)] = v;
  }

  #undef KEY8
  #undef KEY32
```

Could the IS_CPU half take that shape instead, with "KEY8" and "KEY32" defined as the linear form and undefined after it? Expanded they are "(k)", so the generated code is what it is today and this is not a performance question. What it buys is the 2 halves reading alike, and the decision to drop "lid" sitting in the macro where a reader meets it rather than being spread across 3 function bodies.

thatux:
Done in de91ca3d2. The IS_CPU half defines KEY8 and KEY32, the three accessors are one line each,
and both macros are undefined after them, the same shape as the half below.

One difference from what you wrote: with this pull request they expand to `(((t) * 256) + (k))` and
`(((t) * 64) + (k))`, not to `(k)`, because giving each work item its own S box is what the change
is. If you would rather the macros spell master's layout, `(k)`, so that dropping lid is written
down where a reader meets it rather than changed, I can do that instead, and this becomes the
restyle plus a comment saying the layout is only sound at a local size of 1.

The self-test vectors of the RC4 modes still crack on the CPU path:

    for m in 9700 9800 10400 13100 18200; do tools/test.sh -S -m $m -D 1 -f; done

    5 x OK

jsteube:
Keep the macros spelling the layout this pull request introduces. The buffer is declared "LOCAL_VK u32 S[64 * FIXED_LOCAL_SIZE]", so an address that leaves lid out hands the whole group one work item's share of a buffer sized for the group. Putting lid in the address means the header stops carrying an invariant it never stated, and at a local size of 1 it costs nothing. A comment asserting the group holds one work item is weaker than an address that does not need the assertion.

One thing in the new comment:

```
// At FIXED_LOCAL_SIZE 1, which is what every other caller compiles with on a CPU, lid is always 0
// and these are the same addresses as before.
```

"before" is the diff, and so is "nothing in the header said so" in the paragraph above it. Someone opening this file in a year has neither. Could you say what the code does instead, that lid is 0 at a local size of 1 so the address collapses to the bare index, and leave the comparison with master to the commit message?

thatux:
Done

